PR #5, opened 2019-09-03
Added --as-needed flag to the link.

kutschke:
Fixme:
 scons env.MergeFlags did not correctly add this only to the link commands.
So I had to use the -Wl, syntax to tell the compile lines to ignore it.  It
would be good to figure out how to teach MergeFlags to add it only the link
commands.

### DIFF
--- a/scripts/build/python/sconstruct_helper.py
+++ b/scripts/build/python/sconstruct_helper.py
@@ -128,7 +128,7 @@ def libPath(mu2eOpts):
 def mergeFlags(mu2eOpts):
     build = mu2eOpts['build']
     flags = ['-std=c++17','-Wall','-Wno-unused-local-typedefs','-g',
-             '-Werror','-Wl,--no-undefined','-gdwarf-2',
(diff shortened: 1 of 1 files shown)


kutschke:
Without the --as-needed flag the number of libraries linked into Analyses/src/ReadBack0_module.cc is 80; with the flag it is 18. In the commit comments I have a suggestion for a future improvement.

kutschke:
> On Sep 3, 2019, at 11:00 AM, gaponenko <notifications@github.com> wrote:
> 
> @gaponenko approved this pull request.
> 
> I am using the "review changes" control trying to "approve" but not merge the request because there are more reviewers on this one. A comment on "It
> would be good to figure out how to teach MergeFlags to add it only the link
> commands." A standard way to achieve this would be to have separate lists for compiler options and linker options. Usually there is also a separate list for preprocessor options (the -I flags).
> 

The issue is that scons thinks it's smarter than that.  I have not read the scons docs in a few years but the last time that I did, the only way to add options was to use env.MergeFlags("-option1 -option2"), which claims to be smart enough to recognize which options are needed on which build step.  It did not correctly recognize --as-needed as a link-time-only option.

PR #11, opened 2019-09-24
Configuration validation

gaponenko:
The changes are mostly straightforward migration from ParameterSet use to validated FHICL configurations.  There is a change in behavior of FilterG4Out (simplified config), with all fcl files in the repository updated accordingly.  There is also a drive-by bug fix to RootTreeSampler.


kutschke:
I spot checked about half of these and am happy to do the merge.  If I don't hear from anyone by early afternoon I will do it.

PR #12, opened 2019-09-25
add more useful debug print

rlcee:
Add a print Dave requested


kutschke:
I am comfortable merging this branch and will do so after lunch and after merging Andrei's request.

PR #15, opened 2019-09-26
bug fixes in: CalHelixFinderAlg, RobustHelixFit; added new fast track trigger sequences; started migration to Validated parameterSet in CalHelixFinder (I gave up due to a complicated nested structure inherited by the Plugin configuration); changed one parameter in CalHelixFinder trigger configuration to improve the rejection of fake tracks.

gianipez:


brownd1978:
Hi Giani,
There is too much in this pull request to process coherently.  Can you please split this up?  In particular, approving the bug fixes is easy, but approving a partial migration of CalPatRec fhicl is much more problematic.  I suggest you  close this request and make a new request with just the bug fixes, then wait to complete the fhicl migration before requesting that again.
thanks, Dave

gianipez:
Hi Dave, 

I sincerely apologize about the confusion I made on the FHICL-validation code for CalHelixFinder, but it really turned out to be non trivial to di it right because RobustHelixFinder and CalHelixFinder share the same plugin interface. 

Cheers,

Giani

PR #20, opened 2019-10-01
fixed extMonSD code to fix bug.  moved extMonSD ownership and registr…

goodenou:
…ation to SensitiveDetectorHelper


kutschke:
This summarizes a discussion that we had in a private email thread.

1) This is a fix to the ExtMon sensitive detector code that was deferred until needed.  It is now needed.  This and some other things were deferred to reduce the scope of the initial G4MT project to something more manageable.

2) Lisa has requested that we wait until further notice before pulling.  She has verified that this code works in sequential mode but still needs to verify MT mode.

3) Lisa will let us know when she updates.

4) Andrei requests that in the future we separate white space changes from real changes into separate commits.

5) I cloned both repos and diff a whitespace blind diff.  It looks good to me. I have one comment that I will put in a separate comment.

goodenou:
I checked that this code is thread safe, and it is because the EventAction and SensitiveDetectorHelper objects are thread-local.  The SDH now owns the pointer to the ExtMonFNALPixelSD.

I also changed the the stepInstanceNamesToBeProduced() function in SDH from public back to private.  I had made it public for my MT testing.

kutschke:
I merged it.

PR #21, opened 2019-10-01
update Validation to use standard reco patterns

rlcee:
update Validation fcl to use standard reco paths.  We should pull this earlier in a week so I have a  few days to stabilize the overnight validation.  There will be diffs, and there will be some missing plots in the first night after the pull.


kutschke:
I am not in the loop well enough to really check this.  But I propose to pull it at 10:30 PM this evening unless someone objects.   I would like to get started on stabilizing the nightly validation

PR #24, opened 2019-10-07
add plots that look at CPR and TPR separately

rlcee:


kutschke:
I am happy to merge this.   I see that Dave is listed as  a reviewer - do you want to have a look first?

PR #25, opened 2019-10-07
Update ExtMon pixel module geometries and update g4detector* fcl files

mhedges:
This changes the upstream and downstream pixel module mother volumes from G4ExtrudedSolid to G4Box, as geant 4.10.4 has a bug with G4ExtrudedSolids.

This PR does not fix the G4ExtrudedSolid implementation for the muonID in the ExtMon geometries (I propose this be addressed in later PR).


kutschke:
The changes look good to me.  I would like to hear from Andrei and Krzysztof before I merge.  I am very happy that you removed commented out code in a final revision.

gaponenko:
Hi Rob, Dave,  please add me as a reviewer on all ExtMon related pull requests.  I added myself to this one.

gaponenko:
Hi Michael,  the changes introduce a new ad-hoc coordinate system that shows up in the geometry file as plane z-offset changes.  The redefinition of the mother volumes should not be user visible, i.e. the bug workaround should require no changes to Mu2eG4/geom/extmon_fnal_v02.txt  

Please update the code to keep the geometry file parameters in the ExtMon coordinate system,   and compute intermediate numbers for volume placements internally.

kutschke:
Hi Michael and Andrei,

    I agree 100% that things that can be computed should be computed - the idea is that the geometry files should be the
minimal set of numbers to provide a complete description of the *nominal* geometry.   Our intention is that as-installed and
as-calibrated positions will be provided as deltas to the nominal geometry and provided by the conditions system.

       Rob

mhedges:
Ok, I think these latest commits should address the requested changes.

kutschke:
I am happy with this.  I spoke with Krzysztof and he is happy for us to proceed without his input.
I will merge now.

PR #28, opened 2019-10-09
Config val

rlcee:


Analyses/src/PointerCheck_module.cc
```diff
@@ -38,7 +40,54 @@ namespace mu2e {
...
-    explicit PointerCheck(fhicl::ParameterSet const& pset);
+
+    struct Config {
+      using Name=fhicl::Name;
+      using Comment=fhicl::Comment;
+
+      fhicl::OptionalAtom<bool> skipDereference{Name("skipDereference"), 
+	  Comment("skip dereferencing pointers")};
+      fhicl::OptionalAtom<int> verbose{Name("verbose"), 
+	  Comment("verbose flag, 0 to 10")};
```
gaponenko: It would make more sense to have Atom<bool> with a default value rather than OptionalAtom<bool> that requires more lines of code.   Also,  in configuration description art prints defaulted values for Atoms, but does not know what Optional entries imply.    Optional constructs are necessary when the default values is not static but must be determined at run time (e.g. from geometry), or when a default does not make sense (e.g. for a complicated piece of subconfig for a helper class that does not get used).

Analyses/src/PointerCheck_module.cc
```diff
@@ -45,42 +45,50 @@ namespace mu2e {
...
       using Comment=fhicl::Comment;
 
-      fhicl::OptionalAtom<bool> skipDereference{Name("skipDereference"), 
-	  Comment("skip dereferencing pointers")};
-      fhicl::OptionalAtom<int> verbose{Name("verbose"), 
-	  Comment("verbose flag, 0 to 10")};
+      fhicl::Atom<bool> skipDereference{Name("skipDereference"), 
+	  Comment("skip dereferencing pointers check"),false};
+      fhicl::Atom<int> verbose{Name("verbose"), 
+	  Comment("verbose flag, 0 to 10"),1};
```
gaponenko: This line sets verbose=1 by default.   It used to be zero.  Is this change intentional?

rlcee:
Yes, that was the way it was in the original:
_verbose = pset.get<int>("verbose",1);

Ray Culbertson  mu2e/FNAL  rlc@fnal.gov  630-840-6744  WH9E (MS122)

gaponenko:
> Yes, that was the way it was in the original: _verbose = pset.get<int>("verbose",1);

Ah.  In the intermediate OptinalAtom version it was 0:
art::EDAnalyzer(conf),_skipDereference(false),_verbose(0)

So it is back to the original setting now.

PR #32, opened 2019-10-14
preparing to use geant4 10.6

resnegfk:


kutschke:
Hi Krzysztof,

      I will merge this.
  
       For my own information, please explain a little about G4DataQuestionaire and what it does.  What do we loose by commenting out the call.  Also, is it gone for good or just until something is fixed?

 Rob

PR #33, opened 2019-10-16
TrkAna config change

AndrewEdmonds11:
I've changed the TrkAna config a little bit. Now the scripts write out the MC information for the "candidate" track (i.e. de for most scripts, but ue for the upstream script). I also fixed a big with TrkAna trying to get a TrkPID object when TrkPID writing had been turned off.


AndrewEdmonds11:
Yes, I checked TrkAnaReco.fcl and TrkAnaEnsemble*.fcl and saw the MC branches stayed for de. I have not checked TrkAnaDigisReco.fcl but can look into that if needs be.

PR #36, opened 2019-10-22
added filter option to the HelixFilter + added track sequences for low p e-/e+ from the Stopping Target

gianipez:
1) added an option to the HelixFilter that allows prescaling the result based on the phi of the point of closest approach; the function used results from Michael Whalen studies of the IPA e-.

2) added IPA-calib track sequence and track sequences for selecting low-p e-/e+ generated in the Stopping Target


kutschke:
Does he pre-scaling involve randomness?

TrkFilters/src/HelixFilter_module.cc
```diff
@@ -56,31 +56,37 @@ namespace mu2e
     double        _bz0;
     TrkFitFlag    _goodh; // helix fit flag
     std::string   _trigPath;
+    bool          _prescaleUsingD0Phi;
+    std::vector<float> _prescalerPar;
```
kutschke: A small struct would be better here - give the 3 parameters meaning names.   You can write fcl helper functions that allow you to read a table into a struct with a single function call in the calling code.  Let me know if you want to see an example.

gianipez:
Hi Rob,

thanks! Could you please show me an example? I know how to crate a struct with functions, but I’m not sure I know how you want me to implement the code with the parameter-set. 

Thanks,

Gianantonio Pezzullo, Ph.D.
giani.pezzu@gmail.com


kutschke:
Hi Giani,

  Use these as a model:

GeneralUtilities/inc/ParameterSetHelpers.hh 
GeneralUtilities/src/ParameterSetHelpers.cc

For example, the above let's you get a CLHEP::Hep3Vector from a fcl sequence of 3 values.

This model lets you do:

  MyType _foo;

and in the intiailizer:

  _foo( pset.get<MyType>("preScaleParams")),

where preScaleParams might look like:

  preScaleParams : { fraction : 0.5  frequency : 3.0  phase : 1.5707 }

Rob

gianipez:
This is a particular rescaling that, for the IPA e- case, allows to have a uniform occupancy of the tracker volume. No randomness is involved. 
In our current trigger sequences, we use two prescalesr: one at the beginning of the sequence and one at the end. The latter one allows to make the scaling on the “good events” that passed the series of filters within the sequence, while the first one will be needed in case we need to reduce the total execution time of the online reconstruction. I guess, we want to introduce randomness at the first prescaler, but that is not the topic of this pull-request.

Cheers,

Gianantonio Pezzullo, Ph.D.
giani.pezzu@gmail.com


TrkFilters/src/HelixFilter_module.cc
```diff
@@ -95,6 +101,13 @@ namespace mu2e
     return true;
   }
```
bechenard: You could make the parameter const
int HelixFilter::evalIPAPresc(const float &phi0){

kutschke:
Hi Everyone, 

  Giani just told me privately that his code builds but does not run.  I have a guess at the problem and expect a fix soon.
So it is not ready to merge.

   Giani: in the future please ask questions like that using the github tools - that way people know that it's not ready to merge.

    Rob

gianipez:
Hi,

sorry about that. The code crashes when running a specific trigger line (tprHelixCalibIPADeM). We are still investigating how to fix the bug. I just pushed some cleanup I made in the code.

Cheers,


Gianantonio Pezzullo, Ph.D.
giani.pezzu@gmail.com


gianipez:
Hi,

I implemented all the fixes needed. 

Thanks for the patience,

Giani

Trigger/data/cprSeedDeM/main_cprSeedDeMHSFilter.config
```diff
@@ -1,10 +0,0 @@
-MinNStrawHits	15
```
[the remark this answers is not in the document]
gianipez: Hi Dave,

I deleted the old .config files that are no longer used. Only the file with the list of the trigger sequences is a .config, all the others are fcl files

PR #42, opened 2019-10-30
Art3 mt

goodenou:
This branch has had master merged into it a few hours ago.  I did ceRecoSim validation on 5000 events on both this branch and Mu2e/Offline/master and the agreement was perfect in all histograms except the CPU and wall clock time histos.

This branch contains the changes related to the integration of multithreaded Mu2eG4 with multithreaded art3.  There is a new module Mu2eG4MT_module.cc that runs art and G4 in MT mode.  The module Mu2eG4_module.cc has been completely re-worked and now runs only in sequential mode.  No changes to anyone's workflow is needed to run standard Mu2eG4.


rlcee:
I see you have geant 10_5.  We have not decided to allow this on the head yet

goodenou:
Oh, I need to use version 10_5 in order to avoid a seg fault in the destructors of the Worker Run Managers caused by a bug in G4 when using the ShieldingM physics list in MT mode.

kutschke:
I see two ways around this:

1) add a buildopt for g4 10_5 and update setup.sh to respond to it.
2) Add a copy of setup.sh setup_g4_10_5.sh with the one line change

In both cases we need to back out the change when it is no longer needed.

I think that 1) is more elegant but more work and easier to forget to back it out.
So I vote for 2).

Any other ideas?

Rob

ConditionsService/inc/ConditionsService.hh
```diff
@@ -123,5 +123,5 @@ namespace mu2e {
 
 }
 
-DECLARE_ART_SERVICE(mu2e::ConditionsService, LEGACY)
+DECLARE_ART_SERVICE(mu2e::ConditionsService, SHARED)
```
kutschke: I am confused about this change in several ways.  

First, I was surprised to see it made SHARED. 

Second, I cloned both Mu2e::master and goodenou:art3_MT locally.  And in both of my clones it shows as LEGACY?

Does anyone understand what's up?

goodenou:
Any service that is listed in any fcl file within the job that runs a SHARED module has to be a SHARED service, even if that service is not used by that module.  There is no harm in having a SHARED service used by LEGACY modules, as they will always be run in single-threaded mode.  My notes indicate that I modified the Condtions, DB, Proditions, and G4Helper Services to be SHARED for the POT jobs.  We can keep things as they are with the SHARED Services, or we can have a very different set of fcl files for the MT jobs, though some of the services may need to remain SHARED anyway.

As for why goodenou:art3_MT doesn’t show the change to SHARED, I am confused.  I certainly pushed my most recent version to the repo.  And my remote shows CondServ as SHARED.

gaponenko:
Hi Rob,
I also had a copy of Lisa's art3_MT in my working repo.   After doing "git checkout art3_MT; git pull goodenou art3_MT", I see "SHARED" in the ConditionService header file.
Andrei

kutschke:
I will get to this after my talk at the Ops Planning meeting tomorrow.

goodenou:
Is there any news on this pull request?  Do I need to make any changes?

kutschke:
HI Lisa,

  My apologies for being slow getting back to this.  This commit will need a little work before I can merge it in.  The first things are:

1. We do not want the default g4 to change at this time. Please copy your setup.sh to setup_g4_10_5.sh and  then reset setup.sh to it's original state.
2. The two files DAQDataProducts/src/SConscript and DAQ/src/SConscript are deleted in your version.  I vaguely recall that I told you to do this as a development expediency because some dependency was missing/broken.  This needs to be resolved before we merge.
3. I understand why the merge button is no longer green.  When I merged in Roberto's pull request it also changed Sources/src/SConscript.  The conflict resolution is easy - he made a lot of changes but you added only one line to one link list.  Then there is a ton of white space changes.  To see you change in isolation: git diff  b48fc0c1efacc Sources/src/SConscript 

I am looking at a few other things now.

PR #43, opened 2019-10-31
Valfix

brownd1978:
Fix problem with ceMixDigi validation.  This was using old mixin files which were too small for sequential mode.  It also was only testing 1 actual event (out of 10), I increased to 50 (15 actually tested), which takes 10 minutes.  For safety I reduced the skip in the mixin files, but left the mode as sequential.


rlcee:
- Please change back the file names in Validation/fcl/ceMixDigi.fcl
- Please comment on the rules/philosophy of including unrelated TrkDiag changes in this commit.
- Please comment on whether, in this case, and in general, how you want to coordinate merges?  If I am assigned a PR, can I just pull it whenever it is ready?  What if it doesn't fit in the larger plan of merges?  (i.e. preparing for a release, or trying to complete a large unrelated merge.)  I don't sit in your and Rob's planning discussions.  Maybe only you two should merge and I should be a reviewer, and backup merger only.

brownd1978:
item 1: done
item 2: these changes were stashed in my work copy they aren't related but I didn't want to lose them.
item 3: I re-assigned myself as approver.  I will consider you to have approved once you complete your review.

PR #45, opened 2019-11-04
Addition of scripts for cosmic and beam resampling

oksuzian:


kutschke:
One question:  is  Filters/src/FilterCosmicsStage1_module.cc a well chosen name?  Is this useful for any (or at least most) cosmic stage 1 jobs?  If not you should think about a new name.

Otherwise I am ready to approve this request - it's a pure addition.

oksuzian:
Hi Rob,

I plan to use FilterCosmicsStage1_module.cc to filter stage 1 cosmics.
For example, I will use 3 exclusive sub-samples defined  as such:
1) Gap: class of events that deposit less than 10 MeV in CRV
2) Upstream: class of events that deposit more than 10 MeV in CRV and pass through the upstream CRV region
3) Downstream: class of events that deposit more than 10 MeV in CRV and pass through the downstream CRV region.

We can choose unique resampling factors for each subsample:
(1) - 1000 
(2) - 100 
(3) - 10 

Does that answer your question?
I can change the name to something more appropriate. 

   Best, Yuri.

kutschke:
THanks Yuri,

 Does it make sense to use this with any of Daya Bay, CRY, Corksika?

Rob

oksuzian:
Yes, I would recommend using the same approach for all cosmic generators.
CRY is not a special case.

kutschke:
Thanks Yuri.  I will assign it to myself and merge it.

PR #49, opened 2019-11-11
E19

rlcee:
Move the compiler to e19.  All products except pcie_linux_kernel_module are exactly
the same code.  A few minor fixes had to be done due to tighter warnings.  Rob or Lisa
should confirm the change to Mu2eG4/src/PrimaryGeneratorAction.cc


goodenou:
The change made to PrimaryGeneratorAction.cc is fine.

PR #50, opened 2019-11-11
Sequential Mixing Update

AndrewEdmonds11:
I've updated the sequential mixing code so that it will skip a number of background events between 0 and skipFactor*meanEvents*PBI, rather than between two user-defined fcl parameters. skipFactor is a new fcl parameter that defaults to 1. These changes mean that the skipping will work (i.e. not throw an exception) for all background frames. 

I've also double-checked that a background frame with a single file will skip a different number of background events when it wraps around to the beginning again.

### DIFF
--- a/EventMixing/src/MixBackgroundFrames_module.cc
+++ b/EventMixing/src/MixBackgroundFrames_module.cc
@@ -53,13 +53,9 @@ namespace mu2e {
           Comment("control the level of debug output"),
           0u
           };
-      Atom<int> minSkip { Name("minSkip"),
-	  Comment("minimum number of background events to skip at the start of a file"),
-	  0
-	  };
-      Atom<int> maxSkip { Name("maxSkip"),
-	  Comment("maximum number of background events to skip at the start of a file"),
-	  10000
+      Atom<int> skipFactor { Name("skipFactor"),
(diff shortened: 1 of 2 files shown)


EventMixing/src/MixBackgroundFrames_module.cc
```diff
@@ -96,8 +92,7 @@ namespace mu2e {
 
     ProtonBunchIntensity pbi_;
     int totalBkgCount_;
-    int minSkip_;
-    int maxSkip_;
+    int skipFactor_;
```
gaponenko: Why is this an integer and not a real number?  Perhaps I do not quite get the "philosophical"  meaning of this factor.   I understand the technical meaning and see that it is multiplied with real numbers in its use.

AndrewEdmonds11: I was just thinking of it as the number of mean microbunches you want to be able to skip in the background frame. I'm happy to change it to a real number or find a better name :)

gaponenko:
Changing it to a real number would make sense.  One may want to skip half of a  microbunch.

Andrei

AndrewEdmonds11:
Done :)

PR #54, opened 2019-11-12
Fixed ensemble scripts for python3

bonventre:


bonventre:
I was worried for a second when I saw that too! But no, it just changed the default number of jobs for the generate_fcl script which I overwrote anyway.

PR #65, opened 2019-11-23
Added support for compact CORSIKA binaries (40x smaller files)

soleti:
Now it is possible to read from CORSIKA binary files produced with the "compact" option, which produces files on average 40x smaller. With this option the electromagnetic interactions are calculated analytically with the non-linear Klein-Gordon equations and are not simulated with EGS4. This affects only the Cherenkov radiation in the GeV range. This should speed up the processing, especially when the resampling option is on.

### DIFF
--- a/EventGenerator/test/corsikaTest.fcl
+++ b/EventGenerator/test/corsikaTest.fcl
@@ -17,7 +17,7 @@ targetParams: {
 
 source: {
     module_type: FromCorsikaBinary
-    fileNames: ["/pnfs/mu2e/persistent/users/srsoleti/corsika/DAT110001"]
+    fileNames: ["/mu2e/app/users/srsoleti/corsika-77100/run/DAT110001"]
     runNumber          : 1
     firstSubRunNumber  : 0
     firstEventNumber   : 1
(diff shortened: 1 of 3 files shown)


EventGenerator/test/corsikaTest.fcl
```diff
@@ -26,6 +26,7 @@ source: {
     @table::targetParams
     seed: 1
     resample: false
+    compact: false
```
brownd1978: Is the default then not to use the compact format?

Sources/inc/CosmicCORSIKA.hh
```diff
@@ -241,6 +242,7 @@ namespace mu2e {
       unsigned int _primaries = 0;
 
       bool _resample = false;
+      bool _compact = false;
```
brownd1978: Same question here.

EventGenerator/test/corsikaTest.fcl
soleti: I left the full version as default because I wanted to check the differences with a large dataset in the compact format. I did that and I can't see any difference, switching to compact as default.

Sources/inc/CosmicCORSIKA.hh
soleti: Changed as above.

soleti:
I wrote few words on CORSIKA and on this module on the wiki which might be helpful: https://mu2ewiki.fnal.gov/wiki/CORSIKA

PR #67, opened 2019-11-26
Hayman2

rhbob:
This is the Hayman 2 target that is the official replacement for the tier 1.  I am sure this will not be up to standards -- I will be happy to work with people to improve my code, but we do need to have this in for future production running.


resnegfk:
I made a few suggestions which Bob agreed to incorporate.
I suggest that the review continues once Bob makes another pull request.

PR #70, opened 2019-12-04
CORSIKA fcl files and small fixes

soleti:
- I created two FCL files for the production of a CORSIKA sample, one for generation and one for reconstruction. 
- I added some heavy nuclei to the list of CORSIKA particles that weren't included before
- I made a small fix to CosmicCRY: now it prints some settings after assignation and not before (e.g. _cosmicReferencePointInMu2e).


kutschke:
It looks good to me.  I will approve it now.

PR #74, opened 2019-12-18
updates to the Trigger-related code

gianipez:
- improved the TriggerResultsNavigator class with functions that retrieve:
    1) the list of the modules associated to a given trigger_path
    2) the index (and name) of the last module passed in the sequence
- removed the "getMany" function in the Trigger/src/ReadTriggerInfo module
- a few parameters change in the new track sequences under development
- added the calculation of the helix hitRatio (expected to measured) in the HelixFilter module


TrkFilters/src/HelixFilter_module.cc
```diff
@@ -191,6 +201,61 @@ namespace mu2e
     }
     return true;
   }
+
+
+  float   HelixFilter::helixHitRatio(const HelixSeed* helix){
+
+    int      nCH = helix->hits().size();
+    int      nFaces(36); //was 36, now doing stations instead of faces
```
brownd1978: This should come from StrawId.  The name is also misleading, 36 is # of planes not faces or stations.

TrkFilters/src/HelixFilter_module.cc
```diff
@@ -191,6 +201,61 @@ namespace mu2e
...
+    float    expected_hits(0);
+    float    z_step(178.); //are they really equally spaced? was 89 before
+    float    minR_tracker(380.);
+    float    maxR_tracker(860.);
```
brownd1978: The active volume of the tracker extends to 700, did you mean 680 here?

TrkFilters/src/HelixFilter_module.cc
```diff
@@ -191,6 +201,61 @@ namespace mu2e
...
     return true;
   }
+
+
+  float   HelixFilter::helixHitRatio(const HelixSeed* helix){
+
+    int      nCH = helix->hits().size();
+    int      nFaces(36); //was 36, now doing stations instead of faces
+    float    expected_hits(0);
+    float    z_step(178.); //are they really equally spaced? was 89 before
```
brownd1978: Is this an algorithm parameter?  If so it should be in fcl.

TrkFilters/src/HelixFilter_module.cc
```diff
@@ -191,6 +201,61 @@ namespace mu2e
...
+      //now check that we are in the active area of tracker
+      double   hitR = sqrt(pos.x()*pos.x() + pos.y()*pos.y());
+      if (hitR < minR_tracker)  continue;
+      if (hitR > maxR_tracker)  continue;
+
+      ++expected_hits;
+    }
+
+    //each face has two planes. We are assuming 2 ComboHits per face
+    float   hitRatio = nCH / (2.*expected_hits);
```
brownd1978: This assumes every plane produces 2 hits, which may be 'true' statistically, but  It would be more consistent if you compared to the number of planes with hits, as expected_hits is really expected_planes.  In any event the value 2 is an algorithmic parameter and should be in fcl.

Trigger/fcl/templates.fcl
```diff
@@ -212,8 +212,7 @@ Trigger : {
 			     tprSeedDeM_trigger, tprSeedDeP_trigger, cprSeedDeM_trigger, cprSeedDeP_trigger ]
 #	    Compressionlevel: 0
 	    
-	    outputCommands: [ "keep *_*_*_*",
-			      "drop *_daq_*_*" ]
+	    outputCommands: [ "keep *_*_*_*" ]
```
brownd1978: Do we want to keep all the data products by default?  That could be confusing if the trigger occurs at different points (ie prescaled random triggers won't have TimeClusters, etc).   In sim production we always start with 'drop *_*_*_*' and then add.

Trigger/fcl/templates.fcl
```diff
@@ -204,14 +204,21 @@ Trigger : {
     outputs: {
 	triggerOutput : { 
 	    module_type: RootOutput
-	    fileName: "filteredOutput.root"
+	    fileName: "filteredOutput_%#.root"
```
brownd1978: Is the %# replaced with something meaningful in the python code?  If the file isn't valid fhicl we should name it something else: .fclpy?

Trigger/src/ReadTriggerInfo_module.cc
```diff
@@ -744,108 +744,88 @@ namespace mu2e {
...
+	    findTrigIndex(_trigHelix, moduleLabel, index);
+	    _trigHelix[index].label  = moduleLabel;
+	    _trigHelix[index].counts = _trigHelix[index].counts + 1;
+	    const HelixSeed*hseed = trigInfo->helix().get();
+	    if(hseed) {
+	      fillHelixTrigInfo(index, hseed, _helHist);
+	      fillOccupancyInfo(_nTrackTrig+index, sdCol, cdCol, _occupancyHist);
+	    }
+	    
+	  }else if ( moduleLabel.find("TSFilter") != std::string::npos){
```
brownd1978: This code compares to hard-coded strings: shouldn't those be fcl parameters?  or come from the trigger table?  I'm not sure of the architecture to know which.

Mu2eUtilities/inc/HelixTool.hh
```diff
@@ -60,10 +60,42 @@ namespace mu2e {
...
+	//now check that we are in the active area of tracker
+	double   hitR = sqrt(pos.x()*pos.x() + pos.y()*pos.y());
+	if (hitR < minR_tracker)  continue;
+	if (hitR > maxR_tracker)  continue;
+
+	++expected_hits;
+      }
+
+      //each face has two planes. We are assuming 2 ComboHits per face
+      _hitRatio = nhits / (2.*expected_hits);
```
brownd1978: This code block seems to be a copy of what's in the helix filter, please consolidate and make the changes suggested there.

TrkFilters/src/HelixFilter_module.cc
gianipez: Thank you for noticing it. I changed it as suggested.

TrkFilters/src/HelixFilter_module.cc
gianipez: typo... DONE

TrkFilters/src/HelixFilter_module.cc
gianipez: so, the model is considering the tracker as a perfect cylinder with equally spaced planes. This is actually a typo!, because z_sep = 3200. mm / nPlanes = 88.89 mm
Is it reasonable to be hardcoded?

TrkFilters/src/HelixFilter_module.cc
gianipez: Thank you for the suggestion. I'm now testing the plane counting instead oh the comboHits.

Trigger/fcl/templates.fcl
gianipez: The line I removed was not allowing the art::Fragment dataProducts to be stored in the output file, while this is something we want. TBH, Eric wants to use a different module that lives in artdaq ("RootDAQOutput") for handling the creation of the art files. 

I think that the discussion of "which data products to keep and which not" will be handled more in detail once we will have a more robust model of the DQM. For analyzing the Trigger performance my guess is that we might need almost all of them, otherwise we would loose track of the "objects" that fired a given trigger.

Trigger/fcl/templates.fcl
gianipez: to make it short and concise, I started learning how in OTS we will be assigning the subRun number (it will come from the CFO) and how to create an art file per subRun. Kyle pointed me this page: https://cdcvs.fnal.gov/redmine/projects/art_root_io/wiki/Output-file_handling 
so the way I set the RooutOutput, should create a new .art file every 1000 events. The %# should be increased automatically as the files get created.

Trigger/src/ReadTriggerInfo_module.cc
gianipez: I followed the convention I have been using for all the current Track Triggers. I also don't like it, but I'm not sure that making it a fcl parameter would be better, because the code is extremely tight to the output object of each stage.

Mu2eUtilities/inc/HelixTool.hh
gianipez: DONE

kutschke:
HI Giani,

   Please do not hard code any of these numbers.  This is very, very important.
  
  They are all available from the Tracker object in the geometry service.  We commit
to maintaining both the nominal geometry of the tracker and the our best understanding
 of the as-aligned positions of the tracker.  I believe that the ideal tracker you describe is
available from the nominal geometry.

    If you need a number that is not in the nominal geometry description either 
 
 a) Compute it from numbers that are availalbe
 b) add the information to the nominal geometry

Rob

gianipez:
Hi Rob, 

Thank you for the comments. I am working on it.

Thanks,


Gianantonio Pezzullo, Ph.D.
giani.pezzu@gmail.com


gianipez:
Hi Rob,

I just noticed one strange thing; in TrackerGeom/inc/Tracker.hh there is a function declaration “zHalfLength” that is not set in the source file. Isn’t that strange? What do you suggest would be the best way to get the total tracker length for my cylinder approximation?

Thanks,

Gianantonio Pezzullo, Ph.D.
giani.pezzu@gmail.com


gianipez:
Hi,

I just found this:


I’m testing it right now.

Cheers,

Gianantonio Pezzullo, Ph.D.
giani.pezzu@gmail.com


kutschke:
Hi Giani,

  Please explain what you just found?

  I see that the function zHalfLength is declared but it is not implemented anywhere.

  You might mean many different things by the half length:
  1) the half length of the cylinder that exactly encloses all stations
  2) the same plus a little padding
  3) half of the length between the z of the most upstream straw and the most downstream straw
  4) the half length between the centers of the most upstream and downstream stations

 Which of these do you need? Or do you need something else?

  I believe that 2) is available via:

    tracker.mother().zHalfLength();

You can construct 3 and 4 by asking for the z of straws with particular IDs and subtracting.

Rob

kutschke:
Hi Giani,

  Also, if you just want the z of each plane of wires, ask for the nominal value of z for
wires 0 and 1 of a panel on each side of each plane.   I think that this is a better way to
do it instead of back computing spacing parameters and then recomputing straw positions.

Rob

gianipez:
Hi Rob,

Good point. I think I know where to copy the loop over the planes, I will try to implant it now. 
I just found out that the accessor “rOut” of the Tracker class delivers 0 every time. If it’s not used anywhere, should I remove it?

Cheers,

Gianantonio Pezzullo, Ph.D.
giani.pezzu@gmail.com


gaponenko:
On Wed, 18 Dec 2019, Rob Kutschke wrote:

> Hi Giani,
> 
>   Also, if you just want the z of each plane of wires, ask for the nominal value of z for
> wires 0 and 1 of a panel on each side of each plane.  

What will happen here when we introduce wire level misalignments?

Andrei


>  I think that this is a better way to
> do it instead of back computing spacing parameters and then recomputing straw positions.
> 
> Rob
> 
>

kutschke:
HI Andrei,

   There already are two tracker objects.  The one that you get from the Geometry Service
holds nominal information.  The one that you get from the ProditionsService holds aligned
information.

   Both are objects of the same type but differing state.

Rob

gaponenko:
Thank you Rob.  I missed the word "nominal" when reading your suggestion.

Andrei

TrkFilters/src/HelixFilter_module.cc
gianipez: Hi Dave et al,
I tried to get the correct numbers from the "innerTrackerEnvelopeParams" data member of the Tracker class, but it shows the following (see attached picture).
How can I get the actaul outerRadius of the active volume?
Thanks,

Giani
```suggestion
    float    maxR_tracker(860.);
<img width="178" alt="Screen Shot 2019-12-18 at 11 30 30 PM" src="https://user-images.githubusercontent.com/25443728/71132368-8463e780-21f7-11ea-8413-8fd696b0aeec.png">

```

gianipez:
Hi Everyone,

I believe that my last commits addressed all the issues.

Please, let me know any comments.

Cheers,

Giani

kutschke:
Completed Merge and will tag as v07_07_00 (following the proposed new convention).

PR #76, opened 2019-12-19
Use aligned geometry when making TrkStrawHits

brownd1978:
Restore lost functionality.


brownd1978:
There are other strange effects in this code that need further study and probably fixing before this should be pushed back.

kutschke:
Ok the two merges and 2 tags are done.

Should we cherry pick up to the merge of Giani's code into the MDC2018 branch?

Rob

PR #83, opened 2019-12-23
Fix paths

brownd1978:


rlcee:
All looks good and tests OK, please proceed.

PR #92, opened 2020-01-03
Break compile-dependency loop between ProditionsService and TrackerCo…

brownd1978:
…nditions.  Knock-on fixes to add missing includes that were coming indirectly from these


kutschke:
This looks straightforward.  I will merge.

PR #96, opened 2020-01-09
Geometry Updates

michaelmackenzie:
Updates to the Mu2e Geant geometry for the CRY simulation campaign, starting from the 2019 MARS campaign updates.
Includes updates to the overburden and Mu2e building, as well as an implementation of generic trapezoids in the overburden construction.

Validated using output from Validation/fcl/cosmicSimReco.fcl, Validation/fcl/potSim.fcl, and Validation/fcl/simReco.fcl passed to Validation/fcl/val.fcl on master and Geom, and compared with valCompare. Mu2eG4/fcl/g4test_03.fcl was also used for validation with valCompare.

The updated geometry can be used by switching from geom_common_current.txt to geom_CRY_2019.txt.


michaelmackenzie:
I removed the CRY component of the naming, and defined the phase-II configuration by including the phase-I file.

oksuzian:
Hi Michael,

Please check with George on the hatch blocks configuration for phase-ii.

   Thank you, Yuri.

PR #104, opened 2020-01-14
Art3 mt

goodenou:
I have made the changes to the fcl files to allow the MT POT and g3test_03 configurations to run using the Mu2eG4MT module.


kutschke:
Everything looks good.

PR #110, opened 2020-01-16
CORSIKA time fix and fcl file for production

soleti:
I added the CORSIKA GenID to the GenerateCosmicTimes module and I updated the FCL files needed for production.

### DIFF
--- a/Sources/inc/CosmicCORSIKA.hh
+++ b/Sources/inc/CosmicCORSIKA.hh
@@ -229,9 +229,9 @@ namespace mu2e {
 
       GlobalConstantsHandle<ParticleDataTable> pdt;
 
-      const float _GeV2MeV = CLHEP::GeV / CLHEP::MeV;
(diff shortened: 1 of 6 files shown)


Sources/inc/CosmicCORSIKA.hh
```diff
@@ -231,6 +231,7 @@ namespace mu2e {
 
       const float _GeV2MeV = CLHEP::GeV / CLHEP::MeV;
       const float _cm2mm = CLHEP::cm / CLHEP::mm;
+      const float _ns2s = CLHEP::ns / CLHEP::s;
```
kutschke: Please make these:

 static constexpr float

I just checked that the CLHEP definitions are constexpr so this will work. 

In this case it's not very important ( I bet you only ever make on CosmicCORSIKA object per job) - so you are not really wasting memory.   I am just getting it on the radar for when it is important.

soleti: FIxed, thanks.

PR #114, opened 2020-01-17
Mixing updates

gaponenko:
Mixing needs to handle EventIDSequence to support bookkeeping in multi-stage resampling.


kutschke:
HI Andrei,

  Does this need to go in before tonights's validation or can it wait for tomorrow.

Rob

gaponenko:
This PR is independent of anything else as far as I can tell.  It can go 
now or later.  It does not affect validation.

Andrei

kutschke:
Ok.  I will wait till tomorrow.  There is so much going in tonight and there is already one known change in validation.
I feel we are pushing our luck a little.

PR #120, opened 2020-01-23
Put back unused schema till readback problem is understood

brownd1978:


kutschke:
I will pull this - I hope we can take it out again soon.

PR #127, opened 2020-01-30
Art3 mt

goodenou:
I have made much of the output from the G4MT modules optional with a new fcl parameter 'mtDebugOutput' which is set to false by default.  I have added in the ability to modify the number of random number seeds at runtime using a new fcl parameter 'maxEventsToSeed'.  This determines the maximum number of events that can be processed in a given run.  By default this value is set to 10k.  I have added and modified some fcl files for running in MT mode.


gaponenko:
Hi Lisa,

> I have added in the ability to modify the number of random number seeds at runtime using a new fcl parameter 'maxEventsToSeed'. This determines the maximum number of events that can be processed in a given run. By default this value is set to 10k.

Can you please explain the seed issue to me?   In the non-MT simualtions we can use one seed to produce a practically unlimited number of events.   Why is the number of events per run limited in MT?  What does it have to do with multiple seeds?

goodenou:
Hi Andrei.  You have asked a good question.  From a Geant4 perspective, the answer is that in order to have reproducibility of our results, regardless of the number of threads that we run, we need to have a new seed for each event.  Alternatively, if we want to relax the condition to be reproducibility only in the case that we re-run with the same number of threads, we can have a single seed for each thread.  However, TBB can create as many threads as it wants to do a job in which you have required N threads.  TBB guarantees only that the total number of ACTIVE threads is N.  (I have seen this in action, it is currently causing me an issue with the shutdown of my code.)  So, in this case, the marriage of G4 and TBB forces us to have a seed for each event in order to have reproducibility.  It is not an elegant situation, but it is a practical one.

kutschke:
Hi Lisa,

 I would like to understand this too:

> On Jan 30, 2020, at 3:49 PM, Lisa Goodenough <notifications@github.com> wrote:
> 
> Hi Andrei. You have asked a good question. From a Geant4 perspective, the answer is that in order to have reproducibility of our results, regardless of the number of threads that we run, we need to have a new seed for each event. 

I believe that this comment only applies to MT mode when you are running with more than 1 G4 worker thread.  Is that right?

I also think that the issue is that instances of the random number engine are tied to G4 worker threads.  If I rerun the same job many times, the matching of events to G4 worker threads will have at least some, and maybe a lot, of randomness.  If there were a deterministic matching of events to threads then this question would be moot.

Is that right?


> Alternatively, if we want to relax the condition to be reproducibility only in the case that we re-run with the same number of threads, we can have a single seed for each thread. However, TBB can create as many threads as it wants to do a job in which you have required N threads. TBB guarantees only that the total number of ACTIVE threads is N. (I have seen this in action, it is currently causing me an issue with the shutdown of my code.) So, in this case, the marriage of G4 and TBB forces us to have a seed for each event in order to have reproducibility. It is not an elegant situation, but it is a practical one.

This one is over my head.  Are you saying that if I ask for N G4 worker threads I may get more than N?  Or are you saying that there may be more than N threads in the job, of which only N will be G4 worker threads.  Which threads are active at a given time is a separate question.

Rob

goodenou:
> On Jan 30, 2020, at 4:21 PM, Rob Kutschke <notifications@github.com> wrote:
> 
> Hi Lisa, 
> 
> I would like to understand this too: 
> 
> > On Jan 30, 2020, at 3:49 PM, Lisa Goodenough <notifications@github.com> wrote: 
> > 
> > Hi Andrei. You have asked a good question. From a Geant4 perspective, the answer is that in order to have reproducibility of our results, regardless of the number of threads that we run, we need to have a new seed for each event. 
> 
> I believe that this comment only applies to MT mode when you are running with more than 1 G4 worker thread. Is that right? 

Yes, that is correct.


> 
> I also think that the issue is that instances of the random number engine are tied to G4 worker threads. If I rerun the same job many times, the matching of events to G4 worker threads will have at least some, and maybe a lot, of randomness. If there were a deterministic matching of events to threads then this question would be moot. 
> 
> Is that right? 

Yes, that is also correct.

> 
> 
> > Alternatively, if we want to relax the condition to be reproducibility only in the case that we re-run with the same number of threads, we can have a single seed for each thread. However, TBB can create as many threads as it wants to do a job in which you have required N threads. TBB guarantees only that the total number of ACTIVE threads is N. (I have seen this in action, it is currently causing me an issue with the shutdown of my code.) So, in this case, the marriage of G4 and TBB forces us to have a seed for each event in order to have reproducibility. It is not an elegant situation, but it is a practical one. 
> 
> This one is over my head. Are you saying that if I ask for N G4 worker threads I may get more than N? Or are you saying that there may be more than N threads in the job, of which only N will be G4 worker threads. Which threads are active at a given time is a separate question. 
> 
If I ask for N threads (= number of schedules to keep things simple) in my job, I will get N G4Workers active at any one time.  However, TBB feels free to abandon a particular thread and start up a new one at any time in the job at the beginning of any event.  When this happens, the old thread still contains all of it’s G4 objects and ptrs , and the new thread gets a whole new set of G4 objects and ptrs.  When I try to shut down the threads at the end of the job, I cannot access the ones that got abandoned.

gaponenko:
Hi Lisa, 

> From a Geant4 perspective, the answer is that in order to have
> reproducibility of our results, regardless of the number of threads
> that we run, we need to have a new seed for each event.

Thank you for the explanation.  I was the one mourning the loss of
reproducibility with MT, which the individual seeds resolve.

Many older pseudo-random number generators assure "good" statistical
properties of their output from a long run with a single seed, but say
nothing about correlations between multiple chunks started from
different seeds.  I think when we were switching to MIXMAX a statement
was made that the new generator "officially" has good properties
across multiple sequences started with different seeds.  If anyone
knows more about that I would like to hear.  That would "certify" that
RNG use use in the seed-per-event mode.

But then we should be a lot more careful about setting initial seeds.
At the moment generate_fcl makes sure that each generated job gets a
unique SINGLE seed.  If MT jobs use 10k or more seeds, we'll need to
adjust that.  Or we'll have different threads in different jobs
simulating the same event.

Andrei

resnegfk:
Andrei,
From: https://mixmax.hepforge.org/
It is guaranteed that even if the user seeds from sequential seeds, e.g. 1,2,3... 
the initial vectors are independent and are located on the MIXMAX trajectory 
at least 10^100 steps apart. 
Krzysztof

kutschke:
Thanks Lisa,

  A few more things below.

> On Jan 30, 2020, at 4:30 PM, Lisa Goodenough <notifications@github.com> wrote:
> 
> 
> 
> > On Jan 30, 2020, at 4:21 PM, Rob Kutschke <notifications@github.com> wrote:
> > 
> > Hi Lisa, 
> > 
> > I would like to understand this too: 
> > 
> > > On Jan 30, 2020, at 3:49 PM, Lisa Goodenough <notifications@github.com> wrote: 
> > > 
> > > Hi Andrei. You have asked a good question. From a Geant4 perspective, the answer is that in order to have reproducibility of our results, regardless of the number of threads that we run, we need to have a new seed for each event. 
> > 
> > I believe that this comment only applies to MT mode when you are running with more than 1 G4 worker thread. Is that right? 
> 
> Yes, that is correct.
> 
> 
> > 
> > I also think that the issue is that instances of the random number engine are tied to G4 worker threads. If I rerun the same job many times, the matching of events to G4 worker threads will have at least some, and maybe a lot, of randomness. If there were a deterministic matching of events to threads then this question would be moot. 
> > 
> > Is that right? 
> 
> Yes, that is also correct.
> 
> > 
> > 
> > > Alternatively, if we want to relax the condition to be reproducibility only in the case that we re-run with the same number of threads, we can have a single seed for each thread.

I thought that because of race conditions, even if you rerun with the same number of threads reproducibility is not guaranteed.

Is that right or wrong?


> However, TBB can create as many threads as it wants to do a job in which you have required N threads. TBB guarantees only that the total number of ACTIVE threads is N. (I have seen this in action, it is currently causing me an issue with the shutdown of my code.) So, in this case, the marriage of G4 and TBB forces us to have a seed for each event in order to have reproducibility. It is not an elegant situation, but it is a practical one. 
> > 
> > This one is over my head. Are you saying that if I ask for N G4 worker threads I may get more than N? Or are you saying that there may be more than N threads in the job, of which only N will be G4 worker threads. Which threads are active at a given time is a separate question. 
> > 
> If I ask for N threads (= number of schedules to keep things simple) in my job, I will get N G4Workers active at any one time. However, TBB feels free to abandon a particular thread and start up a new one at any time in the job at the beginning of any event. When this happens, the old thread still contains all of it’s G4 objects and ptrs , and the new thread gets a whole new set of G4 objects and ptrs. When I try to shut down the threads at the end of the job, I cannot access the ones that got abandoned.

Wow.   So if you ask for N threads/schedules then you might get more than N of the per-thread objects created?  If you instrument the c'tors can you see them being created?

Rob

goodenou:
Hi Rob,
I have answered your specific questions inline below.

I should add that in the previous version of our MT code, pre-artMT, we were also seeding each event individually, but it was happening ‘behind the scenes’ in G4MTRunManager.  We were providing a single seed to the MTRunManager through our config file, and G4 was creating the #events seeds from that.  The process is the same in the current version, we provide a single starter seed in the config file, and Mu2eG4MTRunManager now creates the #events seeds, one for each event.

The reason we need to explicitly insert a value into the current code is because we now ‘own’ the “InitializeEvent” code in Mu2eG4MTRunManager - previously this code was buried in G4.  The InitializeEvent code needs to know how many events it is producing in order to make the seeds.  Previously this number was given to it in our call to BeamOn in produce(), but now we have to pass the number into the code another way.  Unless there is a way to pass in the result of “-n “ on the command line, we have to do it the way I have set up, which is to hard code a default value (10k, which is good for testing things) and pass in a larger value through the config file for real jobs.


> On Jan 30, 2020, at 9:46 PM, Rob Kutschke <notifications@github.com> wrote:
> 
> Thanks Lisa, 
> 
> A few more things below. 
> 
> > On Jan 30, 2020, at 4:30 PM, Lisa Goodenough <notifications@github.com> wrote: 
> > 
> > 
> > 
> > > On Jan 30, 2020, at 4:21 PM, Rob Kutschke <notifications@github.com> wrote: 
> > > 
> > > Hi Lisa, 
> > > 
> > > I would like to understand this too: 
> > > 
> > > > On Jan 30, 2020, at 3:49 PM, Lisa Goodenough <notifications@github.com> wrote: 
> > > > 
> > > > Hi Andrei. You have asked a good question. From a Geant4 perspective, the answer is that in order to have reproducibility of our results, regardless of the number of threads that we run, we need to have a new seed for each event. 
> > > 
> > > I believe that this comment only applies to MT mode when you are running with more than 1 G4 worker thread. Is that right? 
> > 
> > Yes, that is correct. 
> > 
> > 
> > > 
> > > I also think that the issue is that instances of the random number engine are tied to G4 worker threads. If I rerun the same job many times, the matching of events to G4 worker threads will have at least some, and maybe a lot, of randomness. If there were a deterministic matching of events to threads then this question would be moot. 
> > > 
> > > Is that right? 
> > 
> > Yes, that is also correct. 
> > 
> > > 
> > > 
> > > > Alternatively, if we want to relax the condition to be reproducibility only in the case that we re-run with the same number of threads, we can have a single seed for each thread. 
> 
> I thought that because of race conditions, even if you rerun with the same number of threads reproducibility is not guaranteed. 
> 
> Is that right or wrong? 

The seeds are associated with an event number, so reproducibility is guaranteed.


> 
> 
> > However, TBB can create as many threads as it wants to do a job in which you have required N threads. TBB guarantees only that the total number of ACTIVE threads is N. (I have seen this in action, it is currently causing me an issue with the shutdown of my code.) So, in this case, the marriage of G4 and TBB forces us to have a seed for each event in order to have reproducibility. It is not an elegant situation, but it is a practical one. 
> > > 
> > > This one is over my head. Are you saying that if I ask for N G4 worker threads I may get more than N? Or are you saying that there may be more than N threads in the job, of which only N will be G4 worker threads. Which threads are active at a given time is a separate question. 
> > > 
> > If I ask for N threads (= number of schedules to keep things simple) in my job, I will get N G4Workers active at any one time. However, TBB feels free to abandon a particular thread and start up a new one at any time in the job at the beginning of any event. When this happens, the old thread still contains all of it’s G4 objects and ptrs , and the new thread gets a whole new set of G4 objects and ptrs. When I try to shut down the threads at the end of the job, I cannot access the ones that got abandoned. 
> 
> Wow. So if you ask for N threads/schedules then you might get more than N of the per-thread objects created? If you instrument the c'tors can you see them being created? 
> 

Yes, I see the extra objects being created.

As an example, I ran this command asking for 8 threads/schedules:

	mu2e -c JobConfig/cosmic/dsstops_resampler_lo_forMT.fcl -n 5000 -j 8

I see from the output 8 worker run managers being created right away:

FOR TID: 139661241833216, NO WORKER.  We are making one.
FOR SchedID: 4, TID=139661241833216, workerRunManagers[schedID].get() is:0x44843880

FOR TID: 139661252323072, NO WORKER.  We are making one.
FOR SchedID: 0, TID=139661252323072, workerRunManagers[schedID].get() is:0x44858890

FOR TID: 139661283792640, NO WORKER.  We are making one.
FOR SchedID: 3, TID=139661283792640, workerRunManagers[schedID].get() is:0x44843d80

FOR TID: 139661273302784, NO WORKER.  We are making one.
FOR SchedID: 7, TID=139661273302784, workerRunManagers[schedID].get() is:0x448432a0

FOR TID: 139662353160960, NO WORKER.  We are making one.
FOR SchedID: 1, TID=139662353160960, workerRunManagers[schedID].get() is:0x448a5d70

FOR TID: 139661294282496, NO WORKER.  We are making one.
FOR SchedID: 5, TID=139661294282496, workerRunManagers[schedID].get() is:0x4486cb50

FOR TID: 139661304772352, NO WORKER.  We are making one.
FOR SchedID: 2, TID=139661304772352, workerRunManagers[schedID].get() is:0x44891950

FOR TID: 139661262812928, NO WORKER.  We are making one.
FOR SchedID: 6, TID=139661262812928, workerRunManagers[schedID].get() is:0x448833a0

Then much later in the job I see another worker being created:

FOR TID: 139661206976256, NO WORKER.  We are making one.
FOR SchedID: 1, TID=139661206976256, workerRunManagers[schedID].get() is:0x4e09e3f0

And at the end of the job it is confirmed that I have 9 workers in my map:

At endRun, we have 9 members in the map

kutschke:
Thanks Lisa,

  Do you have a few minutes to talk about this in person - it might converge faster.

  Andrei, Krzysztof, Lisa: can you make a meeting sometime this afternoon after 2:00 ?

  Some more followups in line.

> On Jan 31, 2020, at 9:02 AM, Lisa Goodenough <notifications@github.com> wrote:
> 
> Hi Rob,
> I have answered your specific questions inline below.
> 
> I should add that in the previous version of our MT code, pre-artMT, we were also seeding each event individually, but it was happening ‘behind the scenes’ in G4MTRunManager. We were providing a single seed to the MTRunManager through our config file, and G4 was creating the #events seeds from that.

Just curious: given the one initial seed, what did G4MT do?  Add the G4 event number?  Just increment ... ? Something else?


>  The process is the same in the current version, we provide a single starter seed in the config file, and Mu2eG4MTRunManager now creates the #events seeds, one for each event.


> 
> The reason we need to explicitly insert a value into the current code is because we now ‘own’ the “InitializeEvent” code in Mu2eG4MTRunManager - previously this code was buried in G4. 

I don't quite follow the last two sentences.  The first one says that the code is the same as before the second says that it does something different?  So I did not understand something.


> The InitializeEvent code needs to know how many events it is producing in order to make the seeds.

Is this done in a part of InitializeEvent that is called once at the start of the job?

So my cartoon picture is that once, at the start of the job, you fill a container with a set of seeds.  You give the container to G4. When G4 runs it reaches into the container and uses a seed in a way that's repeatable.  Is that right?


>  Previously this number was given to it in our call to BeamOn in produce(), but now we have to pass the number into the code another way. 

Does this mean that, in the old way, BeamOn produced the big list of seeds?  We just had to give it the number of events we
expected to process?

> Unless there is a way to pass in the result of “-n “ on the command line, we have to do it the way I have set up, which is to hard code a default value (10k, which is good for testing things) and pass in a larger value through the config file for real jobs.

Does G4 support  alternate method by which it calls a function to get the next seed ?

Rob 



> 
>

goodenou:
I can meet at 2 or 2:30.

kutschke:
Hi Lisa,

  This is following up on just the end of job stuff explosion stuff:
 
> > > However, TBB can create as many threads as it wants to do a job in which you have required N threads. TBB guarantees only that the total number of ACTIVE threads is N. (I have seen this in action, it is currently causing me an issue with the shutdown of my code.) So, in this case, the marriage of G4 and TBB forces us to have a seed for each event in order to have reproducibility. It is not an elegant situation, but it is a practical one. 
> > > > 
> > > > This one is over my head. Are you saying that if I ask for N G4 worker threads I may get more than N? Or are you saying that there may be more than N threads in the job, of which only N will be G4 worker threads. Which threads are active at a given time is a separate question. 
> > > > 
> > > If I ask for N threads (= number of schedules to keep things simple) in my job, I will get N G4Workers active at any one time. However, TBB feels free to abandon a particular thread and start up a new one at any time in the job at the beginning of any event. When this happens, the old thread still contains all of it’s G4 objects and ptrs , and the new thread gets a whole new set of G4 objects and ptrs. When I try to shut down the threads at the end of the job, I cannot access the ones that got abandoned. 
> > 
> > Wow. So if you ask for N threads/schedules then you might get more than N of the per-thread objects created? If you instrument the c'tors can you see them being created? 
> > 
> 
> Yes, I see the extra objects being created.
> 
> As an example, I ran this command asking for 8 threads/schedules:
> 
> mu2e -c JobConfig/cosmic/dsstops_resampler_lo_forMT.fcl -n 5000 -j 8
> 
> I see from the output 8 worker run managers being created right away:
> 
> FOR TID: 139661241833216, NO WORKER. We are making one.
> FOR SchedID: 4, TID=139661241833216, workerRunManagers[schedID].get() is:0x44843880
> 
> FOR TID: 139661252323072, NO WORKER. We are making one.
> FOR SchedID: 0, TID=139661252323072, workerRunManagers[schedID].get() is:0x44858890
> 
> FOR TID: 139661283792640, NO WORKER. We are making one.
> FOR SchedID: 3, TID=139661283792640, workerRunManagers[schedID].get() is:0x44843d80
> 
> FOR TID: 139661273302784, NO WORKER. We are making one.
> FOR SchedID: 7, TID=139661273302784, workerRunManagers[schedID].get() is:0x448432a0
> 
> FOR TID: 139662353160960, NO WORKER. We are making one.
> FOR SchedID: 1, TID=139662353160960, workerRunManagers[schedID].get() is:0x448a5d70
> 
> FOR TID: 139661294282496, NO WORKER. We are making one.
> FOR SchedID: 5, TID=139661294282496, workerRunManagers[schedID].get() is:0x4486cb50
> 
> FOR TID: 139661304772352, NO WORKER. We are making one.
> FOR SchedID: 2, TID=139661304772352, workerRunManagers[schedID].get() is:0x44891950
> 
> FOR TID: 139661262812928, NO WORKER. We are making one.
> FOR SchedID: 6, TID=139661262812928, workerRunManagers[schedID].get() is:0x448833a0
> 
> Then much later in the job I see another worker being created:
> 
> FOR TID: 139661206976256, NO WORKER. We are making one.
> FOR SchedID: 1, TID=139661206976256, workerRunManagers[schedID].get() is:0x4e09e3f0
> 
> And at the end of the job it is confirmed that I have 9 workers in my map:
> 
> At endRun, we have 9 members in the map

Let me see if I have the right picture.

We hold at "module scope" a container that holds all of the per thread stuff.   The container is indexed by thread Id and each worker thread get's its correct per thread stuff from this container.  The stuff in the container is all on the heap so, at end of job we need to delete all of the stuff in the container.

I am being a little vague about "module scope" - maybe it's really owned by the master thread, not by the module, or maybe it's held by an object that is a data member of one of them ...  The bottom line is that it holds stuff that is created during the job and that we need to delete at end of job.

In the example you give above there are 9 members in the map.  8 of them are things you need to delete and one of them is a time bomb that if you delete it then it segfaults?

Is that close enough?

Rob

kutschke:
If 2:30 works for the others, then let's do that.  I will see if Kyle is available too.

Rob

goodenou:
> On Jan 31, 2020, at 11:45 AM, Rob Kutschke <notifications@github.com> wrote:
> 
> Hi Lisa, 
> 
> This is following up on just the end of job stuff explosion stuff: 
> 
> > > > However, TBB can create as many threads as it wants to do a job in which you have required N threads. TBB guarantees only that the total number of ACTIVE threads is N. (I have seen this in action, it is currently causing me an issue with the shutdown of my code.) So, in this case, the marriage of G4 and TBB forces us to have a seed for each event in order to have reproducibility. It is not an elegant situation, but it is a practical one. 
> > > > > 
> > > > > This one is over my head. Are you saying that if I ask for N G4 worker threads I may get more than N? Or are you saying that there may be more than N threads in the job, of which only N will be G4 worker threads. Which threads are active at a given time is a separate question. 
> > > > > 
> > > > If I ask for N threads (= number of schedules to keep things simple) in my job, I will get N G4Workers active at any one time. However, TBB feels free to abandon a particular thread and start up a new one at any time in the job at the beginning of any event. When this happens, the old thread still contains all of it’s G4 objects and ptrs , and the new thread gets a whole new set of G4 objects and ptrs. When I try to shut down the threads at the end of the job, I cannot access the ones that got abandoned. 
> > > 
> > > Wow. So if you ask for N threads/schedules then you might get more than N of the per-thread objects created? If you instrument the c'tors can you see them being created? 
> > > 
> > 
> > Yes, I see the extra objects being created. 
> > 
> > As an example, I ran this command asking for 8 threads/schedules: 
> > 
> > mu2e -c JobConfig/cosmic/dsstops_resampler_lo_forMT.fcl -n 5000 -j 8 
> > 
> > I see from the output 8 worker run managers being created right away: 
> > 
> > FOR TID: 139661241833216, NO WORKER. We are making one. 
> > FOR SchedID: 4, TID=139661241833216, workerRunManagers[schedID].get() is:0x44843880 
> > 
> > FOR TID: 139661252323072, NO WORKER. We are making one. 
> > FOR SchedID: 0, TID=139661252323072, workerRunManagers[schedID].get() is:0x44858890 
> > 
> > FOR TID: 139661283792640, NO WORKER. We are making one. 
> > FOR SchedID: 3, TID=139661283792640, workerRunManagers[schedID].get() is:0x44843d80 
> > 
> > FOR TID: 139661273302784, NO WORKER. We are making one. 
> > FOR SchedID: 7, TID=139661273302784, workerRunManagers[schedID].get() is:0x448432a0 
> > 
> > FOR TID: 139662353160960, NO WORKER. We are making one. 
> > FOR SchedID: 1, TID=139662353160960, workerRunManagers[schedID].get() is:0x448a5d70 
> > 
> > FOR TID: 139661294282496, NO WORKER. We are making one. 
> > FOR SchedID: 5, TID=139661294282496, workerRunManagers[schedID].get() is:0x4486cb50 
> > 
> > FOR TID: 139661304772352, NO WORKER. We are making one. 
> > FOR SchedID: 2, TID=139661304772352, workerRunManagers[schedID].get() is:0x44891950 
> > 
> > FOR TID: 139661262812928, NO WORKER. We are making one. 
> > FOR SchedID: 6, TID=139661262812928, workerRunManagers[schedID].get() is:0x448833a0 
> > 
> > Then much later in the job I see another worker being created: 
> > 
> > FOR TID: 139661206976256, NO WORKER. We are making one. 
> > FOR SchedID: 1, TID=139661206976256, workerRunManagers[schedID].get() is:0x4e09e3f0 
> > 
> > And at the end of the job it is confirmed that I have 9 workers in my map: 
> > 
> > At endRun, we have 9 members in the map 
> 
> Let me see if I have the right picture. 
> 
> We hold at "module scope" a container that holds all of the per thread stuff. The container is indexed by thread Id and each worker thread get's its correct per thread stuff from this container. The stuff in the container is all on the heap so, at end of job we need to delete all of the stuff in the container. 
> 
> I am being a little vague about "module scope" - maybe it's really owned by the master thread, not by the module, or maybe it's held by an object that is a data member of one of them ... The bottom line is that it holds stuff that is created during the job and that we need to delete at end of job. 
> 
> In the example you give above there are 9 members in the map. 8 of them are things you need to delete and one of them is a time bomb that if you delete it then it segfaults? 
> 
> Is that close enough? 

The only correction I would make is this:
the 9th thread that has gone ‘offline’ and is no longer active cannot be found by its ID when it is time to delete it.  So as it stands my code just waits, it doesn’t seg fault.

goodenou:
> On Jan 31, 2020, at 11:27 AM, Rob Kutschke <notifications@github.com> wrote:
> 
> Thanks Lisa, 
> 
> Do you have a few minutes to talk about this in person - it might converge faster. 
> 
> Andrei, Krzysztof, Lisa: can you make a meeting sometime this afternoon after 2:00 ? 
> 
> Some more followups in line. 
> 
> > On Jan 31, 2020, at 9:02 AM, Lisa Goodenough <notifications@github.com> wrote: 
> > 
> > Hi Rob, 
> > I have answered your specific questions inline below. 
> > 
> > I should add that in the previous version of our MT code, pre-artMT, we were also seeding each event individually, but it was happening ‘behind the scenes’ in G4MTRunManager. We were providing a single seed to the MTRunManager through our config file, and G4 was creating the #events seeds from that. 
> 
> Just curious: given the one initial seed, what did G4MT do? Add the G4 event number? Just increment ... ? Something else? 

I have not looked into the specifics of how it creates the seeds.

> 
> 
> > The process is the same in the current version, we provide a single starter seed in the config file, and Mu2eG4MTRunManager now creates the #events seeds, one for each event. 
> 
> 
> > 
> > The reason we need to explicitly insert a value into the current code is because we now ‘own’ the “InitializeEvent” code in Mu2eG4MTRunManager - previously this code was buried in G4. 
> 
> I don't quite follow the last two sentences. The first one says that the code is the same as before the second says that it does something different? So I did not understand something. 

The code that makes the seeds IS the same, but how we interact with it is different.

Previously the code was hidden in a part of G4 and was called by one of the functions that BeamOnBeginRun called in beginRun. 

Specifically, the way we determined the number of seeds to set in the job was through the assignment in the MT module:
G4int numberOfEventsToBeProcessed = std::numeric_limits<int>::max();

numberOfEventsToBeProcessed was then used in the call
_runManager->InitializeEventLoop(numberOfEventsToBeProcessed,macroFile,n_select);
which was a G4 function that set the number of seeds among other things.


The functionality of InitializeEventLoop is now in our Mu2eG4MTRunManager::initializeG4(int art_runnumber)which is where the statement that got this whole discussion going is:
int numevents = maxNumEventstoSeed_;
SetNumberOfEventsToBeProcessed(numevents);

So the conclusion is that we are now creating fewer seeds that we were before.

> 
> 
> > The InitializeEvent code needs to know how many events it is producing in order to make the seeds. 
> 
> Is this done in a part of InitializeEvent that is called once at the start of the job?
> 
> So my cartoon picture is that once, at the start of the job, you fill a container with a set of seeds. You give the container to G4. When G4 runs it reaches into the container and uses a seed in a way that's repeatable. Is that right? 
> 

Yes.  The container is a member of G4MTRunManager, which our Mu2eG4RunManager inherits from.


> 
> > Previously this number was given to it in our call to BeamOn in produce(), but now we have to pass the number into the code another way. 
> 
> Does this mean that, in the old way, BeamOn produced the big list of seeds? We just had to give it the number of events we 
> expected to process? 
> 

I just looked at the specifics of the old code.  It turns out that we did not give it the number of events we expected to process, we set the number of seeds in this way:
G4int numberOfEventsToBeProcessed = std::numeric_limits<int>::max();

This was an assignment taken from the original non-MT module.

> > Unless there is a way to pass in the result of “-n “ on the command line, we have to do it the way I have set up, which is to hard code a default value (10k, which is good for testing things) and pass in a larger value through the config file for real jobs. 
> 
> Does G4 support alternate method by which it calls a function to get the next seed ? 
> 
No.



> Rob 
> 
> 
> 
> > 
> >

kutschke:
Thanks Lisa,

   I will make my next suggestion on the thread that Kyle started.

Rob

gaponenko:
Don't we have "Updates from the Scientific Advisory Council" from 14:00 to 15:30 today?
It is on my calendar.

Andrei

kutschke:
Yes.  

Lisa:  does later in the day work for you?

Andrei: how much do you want to go to that meeting?

Rob

goodenou:
Yes, I can do later in the day.  But we will run into the wine and cheese if we wait until 3:30.

kutschke:
Let's stick with 2:30.  Kyle can make it. 

Andrei: let me know if you plan to come.

I also ran into Krzysztof who says that Soon has some experience managing random seeds for GeantV, which may be relevant.

Rob

kutschke:
... Lisa  - can you find a room?

kutschke:
I also approve this pull request.  I will not merge it today so that tonight's validation will look at the Hayman2 merge in isolation.  There are a few other outstanding pull requests and we will decide on an order tomorrow.

I agree with Andrei that the maxEventsToSeed  issue can be resolved in a separate commit.

In a future commit I would like you to address the following issue.  There are many .fcl files for which you have an original file plus a matching MT file.  Please rewrite the MT file to #include the original file and make the necessary redefinitions.   Feel free to ask questions about this last part.

PR #135, opened 2020-02-05
Calo digi

sophiemiddleton:
Changes discussed in the meeting. Further development will be needed in future but I think this is now at a point at which it is useful. There's no huge rush to get it in ASAP so other changes can be merged in ahead if they are more important.


Mu2eG4/test/conditions_01.txt
```diff
@@ -37,7 +37,10 @@ double BirkCorrHadron = 4.0;
 double ROphotostatistic = 30.0;//p.e / MeV
 
 //conversion factor from ADC counts to MeV
-double ADC2MeVConversionFactor = 1.6e-03; //5.46e-04;//0.989e-03;
+double ADC2MeVConversionFactor =  0.0461333;
```
gianipez: This value is wrong, It should be put back to 1.6e-03.

CaloCluster/fcl/prolog.fcl
```diff
@@ -13,6 +13,12 @@ END_PROLOG
 
 BEGIN_PROLOG
 
+CaloClusterCompare : {
```
gianipez: Probably the best location for this analyzer is in the Analysis directory. CaloCluster should only keep modules and algorithms related to the reconstruction.

CaloCluster/fcl/prolog_trigger.fcl
```diff
@@ -41,6 +44,41 @@ CaloClusterFast :
...
+  EminSeed               : 20 
+  ExpandCut              : 1 
+  EnoiseCut              : 1
+  timeCut                : @local::HitMakerBlindTime   
+  deltaTime              : 5.01
+  diagLevel              : 0
+}  
+
+
+NewCaloClusterFast :
```
gianipez: I would move the modules instances within this block in the already present CaloClusterTrigger block.
Also, the reco sequence that uses NewCaloClusterFast should include the sequence from CaloReco/fcl/prolog_trigger.fcl that performs the hit reco (starting from the digi collection.

CaloCluster/src/CaloClusterFast_module.cc
```diff
@@ -229,7 +230,7 @@ namespace mu2e {
          double     hitTime = (seed->index_+offsetT0_)*digiSampling_-timeCorrection_;
          if (includeCrystalHits_) {
            recoCrystalHits.push_back(CaloCrystalHit(seed->crId_, 2, hitTime, 0, seed->val_*adcToEnergy_, 0., caloRecoDigi));
-         }
+	  }
```
gianipez: wrong indentation. please, fix it

CaloCluster/src/CaloProtoClusterFromCrystalHit_module.cc
```diff
@@ -60,79 +61,183 @@ namespace mu2e {
...
+	typedef std::list<const CaloCrystalHit*>    CaloCrystalList;
+
+
+	explicit CaloProtoClusterFromCrystalHit(fhicl::ParameterSet const& pset) :
+	art::EDProducer{pset},
+	caloCrystalToken_{consumes<CaloCrystalHitCollection>(pset.get<std::string>("caloCrystalModuleLabel"))},
+	producerNameMain_(pset.get<std::string>("mainClusterCollName")),
+	producerNameSplit_(pset.get<std::string>("splitClusterCollName")),
+	isOnline_(pset.get<bool>("isOnline",false)),
+	minEnergy_(pset.get<double>("minEnergy",0)),
```
gianipez: this parameter refers to the minimum cluster energy. I would consider calling it in a more meaningful way, like minClusterEnergy_

CaloCluster/src/ClusterFinder.cc
```diff
@@ -20,60 +20,67 @@
...
+
+		while (!crystalToVisit_.empty())
+		{            
+			 int visitId         = crystalToVisit_.front();
+			 isVisited_[visitId] = 1;
+			 
+			 std::vector<int>  neighborsId = cal_->crystal(visitId).neighbors();
+			 if(isOnline_){
+			 	for (const auto& nneighbor : cal_->nextNeighbors(visitId)) neighborsId.push_back(nneighbor);
+			}
```
gianipez: bad indentation.

sophiemiddleton:
sure, thanks Giani - will try to get those done tomorrow. I'll wait a bit until others respond with corrections/comments and then push them all together, as otherwise, things get confusing (I found that out last time)

sophiemiddleton:
Hi Bertrand,

I did change things originally but after speaking with Giani we decided that the existing code pretty much did what I had done so the new code was not necessary. This helped avoid duplicating code. The module I used to have was basically just doing the cluster finder without using that class and had a fault that we could not trace. The logic was identical. Therefore using the existing cluster finder class seemed the best option.

The FastClustering (formerly CaloClusterFast, now NewCaloClusterFast) module has changed from the old - but this is mostly due to improvements in how we get the various fcl parameters and calibration constants. Also before this was just one module, now it is a sequence.

The performance of Offline does not change.

We can discuss today.

Cheers,
Sophie

CaloCluster/fcl/prolog_trigger.fcl
```diff
@@ -40,16 +43,34 @@ CaloClusterFast :
...
 {
    producers : {
        CaloClusterFast  : @local::CaloClusterFast
        CaloTrigger      : @local::CaloTrigger
+       NewCaloClusterFast  : @local::NewCaloClusterFast
    }
 
    Reco : [ CaloClusterFast ]
+   NewReco : [ NewCaloClusterFast ]
+   CrystalReco : [FastCaloRecoDigiFromDigi, FastCaloCrystalHitFromHit]
```
gianipez: This line is duplicated from CaloReco/fcl/prolog_trigger.fcl
I think you should define the NewReco sequence like: NewReco : [ @sequence::OnlineCaloReco.Reco, NewCaloClusterFast ]

kutschke:
Is it important for both Reco and NewReco to be defined?

gianipez:
Sophie, I think Rob is right. I think I have been the only one using CaloClusterFast in writing other prolog_trigger files and I was explicitly calling the CaloClusterFast instance. So I guess you can name the NewReco sequence just Reco and get rid of the old Reco sequence.

bechenard:
Hi Giani,

I just talked to Sophie, and I wanted to better understand the use case 
to keep the recoDigis. I understand we want the clustering info, but not 
sure why we need to keep the underlying recoDigis. Other than that, I 
asked Sophie to split the Online clustering from the offline clustering, 
that should be straightforward to do.

Thanks,

Bertrand

kutschke:
Hi Bertrand,

    Consider running on data: if you drop the reco digis do you have enough information to do things like make an occupancy map of all SiPMs that participate in clusters matched to tracks?  The spirit is that you want to be able to make a high level selection and verify that the low level information does not have unexpected structure.

    When processing MC events, do you need to keep reco digis to have a path back to the complete MC truth?

            Rob

sophiemiddleton:
So - I've pushed the latest code with the new module as I said. The good news is that the time is now comparable to the old fast cluster module. There is the possibility of further optimization as we move forward and conclusions can't really be made without the no primary data as we said before but I think its good that it's now improved a bit.

bechenard:
HI Rob,

We can keep this information in the fast calo cluster itself, i.e. 
record the crystal ID of each hit in the cluster. The "spy nodes" might 
want to inspect the reco digis, but I would argue that we can decompress 
them in those modules. My concern is about the processing speed: 
creating and carrying around information we don't really need slows the 
process down.

I'm not sure about the MC truth, and I need to redesign this anyway, so 
this could be a new design constraint.

Cheers,

Bertrand

kutschke:
Thanks Bertrand,

  As far as I am concerned, it's OK to drop a data product when looking at data but NOT drop the data product when looking at MC.  I would like to hear what others think.

 

   Rob

PS.  I would NOT call the option something-MC.  Call it putDigis or something like that.

bechenard:
Hi Rob,

Not sure I understand the MC statement, but my concern is primarily for 
the trigger. You should feel free to do what you want for MC.

Bertrand

kutschke:
... I mean that if we need to keep the digis for the connection to MC truth, that is only relevant
for MC events, not for data.

Rob

bechenard:
Ok, make sense

Bertrand

gianipez:
Hi,

I want to make a few statement before commenting directly:
- I think it’s good to produce the list of all the reconstructed crystal hits within the event; these could be the inputs of a series of calorimeter-only filters, so it’s important to keep them
- reconstructing a quantity only in the spy job with a different module/algorithm w.r.t. the one used in the online reconstruction would not give us 100% controllable of what is done in the online reco
- the calibration constants will be given per channel, aka they will be applied to the recoDigis
- reconstructing caloCrystalHits w/o pointers to the recodigis will make impossible to analyze the performance of this reconstruction step, which could be non trivial

So, I think that we are still in developing stage and we should try to keep a reconstruction working flow as clean and close to the offline one, to make the life of the analyzers easier in the future. Then, if we will hit a brick wall with the performance, we will think a solution. Just to remind the people, we have known from a long time that the working flow of the straw hit reconstruction is “painful” performance wise, but we didn’t introduce any shortcut, nor different “dirty” data product to speed up the process. 

I propose let’s first see what we will get with this new working flow and then we will re-talk about it.


Gianantonio Pezzullo, Ph.D.
giani.pezzu@gmail.com


bechenard:
Hi Giani,

It is true that it might make sense to keep the reco Digis if we have 
several algorithms based on them, so let's keep them for now (and I 
concur it will make the development phase easier). As far as hitting a 
brick wall, your latest performance measurements shows that we are in 
the wall, so we might have to take pretty drastic measures....

Cheers,

Bertrand

CaloReco/src/FastCaloRecoDigiFromDigi_module.cc
```diff
@@ -0,0 +1,142 @@
...
+		{
+    
+			int    roId     = caloDigi.roId();
+			double t0       = caloDigi.t0();
+			double adc2MeV  = calorimeterCalibrations->Peak2MeV(roId);
+			const std::vector<int>& waveform = caloDigi.waveform();
+
+			size_t index = &caloDigi - base;
+			art::Ptr<CaloDigi> caloDigiPtr(caloDigisHandle, index);
+
```
bechenard: Do we need to fill the x,y vectors here? Could we do instead
time = t0 + (caloDigi.peakpos()+0.5)*digiSampling_+eventWindowOffset;
eDep = waveform.at(caloDigi.peakpos())*adc2MeV;

and avoid those two vectors. They don't seem to be used anywhere else.

sdifalco:
I think we could try to factorize the fast processing from the diagnostic tools: a time performant module can be validated with a subsample of data including MC truth and more detailed low level information. This could be implemented using a flag, similar to the eventmode already existing.

kutschke:
Is it possible to split the diagnostic code into a separate module or does it need access to transients?

  Does the diagnostic code look  at MC truth?

gianipez:
Hi Everyone,

I think the discussion is slightly diverging from the review of this pull request. I would rather organize a discussion at the Soft meeting if necessary. Sophie “defended” her code at two meetings and comments about the general architecture I guess should not be done in a pull-request review process.

Cheers,


Gianantonio Pezzullo, Ph.D.
giani.pezzu@gmail.com


sophiemiddleton:
The old code is still in the pull request. Perhaps that can be optimized and used for the "fast" purposes while this new code is used for Giani's purposes and other studies....(not an expert here but that's why the old code has been kept)

CaloReco/src/FastCaloRecoDigiFromDigi_module.cc
gianipez: I agree. Probably, in the t0 formula, we should already add a  "+calorimeterCalibrations->timeOffset(roId)" to remind us to add the calibrated channel time offset?

sophiemiddleton: yes, just replying to this. I will test but I think you are correct

sophiemiddleton: done

kutschke:
Bertrand, Giani and Sophie:

  There was a lot of activity on this pull request a week ago.  Has it reached an asymptote?  Is it ready to be merged?  Or is the plan to wait for additional work?

Rob

gianipez:
Hi Rob, 
All the changes I asked have been made, so from my side you can go ahead with the merge. 

Giani

bechenard:
Yes, I think we converged. Please merge the changes.

Bertrand

kutschke:
Giani and Bertrand,

  Please check approve on github.

  Sophie:  what differences should we expect to see in validation?

Rob

CaloCluster/fcl/prolog.fcl
```diff
@@ -25,11 +28,12 @@ CaloTrigger : {
     minSeedAmplitude      : 433
     minEnergy             : 50
     timeCorrection        : 20.0
-    adcToEnergy           : 0.0461333
+    adcToEnergy           : 0.0459634#0.0461333
```
kutschke: Hi Sophie - I just noticed this now, after everything is merged in.  I have a strong preference against recording old values in comments - that's what we have a code management system for.  Please fix this sort of thing as targets of opportunity when you touch the code in the future.

sophiemiddleton: ah, sorry, must have missed that, I thought I'd removed it. Will be sure to correct in future.

PR #141, opened 2020-02-11
math.h -> cmath

resnegfk:
math.h -> cmath
Please treat it as a suggestion


ryuwd:
+1

brownd1978:
What validations have you run (besides compile)?

resnegfk:
I run short samples of g4test_03, ceSimReco and potSim with identical results in all 3 cases

kutschke:
Thanks Krzysztof,

    That's good enough for me.

    Dave: is it enough for you?  Also, I had a look at the diffs and they look good to me.  

Rob

PR #145, opened 2020-02-27
G4106

resnegfk:
To restore Geant4 navigator check mode control and to control its verbosity
To use g4run when g4mtrun was used to maintain the same module label to e.g. ease comparisons, and to decrease number of points of maintenance


rlcee:
I see the process name and the file name still have MT.  I'm not sure
if the physical job processing structure should be in these labels.  I think I favor
removing them.  Lets see if any one else comments.

goodenou:
Ray, if we remove the MT fcl files, would we then have an additional fcl layer on top of the base non-MT layer if we want to run MT?

Krzysztof, the changes look fine to me.  Have you run the MT fcl files with -j 2 (or larger) to verify that the MT version runs?

rlcee:
> Ray, if we remove the MT fcl files, would we then have an additional fcl layer on top of the base non-MT layer if we want to run MT?

I would leave the MT fcl files, but change the process name and output file name in the MT fcl files to be the same and nn-MT, or better, the MT fcl file becomes one to two lines, taking the names from non-MT.

resnegfk:
Lisa, yes, I did run with -j>1. When I run with -j 1 the results were identical before and after.
In any case, since it uses different module type it had run in the MT mode.
We have to have a layer on the top of non-MT as we need to have a place to set the MT module type:
physics.producers.g4run.module_type : "Mu2eG4MT"

goodenou:
Will you please run it with -j 2?  We use the MT-complied G4 fro either case, so running with -j 1 will not test the MT case.  I am not convinced that the MT is working until it runs with multiple threads.

resnegfk:
Lisa, as I wrote, I did run with -j>1. It was 5.

goodenou:
Sorry, Krzysztof I misunderstood what you wrote.  I will approve the pull request.

Ray, I understand your point.  I have no opinion on it.  I am sure others do though.

resnegfk:
I modified 
Mu2eG4/fcl/g4test_03MT.fcl
Validation/fcl/potSimMT.fcl
accordingly

rlcee:
Would you mind also fixing up Validation/fcl/ceSimRecoMT.fcl?

PR #147, opened 2020-02-28
Compilation DB integration into SCons

ryuwd:
At the moment the compile_commands.json generator I have written has a number of drawbacks:
- very slow
- if partially or fully built, does not generate a complete list of commands, unless run after `scons -c` (scons doesn't provide a flag to get around this problem)
- slows down Jenkins code checks job by about 1-3min (have to re-generate it every time)

Thankfully there is an existing tool that can do this much quicker and more consistently. MongoDB has open-sourced some of their SCons build tools (Apache 2.0 licence). Original: https://github.com/mongodb/mongo/blob/master/site_scons/site_tools/compilation_db.py

With this Scons tool, and a few additions to SConstruct, a compile_commands.json can be generated much quicker:
```
<setup Offline as usual>

scons -Q compiledb # doesn't perform a build - writes a full compile_commands.json to Offline root folder.
```
It looks like by adding this to the environment it will generate a compile_commands.json automatically during a build anyway, not only if the `-Q compiledb` command is specified. If this behaviour is not desired I can make it toggle-able by adding a new flag toggle. i.e. `scons --gen-compiledb -Q compiledb`

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -65,6 +65,7 @@ env = Environment( CPPPATH = sch.cppPath(mu2eOpts),   # $ART_INC ...
                    LINKCOMSTR = linkcomstr,
                    SHLINKCOMSTR= linkcomstr,
 )
+env.Tool('compilation_db')
 
 # Define and register the rule for building dictionaries.
 # sources are classes.h, classes_def.xml, 
@@ -105,6 +106,10 @@ ss = sch.sconscriptList(mu2eOpts)
 
 # make sure lib, bin and tmp are there
 sch.makeSubDirs(mu2eOpts)
+
+
+compileCommands = env.CompilationDatabase('compile_commands.json')
+compileDb = env.Alias("compiledb", compileCommands)
 
 # operate on the SConscript files
 # regular python commands like os.path() are executed immediately as they are encontered, 
--- a/SConstruct
+++ b/SConstruct
@@ -109,8 +109,7 @@ ss = sch.sconscriptList(mu2eOpts)
 # make sure lib, bin and tmp are there
 sch.makeSubDirs(mu2eOpts)
 
-# Allow a compilation database (compile_commands.json) to be
-# generated if requested (scons -Q compiledb).
+# Allow a compile_commands.json to be generated if requested (scons -Q compiledb).
 compileCommands = env.CompilationDatabase('compile_commands.json')
 compileDb = env.Alias("compiledb", compileCommands)
 
--- a/SConstruct
+++ b/SConstruct
@@ -65,6 +65,8 @@ env = Environment( CPPPATH = sch.cppPath(mu2eOpts),   # $ART_INC ...
                    LINKCOMSTR = linkcomstr,
                    SHLINKCOMSTR= linkcomstr,
 )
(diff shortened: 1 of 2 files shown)


ryuwd:
I've tested clangd, clang-tidy with the output and they are working nicely. Here's an example file: [compile_commands.json](https://github.com/Mu2e/Offline/files/4269119/compile_commands.json.txt)

ryuwd:
Thanks for the review, I agree this is probably the best tool available for SCons at the moment for clang support (P.S. what's Muse? )

Best wishes, Ryun

ryuwd:
Nevermind! - I realise you meant Mu2e :)  Have a good weekend

rlcee:
> Nevermind! - I realise you meant Mu2e :) Have a good weekend

I did mean "Muse".  It is concept we're working on that would allow
users to build multiple repos with a shared environment and cross 
dependencies. See doc 30790

kutschke:
Ryun:

1) Does this thread say what the purpose of this addition is?  What functionality does this buy us?  I don't see it in the conversation but maybe I missed it.

2) I believe that this takes no additional time unless you specifically ask for it with a command line option.  Is that right?

3) You say that a Jenkins build goes up by 1...3 minutes.  Please check to see how much additional time it takes on mu2ebuild01 with -j 24 and on one of the mu2egpvm machines with -j 6 .

( My recommendation is that the -j value should be 1.5 times the number of cores.  The optimum is shallow )

ryuwd:
Hi Rob,

I realise this is a bit confusing. I had a script before called `gen_compile_commands.py` which was slow, however, I'd like to use this Scons-integrated tool instead, as it is more accurate and efficient to use.

> Ryun:
> 
>     1. Does this thread say what the purpose of this addition is?  What functionality does this buy us?  I don't see it in the conversation but maybe I missed it.

Clang-tidy needs a compile_commands.json (which these additions provide) to know what compile flags e.g. includes, etc to give to clang when it is analyzing a file.


> 
>     2. I believe that this takes no additional time unless you specifically ask for it with a command line option.  Is that right?

The time this Tool takes relative to the build is negligible (about 2-3 seconds). At the moment, I realise that it will produce a compile_commands.json whether or not you pass the `-Q compiledb` flag. I can add an extra flag to fix this, however.

> 
>     3. You say that a Jenkins build goes up by 1...3 minutes.  Please check to see how much additional time it takes on mu2ebuild01 with -j 24 and on one of the mu2egpvm machines with -j 6 .

The script I'm proposing in this PR is actually very fast. The slowdown I was referring to was with the script I was using before:  my custom `gen_compile_commands.py`, which scrapes the stdout of scons --dry-run (which can take a long time). It also requires the build to be cleaned before, so it is not a very integrated solution. 

The solution I propose in this PR integrates directly into SCons, so it provides a quicker and more accurate result.

> 
> 
> ( My recommendation is that the -j value should be 1.5 times the number of cores. The optimum is shallow )

I might try changing the number of jobs on jenkins to 24 as per your recommendation

kutschke:
I am merging this in now with the understanding the impact on the default build is small.

PR #151, opened 2020-03-03
Mmax

resnegfk:
Making sure RandomNumberGenerator defaultEngineKind is set properly when setting restoreStateLabel


resnegfk:
AFAIK, the restore will fail, and execution will terminate, so there is no risk of it leading to an incorrect restore.

PR #152, opened 2020-03-05
cleaned up MT fcl files

goodenou:
I cleaned up MT fcl files to reduce points of maintenance.  Also, I added scheduler configuration parameters for MT jobs in anticipation of art 03_05_00.  The latter take the form:

services.scheduler.num_schedules : n
services.scheduler.num_threads   : n

where 'n' is the number of schedules and threads to run.


JobConfig/beam/beam_defs_g4s1.fcl
```diff
@@ -98,7 +98,7 @@ source : {
 services : {
 
    message               : @local::default_message
-   TFileService          : { }
+   TFileService          : { fileName : "beamg4s1.root" }
    RandomNumberGenerator : {defaultEngineKind: "MixMaxRng" }
```
kutschke: I suspect that this field and the output file fields were left blank on purpose because generate_fcl fills them in.  The idea is that this is a deliberately incomplete file.   We do not want to force people to think about certain things when they use this fcl file - so we deliberately leave them blank.

I will add Andrei and Ray as a reviewers since they set up a lot of JobConfig.

Please check the whole pull request for this.

JobConfig/beam/beam_defs_g4s1.fcl
```diff
@@ -329,3 +331,5 @@ services.message.destinations.log.categories.ArtReport.timespan : 300
 // Initialze seeding of random engines: do not put these lines in base .fcl files for grid jobs.
 services.SeedService.baseSeed         :  8
 services.SeedService.maxUniqueEngines :  20
+
+services.GeometryService.simulatedDetector.tool_type : "Mu2e"
```
kutschke: I believe that if you are doing everything else correctly then this should not be necessary - (not 100% certain about that).

JobConfig/beam/beam_defs_g4s1.fcl
goodenou: This was the only file that I made these types of changes in.

JobConfig/beam/beam_defs_g4s1.fcl
goodenou: I removed this line in this file and in one other file.  The other file runs without it.

goodenou:
I have made the changes you requested.

JobConfig/beam/beam_defs_g4s1.fcl
kutschke: Here is the story on that line.  If the fcl for the services contains  @local::Services.Sim  or @local::Services.SimAndReco, then you do not need this line - that definition is already there.  If the fcl for the GeometryService is just:
GeometryService : {  inputFile : "filename.txt" }

Then you need that line.  The preferred long term solution is that the owner of the file should retro-fit it to define services in the recommended way and then remove this line.

JobConfig/beam/beam_defs_g4s1.fcl
gaponenko: > I suspect that this field and the output file fields were left blank on purpose because generate_fcl fills them in. The idea is that this is a deliberately incomplete file. We do not want to force people to think about certain things when they use this fcl file - so we deliberately leave them blank.

Yes, files with "defs" in their names are supposed to be incomplete.  I think they even predate generate_fcl.  The "defs" are similar in spirit to what we call "prolog.fcl" now, but they were written before FHICL had all the functionality we can use now.

JobConfig/cosmic/cosmic_defs_s1s2.fclinc
```diff
@@ -194,11 +193,11 @@ physics :
       cosmicFilter: 
       {
          module_type: FilterG4Out
-         noInstanceName : true
+         #noInstanceName : true
```
gaponenko: The noInstanceName and numSimParticleCollections settings do not exist any more.  One can see the currently supported settings by running "mu2e --print-descriptions FilterG4Out".    Lisa, please remove such commented out lines for the obsolete parameters.

goodenou: I removed the outdated lines from these various fclinc files.

rlcee:
Why does 
 Validation/fcl/potSimMT.fcl 
have 2 threads when the others default to 5?

It looks like there are still "tool_type" lines,
are the remaining ones necessary?

goodenou:
> On Mar 6, 2020, at 12:56 PM, Ray Culbertson <notifications@github.com> wrote:
> 
> Why does
> Validation/fcl/potSimMT.fcl
> have 2 threads when the others default to 5?
> 
> 

I was using potSimMT for some testing, and I didn't need 5 threads.  There is nothing special about 2 or 5.  I just wanted to make it clear that the number can be greater than 1, and at this point I have no preference for a particular number.

> It looks like there are still "tool_type" lines,
> are the remaining ones necessary?
> 
> 

There are two types of tool_type lines remaining:
(1) lines where the tool_type is e.g. “ConstructEnvMuCapture”.  These are in fcl files that are not MT-related, and I don’t know enough about these jobs to know what the correct parameter should be.

(2) lines where the tool_type is “Mu2e”.  These are all in fcl files that have their geometry defined via an input file, e.g. services.GeometryService.inputFile : "Mu2eG4/geom/geom_common_phase1.txt”.  According to Rob’s earlier message, the tool_type parameter must be present in these.

ryuwd:
Hi,
Don't worry about the above message. We just turned FNALbuild on and it prints a 'greeting message' on PR's it sees for the first time. The tests aren't required

goodenou:
So, am I done with this?

goodenou:
Thanks to you all for checking all of the details I am unaware of.

kutschke:
yes - I will merge in a few minutes.

Thanks to everyone for your help reviewing this.

goodenou:
I tested the two fcl files that #include it, JobConfig/cosmic/cosmic_s1_general_cryMT.fcl and JobConfig/cosmic/cosmic_s1_general_dbayMT.fcl

PR #154, opened 2020-03-10
BField Plotting Module

michaelmackenzie:
Added an analyzer that creates a 2D histogram of each magnetic field map, as well as the default map returned by the BFieldManager.

Also added test fcl that shows how to use the module.


kutschke:
@FNALbuild run build test

Analyses/src/BFieldPlotter_module.cc
```diff
@@ -0,0 +1,173 @@
...
+
+  } //end constructor
+
+  void BFieldPlotter::analyze(const art::Event& event) {
+    art::ServiceHandle<art::TFileService> tfs;
+    // Get the magnetic field manager and the map lists
+    GeomHandle<BFieldManager> bf;
+    const BFieldManager::MapContainerType innerMaps = bf->getInnerMaps();
+    const BFieldManager::MapContainerType outerMaps = bf->getOuterMaps();
+
```
kutschke: These should have type:
 const BFieldManager::MapContainerType&

Note the reference.  Not a big deal in this case since your are making a copy of vector of shared pointers to big things, not a copies of big things.  But just do it right every time, even when it does not matter much.

Analyses/src/BFieldPlotter_module.cc
```diff
@@ -0,0 +1,173 @@
...
+    // Get the magnetic field manager and the map lists
+    GeomHandle<BFieldManager> bf;
+    const BFieldManager::MapContainerType innerMaps = bf->getInnerMaps();
+    const BFieldManager::MapContainerType outerMaps = bf->getOuterMaps();
+ 
+    //plot inner maps
+    for(std::shared_ptr<BFMap> map : innerMaps) { 
+      fillHistogram(map, tfs);
+    } 
+
```
kutschke: Again not a big deal in this particular case.   I would write

for ( std::shared_ptr<BFMap> const& map : innerMaps ){ ... }

There is a small overhead in copying rather than just using a reference to the shared_ptr.  Also make everything const that you can.  I never hurts and sometimes it allows the compiler to do better optimization.

I just realized there is a hole in the design of the BFieldManager class - I will open an issue.
  
The first argument to fillHistogram should be a BFMap const* not a copy of the shared pointer. So do the dereferencing of the shared_ptr in the caller.  There are two parts to this story.  In the called function, there is overhead every time you dereference the shared pointer.  So just dereference it once.  Second, inside fillHistogram as written you have non-const access to the map - so again the compiler cannot do some optimizations.

Analyses/src/BFieldPlotter_module.cc
```diff
@@ -0,0 +1,173 @@
...
+	axisTwo += _mapBinSize;
+      } //end axis two loop
+      axisOne += _mapBinSize;
+    } //end axis one loop
+  } //end fillHistorgram
+
+  void BFieldPlotter::endJob() {
+
+  } // end endJob
+
```
kutschke: If you don't need it, just get rid of it.  It's Ok to keep it if, from time to time, you expect to do some debugging that needs it.

Analyses/src/BFieldPlotter_module.cc
```diff
@@ -0,0 +1,173 @@
...
+
+}  // end namespace mu2e
+
+using mu2e::BFieldPlotter;
+DEFINE_ART_MODULE(BFieldPlotter);
```
kutschke: You can write this as one line:

DEFINE_ART_MODULE(mu2e::BFieldPlotter);

In the early days of art you could not do it in one line due to limitations in CPP.  They are long gone.

Analyses/test/bfieldPlotter.fcl
```diff
@@ -0,0 +1,32 @@
+# -*- mode:tcl -*-
+# Example of running BFieldPlotter
+#
+
+#include "JobConfig/primary/flateminus.fcl"
```
kutschke: You don't need this include since you don't need generate or g4run

Analyses/test/bfieldPlotter.fcl
```diff
@@ -0,0 +1,32 @@
...
+    axisTwoMin  : -25000. # -8050. # -25000.
+    axisTwoMax  :  45000. #  8050. #  45000.
+    mapBinSize  :  10.
+}
+
+physics.e1             : [bfieldplotter]
+physics.end_paths      : [e1]
+physics.p1             : [generate, g4run]
+physics.trigger_paths  : [p1]
+
```
kutschke: You should not need either generate or g4run to make this work.  Get rid of them.

Actually I think that the only line that you actually need is:
physics.e1             : [bfieldplotter]

Analyses/test/bfieldPlotter.fcl
```diff
@@ -0,0 +1,32 @@
...
+services.TFileService.fileName: "nts.owner.bfieldplotter.x_-10000.sequencer.root"
+outputs.Output.fileName       : "dig.owner.bfieldplotter.version.sequencer.art" #shouldn't be generated
+
```
kutschke: Remove the outputs line - not needed and it actually makes art do extra work that it is not needed.

Analyses/test/bfieldPlotter.fcl
```diff
@@ -0,0 +1,32 @@
...
+#Use file with altered field maps
+services.GeometryService.inputFile : "Analyses/test/geom_bfieldplotter.txt"
+
+#Don't need full physics to map the field
+physics.producers.g4run.physics.physicsListName : "Minimal" 
+
```
kutschke: This is not needed since you do not need G4.

Analyses/test/geom_bfieldplotter.txt
```diff
@@ -0,0 +1,26 @@
...
+  "BFieldMaps/Mau13/DSMap.header"
+ ,"BFieldMaps/Mau13/PSMap.header"
+ ,"BFieldMaps/Mau13/TSuMap_fix.header"
+ ,"BFieldMaps/Mau13/TSdMap.header"
+ ,"BFieldMaps/Mau13/PStoDumpAreaMap.header"
+ ,"BFieldMaps/Mau13/ProtonDumpAreaMap.header"
+ ,"BFieldMaps/Mau13/DSSweepExtension.header"
+};
+//DSSweepExtension supersedes this
+ // ,"BFieldMaps/Mau13/DSExtension.header"
```
kutschke: My recommendation is to get rid of comments that capture transient state.  git is good at documenting transient state.

Analyses/test/bfieldPlotter.fcl
kutschke: I just realized why you have this line.  You are pulling in some very deeply nested includes.

What you need is:

#include "fcl/minimalMessageService.fcl"
#include "fcl/standardServices.fcl"
services: @local::Services.Core

This defines a little more than you need for this job but it's not a lot more and it's the smallest chunk that's robust against future evolution of our fcl configuration (ie we will always define Services.Core and it will always have a properly configured GeoemtryService ).

Analyses/test/geom_bfieldplotter.txt
michaelmackenzie: Good call, I've removed these comments

Analyses/test/bfieldPlotter.fcl
michaelmackenzie: I had done it that way since it was simple to get the test script going, but I agree the better test file is one simply using the relevant services and less sensitive to outside changes. I've made this switch!

Analyses/src/BFieldPlotter_module.cc
michaelmackenzie: Clearly an inherited aspect, thank you!

Analyses/src/BFieldPlotter_module.cc
michaelmackenzie: Removed this

Analyses/src/BFieldPlotter_module.cc
michaelmackenzie: I've updated this and passed the pointer along in the call rather than creating a new shared_ptr

Analyses/src/BFieldPlotter_module.cc
michaelmackenzie: Good catch, thank you!

Analyses/test/bfieldPlotter.fcl
michaelmackenzie: Fixed this, thank you!

Analyses/test/bfieldPlotter.fcl
michaelmackenzie: Remove this line

Analyses/test/bfieldPlotter.fcl
michaelmackenzie: I've removed this and g4run being used

Analyses/src/BFieldPlotter_module.cc
```diff
@@ -0,0 +1,166 @@
...
+    } 
+
+    //plot the default field returned from the manager
+    fillHistogram(NULL, tfs);
+  } // end analyze
+
+
+  //fill a histogram with the magnetic field
+  void BFieldPlotter::fillHistogram(BFMap* map, art::ServiceHandle<art::TFileService> tfs) {
+    //define histogram binning
```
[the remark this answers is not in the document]
michaelmackenzie: Added this

Analyses/src/BFieldPlotter_module.cc
```diff
@@ -0,0 +1,166 @@
...
+				   nbinsOne, _axisOneMin - _mapBinSize/2.,_axisOneMax + _mapBinSize/2.,
+				   nbinsTwo, _axisTwoMin - _mapBinSize/2.,_axisTwoMax + _mapBinSize/2.);
+    
+    GeomHandle<BFieldManager> bf; //only needed in null map case
+    //Loop through the points in the magnetic field
+    double axisOne = _axisOneMin;
+    while(axisOne <= _axisOneMax) {
+      double axisTwo = _axisTwoMin;
+      while(axisTwo <= _axisTwoMax) {
+	CLHEP::Hep3Vector point;
```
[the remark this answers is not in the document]
michaelmackenzie: I switched to integer for loops, rather than the double comparison while loops, thanks!

kutschke:
@FNALbuild run build test

kutschke:
@FNALbuild run build test

PR #158, opened 2020-03-16
updates to the DAQ modules

gianipez:
- refactored the DAQ directory:
      -> added module for creating binary files from digis (for the calorimeter and the tracker)
      -> added module to convert the artFragments into calo/tracker digis
- removed deprecated DAQDataProducts directory; that was no longer needed. The new schema doesn't need any offline data product for the conversion of artFragment to digi


kutschke:
@FNALbuild run build test

kutschke:
I don't understand this:

> The new schema doesn't need any offline data product for the conversion of artFragment to digi

Do you mean that the module that does the conversion of fragments to digis is not found in Offline?  Or something else?

gianipez:
Hi Rob,

That module is in the Offline. What was doing before was: converting all the fragments (from calo and tracker) to an intermidiate data product, and then using separate modules for converting this tmp data product into tracker or calo digi.
In the new workflow, we use the same module to convert directly fragments (calo or tracker) to digis.

kutschke:
Thanks Giani,

   I understand it now.

   Please point me to am art file on disk that contains artdaq::Fragments consistent with this code.   And also point me to a fhicl file to run the trigger from these files.

  Thanks,
                   Rob

DAQ/src/SConscript
```diff
@@ -18,14 +18,13 @@ rootlibs = env['ROOTLIBS']
 
 mainlib = helper.make_mainlib ( [ 'mu2e_ConditionsService',
                                   'mu2e_GeometryService',
-                                  'mu2e_MCDataProducts',
+                                  # 'mu2e_MCDataProducts',
```
kutschke: It's great that we no longer need MCDataProducts - thanks for looking after that. Please just remove the line instead of comment it out.  Same comment for any other place this shows up.

gianipez:
Hi Rob,

You can run this command after you compiled the code with the trigger parameter in buildopts set to ‘on’.

 mu2e -c DAQ/test/generateDigiFromFragment.fcl -smu2e/data/users/gianipez/data/driver_rootOutput.art

Thanks,


Gianantonio Pezzullo, Ph.D.
giani.pezzu@gmail.com


Phone   : ‭+1 (331) 205-9306‬  	     		

gianipez:
@FNALbuild run build test

kutschke:
Thanks Giani,

  The job ran.   I have a question about the cardinality of the new format data products. 

   In this file the calo artdaq::Fragment data product always has 168 entries - is this fixed, maybe one per some some grouping of crystals? Or is it an artifact of this dataset?  When we look at cosmic rays, will we always have 168 entries, most of which are empty? If so, how big is an empty artdaq::Fragment?
  
   In this file there is a variable number of artdaq::Fragments.  I looks to me as if there is one artdaq::Fragment per StrawDigi.  Is that right?  If you compare the overhead of a fragment compared to the payload of a single digi, how much is this choice costing us in overhead compared to a more compact representation?

  In both cases, is an artdaq::Fragment a fixed size thing?

  In the first two cases I have asked questions about added overheads.  Maybe the answer is that they are small, or even that they are OK when they are big - in all cases all stakeholders need to be in the loop on this.

       Rob

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
```diff
@@ -0,0 +1,1113 @@
+//
+// EDProducer module for converting straw hit digis into DTC formatted
+// packets (which are stored in DataBlock data products)
+//
+//
+// Original author Tomonari Miyashita
+//
```
kutschke: Is Tomo really the original author of this?  It's appropriate to give him credit if this is substantially based on his work but if it's a total rewrite it's your code.

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
```diff
@@ -0,0 +1,1113 @@
...
+#include <fstream>
+#include <stdexcept>
+
+
+// Definitions and typedefs needed for compatibility with HLS codeblock
+#define NUM_PRESAMPLES 4
+#define START_SAMPLES 5 // 0 indexed
+#define NUM_SAMPLES 15
+#define LOWER_TDC 16000
+#define UPPER_TDC 64000
```
kutschke: Two things:

Always prefer to use constexpr over #define.   Remember that #define is NOT part of C++ it is a C-preprocessor directive.   Therefore many of the safety features of the compiler are not available.

As targets of opportunity, please change this everywhere in your code.

In any case these values should not be hard coded.  Other code will want to know them as well.  So  they belong somewhere that many modules have access to them.   Unless you can make a very compelling case that these are fixed for all time, the only place for them is in Proditions.

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
```diff
@@ -0,0 +1,1113 @@
...
+
+typedef uint16_t tdc_type;
+typedef uint8_t  tot_type;
+typedef uint16_t adc_type;
+typedef uint16_t calib_constant_type;
+typedef uint8_t  flag_mask_type;
+typedef uint64_t timestamp;
```
kutschke: Prefer to use the more modern using syntax, as you do below.

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
```diff
@@ -0,0 +1,1113 @@
...
+  //
+  //
+  class ArtBinaryPacketsFromDigis : public art::EDProducer {
+  public:
+
+    explicit ArtBinaryPacketsFromDigis(fhicl::ParameterSet const& pset);
+
+    virtual void beginJob() override;
+    virtual void beginRun(art::Run&);
+
```
kutschke: Add override everywhere it can be used

gianipez:
Hi Rob,

Thanks for making these questions. 168 is the number of ROCs that the calorimeter plans to use. The difference w.r.t. the Tracker is driven by the different data model used to send the data out by the different ROCs. 
I think Bertrand wanted to address this issue of optimization of the Fragment format and their un-packing. As it is right now, it's a 1-to-1 translation of the binary file, which currently reflects the Firmware decisions of the sub-detectors. For the specific questions about the artFragments, I think Eric Flumerfelt is the best person for answering them. 

Cheers,

Giani

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
```diff
@@ -0,0 +1,1113 @@
...
+    // 216 panels
+    //
+    // 6 ROCs per DTC
+    // 36 DTCs
+    //    const size_t number_of_rocs = 216;
+    const size_t number_of_rocs = 240;
+    const size_t number_of_straws_per_roc = 96; // Each panel in the tracker has 96 straws
+    const size_t number_of_rocs_per_dtc = 6;
+    const size_t numADCSamples = 15;
+
```
kutschke: These numbers should not be hard coded but should come from the single authoritative source.

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
```diff
@@ -0,0 +1,1113 @@
...
+    void   fillCalorimeterDataStream(raw_data_list_t& dataStream, calo_data_block_t& caloData);
+
+    void   printCalorimeterData(CaloDataPacket const& curDataBlock);
+
+    size_t waveformMaximumIndex(std::vector<adc_t>& waveform);
+
+    void flushBuffer(raw_data_list_t& buffers);
+    void closeDataBuffer(raw_data_list_t& dataStream, bool openNew = true) {     
+      uint64_t sz = dataStream.back().second;
```
kutschke: Something this complex should not be implemented inside the declaration.  In this case it makes no real difference since, for modules, declaration and implementation are in a single file.  But it's good practice to keep large implementations out of the declarations.

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
```diff
@@ -0,0 +1,1113 @@
...
+      }
+      if (_diagLevel > 3) {
+	std::cout << "[closeDataBuffer] dataStream.size = "<< dataStream.size() << std::endl;
+	if (dataStream.size()> 0){
+	  std::cout <<  "[closeDataBuffer] dataStream.back().second = " << dataStream.back().second <<std::endl;
+	}
+	std::cout<< "[closeDataBuffer] 2) sz = "<<sz<<std::endl;
+      }
+      uint64_t ex_sz = sz - 16;
+
```
kutschke: The above has a lot of numeric literals: 2, 3, 16, 48 ...  Each should be defined as an appropriate symbolic, either constexpr or a variable coming from Proditions.

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
```diff
@@ -0,0 +1,1113 @@
...
+    std::vector<adc_t> generateDMABlockHeader(size_t theCount) const;
+    std::vector<adc_t> generateEventByteHeader(size_t theCount) const;
+  };
+
+  //--------------------------------------------------------------------------------
+  // temporary function used to find the location of the waveform peak in the 
+  // calorimeter digitized waveform
+  //--------------------------------------------------------------------------------
+  size_t ArtBinaryPacketsFromDigis::waveformMaximumIndex(std::vector<adc_t>& waveform) {
+    size_t  indexMax(0), content(0);
```
kutschke: 1) The argument should be const& and the member function should be const; please check for this everywhere.

2) I think that there is a standard library function that will do this for you.  I would implement it using that.  If there is a standard library function to do what you want, the best you can do is equal it - it's highly unlikely you can do better.  Not quite the second law of thermodynamics, but close.

3) This does not need to be a member function. If something does not need to be a member function, make it a free function.  It can live in its own file if it is general interest; if it is so specialized that it really only belongs in this file, put it in an anonymous namespace at the top of the file.

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
```diff
@@ -0,0 +1,1113 @@
...
+      printf("[ArtBinaryPacketsFromDigis::printCaloData] DIRACA        : %i \n", (int)hit.DIRACA);
+      printf("[ArtBinaryPacketsFromDigis::printCaloData] DIRACB        : %i \n", (int)hit.DIRACB);
+      printf("[ArtBinaryPacketsFromDigis::printCaloData] ErrorFlags    : %i \n", (int)hit.ErrorFlags);
+      printf("[ArtBinaryPacketsFromDigis::printCaloData] Time          : %i \n", (int)hit.Time);
+      printf("[ArtBinaryPacketsFromDigis::printCaloData] NumberOfSamples : %i \n", (int)hit.NumberOfSamples);
+      printf("[ArtBinaryPacketsFromDigis::printCaloData] IndexOfMaxDigitizerSample : %i \n", (int)hit.IndexOfMaxDigitizerSample);
+    }
+
+  }
+
```
kutschke: I read this quickly but it looks like all of these print functions can be free functions.  I would put them in a separate file since I can see other modules wanting to call these  - for example we might want to use them in the Print subdirectory of Offline.

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
```diff
@@ -0,0 +1,1113 @@
...
+  void   ArtBinaryPacketsFromDigis::fillTrackerDataStream(raw_data_list_t& dataStream,
+							  tracker_data_block_t const& trackerData) {
+
+    auto sz = sizeof(DataBlockHeader);
+    //check that the trkDataBlock is not empty
+
+    assert(trackerData.first.PacketCount == 0 || trackerData.first.PacketCount == 2); // Tracker DataBlocks have 0 or 2 DataPackets
+    if (trackerData.first.PacketCount > 0) {
```
kutschke: Remember that asserts are ignored in our prof builds.  Is that the intended behaviour?

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
```diff
@@ -0,0 +1,1113 @@
...
+    // Word 5
+    headerData.TimestampHigh = static_cast<adc_t>((timestamp >> 32) & 0xFFFF);
+    // Word 6
+    headerData.Status = 0; // 0 corresponds to "TimeStamp had valid data"
+    headerData.FormatVersion = format_version;
+    // Word 7
+    headerData.DTCID = DTCId;
+    uint8_t  evbMode = 0;//maybe off-spill vs on-spill?
+    headerData.EVBMode = evbMode;
+  }
```
kutschke: Use symbolic names for the numeric literals.  Which of these are used in other modules/functions?  Those should come from a common, authoritative source.

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
```diff
@@ -0,0 +1,1113 @@
...
+    TrkData.SetWaveform(6,  theWaveform[6]);
+    TrkData.SetWaveform(7,  theWaveform[7]);  
+    TrkData.SetWaveform(8,  theWaveform[8]);  
+    TrkData.SetWaveform(9,  theWaveform[9]);  
+    TrkData.SetWaveform(10, theWaveform[10]);  
+    TrkData.SetWaveform(11, theWaveform[11]); 
+    TrkData.SetWaveform(12, theWaveform[12]); 
+    TrkData.SetWaveform(13, theWaveform[13]); 
+    TrkData.SetWaveform(14, theWaveform[14]); 
+
```
kutschke: Why not write this as a loop?

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
```diff
@@ -0,0 +1,1113 @@
...
+    //    TrkData.unused1 = ;
+    if (_enableFPGAEmulation) {
+      // Note: Eventually, there will be a calibration database to provide these sorts of values
+      // For now, these are just placeholders
+      calib_constant_type clockstart = 320; //10 / .03125 (tdclsb)
+      calib_constant_type maxenergyls8 = 583, minenergyls8 = 0;
+      calib_constant_type gainrs15 = 1389;
+      calib_constant_type inverseionizationenergyls26 = 633;
+      calib_constant_type panelTDCoffset = 0, hvoffset = 0, caloffset = 0;
+
```
kutschke: Magic numbers again.

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
```diff
@@ -0,0 +1,1113 @@
...
+    }
+
+    uint8_t max_dtc_id = number_of_calo_rocs / number_of_calo_rocs_per_dtc - 1;
+    if (number_of_calo_rocs % number_of_calo_rocs_per_dtc > 0) {
+      max_dtc_id += 1;
+    }
+
+    // Loop over the DTC/ROC pairs and generate datablocks for each ROC
+    for (uint8_t dtcID = 0; dtcID < max_dtc_id; dtcID++) {
+
```
kutschke: Is there a check anywhere that max_dtc_id < 255 so that it fits in a uint8_t?

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
```diff
@@ -0,0 +1,1113 @@
...
+    return (failed_energy << 1) | failed_time;
+
+  }
+
+}
+
+
+using mu2e::ArtBinaryPacketsFromDigis;
+DEFINE_ART_MODULE(ArtBinaryPacketsFromDigis);
```
kutschke: You can write this in one line:
DEFINE_ART_MODULE(mu2e::ArtBinaryPacketsFromDigis);

In the very old days it had to be done on 2 lines because of limitations in the CPP.

kutschke:
Thanks Giani,

  I don't need these questions resolved in order to approve the pull request but I would like them addressed over the next few months.

  I will ask Eric but I suspect that this is really a Mu2e issue.  As I understand it we subclass from artdaq::Fragment so we control the payload and therefore the size.  Perhaps I have misunderstood the architecture?

       Rob

kutschke:
Giani:  will this pull request  bring online and offline to the point that we are using the same ups products, the same compiler etc?

gianipez:
Hi Rob,

Yes. So probably we want to tag the Offline before and after making the merge.

DAQ/src/StrawAndCaloDigisFromFragments_module.cc
```diff
@@ -70,37 +70,54 @@ StrawAndCaloDigisFromFragments::StrawAndCaloDigisFromFragments(fhicl::ParameterS
...
-  auto trkFragments = event.getValidHandle<artdaq::Fragments>(trkFragmentsTag_);
-  auto calFragments = event.getValidHandle<artdaq::Fragments>(caloFragmentsTag_);
-  size_t numTrkFrags = trkFragments->size();
-  size_t numCalFrags = calFragments->size();
+  art::Handle<artdaq::Fragments> trkFragments, calFragments;
+  size_t numTrkFrags(0), numCalFrags(0);
+  if (parseTRK_){
+    event.getByLabel(trkFragmentsTag_ , trkFragments);
+    if (!trkFragments.isValid()){            
+      return;
```
kutschke: We need to have a policy discussions that involve both online and offline.

Suppose that we have an event in which the tracker is present but the calorimeter is not.  Or vice versa. That might often happen during commissioning and might happen from time to time during physics running when an error happens.  Do you really want to skip the good data just because something else is missing?  It's not useful for physics but it might very well be important for debugging.

The second policy is issue is that this code declares that it will produce something and then never produces it. I believe that for Offline we currently have art configured so that it's an error that art will catch and throw an exception.  There is run-time configuration to allow failure to put to be allowed.  I believe that there is also something like produce_maybe which allows us exempt single modules from this policy.

I am not asking for an action at this time but we need to discuss this the online team.

DAQ/src/StrawAndCaloDigisFromFragments_module.cc
```diff
@@ -405,8 +448,12 @@ void
...
-  event.put(std::move(straw_digis));
-  event.put(std::move(calo_digis));
+  if (parseTRK_){
+    event.put(std::move(straw_digis));
+  }
+  if (parseCAL_){
+    event.put(std::move(calo_digis));
+  }
 
 }  // produce()
```
kutschke: This routine would benefit greatly from major refactoring.   I expect that you inherited a lot of this from the previous author.  As written the code says:

 (loop over the combination of all trk and cal fragments){
     if ( track fragment){
           process the track fragment
     } else if ( cal fragment ) {
          process the cal fragment
     }
 } 

There is no common code between the two modes.  I would write one function to process all of the trk fragments an another to process all of the cal fragments.  I would also extract all of the debug output into helper functions so that the main line of the algorithm is more compact.  It's really hard to tell what this code is doing since it's all debug output with random bits of real code mixed in a few lines at a time.

DAQ/test/BinaryPacketsFromDataBlocksTest.fcl
```diff
@@ -5,62 +5,70 @@
...
+	    includeDMAHeaders: 1
+	    # Cosmic ray veto module is currently incomplete
+
+	    generateTimestampTable : 1
+	    generateBinaryFile : 1
+	    
+	    outputFile      : "DTC_packets_calo.bin"
+	    maxDMABlockSize : 32000
+	    bufferSize      : 1000
+	}
```
kutschke: Consider moving the standard module definitions into a prolog sooner rather than later.

DAQ/src/StrawAndCaloDigisFromFragments_module.cc
gianipez: Hi Rob,

We use two different module instances in the Online reco: one for taking care of the Tracker fragments and a second one for the calorimeter. In this way, we don't break/change the naming convention in the reconstruction sequences in the Online running from fragments or from Digis.

kutschke:
HI Giani,

   Great!  I would like to get the two code bases synched back to the same set of ups products.   

    We will soon need to move to a new art and break it again but having all of the other products the same is important.

 Rob

kutschke:
Then we should just make two modules.   The downstream configuration should not care so long as you keep the module labels the same.

DAQ/src/CrvDigisFromFragments_module.cc
```diff
@@ -64,16 +64,16 @@ CrvDigisFromFragments::CrvDigisFromFragments(fhicl::ParameterSet const& pset)
   , diagLevel_(pset.get<int>("diagLevel",0))
```
kutschke: Why is this class in the art namespace and not the mu2e namespace?  Same question for StrawAndCaloDigisFromFragments_module.cc .  I did not check the others.

gianipez:
The reason I didn't do it (yet) is that I want to measure what would be the reduction in the processing time.

DAQ/src/CrvDigisFromFragments_module.cc
gianipez: I have no clue. I did that because Tomo was doing it. I'm pretty sure the code should work also if I use the mu2e namespace. Probably he was doing that for debugging purposes?

gianipez:
I need to ask Eric.

kutschke:
In an earlier comment I suggested moving a lot of magic numbers into either Proditions or into a header with only a set of constexpr static definitions, much like the block of such definitions in DataProducts/inc/StrawId.hh .

I now remember a discussion from a long time ago about this.  Do I remember correctly that these definitions are also needed inside code that is does not have access to Proditions, either because it is in firmware or because it is in mu2e_artdaq-core?    But we do need to support the possibility that the firmware is changed some day and this code needs to be able to deal with both old and new data.  Are there other parts to this story that we need to think about when coming up with a plan?

kutschke:
> The reason I didn't do it (yet) is that I want to measure what would be the reduction in the processing time.

Are you expecting a reduction in processing time or do you just want to do changes one step at a time and measure with each change?

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
gianipez: I completely agree. The reason why I didn't do it is that I want to do it right. I pinged the people from the sub-detectors group for having a channel map I can use for exercising with the conditions service, but I received nothing yet.

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
gianipez: Hi Rob,

Once we will start using Sophie's improved caloDigi, I will delete this function, because the peak info will be available from it. So, if you agree, I will just implement the conversion to const and const&.

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
gianipez: Should I create a file Mu2eUtilities/inc/DAQTools.hh (src/DAQTools.cc)?

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
gianipez: Hi Rob,

This is on our to-do list. We didn't decide yet with Eric how to set up "EventMode" class in mu2e_artdaq. I was waiting to have some code working for the CRV online reconstruction to start working on it.

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
gianipez: that block of code is something from Tomo. I kept in only for keeping track, but so far I have never used that nor I read anything in Tomo's documentation about it. 
I can either delete it or comment it.
It seems to reproduce the operations performed already on the DRAC. I have never heard about any use-case for doing that.

kutschke:
Thanks Giani,

  I won't insist that this be fixed before approving the pull request - but it needs to stay on your radar.  When the time comes I have some ideas to add to the discussion.

            Rob

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
gianipez: The total number of ROCs and of DTC (for reading the ROCs) was fixed in the project, so that's why we didn't put any check.

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
kutschke: OK.  Please go ahead with that plan.

As you move forward, my advice is that if functions do not need access to member data, make them free functions.    The reason is that you don't want to give a function permission to modify something that it does not need to modify.

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
kutschke: Check with Ray but my suggestion is to put it into something like Print/inc/FragmentPrinter.hh  and Print/src/FragmentPrinter.cc

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
kutschke: Are you saying the when we have real data, you won't need to fill headerData since artdaq will have already filled it for you?  Or something else?

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
kutschke: If you expect that you will want this functionality soon, then keep it.  If you do not expect to need that functionality then please remove it but put enough information in your commit comment that you can find it again if you do need it.

gianipez:
> > The reason I didn't do it (yet) is that I want to measure what would be the reduction in the processing time.
> 
> Are you expecting a reduction in processing time or do you just want to do changes one step at a time and measure with each change?

Yes, I want to measure the variation in the processing time one step at a time.

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
kutschke: All data driven code must assume that it will sooner or later see corrupt data and it needs to be safe when that happens.

gianipez:
Hi Rob,

I did most of the clean up as you suggested. Thanks for the time spent reviewing the code. 
Tomorrow I will ask Eric about the use case for keeping the "EventNumber_t" data product and also about the FGPA emulator code I would like to get rid of. 

Cheers,

Giani

gianipez:
Hi Rob,

Eric told me that we might need in the coming future that "EventNumber_t" data product for debugging purposes when we will develop the way to assign the event number in the Online reco. So far now I guess it makes sense to keep it.

I think I made all the changes you have been asking. Please, let me know if I missed anything.

Cheers,

Giani

kutschke:
> On Mar 23, 2020, at 12:55 AM, Gianantonio Pezzullo <notifications@github.com> wrote:
> 
> 
> Hi Rob,
> 
> Eric told me that we might need in the coming future that "EventNumber_t" data product for debugging purposes when we will develop the way to assign the event number in the Online reco. So far now I guess it makes sense to keep it.
> 

OK - that's a good reason to keep it.

DAQ/test/analyzeTrkFragments.fcl
```diff
@@ -0,0 +1,66 @@
...
+      makeSD:
+      {
+	  module_type: StrawAndCaloDigisFromFragments
+	  diagLevel: 0
+	  parseCAL: 0
+	  parseTRK: 1
+	  
+	  trkTag  : "daq:trk"
+	  caloTag : "daq:calo"
+      }
```
kutschke: Can you use pre-defined configurations of makeSD and readSD here?  Or use them with one or two overrides?  Copying it in full makes it fragile to new parameters added to the standard configuration

DAQ/test/generateBinaryFromDigi.fcl
```diff
@@ -0,0 +1,61 @@
...
+	    # Cosmic ray veto module is currently incomplete
+
+	    generateTimestampTable : 1
+	    generateBinaryFile : 1
+	    
+	    #	    outputFile      : "DTC_packets_from_fragment.bin"
+	    outputFile      : "DTC_packets_NoPrimary.bin"
+	    maxDMABlockSize : 32000
+	    bufferSize      : 1000
+	}
```
kutschke: Is this a candidate for going into a prolog?

DAQ/test/generateDigiFromFragment.fcl
```diff
@@ -18,32 +18,47 @@ source : {
...
+      CaloDigiFromShower:		
+      {
+	  module_type: StrawAndCaloDigisFromFragments
+	  diagLevel: 0
+	  parseCAL: 1
+	  parseTRK: 0
+	  
+	  trkTag  : "daq:trk"
+	  caloTag : "daq:calo"
+      }
```
kutschke: Same comment as above - can you use a predefined standard definition here?

kutschke:
@FNALbuild run build test

DAQ/test/generateDigiFromFragment.fcl
gianipez: Done.

kutschke:
@FNALbuild run build test

kutschke:
There is one unresolved issue that we cannot address until some other work has been done.   I have added it as issue #164 .

PR #159, opened 2020-03-16
Remove bad std::move from CosmicReco

ryuwd:
Correction to one of my suggestions on #157

The `std::move` stopped straw_chits filling properly during outlier hit filtering (my bad - this got put in #157 by my suggestion.)

Testing:
Before: track strawhits after drift fit were empty (0 tracks for 2000 events)
After: track strawhits are now filling properly (tracks are back)

### DIFF
--- a/CosmicReco/src/CosmicTrackFinder_module.cc
+++ b/CosmicReco/src/CosmicTrackFinder_module.cc
@@ -323,7 +323,7 @@ namespace mu2e{
 							tmpHits.push_back(chit);
 						}
 				      }
-				      tmpResult._tseed._straw_chits = std::move(tmpHits);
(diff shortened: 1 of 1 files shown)


ryuwd:
To clarify: this PR doesn't add anything new. It just restores that line to its previous state.

kutschke:
@FNALbuild run build test

kutschke:
I will merge this now.

Do you know if ComboHitCollection satisfies the requirements for the compiler to write a move assignment operator for it?    If so, this assignment is just moving 3 pointers, an art::ProductID and a bool.  So it's fairly lightweight.  But if now, it's making a copy of the payload of the vector.

PR #160, opened 2020-03-21
changes to random number seeding of workers in MT mode

goodenou:
The random number seeds for the worker run managers are seeded by the MT run manager.  10k seeds are created at a time by a MixMax engine.  The initial seed for this generator is set by the fcl parameter 'initialSeed'.  The number of seeds needed for the run, and thus the events that will be processed, does not need to be known at run start time.  The seed algorithm will create seeds in batches of 10k seeds until there are no events left to process.


kutschke:
@FNALbuild run build test

Mu2eG4/inc/Mu2eG4Config.hh
```diff
@@ -200,9 +200,9 @@ namespace mu2e {
       fhicl::Atom<std::string> generatorModuleLabel {Name("generatorModuleLabel"), ""};
 
       fhicl::Atom<bool> G4InteralFiltering {Name("G4InteralFiltering"), false};
-      fhicl::Atom<int>  maxEventsToSeed {Name("maxEventsToSeed"),
-          Comment("Only used by Mu2eG4MT module."),
-          10000 };
+      fhicl::Atom<long> initialSeed {Name("initialSeed"),
+          Comment("Only used by Mu2eG4MTRunManager when running in MT mode"),
+          8 };
```
kutschke: See previous comment about using SeedService.

Mu2eG4/inc/Mu2eG4MTRunManager.hh
```diff
@@ -37,6 +37,10 @@ namespace mu2e {
     void initializeMasterRunAction();
     void stopG4();
     void terminateRun();
+    
+    G4bool SetUpAnEvent(G4Event* evt,
+                        long& s1, long& s2, long& s3,
+                        G4bool reseedRequired);
```
kutschke: For my own edification.  Is this a standard G4 function or one that you added?

Mu2eG4/inc/Mu2eG4WorkerRunManager.hh
```diff
@@ -51,6 +51,7 @@ namespace mu2e {
     void initializeUserActions(const G4ThreeVector& origin_in_world);
     void initializeRun(art::Event* art_event);
     void processEvent(art::Event*);
+    G4Event* generateEvt(G4int i_event);
```
kutschke: Is this a standard G4 function or a Mu2e addition?  The name seems weird to me - it sounds like it is a function that should run an event generator.  So if we control the name, let's talk about a better name.

Mu2eG4/src/Mu2eG4MTRunManager.cc
```diff
@@ -102,10 +106,9 @@ namespace mu2e {
     if(fSDM)
       { currentRun->SetHCtable(fSDM->GetHCtable()); }
     std::ostringstream oss;
-    //WHAT LIBRARY TO USE?
-    //G4Random::saveFullState(oss);
-    //randomNumberStatusForThisRun = oss.str();
-    //currentRun->SetRandomNumberStatus(randomNumberStatusForThisRun);
+    G4Random::saveFullState(oss);
+    randomNumberStatusForThisRun = oss.str();
+    currentRun->SetRandomNumberStatus(randomNumberStatusForThisRun);
```
kutschke: Please fill me in on the plan for storing random number engine state?  The SeedService already has functionality to store the state at the start of each event and I prefer all engines to use it rather than rolling their own.

This brings up another wrinkle - we need to talk more about the management of the engines that are used in the worker threads.  Ideally I would like to manage them with createEngine and the SeedService .  For historical technical reasons, create engine can only be called from the c'tor of a module.  So we would need to make them all in the c'tor, save pointers and pass out the pointers as needed.  I don't know if this causes a big fight with G4 or not.

Mu2eG4/src/Mu2eG4MTRunManager.cc
```diff
@@ -118,71 +121,49 @@ namespace mu2e {
...
-      eventModulo = int(std::sqrt(double(numberOfEventToBeProcessed/numworkers)));
-      if(eventModulo<1) eventModulo = 1;
-    }
-
-    if ( InitializeSeeds(numevents) == false && numevents>0 ) {
+    
+    //if we are using G4's seed filling scheme
+    if ( InitializeSeeds(seedbunchsize) == false && seedbunchsize>0 ) {
 
       G4RNGHelper* helper = G4RNGHelper::GetInstance();
```
kutschke: Is this thread safe?

Mu2eG4/inc/Mu2eG4MTRunManager.hh
```diff
@@ -61,9 +65,7 @@ namespace mu2e {
     G4VUserPhysicsList* physicsList_;
 
     int rmvlevel_;
-
-    G4int maxNumEventstoSeed_;
-
+    long initialSeed_;
```
kutschke: See previous comments about using the SeedService.

Validation/fcl/potSimMT.fcl
```diff
@@ -7,6 +7,7 @@
 # set in the non MT file included above (which includes prolog files)
 
 physics.producers.g4run.module_type : "Mu2eG4MT"
+physics.producers.g4run.initialSeed  : 10
```
kutschke: Another thing on  the list if we can use the SeedService to seed the master engine.

goodenou:
Hi Rob,
I will answer the other questions later, but I want to let you know that we cannot use the Seed Service a shared module, which is what the MT module is.  We must seed the engine ourselves.  In light of that, do you have different questions for me?


Lisa

kutschke:
Hi Lisa,

    We only access the SeedService in the c'tor when everything is still single threaded.   Is there a type of service that can be called from a shared module?  If so we *may* be able to just declare it to be that kind of service instead of a legacy service ( is legacy the proper name?).  Or we could actually make it threadsafe;  I *think* that all we need to do is to put a mutex at the entry and exit of:
 
           seed_t getSeed( std::string const& instanceName );

Since this is a startup only thing, we can afford mutexes.

I also need to go back and look at your answers to a previous question and follow up on that.


     Rob

Mu2eG4/fcl/g4test_pot.fcl
```diff
@@ -67,5 +67,8 @@ services.SeedService.maxUniqueEngines :  20
 # physics.producers.g4run.SDConfig.enableAllSDs : true
 physics.producers.g4run.SDConfig.enableSD     : [tracker, calorimeter, calorimeterRO, CRV, virtualdetector, stoppingtarget ]
 physics.producers.g4run.physics.BirksConsts : { G4_POLYSTYRENE : 0.07943 } // {string : mm/MeV }
+
+services.GeometryService.simulatedDetector.tool_type : "Mu2e"
+
 physics.producers.g4run.physics.physicsListName : "QGSP_BERT"
```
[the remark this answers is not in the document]
goodenou: I made /Mu2eG4/fcl/g4test_pot*fcl for my own testing.  I can delete these if you want.

kutschke:
Hi Lisa,

  If you no longer need them, delete them.

  If you still need them, put a comment at the top that that's what they are for.  If it's reasonable, implement them starting from #include from Validation.

  Let me know your plan.

      Rob

Mu2eG4/fcl/prolog.fcl
```diff
@@ -185,6 +185,7 @@ Mu2eG4MT: {
         g4run : {
             @table::mu2eg4runDefaultSingleStage
             module_type : "Mu2eG4MT"
+            initialSeed : 8 
         }
     }
 }
```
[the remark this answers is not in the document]
goodenou: We can use the Seed Service, but we cannot use the Random Number Engine Service in Mu2eG4MT_module.cc because it is a Shared Module.  I have altered the code to use the Seed Service.  The MTRunManager now gets a seed from the Seed Service.  It uses this seed in its MixMax random number generator to generate seeds for the worker random number generators.

Mu2eG4/inc/Mu2eG4Config.hh
goodenou: I have removed the initialSeed FHICL parameter.  We are now using the Seed Service to provide the seed to the random number generator used by the Mu2eG4 master run manager.

Mu2eG4/inc/Mu2eG4MTRunManager.hh
goodenou: SetUpAnEvent is a virtual member function of the G4MTRunManager.  It is called by the Worker Run Managers once for each event.  I modified the G4 version for our purposes.

Mu2eG4/inc/Mu2eG4MTRunManager.hh
goodenou: I have removed the initialSeed FHICL parameter. We are now using the Seed Service to provide the seed to the random number generator used by the Mu2eG4 master run manager.

Mu2eG4/inc/Mu2eG4WorkerRunManager.hh
goodenou: The G4 Worker Run Managers have a member function called generateEvent on which this is based.  The name is related to the fact that the random number seeds are set and the primary particles are generated, in this way the G4 Event is 'made' or generated.  I can change this name if you wish.  I was trying to keep our code as similar to the G4 code as possible so that we understand the relationship between the two.

Mu2eG4/src/Mu2eG4MTRunManager.cc
goodenou: As I mentioned in an email, we cannot call 'createEngine' from a Shared Module, which is what Mu2eG4MT_module.cc is.  There is a method for storing/reading the state of the Worker Run Managers in/from a file.  We can look into implementing this if necessary.

Mu2eG4/src/Mu2eG4MTRunManager.cc
goodenou: No, but this code is in the call initializeG4 which is called during beginRun when art is still running single-threaded.

Validation/fcl/potSimMT.fcl
goodenou: I have removed the initialSeed FHICL parameter. We are now using the Seed Service to provide the seed to the random number generator used by the Mu2eG4 master run manager.

Mu2eG4/fcl/g4test_pot.fcl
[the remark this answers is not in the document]
kutschke: If you no longer need them, delete them.

 If you still need them, can you implement them by starting from #include the file from Validation?

 Let me know what you plan to do.

Mu2eG4/fcl/prolog.fcl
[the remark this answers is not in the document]
kutschke: I don't see the change.  Did you push it? The last commit I see is 3 days ago.

Mu2eG4/inc/Mu2eG4MTRunManager.hh
kutschke: Thanks.  Once this settles down we should work with the G4 team via Krzysztof to get them to add hooks for us so that we don't need to modify their code.  The way we have it now will be a nuisance to maintain.

Mu2eG4/inc/Mu2eG4WorkerRunManager.hh
kutschke: If it's their name then we should use it.  I just needed to know.

Mu2eG4/src/Mu2eG4MTRunManager.cc
kutschke: Thanks. Let's defer storing and reading state to future work.

Mu2eG4/fcl/g4test_pot.fcl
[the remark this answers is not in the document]
goodenou: I have deleted them.

Mu2eG4/fcl/prolog.fcl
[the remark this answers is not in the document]
goodenou: I have just pushed the change.

goodenou:
I did the following validation tests: 5k ceSimReco events and 1k potSim events.  The agreement between the master branch and my branch was perfect for the non-MT versions of the code.  For the MT versions, there were some differences that may be interesting.  The pdfs showing the plots  are at /mu2e/app/users/goodenou/MT_RNG2_plots/

kutschke:
@FNALbuild run build test

kutschke:
@FNALbuild run build test

kutschke:
Hi Lisa,

  Look in  potSimMT_comp.pdf  at the plot in the upper right on page 11 and at the plot in the lower right on page 12.  It took me a long time convince myself that these are consistent with just random number differences.  They are effectively very low stats and very likely represent very low energy electrons that get into the DS.  The Y plots look Ok since they integrate over all X and Z so the Y plots are really telling you what's happening everywhere along the beam line.


Rob

kutschke:
Hi Lisa,

  I am ready to do the merge but I will hold off, probably until tomorrow. The issues is scheduled all of the merges in a way that the nightly validation helps us.  If we do too many at once it's harder to untangle when something unexpected happens.

Rob

goodenou:
OK.  Thanks for the clarification.  Are they error bars on the eg X plots there statistical?  They appear to be less than Sort(N).

goodenou:
That should say Sqrt(N), spell correct changed it.

kutschke:
I am sure that wars will be started because of autocorrect.

The error bars look like sqrt(N) to me.  Note that there are only about 13000 entries inside the boundary of this plot.  What gets into this plot?  I bet that all of the SimParticles stopped or produced in this region are low energy electrons, positrons and photons.  There might be 1 or 2 muons in there.   Also remember that the 3 "start xyz DS" and the 3 "end xyz DS" plots are not cut on the other variables.

If you still have the art files for these, define a DS region with cuts on all 3 x,y,z variables; use the ranges of xyz used for the plots.   Then make a histograms of the number of StepPointMCs per event that start in that region.  Then make a scatter plot of x vs z for values of y in the range.

PR #161, opened 2020-03-23
Move compile_commands to the gen folder

ryuwd:
Moved `compile_commands.json` to `gen/` and made sure `gen/` is created before executing build targets. Tested on mu2ebuild01 with fresh clone of Offline

**N.B.**
- `compile_commands.json` is created for any build
- `scons -Q compiledb` generates a `compile_commands.json` without executing other build targets
- `compile_commands.json` is removed correctly after `scons -c`


kutschke:
@FNALbuild run build test

PR #163, opened 2020-03-23
Protecting rounding issues and bin size issues in BField plotting

michaelmackenzie:
Adding rounding when making bin numbers, and checking that the number of bins is consistent with the map grid size.

Also adding a small step into the map when at the edge of the histogram, 1% of the map grid size.

### DIFF
--- a/Analyses/src/BFieldPlotter_module.cc
+++ b/Analyses/src/BFieldPlotter_module.cc
@@ -80,15 +80,15 @@ namespace mu2e {
     if(_axisOneMin >= _axisOneMax || _axisTwoMin >= _axisTwoMax) {
       throw cet::exception("BADCONFIG") << "BField mapping plane ill defined: "
 					<< _axisOneMin << " < axisOneValues < " << _axisOneMax << ", "
-					<< _axisTwoMin << " < axisTwoValues < " << _axisTwoMax << endl;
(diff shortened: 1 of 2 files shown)


Analyses/src/BFieldPlotter_module.cc
```diff
@@ -79,15 +80,15 @@ namespace mu2e {
     if(_axisOneMin >= _axisOneMax || _axisTwoMin >= _axisTwoMax) {
       throw cet::exception("BADCONFIG") << "BField mapping plane ill defined: "
 					<< _axisOneMin << " < axisOneValues < " << _axisOneMax << ", "
-					<< _axisTwoMin << " < axisTwoValues < " << _axisTwoMax;
+					<< _axisTwoMin << " < axisTwoValues < " << _axisTwoMax << endl;
     }
```
kutschke: The exception system strips std::endl.  If you want to force a break use "\n".

kutschke:
@FNALbuild run build test

Analyses/src/BFieldPlotter_module.cc
michaelmackenzie: Good to know, I'll simply let it not end the line as it originally was in that case.

kutschke:
@FNALbuild run build test

PR #166, opened 2020-03-31
moved setting Birks' constants to ConstructMaterials

resnegfk:
Moved setting of Birks' constants to ConstructMaterials to enable setting the constants also for Mu2e constructed materials


kutschke:
@FNALbuild run build test

PR #172, opened 2020-04-10
modification to thread treatment at shutdown

goodenou:
I have changed the treatment of the threads at shutdown.  I am now using a 'hybrid' model: I correctly delete the objects (clear the memory) in the Worker Run Managers for threads that are still active at the end of the run, and I let the memory leak in those threads that are no longer active and thus not accessible.

There are additions changes in code involving the fhicl parameter 'mtDebugOutput' which allows for more flexibility in the printouts.  mtDebugOutput was changed from a bool to an int.


kutschke:
@FNALbuild run build test

PR #173, opened 2020-04-10
FromCorsikaBinary: use mu2e SeedService for random seed managment.

gaponenko:
Hello,  this pull request is to switch the FromCorsikaBinary input source to using Mu2e SeedService instead of requiring a separate seed.


kutschke:
@FNALbuild run build test

PR #176, opened 2020-04-15
update products to art v3_05_00

rlcee:
Update products for art v3_05_00.  Changed mu2e_artdaq_core qualifier from online back to offline.  ceSimReco check shows no physics change.


brownd1978:
@FNALbuild run build test

PR #177, opened 2020-04-17
fixed bug in trigger prolog files

gianipez:
by mistake, I committed a change that I was testing on my OTS test.
In OTS I generated a binary file with the correct peak info in the calo data packet. So, I set by default the new Sophies calo-online sequences. In updating the TrkFikter/fcl/prolog_trigger, I forgot to add @sequence::CaloReco.OnlineReco to one of the sequence. That was creating the crash. The reason why these changes got into the merge is that I used git to migrate changes from /mu2e/app to the local disk on the mu2edaq nodes. I did this commit while the merge was still pending.
Sorry about that.

### DIFF
--- a/CaloCluster/fcl/prolog_trigger.fcl
+++ b/CaloCluster/fcl/prolog_trigger.fcl
@@ -25,7 +25,7 @@ CaloTrigger : {
 }
 
 
-OldCaloClusterFast :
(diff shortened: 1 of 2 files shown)


kutschke:
@FNALbuild run build test

PR #178, opened 2020-04-18
Bug fix: fcl/standardServices.fcl referred to an undefined variable.

gaponenko:
Add an include that defines @local::default_message.


kutschke:
@FNALbuild run build test

PR #179, opened 2020-04-20
upgrading geant4 to 10.6.p01

resnegfk:
I suggest to make people mentioned in:
https://mu2ewiki.fnal.gov/wiki/GeantChecklist#v4_10_6_p01
to be reviewers


kutschke:
@FNALbuild run build tests

PR #180, opened 2020-04-20
Top level geometry refactor

kutschke:
Rename some of the top level geometry files to bring them into compliance with the geometry file naming policy.  There only functional change is that geom_common.txt has been updated to point to a new named file.


kutschke:
@FNALbuild run build test

kutschke:
A few more comments on what I changed other than just renaming files to follow the new policy:

Bob had left geom_common.txt unchnaged.  I change it to point at a file that is different from before by only the pbar window change.

Bob made a new top level geometry file Mu2eG4/geom/geom_common_hayman_v2_TS06.txt by #including geom_common_hayman_v2.txt .  I redid this by copying geom_common_hayman_v2.txt to a new file (top_bridge .. ) and editing one line to make it match.  I did this because I would like the main line top level files to be complete geometry files, not #include a base and then add deltas; the second pattern is definitely what I want for files that are not in the main line of development.

This left Mu2eG4/geom/geom_common_hayman_v2_TS06.txt  as obsolete so I deleted it.

Bob's PR had also left  Mu2eG4/geom/geom_2019_PhaseI.txt as obsolete.  So I deleted it.

michaelmackenzie:
Hi Rob,

For the change in the stable geometry (geom_common.txt), I don't think we should be using the transport solenoid version 6. This has several updates for the MARS and CRY campaigns, but doesn't have the matching updates in the beamline, DS, shielding, etc. I mostly saw this top level version added as a file used for validation of the update to the pbar window and wedge.

If we want to include TS_v06, I think this should come with the other updates done along with it. We could either have this top level geometry use the updates from the MARS and CRY campaigns, and nothing beyond this (so no titanium window + aluminum wedge or updates we are working on) or have it stay as the geometry before these updates except for the new production target.

Let me know what you think!

Best,
Michael

kutschke:
Thanks Michael,

   So geom_common_hayman_v2.txt should have been left alone.  Right?

   I see that TS v07 was added the file included in geom_common_current.  Is that the right thing to do?

Rob

michaelmackenzie:
Hi Rob,

That's my opinion at least. I think the titanium window (TS_v07) is our best current idea of the window, but the design isn't set so it should be in the current model, but not the stable model.

I think once we're happy with the MARS/CRY updates, we can pull these into common (which would probably look like 
```c++
#include "top_MDC2020_v01.txt"
#include "TransportSolenoid_v06.txt"
```
Or perhaps making this MDC2020_v01 and the titanium window v02 which would include this and TS_v07...).

Best,
Michael

kutschke:
THanks Michael,

   This is very valuable.

   So we need to keep geom_common_hayman_v2_TS06.txt as Bob created it.  But I would like to
give it a useful name.  Does the following capture how it is used:

      top_MARS_CRY_2019_v01.txt

Or should it be 2020 not 2019?

         Rob

michaelmackenzie:
Hi Rob,

The geom file Bob created doesn't capture all of the CRY or MARS updates, but instead just the TS updates. I would either name it something that indicates this (perhaps top_MDC2018_hayman_v2_TS_v06?), or include the rest of the CRY/MARS updates and name it as you've given. These can be found as top_MDC2020_v01.txt except TS_v07 should be TS_v06.

Best,
Michael

kutschke:
Hi Michael,

  Do you have a few minutes to zoom?

Rob

michaelmackenzie:
Hi Rob,

Sure, I'm free to meet now if that works for you.

Best,
Michael

kutschke:
Thanks Ryan

  I will cancel my pull request and submit a new one.   I have two issues that I need to understand

1) Mu2eG4/geom/geom_2019_PhaseII.txt starts by #include Mu2eG4/geom/geom_2019_PhaseI.txt
   Should this be changed to #include the file pointed to by geom_common_current?  Or does it
   have to stay with the file currently included?  My sense is that it it should be the former.

2) We need to converge on a good name for  Mu2eG4/geom/geom_common_hayman_v2_TS06.txt.
     In an earlier email I suggested: top_MARS_CRY_2019_v01.txt  
     Is this a sensible name or is it for a more restricted study.  Should the year be 2019 or 2020?

On a separate issue:

> 
> I think once we're happy with the MARS/CRY updates, we can pull these into common (which would probably look like
> 
> #include "top_MDC2020_v01.txt"
> #include "TransportSolenoid_v06.txt"
> Or perhaps making this MDC2020_v01 and the titanium window v02 which would include this and TS_v07...).

I would like to keep with the rule that both geom_common and geom_common_current #include exactly
one file.  Their purpose is to minimize the maintenance of production and validation scripts but providing
a layer of indirection - so I want to keep them as just synonyms for named files.

Rob

michaelmackenzie:
Hi Rob,

> 1) Mu2eG4/geom/geom_2019_PhaseII.txt starts by #include Mu2eG4/geom/geom_2019_PhaseI.txt Should this be changed to #include the file pointed to by geom_common_current? Or does it have to stay with the file currently included? My sense is that it it should be the former. 

This would make sense. The only reason I didn't was that I wasn't sure that if when common current points to is updated, one has to check that this update is intended for both run periods, and update PhaseII to remove it if it isn't.

> 2) We need to converge on a good name for Mu2eG4/geom/geom_common_hayman_v2_TS06.txt. In an earlier email I suggested: top_MARS_CRY_2019_v01.txt Is this a sensible name or is it for a more restricted study. Should the year be 2019 or 2020? 

This file represents a mix of updates, where it has an updated production target and transport solenoid, but doesn't include other beamline or shielding updates. I'm not positive if we need this file in the future, as it seemed mostly there to test the difference in stopped muon yields when changing the pbar window and wedge from TS_v03--> v05 --> v06. I wouldn't give it a name with MARS or CRY though, as it only contains a subset of these updates.

> On a separate issue:
>> I think once we're happy with the MARS/CRY updates, we can pull these into common (which would probably look like #include "top_MDC2020_v01.txt" #include "TransportSolenoid_v06.txt" Or perhaps making this MDC2020_v01 and the titanium window v02 which would include this and TS_v07...).

>I would like to keep with the rule that both geom_common and geom_common_current #include exactly one file. Their purpose is to minimize the maintenance of production and validation scripts but providing a layer of indirection - so I want to keep them as just synonyms for named files.

I agree, I more meant a new top level file that contained those lines, which would then be included by the common or common current file.

Best,
Michael

kutschke:
HI Michael,

  FNAL email stopped delivering email to me for about 4 hours and then delivered it all in one big
chunk.  I will reply to your other emails.

  Please let me know if you are still available.

  Rob

kutschke:
Hi Micahel,

> On Apr 20, 2020, at 2:33 PM, michaelmackenzie <notifications@github.com> wrote:
> 
> 
> Hi Rob,
> 
> The geom file Bob created doesn't capture all of the CRY or MARS updates, but instead just the TS updates. I would either name it something that indicates this (perhaps top_MDC2018_hayman_v2_TS_v06?), or include the rest of the CRY/MARS updates and name it as you've given. These can be found as top_MDC2020_v01.txt except TS_v07 should be TS_v06.

Do you understand who plans to use this file for what and when?  

For now I will leave the name as is.  I don't want to rename it multiple times so that's why I say leave it for now.

PR #182, opened 2020-04-24
Refactoring and clean up of the CORSIKA code

soleti:
Following @gaponenko suggestions I implemented a new way to read the CORSIKA binary files. Now there is no need of specifying the format of the input binary file ("compact" or "full"): a common method is able to distinguish between the two and read the information accordingly. The run number is now obtained from the binary file content and not from the filename. Also, the "lowE" and "highE" parameters that were specified in the fcl can be read from the binary file, so they have been removed from the fcl configuration.

### DIFF
--- a/Sources/inc/CosmicCORSIKA.hh
+++ b/Sources/inc/CosmicCORSIKA.hh
@@ -234,10 +234,11 @@ namespace mu2e {
       static constexpr float _GeV2MeV = CLHEP::GeV / CLHEP::MeV;
       static constexpr float _cm2mm = CLHEP::cm / CLHEP::mm;
       static constexpr float _ns2s = CLHEP::ns / CLHEP::s;
-      static constexpr unsigned _fbsize_words = 5733 + 2;
+      static constexpr unsigned _fbsize_words = 5733 + 2; ///< CORSIKA fixed block size plus possible g77 padding per footnote 105
+                                                          ///< in the User Guide https://web.ikp.kit.edu/corsika/usersguide/usersguide.pdf
 
       CLHEP::Hep3Vector _cosmicReferencePointInMu2e;
-      float _fluxConstant = 1.8e4;
+      float _fluxConstant = 1.8e4; ///< Primary nucleon intensity for cosmic rays, as quoted in the PDG. Used for cosmic live-time calculation
       float _tOffset = 0; ///< Time offset of sample, defaults to zero (no offset) [s]
       bool _projectToTargetBox = false;
       float _showerAreaExtension = 0; ///< Extend distribution of corsika particles in x,z by this much (e.g. 1000 will extend 10 m in -x, +x, -z, and +z) [mm]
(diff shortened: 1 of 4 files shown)


Sources/inc/CosmicCORSIKA.hh
```diff
@@ -229,30 +234,35 @@ namespace mu2e {
       static constexpr float _GeV2MeV = CLHEP::GeV / CLHEP::MeV;
       static constexpr float _cm2mm = CLHEP::cm / CLHEP::mm;
       static constexpr float _ns2s = CLHEP::ns / CLHEP::s;
+      static constexpr unsigned _fbsize_words = 5733 + 2;
```
gaponenko, line 237: A comment about 5733 + 2 would be helpful.

Sources/inc/CosmicCORSIKA.hh
```diff
@@ -229,30 +234,35 @@ namespace mu2e {
...
 
       CLHEP::Hep3Vector _cosmicReferencePointInMu2e;
       float _fluxConstant = 1.8e4;
```
gaponenko, line 240: I  wanted to know why 1.8e4, but then realized that the number is always owerwritten by the corresponding setting in the fcl file.   May be just get rid of the number here?

Sources/src/CosmicCORSIKA.cc
```diff
@@ -23,19 +23,74 @@ namespace mu2e {
...
+        throw std::runtime_error("Error: record size not a multiple of 4");
+      }
+
+      // We will be looking at at least 8 bytes to determine the
+      // input file format, and all real CORSIKA records are longer
+      // than that.
+      if(reclen < 2*4) {
+        throw std::runtime_error("Error: reclen too small");
+      }
+
```
gaponenko, line 53: I suggest to also add a protection agains reclen > 4*_fbsize_words,  so we do not overwrite memory on corrupt inputs.

Sources/inc/CosmicCORSIKA.hh
gaponenko: That comment went onto a wrong line.  It was about the _fluxConstant.

brownd1978:
@FNALbuild run build test

Sources/inc/CosmicCORSIKA.hh
soleti: Added, thanks.

Sources/inc/CosmicCORSIKA.hh
soleti: 1.8e4 is the intensity of cosmic rays primary nucleons as quoted by the PDG and it's needed for the live-time calculation. I would still keep it in the code for clarity. I added a comment.

Sources/src/CosmicCORSIKA.cc
soleti: Added, thanks.

brownd1978:
@FNALbuild run build test

PR #185, opened 2020-04-30
generalized addG4Particle (again)

resnegfk:
enabled proper restoration of excited states

### DIFF
--- a/Mu2eG4/fcl/prolog.fcl
+++ b/Mu2eG4/fcl/prolog.fcl
@@ -118,7 +118,6 @@ mu2eg4DefaultDebug: {
     steppingVerbosityLevel : 0
     navigatorVerbosityLevel : 0
     navigatorCheckMode : false
-    preCreateIsomers : false
     pdgIdToGenerate : -1111111111 // change from  -1111111111 to activate for testing
     checkFieldMap : 0
     writeGDML : false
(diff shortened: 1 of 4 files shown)


goodenou:
The changes look fine to me.  Thank you Krzysztof for taking care of this.

kutschke:
@FNALbuild run build tests

kutschke:
Hi Pasha,

   If I remember correctly, Krzysztof discussed this with me and I wanted the exception thrown.  If this exception is thrown, it means that we tried to add the new particle to the table and the addition produced an entry with the wrong PDGID.   This should never happen unless something is badly broken.  If G4 evolves and requires us to modify this code it should usually get caught in validation.

  I feel strongly that, in this case, printing an informational message and continuing is wrong.

Rob

resnegfk:
Pasha, were you able to test if the PR works when you process events where the previous version did not?

gaponenko:
Hi Pasha,

> I'd argue that the production-quality software, and I consider the
> Mu2e MC to be at that level, should not be throwing exceptions in
> logically foreseeable cases. Instead, the code should generate a
> warning, return an error code, do the right thing, and continue,
> letting the error code to be handled at a higher level.

I second Rob here.  Throwing an exception is the right thing to do.

A production quality code should not plunge ahead no matter what to
just report an exit code of zero.  It should try and detect as many
"wrong" situations as possible, and stop there.  So that a real
solution can be implemented.  Warnings are not very useful.  We have
no idea how many and which warnings our millions of production jobs
printed out.  But we did investigate job failures.

Andrei

PR #192, opened 2020-05-10
fix gen directory in pgit

rlcee:
Also skip gen directory when analyzing dependencies for pgit warnings


kutschke:
@FNALbuild run build tests

PR #195, opened 2020-05-14
removed deprecated class TrackSeed

gianipez:


kutschke:
Giani: I believe that this will not affect nightly validation because the validation does not look at objects of this type.  Is that correct?

kutschke:
@FNALBuild run build tests

gianipez:
Correct. There are no reco modules using that data product anymore.

PR #196, opened 2020-05-18
Changes to CRV part of DAQ and analysis tree

ehrlich-uva:
Changes to CRV part of DAQ
-Added the CRV part to the ArtBinaryPacketsFromDigis module
-Updated the CrvDigisFromFragments module

Changes to CRV part of analysis tree
-Added more CRV branches to the analysis tree
-Changes mostly in CRVAnalysis


brownd1978:
@FNALbuild run build test

PR #204, opened 2020-05-29
Added dot to branch names with same struct

### LINKED ISSUE #186: Custom objects in TTrees can't be read in Python
The Python package [uproot](https://github.com/scikit-hep/uproot/blob/master/README.rst) is becoming the _de facto_ standard for converting TTrees into Numpy arrays or Pandas dataframes. Unfortunately, it is not able to read custom objects (and I think PyROOT also have this issue). 
In particular, I am referring to what happens e.g. in `ComboHitDiag` (but also other analyzers) where we store positions and directions as `XYZVec`. In my opinion, we should try to store this type of information as flat objects (for example three floats like `pos_x`, `pos_y` and `pos_z`) or fixed-size arrays (which can be read by uproot) in order to make TTrees as widely accessible as reasonably possible. 

Edit: in the past I opened an issue on the uproot github where the author confirmed that there is no easy way to read these objects https://github.com/scikit-hep/uproot/issues/418.

soleti:
Very small fix: these three branches had the same struct but didn't have the dot in the filename. This prevented them to be read correctly. This also partially fixes #186, since now I am able to access the `fCoordinates` attributes (`fX`, `fY`, `fZ`) with `uproot` (but not the methods e.g. `Theta()`, `Phi()`) .

### DIFF
--- a/TrkDiag/src/ComboHitDiag_module.cc
+++ b/TrkDiag/src/ComboHitDiag_module.cc
@@ -107,8 +107,8 @@ namespace mu2e
       _chdiag=tfs->make<TTree>("chdiag","combo hit diagnostics");
       _chdiag->Branch("evt",&_evt,"evt/I");  // add event id
       _chdiag->Branch("wdist",&_wdist,"wdist/F");
-      _chdiag->Branch("pos",&_pos);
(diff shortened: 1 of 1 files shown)


brownd1978:
@FNALbuild run build test

PR #205, opened 2020-06-08
Scons optimisation suggestions

ryuwd:
Some suggestions to reduce scons overhead. Most/all of these come from the [Scons wiki page](https://github.com/SCons/scons/wiki/GoFastButton).

#### benchmark (mu2egpvm05)
<details><summary>Read more</summary>
<p>

To test these, I ran `time scons` on a fully built, up-to-date build of Offline. There is no actual re-building to do, so all the time is spent scanning dependencies. 

Before recording these times, I ran `scons` a few times to make sure the overhead from reading from network disks, cvmfs, etc didn't slow things down.

##### head of master
```
time scons
scons: Reading SConscript files ...
scons: done reading SConscript files.
scons: Building targets ...
Building compilation database gen/compile_commands.json
scons: done building targets.

real	1m7.813s
user	1m1.551s
sys	0m1.999s
```
##### this branch after `git merge master` (i.e. no change to Offline code)
(The merge commit is not pushed)
```
time scons
scons: Reading SConscript files ...
scons: done reading SConscript files.
scons: Building targets ...
Building compilation database gen/compile_commands.json
scons: done building targets.

real	0m29.145s
user	0m26.707s
sys	0m0.967s
```

</p>
</details>

It's worth noting that changing the number of jobs **`-j N` only affects how many build commands are run in parallel**, the dependency scans+checks are not parallelised. (i.e. the time taken for `scons -j 8` should be roughly the same as for `scons`, given no build targets changed)

#### 1.  `env.Decider('MD5-timestamp')`

SCons uses MD5 file hashes to determine whether a file changed, which can be expensive over many files. `MD5-timestamp` only computes and compares a hash if the time-modified timestamp mismatches the cached timestamp.

#### 2. `SCons.Defaults.DefaultEnvironment(tools = []) `

Don't load any tools into the default environment. This is OK because the build is configured in a custom `Environment` object. The default environment is not used.

#### 3. `SetOption('implicit_cache', 1)`

This isn't the same as `--implicit-deps-unchanged`, and it's not as fast, but it is more accurate. 

There are some caveats worth noting which are best explained at the below link (see bullet point list)
https://scons.org/doc/0.97/HTML/scons-user/x933.html

I suspect the most likely case that would cause issues would be if the ups products changed version after someone merged into an old development branch, for example, and tried to re-build rather than cleanbuild. A simple fix is to run scons with the `--implicit-deps-unchanged` flag.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -70,7 +70,9 @@ env = Environment( CPPPATH = sch.cppPath(mu2eOpts),   # $ART_INC ...
 )
 
 # Make the Compilation DB generator available in the environment
-env.Tool('compilation_db')
+env.Tool('compilation_db', COMPILATIONDB_COMSTR="@")
+
+# Only re-compute an MD5 hash for a build target if the timestamp changed.
 env.Decider('MD5-timestamp')
 
 # Define and register the rule for building dictionaries.
--- a/SConstruct
+++ b/SConstruct
@@ -14,6 +14,10 @@ from mu2e_helper import mu2e_helper
 # this will contain global config info about the build
 mu2eOpts = {}
 
+# Intelligent caching of implicit dependencies
+SetOption('implicit_cache', 1)
+
+
 # add a mu2e debug print option like "--mu2ePrint=5"
 AddOption('--mu2ePrint', dest='mu2ePrint',
           type='int',nargs=1,default=1,
--- a/SConstruct
+++ b/SConstruct
@@ -70,7 +70,7 @@ env = Environment( CPPPATH = sch.cppPath(mu2eOpts),   # $ART_INC ...
 )
 
 # Make the Compilation DB generator available in the environment
-env.Tool('compilation_db', COMPILATIONDB_COMSTR="@")
+env.Tool('compilation_db', COMPILATIONDB_COMSTR=None)
 
 # Only re-compute an MD5 hash for a build target if the timestamp changed.
 env.Decider('MD5-timestamp')
--- a/SConstruct
+++ b/SConstruct
@@ -4,8 +4,8 @@
 import os, re, string, sys
 
 import SCons 
-
 SCons.Defaults.DefaultEnvironment(tools = []) 
+
 # Functions that do small tasks and build lists
 import sconstruct_helper as sch
 # handles how the input files are collected and the output files are named
@@ -13,10 +13,6 @@ from mu2e_helper import mu2e_helper
 
 # this will contain global config info about the build
 mu2eOpts = {}
-
-# Intelligent caching of implicit dependencies
-SetOption('implicit_cache', 1)
-
 
 # add a mu2e debug print option like "--mu2ePrint=5"
 AddOption('--mu2ePrint', dest='mu2ePrint',
--- a/SConstruct
+++ b/SConstruct
@@ -3,6 +3,9 @@
 #
 import os, re, string, sys
 
+import SCons 
+
+SCons.Defaults.DefaultEnvironment(tools = []) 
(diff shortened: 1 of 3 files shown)


kutschke:
@FNALbuild run build tests

rlcee:
Sorry for the delay in a review.  I think 1 is very low risk,
2 is no risk, but 3 has the failure mode that Ryun points out.
I think I would vote to not include 3 since we modify product 
dependencies regularly.

kutschke:
I was also tuned out for the collaboration meeting.

I agree that 1 and 2 look safe.

My first impression is that I don't want 3 to be a default but that experts may use it.   My reason is that I want the default behaviour to have the maximum safety features enabled.   This is the first build system that I have used in which cross-stitched builds were not major waste time.

Ryun:  as best I can tell, implicit_cache does not speed up an initial build.  But it will speed up subsequent rebuilds.    Is that right?

ryuwd:
I agree the 3rd option is better left off by default. I can make the commit now

> 
> Ryun: as best I can tell, implicit_cache does not speed up an initial build. But it will speed up subsequent rebuilds. Is that right?

Yes that's correct. That's why we don't see a speedup when clean building from scratch

kutschke:
Please make the commit.

While you are at it can you make the heartbeat message from the compilation db builder off by default.

Then I will merge.

Thanks.

On Jun 23, 2020, at 9:28 AM, ryuwd <notifications@github.com<mailto:notifications@github.com>> wrote:



I agree the 3rd option is better left off by default. I can make the commit now

Ryun: as best I can tell, implicit_cache does not speed up an initial build. But it will speed up subsequent rebuilds. Is that right?

Yes that's correct. That's why we don't see a speedup when clean building from scratch

kutschke:
@FNALbuild run build tests

kutschke:
The last fix does not get rid of the heartbeat message, it just changes it to "@".   I am happy if you keep the heartbeat message as an option for debugging.

ryuwd:
Changing it to `@` should tell the `SCons.Action` object not to print anything. Have you checked locally?

When I run
```
curl -L https://buildmaster.fnal.gov/buildmaster/job/GitHubPRTests/job/mu2e-offline-build-test//145/artifact/scons.log | grep compile_commands
```

I see that scons does generate the compilation DB but doesn't print a message.

kutschke:
I am building locally. I believe that I am up to date with your repo.  When I build I see:

time scons -j 36 --implicit-cache
scons: Reading SConscript files ...
scons: done reading SConscript files.
scons: Building targets ...
@
scons: done building targets.

I believe that I am up to date with the head of your repo

git log
commit 2594f4f1eb91b922a0916f325aff5dd3633ce835 (HEAD -> scons-optimisations, origin/scons-optimisations)
Author: Ryunosuke O'Neil <roneil@fnal.gov>
Date:   Tue Jun 23 15:43:37 2020 +0100

    comment, and silence the compilation DB generator.

kutschke:
You can also see it in the Jenkins log file that you mentioned above; there 3 lines:

JobConfig/ensemble/genEnsembleFcl.py
@
Command execution time: gen/compile_commands.json: 0.152635 seconds

ryuwd:
Apologies Rob, I had misunderstood the internals of scons.

This latest commit should silence the message properly now.

kutschke:
@FNALbuild run build tests

kutschke:
HI Ray,

  Ryun and I have converged and agree with your point of view.

  When you get a chance can you approve the PR and I will merge.

Rob

PR #206, opened 2020-06-15
Fix: Have MVATools constructor follow the config file lookup policy

AndrewEdmonds11:
This commit corrects something that was missing in the initial TrkQual infrastructure update.


brownd1978:
@FNALbuild run build test

PR #209, opened 2020-06-21
Fix tool path for satellite builds

ryuwd:


kutschke:
@FNALbuild run build tests

PR #212, opened 2020-06-27
Add source module to create ProtonBunchIntensity sequence

brownd1978:
There is still one issue, that the new ProtonBunchIntensitySummary object is not being put into the SubRun as it should.  I don't know if the problem is related with that (new) class or with the art source code.  That functionality is not essential for this pull.

### DIFF
--- a/Sources/test/PBISequence.fcl
+++ b/Sources/test/PBISequence.fcl
@@ -1,0 +1,28 @@
+#include "fcl/minimalMessageService.fcl"
+process_name: PBISequenceTest
+
+services: {
+  message : @local::default_message
+}
+
+source: {
+  module_type :  PBISequence 
+  fileNames: ["/pnfs/mu2e/persistent/users/brownd/SS_2020-05-23.dat"]
+  runNumber : 1
+}
+
+Output : {
+  module_type : RootOutput
+  outputCommands:   [ "drop *_*_*_*", "keep mu2e::ProtonBunchIntensity_*_*_*" ]
+  fileName : "SS_2020-05-23.art"
+}
+
+physics : {
+  producers : {}
+  filters : {}
+  analyzers : {}
+  TriggerPath : []
+  EndPath : [ Output ]
+}
+
+outputs : { Output : @local::Output}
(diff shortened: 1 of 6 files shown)


brownd1978:
Here's what's produced by ProtonBunchIntensityAnalyzer, to be compared with slide10 of doc 33344
[normal_PBIS.pdf](https://github.com/Mu2e/Offline/files/4839793/normal_PBIS.pdf)

brownd1978:
@FNALbuild run build test

Sources/test/PBISequence.fcl
```diff
@@ -0,0 +1,28 @@
...
+
+source: {
+  module_type :  PBISequence 
+  fileNames: ["/pnfs/mu2e/persistent/users/brownd/SS_2020-05-23.dat"]
+  runNumber : 1
+}
+
+Output : {
+  module_type : RootOutput
+  outputCommands:   [ "drop *_*_*_*", "keep mu2e::ProtonBunchIntensity_*_*_*" ]
```
kutschke, line 16: Please extend the keep command so that you keep the Subrun summary object.  For example:

  outputCommands:   [ "drop *_*_*_*",
                      "keep mu2e::ProtonBunchIntensity_*_*_*",
                      "keep mu2e::ProtonBunchIntensitySummary_*_*_*" ]

brownd1978:
@FNALbuild run build test

brownd1978:
Good catches!  The extra empty event is because the ifstream still returns 'good' after reading the last entry, I think because there's a final EOL character the << doesn't read.  I just moved the test to after reading the proton count.  I also replaced the files in /pnfs.

kutschke:
With these fixes it looks good to me and I will merge.   I have few things that we can follow-up on later.

About your comment in the code: "reconstitutes" is the way you spell "produces" in a source module.  I searched for "reconstitutes" in the art wiki - there is nothing useful for people who don't already know.  I will mention it to Kyle.

Online help does not work:

  mu2e --print-description PBISequence

Does not print the parameters of the detail class.   I will ask Kyle if we need to change:
   
    typedef art::Source<mu2e::PBISequenceDetail> PBISequence;

to get a module that is aware of a Config struct defined in the detail class.

I also checked what happens if I read two files sequentially.   We start new counter when we encounter the second file.  I like that as a default behavior.  In the future we may wish to add some configurability here - but no need to do that unless and until we need it.

PR #213, opened 2020-06-30
DB update

rlcee:
1. Richie and Ryun reported some queries were not working.  This is fixed by updating to https and the latest port numbers.   
2. By accident, the default DbService behavior was set to read a test calibration from the real database, but this had no effect because no tables were requested.  When Ryun started using alignment tables, he discovered this mistake.  I set the default  purpose to "EMPTY" which means expect all database tables to come from a database text file.

### DIFF
--- a/DbService/fcl/prolog.fcl
+++ b/DbService/fcl/prolog.fcl
@@ -1,9 +1,13 @@
 BEGIN_PROLOG
 
 DbEmpty : {
-   purpose :  "TRK_TEST"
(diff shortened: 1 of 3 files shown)


kutschke:
@FNALbuild run build tests

kutschke:
Just checking: we expect no changes to the nightly validation.  The only reason I ask is that I want to merge two PRs today and I will only do that if I expect both to make no change.

PR #214, opened 2020-07-01
preparing for a future use of VecGeom

resnegfk:
VecGeom will be enabled by ./buildopts --g4vg=on; requires special ups version of geant4 with vg qualifier, to be made routinely available in future
changes should have no impact in the default -g4vg=off case

### DIFF
--- a/Mu2eG4/src/SConscript
+++ b/Mu2eG4/src/SConscript
@@ -19,6 +19,7 @@ helper=mu2e_helper(env)
 
 g4inc    = os.environ['G4INCLUDE'] + '/Geant4'
 g4libdir = os.environ['G4LIB']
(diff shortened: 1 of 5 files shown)


kutschke:
@FNALbuild run build tests

kutschke:
I have looked at the changes.   I think two things are true - am I right:

1) All of the changes are behind "if" tests and are off by default
2) We expect no changes to nightly validation.

resnegfk:
> I have looked at the changes. I think two things are true - am I right:
> 
>     1. All of the changes are behind "if" tests and are off by default
> 
>     2. We expect no changes to nightly validation.

That is the intent of the changes

PR #215, opened 2020-07-10
Remove old FCL param from prolog

ryuwd:
In master, running my module with this prolog `#include`d will cause validated FCL errors on job startup. I forgot to remove this line when I was removing some obsolete FCL parameters from my module.


kutschke:
@FNALBuild run build tests

PR #220, opened 2020-07-18
Offline0620

oksuzian:
This pull request adds an option to save CRV pulses in TrkAna trees.
Both pulse reco objects and individual waveforms are saved in TrkAna branches:
* crvpulseinfo
* crvpulseinfomc
* crvwaveforminfo

This information will be used by the CRV team to study the sources of the CRV inefficiencies.
These additional branches are not filled by default:
physics.analyzers.TrkAnaNeg.AnalyzeCRVpulses : false

Screen Shot 2020-07-17 at 9 54 51 PM


brownd1978:
@FNALbuild run build test

TrkDiag/src/TrackAnalysisReco_module.cc
```diff
@@ -127,9 +129,13 @@ namespace mu2e {
       fhicl::Atom<std::string> simParticleLabel{Name("SimParticleLabel"), Comment("SimParticleLabel")};
       fhicl::Atom<std::string> mcTrajectoryLabel{Name("MCTrajectoryLabel"), Comment("MCTrajectoryLabel")};
       fhicl::Atom<double> crvPlaneY{Name("CrvPlaneY"),2751.485};  //y of center of the top layer of the CRV-T counters
+      fhicl::Table<SimParticleTimeOffset::Config> timeOffsets{ Name("TimeOffsets"), Comment("Time maps") };
+      fhicl::Atom<std::string> crvWaveformsModuleLabel{ Name("CrvWaveformsModuleLabel"), Comment("CrvWaveformsModuleLabel")};
+      fhicl::Atom<std::string> crvDigiModuleLabel{ Name("CrvDigiModuleLabel"), Comment("CrvDigiModuleLabel")};
       fhicl::Atom<bool> fillmc{Name("FillMCInfo"),true};
       fhicl::Atom<bool> pempty{Name("ProcessEmptyEvents"),false};
       fhicl::Atom<bool> crv{Name("AnalyzeCRV"),false};
+      fhicl::Atom<bool> crvpulses{Name("AnalyzeCRVpulses"),false};
```
AndrewEdmonds11, line 138: For consistency with the other fhicl parameters, I think I'd prefer this was "AnalyzeCRVPulses" with a capital P. This isn't a major issue though...

TrkDiag/src/TrackAnalysisReco_module.cc
```diff
@@ -243,7 +253,8 @@ namespace mu2e {
     art::EDAnalyzer(conf),
     _conf(conf()),
     _trigbitsh(0),
-    _infoMCStructHelper(conf().infoMCStructHelper())
+    _infoMCStructHelper(conf().infoMCStructHelper()),
+    _toff(conf().timeOffsets())
```
AndrewEdmonds11, line 257: We already set the time maps in the InfoMCStructHelper so I'd rather not have them defined twice. It looks like access to "_toff" is needed though. I would just add an accessor to InfoMCStructHelper and get the time maps from there.

oksuzian: Hi Andy,

Thank you for the comment. It's a reasonable request.
The issue I'm struggling with is that the time maps in the InfoMCStructHelper point to compressed objects:
TimeMaps : [ @sequence::RecoCompression.TimeMaps ]

I'd like to use uncompressed CRV data products to study the sources of the inefficiencies.
In this case, I have to provide uncompressed time maps.
Do you have a recommendation on how to 'properly' implement this?

   Best, Yuri.

AndrewEdmonds11: Hi Yuri,

You should be able to give the uncompressed time maps to InfoMCStructHelper and the rest of trkana should work. Let me know if that's not the case and I will rethink.

For getting the time maps from InfoMCStructHelper to pass to `FillCrvPulseInfoCollections`. I would add a function like this:

`const SimParticleTimeOffset& getTimeMaps() const { return _toff; }`

to InfoMCStructHelper.hh and then use `_infoMCStructHelper.getTimeMaps()` in the `FillCrvPulseInfoCollections` call.

Cheers,
Andy

oksuzian: Hi Andy,

I've tried exactly that. It throws an exception:
        ---- BADINPUTS BEGIN
          SimParticleTimeOffset::totalTimeOffset(): the primary (2741432267, 0) is not in an input map
I'll take another look

AndrewEdmonds11: For anyone interested, we resolved this by reconfiguring Yuri's reconstruction job to not compress any CrvDigiMCs for his study.

brownd1978:
@FNALbuild run build test

PR #222, opened 2020-07-29
art v3 06 02 port

kutschke:
Updated from art v3_05_01 to v3_06_02.  Several art bugs were discovered during the port and it required two bug fix art releases to converge.

The changes required by the port were:

1. Update usage of art::Selector to conform to the new interface; removed an unused data member in our custom selector.
2. New #include needed in source modules
3. Root no longer has -lSpectrum; remove from all link lists.
4. BOOST has declared many of the features that we use to be deprecated.  This generates compiler warnings which are turned into errors with our use of -Werror.  I added the build option '-DBOOST_BIND_GLOBAL_PLACEHOLDERS' which tells BOOST to allow the deprecated code without issuing  warning.  We have homework to upgrade our use of BOOST and remove this option.

I ran all of the tests in the nightly validation.  All of the do-they-run tests passed and 3 of the 4 sets of validation histograms compared exactly.  Validation/fcl/reco.fcl had some differences shown at:

        https://home.fnal.gov/~kutschke/Mu2e/compare/reco/result.html

This was traced to the new version of ROOT that came along with the new art.  The module CaloFilters/src/FilterEcalMVATrigger_module.cc loads some saved state for TMVA objects and uses the TMVA objects. The net effect is that one event failed the filter with the new art but passed the filter with the old art.  I removed this module from the two paths on which it is used and reran the tests.   The histograms all matched exactly.

Homework:

1. Update our use of boost to avoid deprecated features
2. Check if CaloFilters/src/FilterEcalMVATrigger_module.cc is working correctly; the tests above had poor coverage.

### DIFF
--- a/Analyses/src/ReadCaloDigi_module.cc
+++ b/Analyses/src/ReadCaloDigi_module.cc
@@ -777,7 +777,7 @@ namespace mu2e {
 
 
       CaloCrystalHit const& hit    = caloCrystalHits->at(ic);
-      CLHEP::Hep3Vector crystalPos = crystalPos = _calorimeter->geomUtil().mu2eToDiskFF(diskId,_calorimeter->crystal(crystalId).position());
+      CLHEP::Hep3Vector crystalPos = _calorimeter->geomUtil().mu2eToDiskFF(diskId,_calorimeter->crystal(crystalId).position());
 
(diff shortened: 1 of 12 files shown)


brownd1978:
@FNALbuild run build test

PR #223, opened 2020-08-05
Revert curl to previous version.

kutschke:
curl version v7_67_0 breaks the command line git command.

Revert to previous version, v7_64_1.


kutschke:
@FNALbuild run build tests

PR #227, opened 2020-08-16
updated MakeCrvRecoPulses

ehrlich-uva:
Changes speed up the CRV pulse fit

### DIFF
--- a/CRVResponse/inc/MakeCrvRecoPulses.hh
+++ b/CRVResponse/inc/MakeCrvRecoPulses.hh
@@ -2,6 +2,7 @@
 #define MakeCrvRecoPulses_h
 
 #include <vector>
+#include <TF1.h>
 
 namespace mu2eCrv
 {
@@ -29,11 +30,14 @@ class MakeCrvRecoPulses
   int          GetPeakBin(int pulse);
 
   private:
+  TF1                 _f;
   std::vector<int>    _PEs, _PEsPulseHeight;
   std::vector<double> _pulseTimes, _pulseHeights, _pulseBetas, _pulseFitChi2s;
   std::vector<double> _fitParams0, _fitParams1, _fitParams2, _t1s, _t2s;
   std::vector<double> _LEtimes, _LEfitChi2s;
   std::vector<int>    _peakBins;
+
+  static double Gumbel(double* xs, double* par);
 };
 
 }
--- a/CRVResponse/inc/MakeCrvRecoPulses.hh
+++ b/CRVResponse/inc/MakeCrvRecoPulses.hh
@@ -10,9 +10,12 @@ namespace mu2eCrv
 class MakeCrvRecoPulses
 {
   public:
-  MakeCrvRecoPulses();
+  MakeCrvRecoPulses(double minADCdifference, double defaultBeta, double minBeta, double maxBeta,
+                    double maxTimeDifference, double minPulseHeightRatio, double maxPulseHeightRatio,
+                    double LEtimeFactor);
   void         SetWaveform(const std::vector<unsigned int> &waveform, unsigned int startTDC, 
-                           double digitizationPeriod, double pedestal, double calibrationFactor, double calibrationFactorPulseHeight, bool darkNoise);
+                           double digitizationPeriod, double pedestal, double calibrationFactor, 
+                           double calibrationFactorPulseHeight);
   unsigned int GetNPulses();
   int          GetPEs(int pulse);
   int          GetPEsPulseHeight(int pulse);
@@ -30,14 +33,21 @@ class MakeCrvRecoPulses
   int          GetPeakBin(int pulse);
 
   private:
-  TF1                 _f;
+  MakeCrvRecoPulses();
+
+  TF1    _f;
+  double _minADCdifference;
+  double _defaultBeta;
+  double _minBeta, _maxBeta;
+  double _maxTimeDifference;
+  double _minPulseHeightRatio, _maxPulseHeightRatio;
+  double _LEtimeFactor;
+
   std::vector<int>    _PEs, _PEsPulseHeight;
   std::vector<double> _pulseTimes, _pulseHeights, _pulseBetas, _pulseFitChi2s;
   std::vector<double> _fitParams0, _fitParams1, _fitParams2, _t1s, _t2s;
   std::vector<double> _LEtimes, _LEfitChi2s;
   std::vector<int>    _peakBins;
-
-  static double Gumbel(double* xs, double* par);
 };
 
 }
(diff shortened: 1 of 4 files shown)


kutschke:
@FNALbuild run build tests

kutschke:
Please document the speed improvement.  Please also compare with tag v09_07_00, with and without these changes.

So I am looking for a table with 4 columns: ( with and without these changes ) x ( old and new art/root )

I want to untangle two things.  I want to understand the effect of the improvement in its own right.  I also want to understand if we understand the timing difference between the old and new versions of art/root or if we still have work to do to understand that.

CRVResponse/src/MakeCrvRecoPulses.cc
```diff
@@ -30,75 +36,106 @@ void MakeCrvRecoPulses::SetWaveform(const std::vector<unsigned int> &waveform, u
...
 
     //set the fit function
-    TF1 f("peakfitter","[0]*(TMath::Exp(-(x-[1])/[2]-TMath::Exp(-(x-[1])/[2])))");
-    f.SetParameter(0, (waveform[maxBin]-pedestal)*TMath::E());
-    f.SetParameter(1, (startTDC+maxBin)*digitizationPeriod);
-    f.SetParameter(2, darkNoise?12.6:19.0);
-    if(peaks[i].second) f.SetParameter(1, (startTDC+maxBin+0.5)*digitizationPeriod);
+    _f.SetParameter(0, (waveform[peakStartBin]-pedestal)*TMath::E());
+    _f.SetParameter(1, peakTime);
+    _f.SetParameter(2, darkNoise?12.6:19.0);
```
kutschke, line 103: Make these magic numbers symbolic constants.  

It would be better to get them from conditions ( or fcl depending exactly on their meaning, but probably conditions.)  You can do this change later.

CRVResponse/inc/MakeCrvRecoPulses.hh
```diff
@@ -29,11 +30,14 @@ class MakeCrvRecoPulses
...
 
   private:
+  TF1                 _f;
   std::vector<int>    _PEs, _PEsPulseHeight;
   std::vector<double> _pulseTimes, _pulseHeights, _pulseBetas, _pulseFitChi2s;
   std::vector<double> _fitParams0, _fitParams1, _fitParams2, _t1s, _t2s;
   std::vector<double> _LEtimes, _LEfitChi2s;
   std::vector<int>    _peakBins;
+
+  static double Gumbel(double* xs, double* par);
```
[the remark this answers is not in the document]
ehrlich-uva: Done

CRVResponse/src/MakeCrvRecoPulses.cc
```diff
@@ -1,16 +1,22 @@
...
-#include <TF1.h>
 #include <TGraph.h>
 #include <TMath.h>
 
 namespace mu2eCrv
 {
 
-MakeCrvRecoPulses::MakeCrvRecoPulses()  
+//MakeCrvRecoPulses::MakeCrvRecoPulses() : _f("peakfitter","[0]*(TMath::Exp(-(x-[1])/[2]-TMath::Exp(-(x-[1])/[2])))")
+MakeCrvRecoPulses::MakeCrvRecoPulses() : _f("peakfitter",Gumbel,0,0,3)
```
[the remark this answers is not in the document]
ehrlich-uva: Done

CRVResponse/src/MakeCrvRecoPulses.cc
ehrlich-uva: Done

ehrlich-uva:
TimeTracker (average time spent in module) comparison for CrvRecoPulsesFinder applied to /mu2e/data/users/kutschke/dig.mu2e.CeEndpoint-mix.MDC2020DEVa.001002_00000000.art
Offline v09_07_00 before code changes: 0.185s
Offline v09_07_00 after code changes: 0.066s
head of Offline before code changes: 0.522s
head of Offline after code changes: 0.239s

kutschke:
It's great that your improvments do such a good job in both versions of the code.  But we still have a significant open issue: why is the module still so much slower in art v3_06_02.  

Have you learned how to use the profiling tools that Kyle talked about?  Do you use root elsewhere in this module, either directly or in called routines?  Do you make histograms?  If so turn them off and see if that is the issue?

ehrlich-uva:
Root is not used anywhere else in this module anymore. I run a small test program which calls TGraph::Fit 100,000 times for the same waveform in both Root versions. I got 6.5s in Root 6.18/04 and 22.4s in Root 6.20/07. That's 3.4 times longer - similar to our previous observations.

#include <vector>
#include <TF1.h>
#include <TGraph.h>
double Gumbel(double* xs, double* par)
{
  double const x = xs[0];
  return par[0]*(TMath::Exp(-(x-par[1])/par[2]-TMath::Exp(-(x-par[1])/par[2])));
}
int main()
{
  std::vector<double> t;
  for(int i=0; i<9; i++) t.emplace_back(200.8+i*12.55);
  double v[9]={0,0,92,430,530,341,149,89,62};
  TGraph g(9, t.data(), v);
  TF1 f("peakfitter",Gumbel,0,0,3);
  f.SetParameter(0, 530*TMath::E());
  f.SetParameter(1, 251.0);
  f.SetParameter(2, 19.0);
  for(int i=0; i<100000; i++) g.Fit(&f,"NQS");
  return 0;
}

kutschke:
Great work.  I have forwarded your standalone test program to the root experts and cc'ed you on that email. 

We don't need to resolve this within this PR so I will give it one last look and decide if I can merge it tonight.

kutschke:
@FNALbuild run build tests

CRVResponse/src/CrvRecoPulsesFinder_module.cc
```diff
@@ -60,21 +59,50 @@ namespace mu2e
...
+    _maxTimeDifference(pset.get<double>("maxTimeDifference",20.0)), //20.0ns largest accepted differences between
+                                                                    //time of largest ADC value and fitted peak
+    _minPulseHeightRatio(pset.get<double>("minPulseHeightRatio",0.7)), //smallest accepted ratio between
+                                                                       //largest ADC value and fitted peak
+    _maxPulseHeightRatio(pset.get<double>("maxPulseHeightRatio",1.5)), //largest accepted ratio between
+                                                                       //largest ADC value and fitted peak
+    _LEtimeFactor(pset.get<double>("LEtimeFactor",0.985)), //time of leading edge is peakTime-LEtimeFactor*beta
+                                                           //e.g. 0.985 for a leading edge at 50% pulse height
+                                                           //e.g. 1.385 for a leading edge at 20% pulse height
+                                                           //e.g. 1.587 for a leading edge at 10% pulse height
```
kutschke, line 94: We have a policy that parameters that affect physics should not have defaults in code.  That is we require them to be present in fcl. Please make this change.

We are also asking everyone to change their modules to use validated fcl - so that typo's in parameter names will get caught at run time.

ehrlich-uva: Done

brownd1978:
@FNALbuild run build tests

brownd1978:
Please use a flag to mark failed reco pulses and overwrite any NANs.  Validation should also add a test of this flag.

PR #228, opened 2020-08-18
Geometry updates

michaelmackenzie:
Adding back in a poly absorber just upstream of the IFB window, and created a named file for the CRY 2020 campaign to use. 
Also took this opportunity to add in the WorldMap to the outer B-field maps, and update the PSAreaMap to Mau13 from Mau9.

Validated using Geant4 surface checking with the updated configuration, and the only code change (constructMBS.cc) doesn't affect any older geometries due to a flag defaulting to false (that isn't set to true in any files in Mu2eG4/geom/).

### DIFF
--- a/Mu2eG4/geom/muonBeamStop_v08.txt
+++ b/Mu2eG4/geom/muonBeamStop_v08.txt
@@ -1,0 +1,11 @@
+// Muon Beam Dump/Stop
+
+//turn on absorber for STM
+bool   mbs.CLV2.absorber.build          = true;
+string mbs.CLV2.absorber.MaterialName   = "Polyethylene0935";
+double mbs.CLV2.absorber.halflength     =  10.0;
+
+// This tells emacs to view this file in c++ mode.
+// Local Variables:
+// mode:c++
+// End:
--- a/Mu2eG4/geom/muonBeamStop_v08.txt
+++ b/Mu2eG4/geom/muonBeamStop_v08.txt
@@ -1,5 +1,7 @@
 // Muon Beam Dump/Stop
 
+#include "Mu2eG4/geom/muonBeamStop_v07.txt"
+
 //turn on absorber for STM
 bool   mbs.CLV2.absorber.build          = true;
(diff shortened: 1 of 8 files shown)


kutschke:
@FNALbuild run build tests

Mu2eG4/geom/muonBeamStop_v08.txt
```diff
@@ -0,0 +1,13 @@
+// Muon Beam Dump/Stop
+
+#include "Mu2eG4/geom/muonBeamStop_v07.txt"
+
+//turn on absorber for STM
```
resnegfk, line 5: Shouldn't mbs.Version be changed?

michaelmackenzie: I didn't increment the version number, as I typically do this when there's a significant enough change that the version needs to be used inside the code. I can increment the number here though and make sure there are no `version == 7` statements in the code though to ensure the version number matches the file version.

michaelmackenzie: I've updated the version numbers and one check in the code for the Saddles and MBS

resnegfk:
Consider replacing 0 with nullptr where appropriate (it would increase the code readability)
Prefer pre-increment over post-increment when the return value is not used
Consider using newer nest functions taking lookup token (or consider a special project to convert the code to use them;
  It would make it easier to control the geometry options/attributes in a more uniform way)
Consider replacing tabs with spaces and removing trailing spaces (or wait till we have a style checker)

Mu2eG4/src/constructSaddles.cc
```diff
@@ -314,7 +314,7 @@ namespace mu2e {
 	
 	} // end of for loop over saddles
     
-    } else { // end of version 1, 2.  At this time, other versions are not
+    } else { // end of version 1, 2, 3.  At this time, other versions are not
       // implemented.
```
resnegfk, line 318: The printout needs to be updated (add 3); Use G4cout, G4endl in code called by Geant4 such as this one

kutschke:
@FNALbuild run build tests

PR #229, opened 2020-08-18
TrkAna Updates

AndrewEdmonds11:
This pull request includes two new options, a new example, and two bug fixes. All options are off by default.

- new option: hit-level diagnostics for supplemental tracks
    - now diagLevel > 1 will turn on hit-level diagnostics for all track-types
    - hit-level diagnostics for individual track-types can be turned on/off with the `fillHits` option
 - new option: require a supplemental track to be in the event (`required`)
    - i.e. only write out events with both a candidate track and a supplemental track
 - new example: TrkAnaReco_BothCharges.fcl gives an example fhicl configuration to write out positive and negative tracks to the same tree
 - bug fix: `ProcessEmptyEvents` wasn't working as intended
 - bug fix: the input tag suffix that we use to identify collections (e.g. DeM) was being found in the middle of module labels

It looks like this will not merge automatically. Do I just merge in the head of master into my branch and resolve the conflicts myself, or is there a different way to manage this?


brownd1978:
Conflict resolution is described starting in step 10 of the githubworkflow wiki page:
https://mu2ewiki.fnal.gov/wiki/GitHubWorkflow#Developer_Workflow

brownd1978:
@FNALbuild run build test

PR #238, opened 2020-09-02
Cvs comment removal

kutschke:
Remove relic CVS and RCS metadata comments: Id/Author/Date/RCSfile .   Add missing newline to the end of a few files. No functional changes.   This touched about 930 files.

The only tests I have done are the build and g4test_03.fcl, which compares exactly.  I propose that we let nightly validation be the next comparison; I except exact matches.


kutschke:
@FNALbuild test

brownd1978:
Since this is a comment-only change, I think we can conduct a minimal review, and let the nightly check the results.

kutschke:
I agree

PR #243, opened 2020-09-14
Mixing Straw Gas Steps

AndrewEdmonds11:
This PR makes the necessary fcl changes to write StrawGasSteps out at the end of each individual background frame simulation, and to mix them in the mixing stage. StrawGasSteps were already added to the Mu2eProductMixer in PR#105. This PR would be a breaking change for *-mix jobs that are reading MDC2018 background frames, which contain StepPointMCs. If we still need this functionality, we can reprocess the MDC2018 files to create StrawGasSteps.

I have made this a draft pull request because I want to ask whether it is worth merging this branch with PR#231 and PR#234 to make sure all these related developments work together before merging into master.

### DIFF
--- a/JobConfig/beam/deuteron.fcl
+++ b/JobConfig/beam/deuteron.fcl
@@ -16,6 +16,7 @@ outputs: { @table::mu2e.outputs.g4s4Defs }
 
 physics.producers.generate.physics: {
     pdgId : 1000010020 # deuteron
+    genId : StoppedParticleReactionGun
     spectrumShape: ejectedProtons # same spectrum for protons and deuterons, as in the MECO note
     spectrumVariable: kineticEnergy
     nbins: 1000
--- a/JobConfig/beam/dio.fcl
+++ b/JobConfig/beam/dio.fcl
@@ -16,6 +16,7 @@ outputs: { @table::mu2e.outputs.g4s4Defs }
 
 physics.producers.generate.physics: {
     pdgId : 11
+    genId : StoppedParticleReactionGun
     spectrumShape: Czarnecki
     spectrumVariable: totalEnergy
     elow: 1.  // MeV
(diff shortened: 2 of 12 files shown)


brownd1978:
@FNALbuild run build test

brownd1978:
I am happy to have this as a separate PR.  Should 231 and/or 234 be closed?

brownd1978:
Andy, from what I understand of this PR we expect ceSimReco to fail, but not g4test_03mt or transportOnly.  Can you have a look at those?  Also I find I cannot follow the 'Details' list of the build test, I guess they are kept on a protected server.

kutschke:
I believe that you need two things to be able to see that page: VPN and a certificate.  I have it working on my work laptop but not on my home iMac.  Let me see what the difference is.

ryuwd:
> Also I find I cannot follow the 'Details' list of the build test, I guess they are kept on a protected server.


The links need Fermilab VPN or access from inside the fermilab network. If you have kerberos and SSH set up for `mu2egpvm` machines I prefer to inspect logs using `curl` and `less` on mu2egpvm
```bash

curl -L <paste log file link> | less

```

AndrewEdmonds11:
The failures in all 3 tests were because I had separated makeSGS and makeSD. I have fixed the fcl cinfigurations and the test should now pass.

AndrewEdmonds11:
The last remaining task for this PR is to make sure it now passes the FNALbuild tests. Once that's done, I will close this PR and merge it with PRs #231 and #234. Here's the run down of this PR:

 - changed fcl configuration to split StrawGasStep creation from the beginning of the digi stage to the end of the simulation stage
 - this is a breaking change for jobs that use MDC2018 background frames (i.e. any mixing job)

brownd1978:
@FNALbuild run build test

PR #244, opened 2020-09-14
TEveMu2e Event Display 

sophiemiddleton:
This PR is to put in the first iteration of our TEve based Event Display for Mu2e. The display can be used for several Mu2e Data Products and we encourage feedback from anyone using it so we can build the code into something everyone can use. In parallel, we will now begin adapting the code for Online use.

This code was written by both me - Sophie Middleton (@sophiemiddleton) and Aditi Venkatesh (https://github.com/avenkatesh20/).

Please contact me (sophie) with suggestions/comments.


brownd1978:
@FNALbuild run build test

TEveMu2e.sh
```diff
@@ -0,0 +1,65 @@
...
+	fi
+
+	if [ "$var" == 'cosmictracks' ]; then
+	    sed -i 's/addCosmicSeedFit.*/addCosmicSeedFit : true/' TEveEventDisplay/fcl/prolog.fcl
+	fi
+
+	if [ "$var" == 'mctraj' ]; then
+	    sed -i 's/addMCTraj.*/addMCTraj : true/' TEveEventDisplay/fcl/prolog.fcl
+	fi
+done
```
ryuwd, line 63: Perhaps you could generate a new FCL file based on the `prolog.fcl` in the current working directory, based on the prolog.fcl, then run on that, rather than modifying FCL inside Offline.

A problem I see with this is that this would only work if the user's current working directory was the Offline folder. Also if the user is sourcing a release from CVMFS the `sed` commands would fail because the mount is readonly

ryuwd:
```bash

cp ${MU2E_BASE_RELEASE}/TEveEventDisplay/fcl/prolog.fcl EDcfg.fcl

...
sed -i '...' EDcfg.fcl
...

mu2e -c EDcfg.fcl

```

ryuwd: If `${MU2E_BASE_RELEASE}/bin` is added to the `PATH` when a user runs `source setup.sh`, then:
- moving this script to `bin/`
- renaming the file to `TEveMu2e`
- running `chmod +x TEveMu2e`
- then git add && committing the file

the script should be accessible from any working directory in a mu2e environment by running
```
TEveMu2e <options>
```

sophiemiddleton: Thanks Ryun, yes that's true. I had some similar ideas - this part of the infastructure is very much in the development stages and SED is probably a better bet!

sophiemiddleton: Ah ok, I see what you mean now - I'll investigate and get this part of the code in better shape

sophiemiddleton: One thing we could so is maybe remove the TEveMu2e.sh from this PR and develop a better interface for the next update we do. This was something aditi made and I thought it would be useful but now thinking about it there are better ways to do this. I'll see what the reviewers say but we can pull it out for now and add a new interface when we do our next PR

sophiemiddleton: I have enacted this version- tested and it worked will await full review to see whether its OK with Rob/Dave

kutschke:
@FNALbuild run build tests

kutschke:
I have several comments on this approach.  I will write each as a separate comment.

The first is along the lines of Pasha's comments yesterday.  I have seen many people go this route road before and it has traps.   It works OK so long as the number of supported command line options remains small and so long as you do not modify anything that affects physics results.   As you start to support more and more options, people will write really long command lines and this makes it hard to reproduce what they did; if they did not write an output file then their command line options are simply lost. If they did write an output file then it's possible to tedious to recover the information from the parameter sets stored in the output file.  Actually I am not 100% certain of that; fi the module did not write any data products I think that the parameter set for that module might be pruned from the output file.

Once the command line gets really long, or when people start to worry about documenting what they did, then they write a shell script and they run the shell script.  So you are back to editing a file, which is what you were trying to avoid but now there is an extra layer.

Continued ...

kutschke:
Comment 2.  

I have not looked too carefully at your code but you seem to be editing the fcl file, in place.  I most strongly recommend that you not do this; trivial formatting changes may break your scripts; if the line you are looking for is #included into the fcl, file then you will not find it.    If you don't use #include, it's very, very hard to avoid multiple points of maintenance.

The preferred method is to write a new fcl file that #includes the base fcl file and appends overriding definitions.  See, for example,
  
    JobConfig/mixing/NoPrimaryPBISequence.fcl

This job takes a separately maintained, and very complex, workflow and makes a small number of changes.  Using this method the changes are human visible, not buried inside the deeply nested include hierarchy.





 strongly recommend that write

kutschke:
Comment 3:

You have a lot of options that I guess are mutually exclusive, for example -3DOnly and -2Dand3D, or -DSOnly or -DSandCRV . I should make this manifest in the grammar of the command line, for example:

   --Dim=3DOnly or --Dim=2Dand3D

You also, hits, clusters, tracks etc are really part one one option. For example:

   --elements="(hits, clusters,tracks)"  

The idea is to make the grammar reinforce the logical organization of the information.

Bash has a powerful getopts package for parsing command line options that would greatly simplify the implementation of this sort of thing.  

That's the end of the comments.

I should have said that the beginning that the event display itself looks great and that's the most important thing.

TEveMu2e.sh
ryuwd: I realise now after my comments that `prolog.fcl`, the file that is edited in place, is actually included by another `fcl` file which is actually passed to `mu2e`. 

I agree with Rob that overriding the relevant fcl parameters makes more sense than editing them in place further up an include chain.

I made a similar command line utility while I was developing tracker alignment. It generates new FCL files, rather than editing existing files so there's sufficient bookkeeping.

sophiemiddleton:
Hi @kutschke 

It seems all the comments are about the command line interface. This is an optional thing. Should we just remove it from this PR?

Thanks.
Sophie

sophiemiddleton:
I can also remove the CallerExample directory that is just for people to see use cases but there is not need for it

sophiemiddleton:
Did anyone have any further issues with this PR?

kutschke:
@FNALbuild run build tests

kutschke:
What is the purpose of fix.gdml?  Will the code not work on a standard cdml file?  If so we should either fix the upstream problem or add an issue that it needs to be fixed.  In the long run a I don't think that a gdml file belongs here.

sophiemiddleton:
> What is the purpose of fix.gdml? Will the code not work on a standard cdml file? If so we should either fix the upstream problem or add an issue that it needs to be fixed. In the long run a I don't think that a gdml file belongs here.

Hi Rob,

This came out of something that we discussed way back when we first started - do you recall we had some bug in the mu2e gdml and you said you helped fix it and said that you would contact the responsible person to get a full fix implemented in Offline - I didnt follow up on this

sophiemiddleton:
I will begin switching to use toDetector instead of our custom transform.

sophiemiddleton:
Ok, I have switched the transforms to use the GeomService with the exception of the caloclusters which required there own transform still (center of tracker when use toMu2e from DetectorSystem). Let me know if there is another transfrom in existance which we can re-use there.

kutschke:
@FNALbuild run build tests

kutschke:
There is still some work that needs to be done with this PR but it's important to get it in front of people sooner rather than later.  I will open some issues to track the outstanding work.  If the comment in the code feature still works after the PR is approved, I will use that too.

sophiemiddleton:
Thanks Rob - we have an issue tracker here: https://github.com/sophiemiddleton/EventDisplayDevelopment/issues or you could do it in the main Mu2e one and assign it to us - either way we are pleased to see this getting done!

kutschke:
Hi Sophie,

  I  will use this one.  Thanks for telling me about it.

Rob

On 9/21/20, 5:10 PM, "Sophie Middleton" <notifications@github.com<mailto:notifications@github.com>> wrote:


Thanks Rob - we have an issue tracker here: https://github.com/sophiemiddleton/EventDisplayDevelopment/issues<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_sophiemiddleton_EventDisplayDevelopment_issues&d=DwMCaQ&c=gRgGjJ3BkIsb5y6s49QqsA&r=Zi5IH-k2_JiK_Wipiv3pwwSXO9UNaHiLP8sisfPz__k&m=4kjnXvil4i2Z0esedSsjSVa00o0hMSvYVsRHQTUGgHY&s=_gDJwFbrde8R5xTQiyCOnWhHtf_4hGZeOQR_BOitt_c&e=> or you could do it in the main Mu2e one and assign it to us - either way we are pleased to see this getting done!

PR #245, opened 2020-09-15
(CI test) upgraded to python 3

ryuwd:
Much to my surprise, the CI scripts were not being run in python 3, but in python 2.7 (which is not very stable in the long-term) and were being run with old versions of github api libraries

Should be fixed soon, just testing it out here


ryuwd:
@FNALbuild test

ryuwd:
@FNALbuild run code checks

ryuwd:
Looks good

PR #249, opened 2020-09-16
Define which art jobs to run in a build test inside a versioned config file

ryuwd:
This is mostly to help facilitate branches like `Mu2eII_SM21` which may use other FCL files to test or validate a version of Offline.

Generally speaking though this makes it much simpler to configure which FCL files to run during a build test, in the case that new validation job configurations are introduced or retired.

### DIFF
--- a/.build-tests.sh
+++ b/.build-tests.sh
@@ -1,0 +1,29 @@
+
+# This configuration file is sourced from 
+# https://github.com/Mu2e/codetools/blob/master/bin/github/jenkins_tests/mu2e-offline-build-test/build.sh
+
+# These environment variables influence which art jobs which must return status 0 for a build test to pass.
+# For each array index, the elements from each of the 3 arrays 
+# are combined as below:
+#  mu2e -n "${NEVTS_TJ[$i]}" -c "${FCLFILES[$i]}" >> "${JOBNAMES[$i]}.log" 2>&1
+# where $i is the index corresponding to an element of each of the 3 arrays.
+
+# an easily identifiable, human-recognisable label for the art job being run
+declare -a JOBNAMES=( \
+  "ceSimReco" \
+  "g4test_03MT" \
+  "transportOnly" \
+  "PS" \
+  "g4study" \
+  "cosmicSimReco")
+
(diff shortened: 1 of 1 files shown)


ryuwd:
If the location is not suitable I'm happy to move it to another folder.

rlcee:
Why does this have to be in Offline instead of codetools?
Without knowing the answer.. I'd suggest putting it in scripts/build/bin
We have a long-term plan to eliminate bin

kutschke:
I agree with Ray's comment:  the choice of which tests to run does not belong in Offline itself.  Do we already have a use case for new functionality needed for Mu2eII_SM21 or are you anticipating a possible future need?

ryuwd:
The basic argument is there's no control over which version of `codetools` is used to test at a given commit in Offline, as the latest version will always be checked out.

As an example, if more validation / test FCLs are added to Offline, or perhaps even a unit testing system to increase test coverage, the commands to run them would be added to `codetools`. Those changes would be compatible with the latest commits in `master` , but not `MDC2018`, `Mu2eII_SM21` branches, or older versions of offline.

But on second thought I'm not sure this PR will address that. Will close.

kutschke:
When we do a release that we want to test carefully we have other procedures that run much longer jobs and do more work to validate the output.  At this time I am not very concerned with knowing in detail which tests were active on which night in the past.   I think it would be good to make sure that the relevant information is recorded in the detailed longs on Jenkins.  But I don’t think that the information is useful after those logs have been deleted.


On 9/17/20, 11:16 AM, "ryuwd" <notifications@github.com<mailto:notifications@github.com>> wrote:


The basic argument is there's no control over which version of codetools is used to test at a given commit in Offline, as the latest version will always be checked out.

As an example, if more validation / test FCLs are added to Offline, or perhaps even a unit testing system to increase test coverage, the commands to run them would be added to codetools. Those changes would be compatible with the latest commits in master , but not MDC2018, Mu2eII_SM21 branches, or older versions of offline.

But on second thought I'm not sure this PR will address that. Will close.

ryuwd:
Test ROOT overlap check.

@FNALbuild test

PR #250, opened 2020-09-17
Mu2eII_SM21: Add Mu2e-II era target and remove pbar absorbers

michaelmackenzie:
Mu2eII_SM21:
Add a Mu2e-II era target configuration (conveyor), with a placeholder for a second target type (rotating).
Also add a flag to remove pbar absorber elements, which are not needed for 800 MeV proton beams.
The target implementation is preliminary, but gives a starting point for understanding POT in the PIP-II era.
Updated geometry is by default on in Mu2eG4/geom/geom_common_current.txt (which includes Mu2eG4/geom/geom_Mu2eII_2020.txt).

Validation run:
ROOT overlap checker on gdml file, G4 overlap checker

### DIFF
--- a/GeometryService/inc/GeometryService.hh
+++ b/GeometryService/inc/GeometryService.hh
@@ -34,6 +34,8 @@ namespace mu2e {
   class Mu2eG4Study;
   class Mu2eHall;
   class G4GeometryOptions;
+  class ProductionTarget;
+  class ProductionTargetMu2eII;
 
   class GeometryService {
 public:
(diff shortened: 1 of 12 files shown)


oksuzian:
@FNALbuild run build test

sophiemiddleton (approved):
All looks good to me

oksuzian (approved):
Looks good. An impressive amount of work on short notice.

resnegfk (approved):
Have very little time to look at it now. Seems fine though.

PR #257, opened 2020-09-25
Changed geometry file to geom_common_current.txt

kutschke:
The standard overlap check fcl file was using geom_common.txt.  The nightly job had code to swap this to geom_common_current.txt.  After discussion with Ray, Krzysztof and Ryun, we decided to change the fcl file to use geom_common_current.txt  .  This will also be used by the overlap check in the CI tests.

### DIFF
--- a/Mu2eG4/test/geom_SurfaceCheck.txt
+++ b/Mu2eG4/test/geom_SurfaceCheck.txt
@@ -3,7 +3,7 @@
 //
 //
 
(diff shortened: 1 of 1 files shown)


kutschke:
@FNALbuild run build tests

ryuwd:
Thanks Rob!

rlcee (approved):
As expected, please proceed

brownd1978 (approved):

PR #259, opened 2020-10-02
Added back in tracker digitization deadtime, adjusted to 400 ns

bonventre:
Current sims only include TDC deadtime and not the time required to fully digitize a tracker hit. This adds back in the digital deadtime of 400 ns, which causes an approximately 3% loss of hits for CeEndpointMix (both signal and background hits), and an approximately 3% loss of efficiency for signal tracks in the same sample.


kutschke:
@FNALbuild run build tests

brownd1978 (approved):

PR #260, opened 2020-10-03
Pbt

brownd1978:
Add module to reconstructed proton bunch time from early tracker hits, and a diagnostic module to test it.  Put PBT reconstruction into the Tracker hit preparation sequences.  Replace use of MC truth EWM offset in TrkHitReco with reconstruted proton bunch time.  Update TrkHitReco to validated fhicl.


brownd1978:
@FNALbuild run build test

kutschke:
I see that there are matching changes required in prolog.fcl and prolog_trigger.fcl .  Does it make sense to reorganize the prologs so that one is a delta on the other?  Or that both are deltas on an original?   Or are there good reasons to keep two points of maintenance?

As I read the code the behaviour of StrawHitReco_module will change because the time Offset comes from the data, not the MC truth value of the event window marker.    I think you said on a separate thread that this will not change the nightly tests - what I am missing here?

kutschke:
@FNALbulid run build tests

brownd1978:
@FNALbuild run build test

kutschke:
Are you finished with this iteration?  If so I think it's ready to merge.

PR #264, opened 2020-10-06
Corsika campaign

soleti:
This updated campaign configuration includes:
- A concatenation stage after `resampler_lo`
- Use of Offline tagged version v09_09_00
- Updated CORSIKA version
- Login information to run in production mode as `mu2epro`

Also, we keep the `CosmicLivetime` data product in `JobConfig/cosmic/cosmic_s1_dsstops.fcl` containing the amount of simulated livetime.

### DIFF
--- a/CampaignConfig/corsika.cfg
+++ b/CampaignConfig/corsika.cfg
@@ -20,13 +20,19 @@ streamname = only
 fcl_list = override_me
 stage_name = override_me
 project_name = corsika_dsstop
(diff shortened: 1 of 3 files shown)


kutschke (approved):
The fcl changes look good to me.  Thanks for zapping the trailing white space

I don't think that any of us can meaningfully review the .cfg and .ini files.

 I will approve and merge.

PR #273, opened 2020-11-04
Rpc

sophiemiddleton:
Mu2eII_SM21: RPC Job config files + TTree Analysis


EventGenerator/src/StoppedParticleReactionGun_module.cc
```diff
@@ -136,7 +141,12 @@ namespace mu2e {
       _hPdgId  = tfs->make<TH1F>("hPdgId" , "PDG ID"      ,  500,  -250, 250);
```
kutschke, line 141: In the early days we often made histograms/TTrees within producer modules.  After some experience we decided that, if all of the information being added to the histograms/TTrees was available in the data products made by the module, then we recommend removing the histograms/TTrees from the producer module and making a separate analyzer module to make the histograms/TTrees.   We have started to move existing histograms/TTrees to separate modules as targets of opportunity.  I would like you to do that here.

The only major exception to this is if there is transient information within the module that never makes it into a data product and you need to monitor the transient information.  In that case be sure that you can complete disable creating and filling the histograms with a fcl parameter.

JobConfig/pions/PionsStage1_2020.fcl
```diff
@@ -0,0 +1,146 @@
...
+    TFileService           : { fileName : "nts.owner.pions-g4s1.version.sequencer.root" }
+    RandomNumberGenerator  : {defaultEngineKind: "MixMaxRng"}
+    GeometryService        : { inputFile      : "JobConfig/common/geom_baseline.txt"    }
+   
+    ConditionsService      : { conditionsfile : "Mu2eG4/test/conditions_01.txt"      }
+    GlobalConstantsService : { inputFile      : "Mu2eG4/test/globalConstants_01.txt" }
+    G4Helper               : {}
+    SeedService            : @local::automaticSeeds
+}
+
```
kutschke, line 38: Please use the official definitions of the standard services for Sim or SimReco as appropriate; then override anything that you need to.  If you don't do this then, when the standard changes you need to edit every file that you just made.

JobConfig/pions/PionsStage1_2020.fcl
```diff
@@ -0,0 +1,146 @@
...
+                    { type: inVolume pars: [ HallAir ] },
+                    {
+                        type  : inVolume
+                        pars  : [ TS3Vacuum, TS4Vacuum, TS5Vacuum, DS1Vacuum, DS2Vacuum, DS3Vacuum ]
+                        write : vacuuas1
+                    }
+                ]
+            }
+        } // g4run
+
```
kutschke, line 82: As with the services, please use the standard definitions for modules, or even better for groups of modules.  Again the purposes is to minimize the points of maintenance in the future.

JobConfig/pions/PionsStage1_2020.fcl
```diff
@@ -0,0 +1,146 @@
...
+        g4filter : {
+            module_type               : FilterG4Out
+            mainHitInputs             : [ "g4run:vacuuas1" ]
+            extraHitInputs            : [ "g4run:virtualdetector" ]
+            vetoDaughters             : []
+        }
+    } // filters
+
+    trigFilter    : [ generate, genCounter, g4run, g4filter, compressPV ]
+    trigger_paths : [ trigFilter ]
```
kutschke, line 113: Please check with Dave to learn if compressPV is still part of the standard workflows.  If it is not, then you should remove it unless you have a good reason to keep it.

kutschke (changes requested):

kutschke:
Sophie: can you suggest anyone else who should review this?

kutschke:
@FNALbuild run build tests

EventGenerator/src/StoppedParticleReactionGun_module.cc
sophiemiddleton: Hi Rob, We wanted the generated momentum spectra pre any filtering and reco conditions. Its the same information whcih is stored in the TH1F's which come out here but storing a TTree means its portable to python for anlaysis

sophiemiddleton (commented):
Does anyone know why the build failed? These were just FCL files and a few small additons to C++ code

kutschke:
There is a known, but intermittent and not understood, problem with g4test_03MT .  It is safe to ignore it.  I have kept it in the tests to accumulate some experience with how often it fails.  It is now happening with great regularity so I will ask that it be taken out.

JobConfig/pions/PionsStage1_2020.fcl
kutschke: Look, for example at: JobConfig/mixing/NoPrimary.fcl

The services should be defined like:
services: @local::Services.Sim

with overrides for the TFileService file and geometry file if necessary.

The configuration of g4 should be:

producers: {
  g4run : @local::g4run
}

with overrides as needed.  Note that the module label "g4un" has an important meaning since downstream code uses it to find their data products.

For most of the other modules you use there are predefined standard definitions.  Search the prolog.fcl file to find them.  



Also look at JobConfig/mixing/NoPrimaryPBISequence.fcl   . Note that it says to do the same job as the NoPrimary.fcl one but with a small list of changes.  

This reason for the pattern of deep includes is that it is the only solution to minimizing multiple points of maintenance when a fcl parameter set need so change.  The disadvantage of deep includes is that it makes it more difficult to understand what the fcl file is doing; we have tools to help with this.  Do you know how to use:

 mu2e --debug-config --annotate
 mu2e --config-summary 

JobConfig has files with a mix of vintages.   The older files still have practices that we have long improved on.  We have been updating them as we need to use them again - the cost in people time of systematically changing everything, and testing it, is just too high.

EventGenerator/src/StoppedParticleReactionGun_module.cc
kutschke: I looked at the module.  The histograms come *after* the event.put.  So if you move them to an analyzer module and add that module to one of the end paths they will have exactly the same content that they currently have.  Analyzer modules see every event, even the ones that fail some or all of the filters.  If you want an analyzer module to see events that pass a certain filter, or a combination of filters, you can add a SelectEvents parameter to the analyzer's parameter set to choose what events to see; art looks after the selection.

kutschke:
Hi Sophie,

  I think that I have answered your outstanding questions.  Have I missed any?  

  As I said at the beginning we are moving histograms out of producer modules when we can - you just happened to be the person who touched this one first.

Rob

kutschke:
I just remembered that his pull is to Mu2eII_SM21.  This code will never be merged back into master so I don't are about updating the histograms/TTree to their own module.

So I think this is ready to merge.  Do have other planned work before the merge?

sophiemiddleton:
Hi Rob, yes I agree with this sentiment, although I can make an analyzer for future Mu2e/Offline when I have some spare time again. thanks, Sophie

kutschke:
Hi Sophie,

  I merged the PR.   We are make the change as modules are touched for other reasons.  So there is no need to do it as a one off.   But thanks for asking.

Rob


On 11/6/20, 2:00 PM, "Sophie Middleton" <notifications@github.com<mailto:notifications@github.com>> wrote:


Hi Rob, yes I agree with this sentiment, although I can make an analyzer for future Mu2e/Offline when I have some spare time again. thanks, Sophie

PR #274, opened 2020-11-05
flag to turn off csv cache

rlcee:
Finish installing the flag to turn off saving the csv version of tables.  This saves 30-40% of memory or so.  Run with the csv save turned off in art, but on in bins like dbTool where is it most likely needed to write out tables as they are in the database.


brownd1978 (approved):

kutschke (approved):

kutschke:
@FNALbuild run build tests

PR #277, opened 2020-11-06
AnalysisConditions MVACatalog Fix

AndrewEdmonds11:
Hi Everyone,

@macndev was having some jobs fail on the grid and, with @rlcee, worked out that the jobs were crashing at ~MVAEntry(). The crash was not occurring locally. 

It wasn't entirely clear why the jobs were crashing here but there is some dynamically-allocated memory in the code. I've removed all these new/delete/reserve calls and @macndev has successfully run jobs on the grid with these changes so this issue seems to be resolved. This is in commit 4790350.

I've also taken the opportunity to remove problematic whitespace. This is in a separate commit (b3dbf1a) so hopefully it's easier to review the actual changes.

Thanks,
Andy

### DIFF
--- a/AnalysisConditions/inc/MVACatalog.hh
+++ b/AnalysisConditions/inc/MVACatalog.hh
@@ -16,7 +16,7 @@ namespace mu2e {
 
   template <class T>
   struct MVAEntry {
-    MVAEntry(std::string trainName, std::string xmlFileName, bool calibrated = false) : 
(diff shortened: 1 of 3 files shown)


brownd1978 (approved):

kutschke (approved):
This is a good cleanup.  It defeats  the ROOT and G4 anti-training we all got.

kutschke:
@FNALbuild run build tests

PR #279, opened 2020-11-11
Using new functionality to change G4 MscEnergyLimit

resnegfk:
Using a new way to change the transition energy between multiple scattering (msc) models.
This makes a lot of old code unnecessary and removes the need to maintain it.
Plus minimal changes removing very obsolete code (QGSP.hh) or changing to a more appropriate/consistent configuration approach in case of BirksConst.
None of the changes should change the current default behavior, so one should not expect any of the validation plots to change.


kutschke:
Krzysztof:  Do you have validation plots that we can look at? If not, we can make them.   Please also prepare a short talk for an upcoming Wednesday meeting.

resnegfk:
Rob, Since the results of g4test_03 & transportOnly jobs do not change, validation plots should not change either. See my initial comment above. This PR enables a new way of adjusting the limit mentioned, but does not change it.

Mu2eG4/fcl/prolog.fcl
```diff
@@ -26,8 +26,8 @@ mu2eg4DefaultPhysics: {
     setMuHadLateralDisplacement: true // improves accuracy of boundary
                                       // crossing for muons and common charged hadrons
 
-    modifyEMOption4: false // to set 115 MeV transition energy for the msc models when using _EMZ
-    modifyEMOption0: false // to set 115 MeV transition energy for the msc models with "opt0"
+    // mscTransitionEnergy = 115. // uncomment to change the Geant4 deafult
```
gaponenko, line 29: The commented out example setting should use the colon, not the equal sign.

Mu2eG4/inc/Mu2eG4Config.hh
```diff
@@ -139,6 +137,8 @@ namespace mu2e {
                   "or one of \"PDG\", \"All\", \"None\" pre-defined settings.")
           };
 
+      OptionalDelegatedParameter mscTransitionEnergy {Name("mscTransitionEnergy")};
```
gaponenko, line 140: Is there a reason to use OptionalDelegatedParameter here?  It looks like mscTransitionEnergy can be a more explicit OptionalAtom<double>

Mu2eG4/src/physicsListDecider.cc
```diff
@@ -122,36 +115,30 @@ namespace mu2e{
...
 
-#if G4VERSION>4103
-    // for version 4105 it will need to be rplaced with
-    // emParams->SetMscEnergyLimit(115.0*CLHEP::MeV);
-    if ( phys.modifyEMOption4() && (name.find("_EMZ") != std::string::npos) ) {
-      tmpPL->RemovePhysics(("G4EmStandard_opt4"));
+#if G4VERSION>4104
+
+    // changing msc model transition energy if requested
+    double mscTransitionEnergyValue;
```
gaponenko, line 121: Please initialize mscTransitionEnergyValue.

gaponenko (changes requested):
Hi Krzysztof,
This looks like a nice cleanup.   I have added a few line comments with suggested changes.
Andrei

Mu2eG4/fcl/prolog.fcl
resnegfk: done

Mu2eG4/inc/Mu2eG4Config.hh
resnegfk: I was looking for a way to take an action only if the keyword was present; This was the best way I could find. 
Maybe I missed it, but what I wanted, was to have an equivalent of has_key in Mu2eG4Config::Physics , but could not find it.

Mu2eG4/src/physicsListDecider.cc
resnegfk: I have mixed feelings about initializing something if it is not necessarily going to be used, or if the initial value can cause problems, so I added a scope to prevent its use outside of it;

kutschke:
@FNALbuild run build tests

Mu2eG4/src/physicsListDecider.cc
gaponenko: > I have mixed feelings about initializing something if it is not necessarily going to be used [...]

If it is really not used then the initializer value does not matter; just assign zero or max() as you did,  but I think it is good to have deterministic code behavior.   I remember hearing (a while ago) about getting floating point exceptions because occasionally an ununitialized floating var would get a signalling NaN as its content by chance.  The code logic was correct and the ununitialized value was not used  in computations, but still caused problems when objects containing uninitialized varialbles were copied.   I think it is a good practice to require all variables in our "normal" code to be initialized at the point of declaration, as many C++ guides suggest.   (There are exceptions to always initialize to gain some efficiency; I think the STL uses uninitialized memory internally.  But this is not for user level code.)

Mu2eG4/inc/Mu2eG4Config.hh
gaponenko: DelegatedParameter objects are for the cases when the FHICL content can not be represented as a C++ structure known at the compilation time.  For example our G4Cuts with their arbitrary nesting.    Or BirksConsts, where the keys in key:value pairs are arbitrary and can not be handled directly without either DelegatedParameter or modifying the legacy FHICL syntax.   In our case just declare mscTransitionEnergy as OptionalAtom of double.    There may be no exact equivalent for has_key, rather OptionalAtom behaves like get_if_present.   You will still need to declare

   double mscModelTransitionEnergy(std::numeric_limits<double>::max()); 

then the call

   phys.mscModelTransitionEnergy(mscModelTransitionEnergy)

returns a boolean: true means the key was present and the variable mscModelTransitionEnergy now has the value from FHICL.  False means it was not specified, the value is unchanged from before the call.  An example of OptionalAtom use is defaultMinPointDistance() in Mu2eG4/src/Mu2eG4TrajectoryControl.cc

resnegfk: Thanks Andrei. Done.
This is actually documented, but I missed it the other day. I put the links here for a future reference:
https://cdcvs.fnal.gov/redmine/projects/fhicl-cpp/wiki/Configuration_validation_and_fhiclcpp_types#Optional-parameters
https://cdcvs.fnal.gov/redmine/projects/fhicl-cpp/wiki/Fhiclcpp_types_in_detail#Optional-parameters-argument-types

kutschke:
@FNALbuild run build tests

gaponenko (approved):
Thank you, Krzysztof.

resnegfk:
@FNALbuild run build tests

kutschke:
@FNALbuild run build tests

kutschke:
@FNALbuild run build tests

PR #286, opened 2020-12-04
DAQ related bug fixes + improvement of the TrigResultsNavigator class + new Menus for onSpill,offSpill,extrPos

gianipez:
- some bug fixes in a module and config of the DAQ package
- TriggerResultsNavig now gets the list of the paths in the trigger_paths also in a job where the triggers are not run 
- added menu files created at compilation level for onSpill, offSpill, extrPos
- added UeP and UeM CosmicHelix trigger paths (included in the offSpill Menu)
- converted the CosmicSeedDeM/P to CosmicHelixDeM; the new sequences stop at the pattern_reco stage (no KalSeedFit anymore)
- added the StraightTrack trigger path in the extrPos, offSpill and allPaths menus


gianipez:
Hi folks, I didn't apply yet the change for setting the trigger_bits in the Menu files. I believe these files will be mostly used by myself and people running MC sim+reco together with the trigger reco. Incorporating the trigger_paths with a bit set with an additional one for the MC+reco path is possible, but we need to agree to a convention. For example, if we agree to run the MC production in two separate stages: 
1. MC sim + Offline reco
2. Trigger & ntupling

In this way, in the Menu files, I can set my trigger bits from 0 to N-1 without any problem. This solution although would impose us to use more disk storage I guess.

 In principle, I can also set the bits from 1 to N, and leave the first one "free", BUT I can imagine scenarios where maybe we want to run different paths with different MC selections...

scripts/build/python/sconstruct_helper.py
```diff
@@ -165,7 +165,7 @@ def sconscriptList(mu2eOpts):
     # If we are making a build for the trigger, do not build everything.
     if mu2eOpts["trigger"] == 'on':
         notNeeded = ["Mu2eG4/src/SConscript",
-                     "CRVResponse/src/SConscript",
+                     #"CRVResponse/src/SConscript",
```
kutschke, line 168: Why do we need to build CRVResponse for code to be shipped to the trigger.  The whole point of the notNeeded list is that we do not want the trigger builds to contain MC code and CRVResponse is a MC package.  Please let me know what you need from that package and we can discuss refactoring what lives in which directory.

kutschke (changes requested):
My apologies for the delay in getting to this. 

Please send a list of the names of all files that you modified in a text file so that I can manipulate the names programatically.  I want to get a systematic view of what's been touched.

Also, please suggest people who can review this PR.

scripts/build/python/sconstruct_helper.py
gianipez: Hi Rob, CRVRsponse house MC and Reco code, that's why I need it. I agree that it would be good if someone from the CRV group would implement the separation. I remember I put this comment in a MR a long time ago, but maybe it got ignored.

kutschke: Thanks Giani.  I will speak with Ralf and cc you - I agree that he should do the work.

Trigger/src/ReadTriggerInfo_module.cc
```diff
@@ -1264,6 +1301,10 @@ namespace mu2e {
 	indexMother = 80;
       }else if (pdgM == 211){ //case 8: pi plus
 	indexMother = 90;
+      }else if (pdgM == 11){ //case 9: electrons
+	indexMother = 100;
+      }else if (pdgM == -11){ //case 10: positrons
```
kutschke, line 1306: Please use an enum rather than than hard coded numbers.   Ask Dave if he supports an enum within BTrk and if it will be in the future KinKal.   If not, it is OK to use PDGCode::e_minus.  We moved PDGCode.hh out of MCDataProducts and into DataProducts so that it was no longer considered "MC" and could be used with data.  You should do this everywhere in the code base.

Trigger/src/ReadTriggerInfo_module.cc
```diff
@@ -83,14 +83,14 @@ namespace mu2e {
       kNTrigInfo     = 40,
       kNTrackTrig    = 20,
       kNTrackTrigVar = 50,
-      kNHelixTrig    = 10,
+      kNHelixTrig    = 40,
```
kutschke, line 86: When you change these numbers, do you need to make the corresponding changes in other parts of the code or in fcl or config files?  If so, what would it take to get a single point of maintenance?

TrkHitReco/src/StrawHitReco_module.cc
```diff
@@ -52,7 +52,7 @@ namespace mu2e {
       struct Config {
 	fhicl::Atom<int> diag{ Name("diagLevel"), Comment("Diag level"), 0};
 	fhicl::Atom<int> print{ Name("printLevel"), Comment("Print level"), 0};
-	fhicl::Atom<int> fittype { Name( "FitType"), Comment("Waveform Fit Type"), TrkHitReco::FitType::peakminuspedavg};
+	fhicl::Atom<int> fittype { Name( "FitType"), Comment("Waveform Fit Type"), TrkHitReco::FitType::firmwarepmp};
```
kutschke, line 55: If a parameter is normally provided by fcl, we do not want default values in the code.   Same comment in many places.

DAQ/src/StrawAndCaloDigisFromFragments_module.cc
```diff
@@ -452,7 +462,7 @@ void art::StrawAndCaloDigisFromFragments::analyze_calorimeter_(
         uint16_t apdID = hits[hitIdx].first.DIRACB >> 12;
 
         // FIXME: Can we match vector types here?
-        std::vector<int> caloHits(hits[hitIdx].second.size());
+        std::vector<int> caloHits;
         std::copy(hits[hitIdx].second.begin(), hits[hitIdx].second.end(), std::back_inserter(caloHits));
```
kutschke, line 466: Does the copy does an implicit reserve or do we need to do that ourselves?  I bet it does but I don't know.  Please check.   I agree that you don't need to default construct everything in the vector.

kutschke:
@FNALbuild run build tests

TrkHitReco/src/StrawHitReco_module.cc
gianipez: I'm ok in removing it. I only changed the value of the default previously set by @bonventre

kutschke: Please also check that the relevant fcl files do set it.

DAQ/src/StrawAndCaloDigisFromFragments_module.cc
gianipez: Thanks for checking this. `std::copy` doesn't call any reserve internally. I'll add it now.

Trigger/src/ReadTriggerInfo_module.cc
gianipez: These numbers cannot be part of the list of data-members of the class, because they define the size of the arrays of histogram I set in my `struct` instances. I usually change it when the Menu I test gets longer, which happens not that frequently.

Trigger/src/ReadTriggerInfo_module.cc
kutschke: I will also leave this to Dave to decide whether to do it now or to defer for a later PR

brownd1978: Synchronizing pdgcodes across Offline, BTrk and KinKal is a bigger issue.  Fortunately they all use the same values.  It's clear though this code should use PDGCode values, not hardcoded numbers.

TrkHitReco/src/StrawHitReco_module.cc
brownd1978: Lets defer any changes here.  We can create an issue that the default values should be re-evaluated.

brownd1978 (changes requested):
This should be good to go after fixing the PDGCodes issue.

Trigger/src/ReadTriggerInfo_module.cc
kutschke: I added the issue #292 and will mark it resolved.

TrkHitReco/src/StrawHitReco_module.cc
kutschke: Created issue #293

kutschke (approved):
I have made issues for the outstanding comments and will approve.

Trigger/src/ReadTriggerInfo_module.cc
gianipez: OK, no problem. Done.

gianipez:
@FNALbuild run build tests

gianipez:
Hi folks,

Thanks again for the comments. I integrated the required changes. 

Cheers,

Giani

brownd1978 (approved):

brownd1978:
Thanks Giani.  I want to check a few things about code merged last night, if that comes out as expected I will merge this today.

PR #295, opened 2020-12-11
s104at286, s104 tag updates and relevant header modifications

macndev:


kutschke:
Thanks Mackenzie.  The Jenkins tests auto launched.  They usually take about 25 minutes to come back.

PR #299, opened 2020-12-15
RN number control in G4MT

goodenou:
RNs seeds in G4MT are now handled by us. We use Crypto++ to create hashes for seeds based on art::event ID (Run#, Subrun#, Event#).


gaponenko:
@FNALbuild run build tests

gaponenko:
@FNALbuild run code checks

Mu2eG4/inc/Mu2eG4WorkerRunManager.hh
```diff
@@ -51,7 +51,7 @@ namespace mu2e {
     void initializeUserActions(const G4ThreeVector& origin_in_world);
     void initializeRun(art::Event* art_event);
     void processEvent(art::Event*);
-    G4Event* generateEvt(G4int i_event);
+    G4Event* generateEvt(std::string i_event, std::string i_subrun, std::string i_run);
```
gaponenko, line 54: The signature of generateEvt() strikes me as non-conventional.  If we pass event,run,subrun, I expect three integer arguments.  Or a const art::EventID reference.   Alternatively, if the function only needs a unique string, pass it a string.  But I see in the implementation that i_events is converted from string back into integer.  So the single string option is not what we want, but please consider using one of the first two options I suggested.

Mu2eG4/src/Mu2eG4WorkerRunManager.cc
```diff
@@ -11,6 +11,7 @@
 
 //Framework includes
 #include "cetlib_except/exception.h"
+#include "cstddef"
```
gaponenko, line 14: suggest angular brackets on cstddef include

Mu2eG4/src/Mu2eG4WorkerRunManager.cc
```diff
@@ -40,37 +41,54 @@
...
+  
+  
+  string CryptoPP_Hash(const string msg){
+    
+    CryptoPP::SHA3_224 hash;
+    string digest, str_hex;
+    
+    //digest is 56 hex numbers, or 28 pairs of hex numbers = 28 bytes
+    //compute the digest and keep 8 bytes of it
+    CryptoPP::StringSource(msg, true, new CryptoPP::HashFilter(hash, new CryptoPP::StringSink(digest), false, 8));
```
gaponenko, line 70: Does the cryptopp library delete all the objects that are newed here?

Mu2eG4/src/Mu2eG4WorkerRunManager.cc
```diff
@@ -313,25 +318,32 @@ namespace mu2e {
     
     if(eventHasToBeSeeded)
     {
-      long seeds[3] = { s1, s2, 0 };
+      string msg = "r" + i_run + "s" + i_subrun + "e" + i_event + salt_;
+
+      string hash_out = CryptoPP_Hash(msg);
+      vector<string> randnumstrings = {hash_out.substr(0,8), hash_out.substr(8,8)};
```
gaponenko, line 324: Making sure I understand...   The CryptoPP_Hash() function uses 8 bytes of the digest, but returns it as a hex number.  That is, 16 characters long.  So substr(8,8) does not go out of bounds.  Is that right?

gaponenko (changes requested):
This PR implements an important functionality - ensuring reproducibility of results when G4 is run in the MT mode.    It also looks like a nice code cleanup.     It could be merged as is, but I think changing the signature of generateEvt() would be an improvement.

Mu2eG4/inc/Mu2eG4WorkerRunManager.hh
goodenou: I do need the run, subrun, and event numbers as strings in the function generateEvt in order to make the input string for the hash:
string msg = "r" + i_run + "s" + i_subrun + "e" + i_event + salt_;

However, I can use the const art::EventID reference as you suggest for the argument of the function and convert that to a string inside the function.

gaponenko:
On Wed, 16 Dec 2020, Lisa Goodenough wrote:

> @goodenou commented on this pull request.
> 
> 
> 
> > @@ -51,7 +51,7 @@ namespace mu2e {
>      void initializeUserActions(const G4ThreeVector& origin_in_world);
>      void initializeRun(art::Event* art_event);
>      void processEvent(art::Event*);
> -    G4Event* generateEvt(G4int i_event);
> +    G4Event* generateEvt(std::string i_event, std::string i_subrun, std::string i_run);
> 
> I do need the run, subrun, and event numbers as strings in the function generateEvt in order to make the input string for the hash:
> string msg = "r" + i_run + "s" + i_subrun + "e" + i_event + salt_;
> 
> However, I can use the const art::EventID reference as you suggest for the argument of the function and convert that to a string inside the function.

Yes, that's what I mean.  Make the string inside the function, use 
EventID in the interface.

Thank you!
Andrei

Mu2eG4/src/Mu2eG4WorkerRunManager.cc
goodenou: Honestly, I don't remember ever adding that include and I don't see why we need it now, so I am going to take it out.

kutschke (changes requested):

Mu2eG4/src/SConscript
```diff
@@ -19,7 +19,10 @@ helper=mu2e_helper(env)
 
 g4inc    = os.environ['G4INCLUDE'] + '/Geant4'
 g4libdir = os.environ['G4LIB']
-g4LibInc = [ '-L'+g4libdir, '-I'+g4inc ]
+g4LibInc = [ '-L'+g4libdir, '-I'+g4inc]
+cryptoppinc = os.environ['CRYPTOPP_INC']
+cryptopplibdir = os.environ['CRYPTOPP_LIB']
+g4LibInc = [ '-L'+g4libdir, '-I'+g4inc, '-L'+cryptopplibdir, '-I'+cryptoppinc ]
 g4_version = os.environ['GEANT4_VERSION']
```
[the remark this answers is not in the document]
kutschke: Thanks.

goodenou (commented):
I have responded to all of the suggestions and comments.

PR #301, opened 2020-12-17
RNs seeds in G4MT are now handled by us. We use Crypto++ to create ha…

goodenou:
…shes for seeds based on art::event ID (Run#, Subrun#, Event#).

### DIFF
--- a/Mu2eG4/inc/Mu2eG4WorkerRunManager.hh
+++ b/Mu2eG4/inc/Mu2eG4WorkerRunManager.hh
@@ -49,9 +49,9 @@ namespace mu2e {
     void initializeThread(Mu2eG4MTRunManager* mRM,
                           const G4ThreeVector& origin_in_world);
     void initializeUserActions(const G4ThreeVector& origin_in_world);
-    void initializeRun(art::Event* art_event);
-    void processEvent(art::Event*);
-    G4Event* generateEvt(G4int i_event);
+    void initializeRun(art::Event* const art_event);
+    void processEvent(art::EventID);
+    G4Event* generateEvt(art::EventID);
 
     inline bool workerRMInitialized() const { return m_managerInitialized; }
 
@@ -67,6 +67,7 @@ namespace mu2e {
     bool m_steppingVerbose;
     int m_mtDebugOutput;
     int rmvlevel_;
+    std::string salt_;
 
     std::unique_ptr<Mu2eG4PerThreadStorage> perThreadObjects_;
 
--- a/Mu2eG4/inc/Mu2eG4WorkerRunManager.hh
+++ b/Mu2eG4/inc/Mu2eG4WorkerRunManager.hh
@@ -50,8 +50,8 @@ namespace mu2e {
                           const G4ThreeVector& origin_in_world);
     void initializeUserActions(const G4ThreeVector& origin_in_world);
     void initializeRun(art::Event* const art_event);
-    void processEvent(art::EventID);
-    G4Event* generateEvt(art::EventID);
+    void processEvent(const art::EventID&);
+    G4Event* generateEvt(const art::EventID&);
 
     inline bool workerRMInitialized() const { return m_managerInitialized; }
 
--- a/Mu2eG4/src/Mu2eG4WorkerRunManager.cc
+++ b/Mu2eG4/src/Mu2eG4WorkerRunManager.cc
@@ -276,7 +276,7 @@ namespace mu2e {
   }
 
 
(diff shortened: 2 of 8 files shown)


kutschke:
Lisa: thanks for closing this PR making this second PR - the history post merge is infinitely less confusing now.

Andrei: you added review comments in the "File Changed: tag of the earlier, now closed, PR.  Please go back to that tab and check each issue - if they are resolved to your satisfaction, please click on the "Resolve conversation" button for each of them.   When you are done, please approve this PR.

setup.sh
```diff
@@ -103,7 +103,7 @@ build=$($MU2E_BASE_RELEASE/buildopts --build)
 # and is therefore different from the value shown in
 # SETUP_<productname> environment vars, or by the "ups active" command.
 export MU2E_UPS_QUALIFIERS=+e20:+${build}
-export MU2E_ART_SQUALIFIER=s104
+export MU2E_ART_SQUALIFIER=s101
```
gaponenko, line 106: Please undo changes to setup.sh - I do not think these should go in.

kutschke:
Lisa: I should also have added that you should look at Andrei's comments in the old PR and, if they need a response, reply to them so that he can resolve them.

I resolved the conversation on my SConscript comment.

goodenou:
I believe I answered all of his comments, but I will check again.

kutschke:
@FNALbuild run build tests

Mu2eG4/inc/Mu2eG4WorkerRunManager.hh
```diff
@@ -49,9 +49,9 @@ namespace mu2e {
     void initializeThread(Mu2eG4MTRunManager* mRM,
                           const G4ThreeVector& origin_in_world);
     void initializeUserActions(const G4ThreeVector& origin_in_world);
-    void initializeRun(art::Event* art_event);
-    void processEvent(art::Event*);
-    G4Event* generateEvt(G4int i_event);
+    void initializeRun(art::Event* const art_event);
+    void processEvent(art::EventID);
+    G4Event* generateEvt(art::EventID);
```
gaponenko, line 54: Suggest const art::EventID& instead of art::EventID here to avoid making copies, for both functions.

Mu2eG4/src/Mu2eG4WorkerRunManager.cc
```diff
@@ -40,37 +40,53 @@
...
+  string CryptoPP_Hash(const string msg){
+    
+    CryptoPP::SHA3_224 hash;
+    string digest, str_hex;
+    
+    //digest is 56 hex numbers, or 28 pairs of hex numbers = 28 bytes
+    //compute the digest and keep 8 bytes of it
+    CryptoPP::StringSource(msg, true, new CryptoPP::HashFilter(hash, new CryptoPP::StringSink(digest), false, 8));
+    CryptoPP::CRYPTOPP_DLL HexEncoder encoder_out(new CryptoPP::StringSink(str_hex));
+    CryptoPP::StringSource(digest, true, new CryptoPP::Redirector(encoder_out));
```
gaponenko, line 70: Same question as before:  this looks like new-ing some objects and never deleting them.  How is memory managed here?

gaponenko (changes requested):
I would like to understand whether there is a memory leak, and also ask for code changes.  Please see the inline comments.

kutschke:
> I believe I answered all of his comments, but I will check again.
> […](#

I am still learning about the best ways to use github.  There are at least two ways to make comments: open a comment a line in a code listing on the “files changed” tab or comment in the box at the bottom of the conversation tab.   If you do the former it starts a thread that github organizes as a thread – it’s the thread that has the resolve conversation button.  If you comment in the box at the bottom of the conversation tab, it is not part of any thread, just a part of the whole conversation.   If you reply that way we don’t have the assistance of github to match it to the original comment.

I was asked that you reply to Andrei's in code comments within the thread that each started.

goodenou:
Rob,
I see three comments from Andrei and I replied to them all. Two were in-line/thread comments and one was not. I don’t know how to make the one that was not a thread comment into a thread comment.


Lisa

kutschke:
@FNALbuild run build tests

kutschke:
Lisa and Andrei: what do each of you think is left to resolve, if anything?

goodenou:
I think I have addressed all of Andrei’s suggestions and questions.

gaponenko (approved):
Lisa, thank you for updating the PR.   Rob, please merge.

PR #320, opened 2021-01-23
Trkfix

brownd1978:
Fix a self-canceling bug in sim and reco due to the recent straw direction convention change, needed for VST analysis.  Simplify StrawGasStep creation to use StepPointMC  end position.


bonventre (approved):

kutschke:
@FNALbuild run build tests

PR #321, opened 2021-01-26
add CrvStep

rlcee:
Add CrvStep to Print and Validation

### DIFF
--- a/Print/fcl/printLong.fcl
+++ b/Print/fcl/printLong.fcl
@@ -49,7 +49,7 @@ physics :{
       }
 
       caloShowerStepPrinter : {
-        verbose: 0   # 0, 1
+        verbose: 1   # 0, 1
         #eCut : 10.0          # cut on hits
 	#inputTags : [ "CaloShowerStepFromShowerStep" ] # select instances
       }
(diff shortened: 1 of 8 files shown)


brownd1978 (approved):
looks good

Validation/src/ValCrvStep.cc
```diff
@@ -0,0 +1,40 @@
...
+  _hN->Fill(coll.size()); 
+  for(auto cs : coll) {
+    _hb->Fill(cs.barIndex().asInt());
+    _ht->Fill(cs.startTime());
+    _hE->Fill(cs.visibleEDep());
+    _hposx->Fill(cs.startPosition().x());
+    _hposy->Fill(cs.startPosition().y());
+    _hposz->Fill(cs.startPosition().z());
+    _hp->Fill(cs.startMomentum().mag());
+    _hp2->Fill(cs.startMomentum().mag());
```
ehrlich-uva, line 36: _hp and _hp2 seems to be filled with the same number.

Print/fcl/printLong.fcl
```diff
@@ -72,12 +72,17 @@ physics :{
 	#inputTags : [ "CaloClusterMaker" ] # select instances
       }
 
+      crvStepPrinter : {
+        verbose: 1      # 0.1
+        eCut : -10.0          # cut on hits
```
ehrlich-uva, line 77: Is this negative sign correct (no energy cut)?

Validation/src/ValCrvStep.cc
rlcee: Yes, the two plots have very different scales - beam scale and cosmics scale.

Print/fcl/printLong.fcl
rlcee: Yes, negative means effectively no cut.  It is here as a reminder that the cut is possible.

Validation/src/ValCrvStep.cc
ehrlich-uva: Ok. Thank you.

Print/fcl/printLong.fcl
ehrlich-uva: Ok

ehrlich-uva (approved):

PR #329, opened 2021-01-28
Jcfix2

brownd1978:
More small tweaks to production and validation scripts.  Add CrvStep to product mixing.


kutschke (approved):
It looks good.

As we move forward I suggest that we switch from c-style FLT_MAX to c++style 

#include <limits>
std::numeric_limits<float>::max()

No need to do it on this PR.

PR #330, opened 2021-01-29
enable detailed physics list

brownd1978:


kutschke:
Hi Krzysztof, 

  I added you to this to have a look at the log file from ceSimReco in the CI tests.  That file contains these 3 lines:

physicsListDecider invoked with ShieldingM_EMZ
<<< Geant4 Physics List simulation engine: ShieldingM
<<< Reference Physics List ShieldingM_EMZ is built

do you understand why the middle line says ShieldingM?  Is this something we should care about?

I would lie to merge later today .

Thanks,

Rob

kutschke (approved):
Dave replied privately that he had asked Krzyszof the same question as I had asked and that this is the expected behaviou: "It is expected. _EMZ changes the EM models of whatever lists it is used with."

PR #332, opened 2021-02-01
Use longer G4 range in calo volume

brownd1978:


kutschke:
@FNALbuild run build tests

kutschke (approved):
I resubmitted the tests after merging PR 331.  I full expect the tests to pass again and will merge after they do.

JobConfig/primary/epilog.fcl
```diff
@@ -1,7 +1,7 @@
 # use detailed physics list
 physics.producers.g4run.physics.physicsListName:ShieldingM_EMZ
-# use 'coarse' stepping in the calorimeter to save time  disble for now FIXME!
-#physics.producers.g4run.physics.minRangeRegionCuts: { CalorimeterMother : 0.1}  // {RegionName : mm }
+# use 'coarse' stepping in the calorimeter to save time
```
resnegfk, line 3: This comment should be changed as it is misleading. It is production range cut for that region

PR #333, opened 2021-02-02
upgrading Geant4 to v10.6.p03

resnegfk:
This request is to start the final validation and upgrade process to Geant4 v10.6.p03 from v10.6.p02


brownd1978 (approved):

kutschke (approved):
Mackenzie reported at the  Computing and Software meeting today that the validation plots were on the wiki:

  https://mu2ewiki.fnal.gov/wiki/GeantChecklist#v4_10_6_p03

The plots are exact matches.   She reported that the log files show that the she indeed ran two different versions of Geant and did not accidentally run the same version twice.  Krzysztof reported that he looked again at the release notes and it is reasonable that they comparison is an exact match. 

We agreed to merge after the meeting.

PR #334, opened 2021-02-04
Remove obsolete fcl files.

kutschke:
This was discussed in:  https://mu2e-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=36656

This PR does not delete g4test_03(MT); after they are removed from the CI and from the nightly valJob I will make a separate PR.


brownd1978 (commented):
We eventually need to rename and relocate g4test_03.fcl as well, that invokes digitization and reconstruction so is still a maintenance burden.  Is there a plan for making that conform to JobConfig?  It could be outside this PR.

brownd1978:
sorry, I missed the statement at the top the g4test_03 is explicitly excluded from this PR.

brownd1978 (approved):

PR #335, opened 2021-02-04
Fix primary finding

brownd1978:
This is a minor bug fix, needed for MDC2020


oksuzian (approved):
Previous version contained an exception that was breaking CRY stage-2 simulation for events with multiple gen particles that result in a single sim particle in the filtered output.
I've tested that the new version works without throwing an expection.

kutschke:
Yuri: thanks for signing off on this quickly.  Merging now.

PR #337, opened 2021-02-06
Cosmic job config

oksuzian:


oksuzian:
I’d recommend we test the entire cosmic sim workflow using new configuration files.
There were multiple changes to the workflow.
We need to make sure they don’t introduce a bias.

JobConfig/cosmic/cosmic_s1_dsstops.fcl
```diff
@@ -1,216 +1,15 @@
...
-   TFileService          : { }
-   RandomNumberGenerator : {defaultEngineKind: "MixMaxRng" }
-
-   GeometryService        : { inputFile      : "JobConfig/cosmic/geom_cosmic_PhaseI.txt" }
-   ConditionsService      : { conditionsfile : "Mu2eG4/test/conditions_01.txt"      }
-   GlobalConstantsService : { inputFile      : "Mu2eG4/test/globalConstants_01.txt" }
-   G4Helper               : { }
-   SeedService            : @local::automaticSeeds
-}
+process_name :  CosmicS1
```
brownd1978, line 8: Please rename the process 'primary'.  Otherwise this exposes the origin process in Ensemble files.

JobConfig/cosmic/cosmic_s1_dsstops.fcl
```diff
@@ -229,86 +27,28 @@ physics :
...
-         module_type: FilterStatusG4
-         input: "g4run"
-      }
-
-      g4consistent: {
-         module_type: FilterStatusG4
-         input: "g4run"
-	 maxAcceptedStatus: 9  #  status 10 and above means StepPointMCCollection may have non-dereferencable pointers
-      }
+     g4status: @local::Primary.filters.g4status
```
brownd1978, line 30: This is outside the scope of this PR, but since we  are using a common g4status configuration it should move to JobConfig/common/mu2eg4.fcl

JobConfig/cosmic/cosmic_s1_dsstops.fcl
```diff
@@ -229,86 +27,28 @@ physics :
...
-         "keep mu2e::GenParticles_*_*_*",
-         "keep mu2e::GenEventCount_*_*_*",
-         "keep mu2e::StatusG4_*_*_*",
-         "keep mu2e::CosmicLivetime_*_*_*",
-         "keep *_g4run_*_*",
-         "drop uintmu2e::PhysicalVolumeInfomvstd::pairs_g4run_*_*",
-         "keep *_compressPVFull_*_*"
-      ]
-   }
+outputs.Output.outputCommands : [@sequence::Primary.KeptProducts, "keep *_cosmicFilter_*_*"]
```
brownd1978, line 35: I don't think Primary.KeptProducts will work here after Andy's step compression is merged, since that will look for SimParticles from compression, not g4run.  I think you should define the s1 output either here or in JobConfig/common/mu2eg4.fcl

JobConfig/cosmic/cosmic_s1_dsstops.fcl
```diff
@@ -229,86 +27,28 @@ physics :
...
+services.TFileService.fileName  : "nts.owner.cosmic-s1-dsstops.version.sequencer.root"
```
brownd1978, line 36: We are eliminating nts output for production files unless they serve a demonstrated purpose that cannot be performed on art output.  to disable nts output, assign the filename to '/dev/null'.

JobConfig/cosmic/cosmic_s2_resampler.fcl
```diff
@@ -0,0 +1,48 @@
+// Re-sample cosmics from stage-1
+// Yuri Oksuzian, 2020
+
+#include "JobConfig/primary/primary.fcl"
+#include "JobConfig/cosmic/epilog.fcl"
+
+// Bring in info from the previous stage
+physics.filters.dsResample: {
```
brownd1978, line 8: Is there a reason dsResample isn't defined in prolog?

JobConfig/cosmic/cosmic_s2_resampler.fcl
```diff
@@ -0,0 +1,48 @@
...
+// Point Mu2eG4 to the pre-simulated data
+physics.producers.g4run.MultiStageParameters: {
+   simParticleNumberOffset: 300000  # FIX ME!!!
+   genInputHits:  [ "dsResample:crvStage1" ]
+   inputSimParticles: "dsResample"
+   inputMCTrajectories: "dsResample"
+   inputPhysVolumeMultiInfo: "" 
+}
+// Kill cosmics outside of the CRV region
+physics.producers.g4run.Mu2eG4CommonCut:{
```
brownd1978, line 31: Is there a reason Mu2eG4CommonCut isn't defined in prolog?

JobConfig/cosmic/cosmic_s2_resampler_all.fcl
```diff
@@ -0,0 +1,8 @@
+// Re-sample cosmic dsstops particles from stage-1
```
brownd1978, line 1: What is the value added by this script over its parent?  It looks like a test file.  I suspect the hard-coded input will go out of date quickly.

JobConfig/cosmic/cosmic_s2_resampler_hi.fcl
```diff
@@ -0,0 +1,17 @@
+// Re-sample cosmic dsstops particles from stage-1
```
brownd1978, line 1: Is 'hi' not the default?  If so, can this configuration be made the default, with 'lo' overriding it?

JobConfig/cosmic/genconfig_cosmic_general.txt
```diff
@@ -0,0 +1,27 @@
...
+bool cosmicDYB.do = true;
+bool picapture.do = false;
+bool ejectedProtonGun.do = false;
+bool decayinorbitGun.do = false;
+bool piEplusNuGun.do = false;
+
+double cosmicDYB.mean =  -1.1; 
+double cosmicDYB.muEMin = 0.5e3;
+double cosmicDYB.muEMax = 5000e3;
+double cosmicDYB.dx     = 10000;
```
brownd1978: Do these geometric parameters have any relation to the 'CrvBox' parameters?  Does one have to nest or otherwise correspond to the other?

JobConfig/cosmic/genconfig_cosmic_general.txt
```diff
@@ -0,0 +1,27 @@
...
+double cosmicDYB.dy     = 0;
+double cosmicDYB.dz     = 20000;
+double cosmicDYB.direction = "Negative_y";
+vector<double> cosmicDYB.productionCenterInMu2e = {-3904,0,10200};
```
brownd1978: Is this not the Mu2e detector coordinate center?  Could it instead be read from the GeometryService?  Fixing this might be outside this PR.

brownd1978 (commented):
Thanks a lot for this consolidation, especially cleaning out the obsolete files.  there is a conflict with Validation/fcl/cosmicSimReco, I suggest you use this as an opportunity to update that to get its configuration from JobConfig/cosmic.

JobConfig/cosmic/genconfig_cosmic_general.txt
ehrlich-uva: This is for the Daya Bay cosmic ray generator, which can be configured with 2D production planes or a 3D "production box". The center can be placed anywhere, and we have used many different configurations over the years. So reading one particular set of coordinates from the GeometryService may only be useful, if we used it as default parameter which can be overwritten in the config file.

JobConfig/cosmic/genconfig_cosmic_general.txt
ehrlich-uva: No, these values are for the production planes / production box of the Daya Bay cosmic ray generator. The other CrvBoxes in the fcl files are used to define G4 energy cuts.

JobConfig/cosmic/cosmic_s1_dsstops.fcl
oksuzian: Do you need the process name to be 'primary' for stage-1 as well?
I think stage-1 process name is lost after stage-2 job.
I'm sure I'm missing a point. 
I'll change the name to 'Primary' though.

brownd1978:
You’re probably right, but if we ever decided to keep even 1 data product from stage 1 it would reveal the physics origin. It’s safer to set this to Primary

David Brown

JobConfig/cosmic/cosmic_s1_dsstops.fcl
oksuzian: Will do.

JobConfig/cosmic/cosmic_s2_resampler.fcl
oksuzian: Are you suggesting to put dsResample in cosmic/prolog.fcl
I'm not sure what's the purpose of doing that.
dsResample is only used for stage-2.

JobConfig/cosmic/cosmic_s2_resampler.fcl
oksuzian: Same reason

JobConfig/cosmic/cosmic_s2_resampler_all.fcl
oksuzian: I'll configure cosmic_s2_resampler.fcl as cosmic_s2_resampler_all.fcl and delete cosmic_s2_resampler_all.fcl
I prefer to keep a sample of input file for debugging purposes.
I can update the input file if it goes out of date.

JobConfig/cosmic/cosmic_s2_resampler_hi.fcl
oksuzian: cosmic_s2_resampler_all.fcl is default.

JobConfig/cosmic/cosmic_s1_dsstops.fcl
oksuzian: I've deleted the the lines 'services.TFileService.fileName'.
Filenames are set to null in primary/prmary.fcl, which I include.

JobConfig/cosmic/cosmic_s1_dsstops.fcl
oksuzian: I'll add "keep mu2e::GenParticles_*_*_*",
I think the rest should work.

brownd1978:
@FNALbuild run build test

brownd1978 (approved):
I think this is an important step towards cleaning up JobConfig.

ehrlich-uva (approved):

PR #338, opened 2021-02-09
remove empty dictionary build calls

rlcee:
code cleanup only, no change to build output

### DIFF
--- a/scripts/build/python/mu2e_helper.py
+++ b/scripts/build/python/mu2e_helper.py
@@ -140,29 +140,28 @@ class mu2e_helper:
     #   Make the dictionary and rootmap plugins.
     #
     def make_dict_and_map( self, userlibs=[], pf_dict=[] ):
-        if os.path.exists('classes.h') and os.path.exists('classes_def.xml'):
-            sources = ["classes.h","classes_def.xml"]
-            targets = ["#/"+self.dict_file(),
-                       "#/"+self.rootmap_file(),
-                       "#/"+self.pcm_file() ]
+        sources = ["classes.h","classes_def.xml"]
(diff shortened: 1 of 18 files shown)


scripts/build/python/mu2e_helper.py
```diff
@@ -140,29 +140,28 @@ def make_plugins( self, userlibs, exclude_cc = [], cppf = [], pf = [] ):
     #   Make the dictionary and rootmap plugins.
     #
     def make_dict_and_map( self, userlibs=[], pf_dict=[] ):
-        if os.path.exists('classes.h') and os.path.exists('classes_def.xml'):
```
kutschke, line 143: What happens if someone (inappropriately) adds a make_dict_and_map stanza to an SConscript file in a directory that does not have classes.h and classes_def.xml files?   Will there be an understandable error message?  Silently do nothing?

kutschke:
Ray and I discussed this PR before he did the work.  I have assigned it to myself and I don't think we need additional reviewers.

scripts/build/python/mu2e_helper.py
rlcee: you get a message like "don't know how to build classes.h required for .."

kutschke: Thanks. Does it stop the build or is it informational only?

rlcee: It is a fatal scons error and the build stops

kutschke: Good.  That's what I hoped.  I will merge.

kutschke (approved):

PR #340, opened 2021-02-11
Snmall tweaks to primary and digitize.  Add a script for digitizing p…

brownd1978:
minor update to get steps in the stoppingtarget, and to digitize pileup.


kutschke (approved):
This looks OK and I don't think that it needs additional review.  Will merge now,

PR #342, opened 2021-02-17
Allow CompressDetStepMCs to take multiple SimParticleCollections

AndrewEdmonds11:
For cosmic ray resampling, we generate CrvSteps in "dsResample" and StrawGasSteps and CaloShowerSteps in "g4run". For this reason, we need to allow CompressDetStepMCs to take multuple SimParticleCollection tags. This PR implements this feature.

One knock-on effect is that we need to rekey the output SimParticleCollection to avoid SimParticles with the same ID but from different collections overwriting each other. We do the same thing in CompressDigiMCs for mixed datasets through a fhicl parameter. Here I haven't put in the option to turn the rekeying off since I don't see a benefit for keeping the IDs and the cost is extra configuration differences between subtly different jobs. I'm happy to implement a switch if we really need it.

Cheers,
Andy

### DIFF
--- a/Compression/src/CompressDetStepMCs_module.cc
+++ b/Compression/src/CompressDetStepMCs_module.cc
@@ -104,7 +104,7 @@ public:
     fhicl::Atom<art::InputTag> caloShowerStepTag{Name("caloShowerStepTag"), Comment("InputTag for CaloShowerSteps")};
     fhicl::Atom<art::InputTag> crvStepTag{Name("crvStepTag"), Comment("InputTag for CrvSteps")};
     fhicl::Sequence<art::InputTag> stepPointMCTags{Name("stepPointMCTags"), Comment("Sequence of InputTags for StepPointMCCollections (e.g. virtualdetector)")};
-    fhicl::Atom<art::InputTag> simParticleTag{Name("simParticleTag"), Comment("InputTag for the SimParticleCollection")};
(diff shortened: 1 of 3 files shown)


AndrewEdmonds11:
So it turns out we might not need this for the cosmic resampling since the StepPointMCs used to create the CrvSteps should have been moved to "g4run" so the following change should be made instead:

`physics.producers.CrvSteps.stepPointsModuleLabels : ["dsResample"] `

to 

`physics.producers.CrvSteps.stepPointsModuleLabels : ["g4run"]`

However, we may need this for flash and OOT resampling. But it's unclear to me if these could be done in the same way as the cosmic resampler so I've converted this to a draft PR so that it can be merged at a later date if required.

kutschke:
Does your last comment mean that we don't need this PR or that we don't need the follow on work that you mentioned in the PR submission comment?

AndrewEdmonds11:
I mean that we don't need this PR yet, and it's possible that we might not need it at all.

AndrewEdmonds11:
It looks like we don't (and won't) need this feature for MDC2020. I will close this PR and if it turns out we do need this feature later, I will reopen it.

kutschke:
Thanks Andy,

  That works for me.

Rob

On 2/18/21, 11:50 AM, "Andrew Edmonds" <notifications@github.com<mailto:notifications@github.com>> wrote:


It looks like we don't (and won't) need this feature for MDC2020. I will close this PR and if it turns out we do need this feature later, I will reopen it.

PR #344, opened 2021-02-18
more dev database to a new sl7 server

rlcee:
should have no effect on any validation or user


kutschke (approved):
For a future PR should we make these fcl parameters?

rlcee:
> For a future PR should we make these fcl parameters?

I propose to keep these pointers on cvmfs, for even faster switchovers.  fcl is still mostly captured in releases or in generatefcl.  Technically, it would create an documented input to a job.  Practically though, I think the risk is tiny.

rlcee:
> > For a future PR should we make these fcl parameters?
> 
> I propose to keep these pointers on cvmfs, for even faster switchovers. fcl is still mostly captured in releases or in generatefcl. Technically, it would create an documented input to a job. Practically though, I think the risk is tiny.

that should be "undocumented"

PR #354, opened 2021-02-25
lower verbosity

rlcee:
turn off verbosity left on from previous work


kutschke (approved):
This looks good to me.  I was about to ask about some verbosity issues but I will test this first.

PR #356, opened 2021-02-25
reorganize trigger option

rlcee:
This takes the implementation of the Trigger buildopt option out of the control script and into the SConscripts.  I think this is a bit simpler, more transparent, and trivially Muse-ready.

### DIFF
--- a/Mu2eG4/src/SConscript
+++ b/Mu2eG4/src/SConscript
@@ -6,6 +6,10 @@
 import os, re
 Import('env')
 
+# if building for the trigger exe, don't build this lib
+if env['TRIGGER'] == 'on':
+    Return()
+
 Import('mu2e_helper')
 
 helper=mu2e_helper(env)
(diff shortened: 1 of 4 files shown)


rlcee:
> I see the idea that having each SConscript know what to do is a cleaner separation. But are we making multiple points of maintenance down the road?
> 
> Is the primary driver the Muse compatibility? I can see that controlling.
> 
> Should we take the opportunity to come up with a better name?

To add or remove a lib from the trigger option, you are editing one file either way.  Now, which file you are editing makes  more sense.  I am working towards making SConstruct and scons_helper more generic, holding only global functions, and keeping individual repo build customizations in the repo where it is used. I find this point compelling.  A simpler Muse design is the driving reason for the commit.  I find "trigger" a logical name, but if someone suggest something better, I can make the change, but that is a few more steps - we would have to update the buildopts mechanism.

brownd1978 (approved):

kutschke:
OK.  I tried inventing new names but they all turned into novels.  Let's just go with this.

kutschke (approved):

PR #362, opened 2021-03-02
An option to switch to totalEDep() from the default of visibleEDep()

oksuzian:
CrvStep uses visibleEDep() by default.
Visible energy is not present in older CRY samples.
This option will be used to convert older CRY samples to CrvStep format.

### DIFF
--- a/CRVResponse/src/CrvStepsFromStepPointMCs_module.cc
+++ b/CRVResponse/src/CrvStepsFromStepPointMCs_module.cc
@@ -59,6 +59,8 @@ namespace mu2e
 	  Comment("Starting size for crvIndex-particle vector"),4};
 	fhicl::Atom<bool> removeNeutralParticles{ Name("removeNeutralParticles"),
 	  Comment("Removes steps of neutral particles"),true};
+	fhicl::Atom<bool> useTotalEDep{ Name("useTotalEDep"),
+	  Comment("Use total energy deposition instead. Visible is a new default"),false};
       };
       using Parameters = art::EDProducer::Table<Config>;
       explicit CrvStepsFromStepPointMCs(const Parameters& conf);
@@ -82,6 +84,7 @@ namespace mu2e
 
       int      _debug, _diag;
       bool     _removeNeutralParticles;
+      bool     _useTotalEDep;
       unsigned _csize, _ssize;
       // StepPointMC selector
       // This selector will select only data products with the given instance name.
@@ -101,6 +104,8 @@ namespace mu2e
     _debug(config().debug()),
     _diag(config().diag()),
     _removeNeutralParticles(config().removeNeutralParticles()),
+    _useTotalEDep(config().useTotalEDep()),
+
     _csize(config().csize()),
     _ssize(config().startSize()),
     _selector(art::ProductInstanceNameSelector(config().stepPointsInstance())),
@@ -260,7 +265,12 @@ namespace mu2e
         first     =spmcptr;
         spmcIndices.emplace_back(i,i);
       }
-      edep   +=spmcptr->visibleEDep();
+
+      if(_useTotalEDep)
+        edep+=spmcptr->totalEDep();
+      else
+        edep+=spmcptr->visibleEDep();
+
       pathlen+=spmcptr->stepLength();
       last    =spmcptr;
       spmcIndices.back().second=i;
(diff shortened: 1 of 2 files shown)


CRVResponse/src/CrvStepsFromStepPointMCs_module.cc
```diff
@@ -59,6 +59,8 @@ namespace mu2e
 	  Comment("Starting size for crvIndex-particle vector"),4};
 	fhicl::Atom<bool> removeNeutralParticles{ Name("removeNeutralParticles"),
 	  Comment("Removes steps of neutral particles"),true};
+	fhicl::Atom<bool> useTotalEDep{ Name("useTotalEDep"),
+	  Comment("Use total energy deposition instead. Visible is a new default"),false};
```
kutschke, line 63: Normally I prefer that parameters that affect physics not have defaults in code and that they be required fcl parameters.

Is this seen as a temporary feature that will go away after the older CRY files are retired? If we insist on this policy will there be a lot of fcl files to update?

Will this change modify the behaviour of the nightly comics tests?

oksuzian: > Normally I prefer that parameters that affect physics not have defaults in code and that they be required fcl parameters.
> 
> Is this seen as a temporary feature that will go away after the older CRY files are retired?
Yes. After the current CRY files are retired we can remove this option.
We plan to use visible energy going forward  

> If we insist on this policy will there be a lot of fcl files to update?
I think it's not a large number, but I'll redirect this question to Ralf.
 
> Will this change modify the behaviour of the nightly comics tests?
No, the default behavior has not changed.
We only use non-standard config (useTotalEDep: true) in JobConfig/cosmic/cosmic_s2_converter.fcl

oksuzian:
Can you please assign Ralf?
Somehow, I can't do that.

CRVResponse/src/CrvStepsFromStepPointMCs_module.cc
kutschke: Since this is a transient thing, and you only enable the alternate behaviour in one place,  I am comfortable leaving it as is.  There is no need to modify the other files to explicitly chose the default.

I will resolve this comment.

kutschke:
About adding Ralf.  After you answered my questions, I don't think that we need to involve Ralf.  If you would like him involved for other reasons I am happy to add him.

oksuzian:
@FNALbuild run build test

CRVResponse/src/CrvStepsFromStepPointMCs_module.cc
brownd1978, line 63: Beyond the patch, is it ever possible TotalEDep would be the right answer?  If not, I'd prefer this option be labeled as a 'backwards compatibility' mode, together with the postPosition kludge.

CRVResponse/src/CrvStepsFromStepPointMCs_module.cc
```diff
@@ -61,6 +61,9 @@ namespace mu2e
 	  Comment("Removes steps of neutral particles"),true};
 	fhicl::Atom<bool> useTotalEDep{ Name("useTotalEDep"),
 	  Comment("Use total energy deposition instead. Visible is a new default"),false};
+	fhicl::Atom<bool> noPostPositionAvailable{ Name("noPostPositionAvailable"),
+	  Comment("No postposition is available for CRY3/4 samples"),false};
+
```
brownd1978, line 66: I'd like this bundled with the above to be a single 'backwards compatibility' mode.  Also please put comments in the code that this is a deprecated feature and will be eliminated soon.

brownd1978 (changes requested):

CRVResponse/src/CrvStepsFromStepPointMCs_module.cc
oksuzian: > I'd like this bundled with the above to be a single 'backwards compatibility' mode. Also please put comments in the code that this is a deprecated feature and will be eliminated soon.

We might want to toggle this flags individually for debugging purposes.
I don't have a strong preference though. I can do that if you insist.

brownd1978: OK.  I see the comment basically defines them as deprecated.

brownd1978 (approved):

ehrlich-uva (approved):
Approved

kutschke (approved):

kutschke:
@FNALbuild run build tests

PR #365, opened 2021-03-04
Adding CORSIKA input card

soleti:
I include the CORSIKA input card in the repository, which is needed to create the CORSIKA binaries.


kutschke (commented):
Have you thought a head to future uses of CORSIKA?  Will we need different input card files for that?  If so, is this a good choice of name or should it be more descriptive?   I have no insight into the answer - just asking the question.

soleti:
The only parameters we could reasonably change in the configuration are the particle type (in case we want to simulate heavy nuclei as primary particles and not protons) and the energy (in case we want to make targeted studies). But I think this is a good name for the default input card.

kutschke:
Thanks Roberto.  I will merge it later today.

kutschke (approved):

PR #367, opened 2021-03-04
Val update 210304

rlcee:
add stops time plot and reduce the number of products compared, to reduce clutter, as suggested by Dave


kutschke (approved):
I will approve this.  I will relaunch tests after I merge Dave's PR.  If all goes well I will merge tonight.

In a later PR can you remove magic numbers in  Validation/src/ValSimParticle.cc .  Use the  PDG code enum and, when possible, get geometry numbers from the geometry service objects. 

And you might consider using StoppingTargetGeom/inc/zBinningForFoils.hh and .cc which computes a sensible binning for plots involving the stopping target.  It ensures foils fit cleanly into bins and provides several options.

kutschke:
@FNALbuild run build tests

kutschke:
@FNALbuild run build tests

PR #368, opened 2021-03-05
new module for converting the calo fragments directly into caloHits w/o producing caloDigis

gianipez:
- This is the preliminary version of a new module used in the DAQ for converting art Fragments into CaloHits. We will still need the fragment-to-digi converter for debug and in the  DQM.
- added also a utility class to avoid cloning of shared rotines

I would like to push this MR before @miscetti pushed the caloDMAP, so that he would be capable to integrate the channel map also in this module.


kutschke:
Hi Giani,

  Please look into the failure in ceSimReco.  If you follow the "Logfiles may also be accessed ... " link at the bottom of the posted build/test report, the ceSimReco log file is the top item on the page.   It looks like something in the fcl #include chain within ceSimReco.fcl is still using the old NewCaloClusterFast and not the new CaloClusterOnline.    

   When you change a module label in JobConfig, please check for all uses of the old module label.  The way I do is to do the merge in a scratch git clone and then:

   find . -name \*.fcl -exec grep -H NewCaloClusterFast {} \;

  In this case it finds:

./CaloCluster/fcl/prolog_trigger.fcl
./Analyses/fcl/prolog.fcl

 Note that you also changed the name FastCaloHitMaker.  It is still present in 3 places in:
./CaloReco/fcl/prolog_trigger.fcl

I have no idea if that's a problem or not

Rob

gianipez:
Hi Rob,

I’ll do it. I was not aware that this module was used anywhere already. 

Cheers,

Giani

kutschke:
Hi Giani,

    No worries.   I think it’s Ok to put in a pull request and see what the CI system turns up.

   Rob


On 3/5/21, 11:47 AM, "Gianantonio Pezzullo" <notifications@github.com<mailto:notifications@github.com>> wrote:

Hi Rob,

I’ll do it. I was not aware that this module was used anywhere already.

Cheers,

Giani

gianipez:
please Jenkins try a build

kutschke:
@FNALbuild run build tests

RecoDataProducts/inc/CaloHit.hh
```diff
@@ -37,9 +42,9 @@ namespace mu2e {
           int    crystalId_;
           int    nSiPMs_;
           float  time_;             
-          float  timeErr_;             
           float  eDep_;        
           float  eDepErr_;        
+          float  timeErr_;
```
kutschke, line 47: Is this change just for aesthetic reasons?

I think that root's automatic schema evolution will manage this correctly but it may mean that reading old files with new code will be slower.  Is that an issue that we need to think about?  If that's only a small transient problem then this is a good time for the change.

gianipez: I don't think we will have an issue with older datasets as there were other breaking changes, no? I changed that because in the online reco we don't fill the *Err_ parameters and @bechenard suggested me to make this change to optimize the memory allocation.

DAQ/src/CaloRecoFromFragments_module.cc
```diff
@@ -231,26 +234,26 @@ void art::CaloRecoFromFragments::analyze_calorimeter_(
...
+        calo_digis->emplace_back((sipmID), hits[hitIdx].first.Time, caloHits,
                                  hits[hitIdx].first.IndexOfMaxDigitizerSample);
 
         if (diagLevel_ > 1) {
           // Until we have the final mapping, the BoardID is just a placeholder
           // adc_t BoardId    = cc.DBC_BoardID(pos,channelIdx);
 
           std::cout << "Crystal ID: " << (int)crystalID << std::endl;
-          std::cout << "APD ID: " << (int)apdID << std::endl;
+          std::cout << "SiPM ID: " << (int)sipmID << std::endl;
```
kutschke, line 256: Why do you need this cast?  I don't think that it does anything?  

In cases where you do need a cast, please use c++ style casts.

RecoDataProducts/inc/CaloHit.hh
kutschke: Yes, we are making other breaking changes.  That's why I said this is a good time for a change.   I will mark this resolved but I have a  follow up offline.

DAQ/src/CaloRecoFromFragments_module.cc
gianipez: done.

kutschke: There are many others of these in this file.  Please fix them all and I will merge right after.

gianipez: those I left are needed because they are uint8/16 and when I was printing them with cout was not readable.

kutschke: OK.  Merging now.

kutschke (approved):

DAQ/src/CaloHitsFromFragments_module.cc
```diff
@@ -0,0 +1,327 @@
...
+void art::CaloHitsFromFragments::analyze_calorimeter_(const artdaq::Fragment& f,
+						      std::unique_ptr<mu2e::CaloHitCollection> const& calo_hits,
+						      std::unique_ptr<mu2e::CaloHitCollection> const& caphri_hits) {
+  mu2e::CalorimeterFragment cc(f);
+
+  if (diagLevel_ > 1) {
+    caloDAQUtil_.printCaloFragmentInfo(f, cc);
+  }
+
+  for (size_t curBlockIdx = 0; curBlockIdx < cc.block_count(); curBlockIdx++) {
```
bonventre, line 209: Can this part not go in the utility so its not duplicated from StrawAndCaloDigisFromFragments?

gianipez: The operations inside the loop are different. Both modules are looping over the same object, but they produce different outputs that need different processing.

kutschke: I see that I jumped the gun on the merge.  Sorry about that.  If another small PR is warranted I will watch for it.

I am not sure that I understand.  I think that Richie is talking about the call to the print utility at line 206 but Giani is talking about code in the loop that starts at 209?  Can you straighten me out?

bonventre: We're talking about the loop - we'll just have to be careful to keep these both in sync

PR #370, opened 2021-03-09
New calorimeter noise generation

bechenard:
New version of calorimeter noise generation with several (fast) options. Also moved a variable in the calo conditions.

### DIFF
--- a/CaloMC/fcl/prolog.fcl
+++ b/CaloMC/fcl/prolog.fcl
@@ -47,7 +47,7 @@ CaloShowerROMaker :
 #See doc-db 35519 for noise level discussion
 NoiseGenerator : 
 {
-    noiseWFSize             : 2500
+    noiseWFSize             : 10000
     elecNphotPerNs          : 0.18
     rinNphotPerNs           : 0.36
     darkNphotPerNs          : 0.60
--- a/CaloMC/fcl/prolog.fcl
+++ b/CaloMC/fcl/prolog.fcl
@@ -7,7 +7,6 @@
 BEGIN_PROLOG
   HitMakerBlindTime      : 500. 
   HitMakerDigiSampling   : 5    
-  HitMakerMeVToADC       : 16   
   HitMakerMinPeakADC     : 16
 END_PROLOG
 
@@ -44,17 +43,29 @@ CaloShowerROMaker :
     diagLevel                : 0
 }
 
+
 #See doc-db 35519 for noise level discussion
+NoiseGenerator : 
+{
+    noiseWFSize             : 2500
+    elecNphotPerNs          : 0.18
+    rinNphotPerNs           : 0.36
+    darkNphotPerNs          : 0.60
+    digiSampling            : @local::HitMakerDigiSampling
+    enableAR                : true
+    nParAR                  : 5
+    nMaxFragment            : 1000
+    diagLevel               : 0
+}
+
 CaloDigiMaker : 
 {
     module_type             : CaloDigiMaker
     caloShowerROCollection  : CaloShowerROMaker
     blindTime               : @local::HitMakerBlindTime
     addNoise                : true
-    elecNphotPerNs          : 0.18
-    rinNphotPerNs           : 0.36
-    darkNphotPerNs          : 0.60
-    MeVToADC                : @local::HitMakerMeVToADC
+    generateSpotNoise       : true
+    NoiseGenerator          : { @table::NoiseGenerator  }
     digiSampling            : @local::HitMakerDigiSampling
     nBits                   : 12
     minPeakADC              : @local::HitMakerMinPeakADC
@@ -72,7 +83,6 @@ CaloHitTruthMatch:
     caloShowerSimCollection   : CaloShowerROMaker
     caloHitCollection         : CaloHitMaker
     primaryParticle           : FindMCPrimary
-    MeVToADC                  : @local::HitMakerMeVToADC
     digiSampling              : @local::HitMakerDigiSampling
     minAmplitude              : 32
     fillDetailedMC            : false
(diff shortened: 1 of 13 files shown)


brownd1978:
@FNALbuild run build tes

CaloMC/src/CaloNoiseSimGenerator.cc
```diff
@@ -0,0 +1,220 @@
...
+   void CaloNoiseSimGenerator::generateFragments(const CaloWFExtractor& wfExtractor)
+   {
+       unsigned nwf(0), nfound(0), length(255);
+       for (;nwf<nMaxFragment_;++nwf)
+       {
+          std::vector<double> temp(length,0.0);
+          addFullNoise(temp,false);
+
+          std::vector<int> wf;
+          for (const auto& val : temp) wf.emplace_back(val - pedestal_);
```
kutschke, line 111: Please do a reserve before filling the vector.  If compilers have got smarter and know to do the reserve on it's own, let me know.   Please check other parts of the code for the same issue - I noticed a few more spots.

CaloMC/src/CaloNoiseSimGenerator.cc
```diff
@@ -0,0 +1,220 @@
...
+          std::vector<int> wf;
+          for (const auto& val : temp) wf.emplace_back(val - pedestal_);    
+
+          std::vector<unsigned> starts, stops;         
+          wfExtractor.extract(wf,starts,stops);
+          if (starts.empty()) continue;
+
+          std::vector<double> fragment;
+          std::copy(temp.begin()+starts[0], temp.begin()+stops[0]+1, std::back_inserter(fragment));
+          digiNoise_.push_back(fragment);
```
kutschke, line 119: Please check if std::copy is smart enough to figure out to do the reserve.  I bet it is but I don't know for sure.

CaloMC/src/CaloNoiseSimGenerator.cc
```diff
@@ -0,0 +1,220 @@
...
+
+          ++nfound;         
+          if (nfound==20) break;
+       }
```
kutschke, line 123: Down the road I presume that "20" should come from somewhere?  Do you know where?  In any case make it constexpr and give it a good name.

CaloMC/src/SConscript
```diff
@@ -73,7 +74,7 @@ helper.make_plugins( [ mainlib,
                        'HepPDT',
                        'boost_filesystem',
                        'boost_system',
-                       rootlibs
+                       rootlibs  #needed for reco hit fitting
                      ],
```
kutschke, line 78: The comment about "reco" triggered this.   Something to think about for later: usually root fits are slow; they are great for development.  If you are doing simple fits, and if time an issue, it might make sense to have a project down the road to replace the root fits with something faster.

CaloMC/fcl/prolog.fcl
```diff
@@ -44,17 +43,29 @@ CaloShowerROMaker :
     diagLevel                : 0
 }
 
+
```
brownd1978, line 46: Which of these parameters specifies which noise generation algorithm (full, AR, sample) is being used by default?

brownd1978 (commented):
I am happy with the performance you described in your presentation but have questions about how the configuration choice and fitting procedures.

CaloMC/src/CaloNoiseSimGenerator.cc
bechenard: I don't believe it is (stl functions tend to do the minimum amount of work required). I have added a reserve. Note that this function was only called at the beginning of a subrun, and you would only copy a few short fragments, so the penalty in reallocating a few small vectors was small.

CaloMC/src/CaloNoiseSimGenerator.cc
bechenard: This 20 comes from me guessing a number. I can make it a const expr names MyGuessToStop if you want.

bechenard: Actually I called it enoughFragments

CaloMC/src/SConscript
bechenard: My answer has two parts:
1) I use TMinuit directly, which isn't the fastest algorithm on the market but its faster than the root wrapper 
2) The auto-regressive fit is only performed once at the beginning of the run, so the performance penalty is limited

But if we want to switch to a faster minimizer down the road, let me know

CaloMC/inc/CaloNoiseSimGenerator.hh
```diff
@@ -0,0 +1,84 @@
+#ifndef CaloNoiseSimGenerator_HH
+#define CaloNoiseSimGenerator_HH
+//
+// Generate long noise waveform to use for calorimeter digitization 
+//
+#include "fhiclcpp/types/Atom.h"
```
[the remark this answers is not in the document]
bechenard: The long waveform is generated once per run, at the beginning. I have added method to regenerate it at any point in the execution if we thing we need more randomness.

CaloMC/src/CaloNoiseARFitter.cc
```diff
@@ -0,0 +1,149 @@
+#include "CaloMC/inc/CaloNoiseARFitter.hh"
```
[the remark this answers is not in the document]
bechenard: The noise fit is performed once per run right after generating the long waveform (or every time the long waveform is generated if you want).

CaloMC/fcl/prolog.fcl
bechenard: generateSpotNoise = true will use the spot noise technique, otherwise you'll add the full noise for every waveform
enableAR = true  generates a full waveform with the auto-regressive fit if generateSpotNoise =false. if enableAR is false and generateSpotNoise =false, then you generate full noise waveform with simulating all PE (the old method)

CaloMC/src/CaloNoiseSimGenerator.cc
kutschke: Thanks.  My advise it just always do it, even beginJob, beginRun .... .  Then you don't forget when it's actually important.

CaloMC/src/SConscript
kutschke: Thanks.   Since it's once at begin run, it is a super low priority to bother with switching to a faster minimizer.

kutschke:
@FNALbuild run build tests

brownd1978 (approved):

kutschke:
@FNALbuild run build tests

PR #374, opened 2021-03-15
Updated event display

sophiemiddleton:
Updates are minor but previous bug is fixed


kutschke:
Are there any files in this PR that are not in TEveEventDisplay .  I scanned the file list but that's not very reliable.

sophiemiddleton:
Hi Rob,

I don't think so now, I just fixed the one thing which was not.

kutschke:
Thanks for fixing the spurious change.  Since everything is within TEveEventDisplay, if the CI tests all pass I will merge this.

kutschke:
This still shows as build pending - because of a commit that removed a blank line from setup.sh .  It's now identical to mu2e/master.  So I am merging without rerunning the tests.

PR #386, opened 2021-03-22
Update g4study fcl files

resnegfk:
Updated g4study fcl files by removing obsolete parameters from some to restore them to a working state, while relying on "@local" more


kutschke:
A comment to force the update of the test status.

kutschke:
@FNALbuild run build tests

PR #389, opened 2021-03-24
Renamed G4Helper to Mu2eG4Helper

resnegfk:
The name says it all


JobConfig/validation/stoppedMuonsSingleStageMT.fcl
```diff
@@ -0,0 +1,4 @@
+#include "JobConfig/validation/stoppedMuonsSingleStage.fcl"
```
kutschke, line 1: This is a new file that does not appear to be a part of the big rename.  Is this supposed to be part of it?  I have no objections - just checking.

kutschke:
I see that there are a lot of files that include one of the headers that were renamed but then never use them!   We will address that whenever we decide to let loose include-what-you-use.

kutschke (approved):

JobConfig/validation/stoppedMuonsSingleStageMT.fcl
resnegfk: Yes, it is slightly off topic, but likely a useful addition for a future use.

PR #392, opened 2021-03-25
Removed unused and outdated files from Mu2G4

resnegfk:
Removed unused and and by now outdated files from Mu2G4


kutschke:
Krzysztof: are all of these files part of Zhengyun's modifications for producing pbars? Or are some from other projects?

kutschke:
I am adding information from an offline discussion with Bob Bernstein and Krzysztof Genser.   Based on that discussion I am merging this PR.

 The code in this PR falls into two categories:

1.  Zhengyun You's code to improve the agreement between p-bar differential cross-sections models and data
2. Early versions of the tuned-for-Mu2e physics lists, the ones replaced by the G4 supported ShieldingM physics list.

All files contain code that relies on G4 internals that are allowed to change from version to version. While this code compiles and links it would require a major effort to certify it for use with modern versions of G4; both hand inspection and comparing the output of before and after jobs will be needed.

In Bob's view Zhengyun's code has served it's purpose and will not be used again.  Bob will not be available for a more detailed review until next week.   I have asked him to do that review when he can and we can restore any code that was inappropriately deleted.

PR #393, opened 2021-03-25
Adding Mu2eG4 prefix to more files in Mu2eG4

resnegfk:
Prefixed another group of file names with Mu2eG4


kutschke (approved):
Everything is what I expected.  And the CI tests passed.  That's enough for me.

PR #394, opened 2021-03-25
update to new sl7 production server

rlcee:
This was already updated on cvmfs


kutschke (approved):

rlcee:
> I thought this information was going to be indirected through cvmfs so it could be updated in realtime. Is this just a temporary fix?

  The cvmfs connection parameters system is live.   This PR is just saving the current version of the file on cvmfs.

PR #396, opened 2021-03-25
Update stable geometry to the most recent test geometry.

kutschke:
With this PR geom_common.txt now points to the same file as does geom_common_currnent:  

     Mu2eG4/geom/geom_2021_PhaseI.txt

We need to discuss timing of merging this PR and  tagging. 

Note that JobConfig/common/geom_baseline.txt includes Mu2eG4/geom/geom_common.txt

For the record here are the geometries used by the validation various jobs:

<pre>
ceSimReco        geom_common_current
transportOnly    geom_common
PS               geom_common               via geom_baseline
g4study          g4study_Tube_geom
cosmicSimReco    geom_common_current
rootOverlaps     geom_common
g4surfaceCheck   geom_common_current       via geom_SurfaceCheck
g4test_03MT      geom_common
reco             geom_common               via geom_baseline
potsim           geom_common_current
</pre>

### DIFF
--- a/Mu2eG4/geom/geom_common.txt
+++ b/Mu2eG4/geom/geom_common.txt
@@ -1,9 +1,9 @@
 //
-// Redirection to the current Mu2e top level geometry file.
+// Redirection to the current stable Mu2e top level geometry file.
 // The file that is included will be time dependent.
 //
 
-#include "Mu2eG4/geom/geom_common_hayman_v2.txt"
+#include "Mu2eG4/geom/geom_2021_PhaseI.txt"
 
 // This tells emacs to view this file in c++ mode.
 // Local Variables:


kutschke:
I have added Michael Mackenzie and Dave (Lou) Brown as reviewers.  I would like you both to sign off that I have actually chosen the latest and greatest geometry file.   To sign off, just approve the PR.

Michael:   Please write a short summary of how the geometry changes with this commit.   For example I see that the stopping target foils are thicker ( 0.05 -> 00528 mm  half thickness to match vendor spec); and there are changes in the details of the support structure.  That's the level of detail I am looking for.

Ray: any thoughts on the right place for Michael to put this?  Somewhere on the wiki?  Maybe repeated to MDC2020?  Or do we already have a geometry history page?

For everyone: the nightly ceSimReco uses geom_common_current – so it should not change.  The same for potsim and cosmicSimReco.  However reco uses geom_common.txt.  So it will see a change.  I ran on one of the files from the existing reco input file list and the valCompare (old vs new) is here:

         https://home.fnal.gov/~kutschke/Mu2e/compare_reco/results.html

There are fewer events overall ( 87 drops to 75 in the Pres plot ).  But everything looks consistent with changes to random numbers from small geometry changes.  I recommend that we let the nightly reco make the full stats plots once we have decided to merge.

dnbrow01 (approved):
I concur that geom_common.txt will now point to the most recent geometry configuration file, added by Michael, which should be exactly as intended.  I have not had a chance to exercise the new configuration file in any detail, but if Michael is happy with it, then I approve it.

michaelmackenzie (approved):
This looks good to me. This geometry should be ready for moving to "stable."

Thanks for doing this Rob!

PR #399, opened 2021-04-01
add KinKal to Muse setup

rlcee:
This follows the addition of KinKal to setup.sh, in the Muse build system.


rlcee:
The way that this dependence upgrade was merged did not preserve the Muse build, but with this merge, and resetting up muse, it should get back to working.  For the future, there is a natural way to do it.

brownd1978 (approved):

PR #400, opened 2021-04-05
switch from getenv to data file lookup

rlcee:
Muse doesn't define MU2E_BASE_RELEASE, so switch to data file lookup policy.  But this is the right thing to do even without Muse.


kutschke (approved):
I agree.  This is the right thing to do, even without Muse.

PR #401, opened 2021-04-06
rework bin scripts

rlcee:
rework scripts to remove reference to MU2E_BASE_RELEASE, which is not set in Muse


kutschke (approved):
I just looked at it - did not try anything.  It looks good.

kutschke:
@FNALbuild run build tests

PR #403, opened 2021-04-07
Mu2ekk2

brownd1978:
Bug fixes to support magnetic field correction.  Add KinKal to reco and validation/reco sequences.

### DIFF
--- a/.muse
+++ b/.muse
@@ -1,5 +1,5 @@
 # prefer to build with this environment
-ENVSET p001
+ENVSET p002
 # add these to python path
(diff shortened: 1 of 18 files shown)


brownd1978:
@FNALbuild run build test

kutschke:
@FNALbuild run build tests

kutschke (approved):
Everything looks good.

If we already had a user base I would ask if the change of nbend to nseg will break downstream code, since the meaning differs by 1.  But we don't have users yet,

PR #405, opened 2021-04-08
restore dependency lists in Muse build

rlcee:
only run in release build, so no effect on users or validation


kutschke (commented):
This looks Ok but I need to get up to date on some of the background.

Can you remind me how one invokes the dummy target functionality?

kutschke (approved):
Thanks.  Next time I have a muse environment working I will run it and watch.

PR #407, opened 2021-04-09
update to p003 for update of mu2e_artdaq_core

rlcee:


gaponenko:
Not having this change merged impedes the use of Muse, which is now the official build system.
In the future I suggest that any PRs that affects the required setup are not merged until
they include an appropriate update of the .muse file.
Andrei

$ muse setup
WARNING - local or linked packages have conflicting ENVSET recommendations
                 in .muse files.  Using p002 selected by search algorithm.
WARNING - linked repo Offline does not have the sl7-prof-e20-p002 build
     Build: prof     Core: sl7 e20 p002     Options:

kutschke (changes requested):
Ray.  I just checked envset p003.  The new version of the  ups product mu2e_artdaq_core no longer requires the :offline qualifier.

I should have called this out explicitly when I made the merge.  Sorry about that. 

So we need to change the envset on disk.  Does that p004, which would require an update to this PR?  Note that setup.sh is correct.

kutschke:
Andrei: thanks for the heads up.  I looked it over and encountered the issue that I raised a minute ago.

gaponenko:
I think it was a driver's error.  The backing build link was not pointing 
to the correct place, even thogh I thought I did "muse link master/7a9b4fc4"
Once the link is updated, the latest backing build and master head agree
on p002.

Andrei

rlcee:
> Ray. I just checked envset p003. The new version of the ups product mu2e_artdaq_core no longer requires the :offline qualifier.
> 
> I should have called this out explicitly when I made the merge. Sorry about that.
> 
> So we need to change the envset on disk. Does that p004, which would require an update to this PR? Note that setup.sh is correct.

the "offline" qualifier did not have a "+" so setup works either way.  I removed the qualifier in p003 on disk.  This should not disturb anything, since you get the same product either way.  You can proceed to merge.

kutschke (approved):

PR #408, opened 2021-04-10
Updated time wrapping and implemented the EventWindow in CRV MC and Reco

ehrlich-uva:
Additional changes:
-split the CRV-R1 sector into two sectors (older files can still be read)
-set the default scintillation yield of the CRV-Cryo-Inner modules to 0
-added additional variables to CrvRecoPulse
-added pulse overlap option for CRV coincidence checker 
-removed the old getManyByType and getMany from the CRV and EventDisplay code
CrvDigis from older files can't be used anymore, because their TDC time stamps were with respect to 0 and not with respect to the EventWindow start.


CRVResponse/src/CrvDigitizer_module.cc
```diff
@@ -79,6 +87,23 @@ namespace mu2e
 
   void CrvDigitizer::produce(art::Event& event) 
   {
+    art::Handle<EventWindowMarker> eventWindowMarker;
+    event.getByLabel(_eventWindowMarkerLabel,"",eventWindowMarker);
+    EventWindowMarker::SpillType spillType = eventWindowMarker->spillType();
+
+    ProditionsHandle<EventTiming> eventTimingHandle;
+    const EventTiming &eventTiming = eventTimingHandle.get(event.id());
+    double TDC0time = eventTiming.timeFromProtonsToDRMarker();
```
bonventre, line 96: Here you should be using
```
 const ProtonBunchTimeMC& pbtmc(*pbtmcHandle);
      TDC0time = -1*pbtmc.pbtime_;
```
pbtime is the time of the protons wrt the marker (so -1*pbtime is the marker time wrt to proton time). timeFromProtonsToDRMarker is only the fixed part of pbtime and doesn't include the bunch to bunch jittering. timeFromProtonsToDRMarker should only be used to convert the digitizationTime fcl parameter. (I.e. if timeFromProtonsToDRMarker is 100 ns and digitizationStart is 400 ns, then the cut applied will be 400-100=300 ns after the marker)

CRVResponse/src/CrvPhotonGenerator_module.cc
```diff
@@ -209,32 +298,54 @@ namespace mu2e
...
+            {
+              double timeTmp=times[itime];
+              if(spillType==EventWindowMarker::SpillType::onspill)
+              {
+                timeTmp = fmod(timeTmp,_microBunchPeriod);
+                //photons before the digitization start get removed except photons 
+                //in the first 55ns which get moved to the interval between the end of 
+                //the microbunch period and the digitization end
+                if(timeTmp<_digitizationEnd-_microBunchPeriod) timeTmp+=_microBunchPeriod;  
+                if(timeTmp<_digitizationStart) continue;
```
bonventre, line 340: I think you need to be careful here - this is all in times wrt protons I believe and so doesn't have any information about the jitter in the event marker time - you will lose the photons that get moved back into the digitization window by the jitter

bonventre (changes requested):

CRVResponse/src/MakeCrvRecoPulses.cc
```diff
@@ -11,47 +10,43 @@ namespace
     double const x = xs[0];
     return par[0]*(TMath::Exp(-(x-par[1])/par[2]-TMath::Exp(-(x-par[1])/par[2])));
   }
+  double Gumbel2(double* xs, double* par)
```
oksuzian, line 13: Do we plan to use Gubel2 or NoFit functions?
What's the default?

JobConfig/common/geom_common_phase1.txt
```diff
@@ -80,7 +80,7 @@ double mu2e.detectorSystemZ0 = 10171.;   // mm  G4BL: (17730-7292=9801 mm)
 #include "Mu2eG4/geom/calorimeter_CsI.txt"
 
 //CRV counters
-#include "Mu2eG4/geom/crv_counters_v07.txt"
+#include "Mu2eG4/geom/crv_counters_v08.txt"
```
oksuzian, line 83: We need to make sure this configuration propagates to Mu2eG4/geom/geom_common_current.txt, which is currently included by geom_common_phase1.txt

Mu2eG4/geom/crv_counters_v08.txt
```diff
@@ -0,0 +1,408 @@
...
+                       "T3",  //10   CRV-T modules (four upstream modules)
+                       "T4",  //11  CRV-T modules (remaining 16 downstream modules)
+                       "E1",  //12  CRV-TS-Extension upstream module
+                       "E2",  //13  CRV-TS-Extension downstream module 
+                       "U",   //14  CRV-U modules
+                       "D1",  //15  CRV-D modules above beam pipe
+                       "D2",  //16  CRV-D module at -x of beam pipe
+                       "D3",  //17  CRV-D module at +x of beam beam pipe
+                       "D4",  //18  CRV-D module below beam pipe
+                       "C1",  //19  CRV-Cryo-Inner module upstream of cryo hole
```
oksuzian, line 29: Do you want to disable C1 and C2?
Maybe in a prolog by setting LY to 0,  if you're not comfortable doing that here?

RecoDataProducts/inc/CrvRecoPulse.hh
```diff
@@ -16,9 +17,11 @@ namespace mu2e
     public:
 
     CrvRecoPulse() : _PEs(0), _PEsPulseHeight(0), _pulseTime(0), _pulseHeight(0), _pulseBeta(0), _pulseFitChi2(0), _LEtime(0), 
-                     _scintillatorBarIndex(0), _SiPMNumber(0) {}
+                     _flags(0), _scintillatorBarIndex(0), _SiPMNumber(0) {}
 
-    CrvRecoPulse(int PEs, int PEsPulseHeight, double pulseTime, double pulseHeight, double pulseBeta, double pulseFitChi2, double LEtime, 
+    CrvRecoPulse(float PEs, float PEsPulseHeight, double pulseTime, float pulseHeight, float pulseBeta, float pulseFitChi2, double LEtime,
```
oksuzian, line 22: Is pulseFitChi2 meaningful now?

RecoDataProducts/inc/CrvRecoPulseFlags.hh
```diff
@@ -0,0 +1,14 @@
+#ifndef RecoDataProducts_CrvRecoPulseFlags_hh
+#define RecoDataProducts_CrvRecoPulseFlags_hh
+
+#include <bitset>
+
+namespace mu2e 
+{
+  enum CrvRecoPulseFlagEnums{failedFit=0, unused=1, unused2=2};
```
oksuzian, line 8: How do we plan on using this?
How do you define failedFit?
What's unused and unused2?

oksuzian (commented):
Significant changes.
It would be great to have (maybe I missed it) a presentation on this and run a test on one of the CRY samples.

CRVResponse/src/MakeCrvRecoPulses.cc
ehrlich-uva: The default is Gumbel. Gumbel2 can be turned on in the fcl file. The "NoFit parameters" are always extracted: the start time and end time of a pulse, and the time of the maximum ADC value. These additional parameters can be used (via a fcl file option) in the coincidence finder.

JobConfig/common/geom_common_phase1.txt
ehrlich-uva: It is included in Mu2eG4/geom/geom_common_current.txt via Mu2eG4/geom/geom_2021_PhaseI.txt.

Mu2eG4/geom/crv_counters_v08.txt
ehrlich-uva: I set the LY to 0 in the fcl file. Removing C1 and C2 entirely would make older art files unreadable.

RecoDataProducts/inc/CrvRecoPulse.hh
ehrlich-uva: Yes, it is.

RecoDataProducts/inc/CrvRecoPulseFlags.hh
ehrlich-uva: I'm not using it yet. But I can imagine that we may want to use it later at the coincidence finder.
The unused flags are place holders for additional flags which I may add in the future. I can remove them.

CRVResponse/src/CrvDigitizer_module.cc
ehrlich-uva: I changed it here and in Crv reco (where I used ProtonBunchTime instead of ProtonBunchTimeMC).

CRVResponse/src/CrvPhotonGenerator_module.cc
ehrlich-uva: I added the jitter to the digitization window.

CRVResponse/src/CrvPhotonGenerator_module.cc
```diff
@@ -181,7 +265,23 @@ namespace mu2e
...
+          //see explanation above
+          //On-spill: No photons in the blind time between before digitizationStart and up to 55ns after 0
+          //(which will moved past the microbunch end after time folding) but allow a little bit more for the crvSteps
+          //-record CrvSteps before digitizationEnd-microBunchPeriod (i.e. 1750ns-1695ns=55ns)
+          //-record CrvSteps after digitizationStart-crvStepMargin (i.e. 400ns-50ns=350ns)
+          //Off-spill: Only photons within eventWindow, but allow a little bit more for the crvSteps
+          //-record CrvSteps after eventwindowStart-crvStepMargin until eventWindowEnd
+          if(spillType==EventWindowMarker::SpillType::onspill)
+          {
+            if(t1>digitizationEnd-_microBunchPeriod && t2<digitizationStart-_crvStepMargin) continue;
```
bonventre, line 279: Is step.start and endTime() wrapped by the microbunchPeriod yet at this point?

ehrlich-uva: No, time wrapping happens later after the photons are produced.

brownd1978:
@FNALbuild run build test

oksuzian (approved):

bonventre (approved):

RecoDataProducts/inc/CrvRecoPulseFlags.hh
ehrlich-uva: Fits are considered "failed fits", if the ROOT fit did not succeed, or if one or more fit parameters are at their limits.

PR #413, opened 2021-04-21
using newer Geant4 verbosity controls etc.

resnegfk:
This PR uses Geant4 10.7 hadronic verbosity controls and corrects statement ordering in Mu2eG4 SConscript


kutschke (approved):
This look good.  

Krzysztof tells me that this is only effective after the upgrade to geant v4_10_7 but that it will work prior to the upgrade.

I will merge at the first reasonable opportunity.

kutschke:
@FNALbuild run build tests

PR #416, opened 2021-04-23
fix bug in text override

rlcee:
The code worked for tables which were only in code and text, but did not work for the case of a text file actually overriding a database-supplied table.  The failure mode was to not override.   My guess is that when I tested this before I checked too superficially, and didn't see that the override failed silently.  Thanks and apologies to Richie.


brownd1978 (approved):

bonventre (approved):

brownd1978:
@FNALbuild run build test

PR #417, opened 2021-04-24
Add MDC2020 unit tests

rlcee:
For now these will be unit tests, run for a limited number of events in the nightly
- cdSteps - generate CE, sim to steps
- ceDigi - take ceSteps as input, make digis
- muSteps - generate all daughters of stopped muons, sim to steps
- ceMix  - mix ceSteps and muSteps

### DIFF
--- a/Validation/test/ceDigi.fcl
+++ b/Validation/test/ceDigi.fcl
@@ -1,0 +1,16 @@
+#include "JobConfig/digitize/OnSpill.fcl"
+
+services.scheduler.wantSummary: true
+services.TimeTracker.printSummary: true
+
+services.SeedService.baseSeed: 8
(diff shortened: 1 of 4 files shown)


brownd1978 (commented):
thanks for setting these up.  I would like to see these setup eventually as a POMS campaign, that should clarify how inputs chain to outputs.  Can you please also comment on where these scripts are invoked: is it only the nightly validation?  I would like a minimal version also run for every pull request, as part of the build test.

rlcee:
Once these are in the repo, I can add them to the nightly and I can work with Mackenzie to add them to git CI.  Creating a POMS test is more complex, we should probably talk about that.

brownd1978 (approved):

PR #419, opened 2021-04-26
Used  AntiLeakRegistry for rotation matrices and nullptr for identity rotations

resnegfk:
Used AntiLeakRegistry for rotation matrices where it was missing to avoid memory leaks and used nullptr for identity rotations in Geant4 volume placements to avoid unnecessary calculations.
(The whitespace change was replacing tabs with spaces and trailing space removal)

The code change should not modify any actual volume placements, but it should be tested as a precaution.


dnbrow01 (approved):
Does not look like any functional changes.  Very helpful leak protection and general cleanup.  Thanks, Krzysztof, for doing this!

brownd1978:
@FNALbuild run build test

brownd1978:
@FNALbuild run build test

PR #420, opened 2021-04-27
VolumeInfo is supposed to be optional in product mixinig

gaponenko:
A recent update to mu2e product mixing code defined VolumeInfo mixing configuration as required.
This was not intentional,  VolumeInfo is supposed to be an optional data product like everything else.
The PR is to make VolumeInfo optional.


brownd1978 (approved):

brownd1978:
@FNALbuild run build test

PR #421, opened 2021-04-27
fixed names of the module instance

gianipez:


gianipez:
@FNALbuild run build tests

brownd1978:
@FNALbuild run build tests

brownd1978 (approved):
This should fix the immediate problem, but I think its time to go over the naming conventions for calo trigger modules more systematically.

PR #427, opened 2021-04-28
KinKal fit extension

brownd1978:
Factorize common fit functions out of LoopHelixFit into KKFit, to facilitate implementing other kinds of fits.  Implement fit extension; ie hit and cluster adding and refit.  Remove internal references to HelixSeed from KalSeed, and replace with an external Assns, to support other kinds of fit seeds.  Add an explicit KKTrack subclass of KinKal::Track to hold mu2e-specific hit and material payload.   Also add a mu2e-specific KKStraw subclass of KinKal::Straw to allow storing strawId, needed to define material uniqueness.


JobConfig/reco/prolog.fcl
```diff
@@ -100,7 +100,8 @@ Reconstruction : {
 	HelixSeedCollections : [ "MHDeM" ]
         ComboHitCollection : "makeSH"
         StrawHitFlagCollection : "FlagBkgHitss:StrawHits"
```
gianipez, line 102: `FlagBkgHitss` is this a typo? 2 's'?

brownd1978: good catch!  Not sure why, but this worked.  I'll investigate.

AndrewEdmonds11 (approved):
The TrkAna-related changes look good to me. I like that InfoStructHelper::fillHelixInfo now takes a HelixSeed rather than a KalSeed - much more sensible!

JobConfig/reco/prolog.fcl
brownd1978: This collection isn't currently being used so the config error was irrelevant.  I fixed it anyways.

brownd1978:
@FNALbuild run build test

sophiemiddleton (approved):
All looks good to me

Mu2eKinKal/inc/KKFitSettings.hh
```diff
@@ -0,0 +1,62 @@
...
+#include "DataProducts/inc/PDGCode.hh"
+#include "RecoDataProducts/inc/TrkFitDirection.hh"
+namespace mu2e {
+  namespace Mu2eKinKal{
+    using Name    = fhicl::Name;
+    using Comment = fhicl::Comment;
+
+    // struct for defining the KinKal Config object
+    struct KKConfig {
+      fhicl::Atom<int> maxniter { Name("MaxNIter"), Comment("Maximum number of algebraic iteration steps in each fit meta-iteration"), 10 };
```
gianipez, line 20: are all these default necessary? or they could live in the corresponding prolog file?

RecoDataProducts/inc/KalSeed.hh
```diff
@@ -40,12 +40,12 @@ namespace mu2e {
     UInt_t nTrajSegments() const { return _nseg; }
     bool hasCaloCluster() const { return _chit.caloCluster().isNonnull(); }
     art::Ptr<CaloCluster> const& caloCluster() const { return _chit.caloCluster(); }
-    art::Ptr<HelixSeed> const& helix() const { return _helix; }
```
gianipez, line 43: I understand why you removed it; users where using it improperly. BUT I have thinking quite a bit about it and I remembered its real need; I'm using it to associate a reco track with the corresponding Helix that started it. 
Looking at the KalSeedFit_module I don't see a parallel structure of HelixSeedCollection pushed into the event.
Is there a different way I'm not aware of to match a track with the seed Helix?

RecoDataProducts/inc/KalSeedAssns.hh
```diff
@@ -0,0 +1,7 @@
+#include "RecoDataProducts/inc/KalSeed.hh"
+#include "RecoDataProducts/inc/HelixSeed.hh"
+#include "canvas/Persistency/Common/Assns.h"
+namespace mu2e {
+  // Assns between a KalSeed and the HelixSeed it was created from
+  typedef art::Assns<KalSeed,HelixSeed> KalHelixAssns;
```
gianipez, line 6: here it is! please, ignore my previous comment.

gianipez (changes requested):
Hi Dave,
My last comment is only about those default values in the Config struct. Other than that, the PR looks good to me!

brownd1978:
@FNALbuild run build test

brownd1978:
I moved the parameter defaults all into prolog.

gianipez (approved):
It looks great now!

PR #430, opened 2021-04-29
Fix test dir dependence

rlcee:
clean up test directory dependence for surfaceCheck and g4study, as recommended by Rob


brownd1978 (approved):

brownd1978:
@FNALbuild run build test

resnegfk (changes requested):
Changes to the location of the g4studyParticleGun.txt will break the other g4study_*.fcl files.
If we are deciding to move g4study_Tube_geom.txt we should also move the other g4study geometry files there as well.
Is there a document which describes the future locations of the files currently in Mu2eG4/test ?

rlcee:
> Changes to the location of the g4studyParticleGun.txt will break the other g4study_*.fcl files.
> If we are deciding to move g4study_Tube_geom.txt we should also move the other g4study geometry files there as well.
> Is there a document which describes the future locations of the files currently in Mu2eG4/test ?

I'll fix the other dependences on  g4studyParticleGun.txt, that was just an  oversight.

How about if I move  g4study_Tube_geom.txt back so there continues to be a complete g4study suite in test and I will create a new fcl and geom called g4val which will be used for validation?  The inconsistency comes from the dual purpose.  If you want, I can do the same with g4studyParticleGun.txt too.

The main idea we are going after is that validation and production should not depend on any test dirs, so that code management does not need to be immediately concerned with anything in a test dir.  The test dirs are them maintained by subject experts, as they see fit, when they have time.  The code and fcl in test is often not well maintained because it is used only occasionally, and this causes a drag on production code development.  This came up most recently when I was trying to parse code dependencies, and found problems in obsolete test code.  Shortly, we will delete all test dirs before running validation and production tests to enforce this separation.  At the highest conceptual level, most everything that is not involved in production simulation and reconstruction should be split into supplementary repos, to further streamline Offline.  A good example is that we just split TrackerAlignment into its own repo.

resnegfk:
Perhaps we should put all the self contained study files in say Mu2eG4/studies instead of the current test? And the validation ones elsewhere. However, the validation should test if the study still works, hence g4study in validation.

PR #434, opened 2021-04-30
Jobconfig dependence cleanup

rlcee:
- implement what I can from recommendations in Issue #426 
- Probably only remaining important item before splitting JobConfig is resolving the event display


brownd1978 (approved):

resnegfk (changes requested):
If we move  Mu2eG4/fcl/iontest_g4s2.fcl, we should also move the two other iontest*.fcl files
I would really suggest to make list of what goes where before doing the actual moves

rlcee:
@FNALbuild run build test

resnegfk (approved):

brownd1978:
@FNALbuild run build test

PR #435, opened 2021-05-05
Updated tracker TDC LSB and retuned calibration

bonventre:
Tracker sim updated to match hardware, which uses 200 MHz clock divided by 256 for TDC. TOT LSB also adjusted to 200 MHz clock. Retuned TOT drift time calibration and time division calibration, which now uses constant size bins.


TrackerConditions/src/StrawResponse.cc
```diff
@@ -129,19 +129,37 @@ namespace mu2e {
   }
 
   double StrawResponse::halfPropV(StrawId strawId, double kedep) const {
-    return PieceLine(_edep,_halfvp,kedep);
```
brownd1978, line 132: I can see it was useful to keep the uneven binning code while developing and debugging, but I don't think we will use it again.  Once you've verified the even binning gives equivalent results I suggest to remove all the code for uneven binning (and the eDep array).

TrkDiag/test/TOTDriftTime.py
```diff
@@ -0,0 +1,71 @@
+import ROOT
```
brownd1978, line 1: I think it's great you wrote this script.  For this PR it's fine to keep it in Offline, longterm I think we want to move this and other similar analysis outside of Offline, maybe into a TrackerCalib repo.

rlcee (commented):
A couple of quick comments.  
1) Consider doing something to check the length of the fcl arrays.  Either make them fixed length in the [fcl](lhttps://cdcvs.fnal.gov/redmine/projects/fhicl-cpp/wiki/Fhiclcpp_types_in_detail#SequenceltTstdsize_tgt) or check them against the nbins parameter when they are stored.
2) A couple of places you have the construct where you calculate ibin, with a limit of nbin-1, then refer to array entry ibin+1.  Taken at face value, nbin-1 is the last entry so ibin+1 would overrun, but I may be missing something in  the details.

brownd1978 (approved):
looks good

rlcee (approved):
Thanks for addessing my comments

brownd1978:
@FNALbuild run build test

PR #437, opened 2021-05-06
Fixed cosmic resampling

soleti:
This PR adds the `compressPV` producer to the stage-1 cosmic production and the `volumeInfoMixer` to the stage-2 DS resampling.

### DIFF
--- a/JobConfig/cosmic/S1DSStops.fcl
+++ b/JobConfig/cosmic/S1DSStops.fcl
@@ -2,15 +2,52 @@
 // Cosmics hitting crvStage1End volumes are stopped and saved for later resampling.
 //
 // Yuri Oksuzian, 2021
-#include "JobConfig/primary/primary.fcl"
+#include "JobConfig/primary/prolog.fcl"
 
 process_name : Primary
 
-physics.TriggerPath: [generate, genCounter, g4run, g4status, stepPointMomentumFilter, cosmicFilter, compressPV]
+services: @local::Services.Sim
+
+physics: {
+  producers : {
+    generate: @local::generate
+    g4run : @local::mu2e.physics.producers.g4run.muons
+    genCounter: { module_type: GenEventCounter }
+    compressPV: @local::Cosmic.compressPV
+  }
+
+  filters : {
+    g4status: @local::Cosmic.g4status
+    stepPointMomentumFilter : @local::Cosmic.stepPointMomentumFilter
+    cosmicFilter : @local::Cosmic.cosmicFilter
+  }
+
+  analyzers: {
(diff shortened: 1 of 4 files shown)


JobConfig/cosmic/S1DSStops.fcl
```diff
@@ -6,8 +6,8 @@
 
 process_name : Primary
 
-physics.TriggerPath: [generate, genCounter, g4run, g4status, stepPointMomentumFilter, cosmicFilter]
-outputs.Output.outputCommands : [@sequence::Primary.KeptProducts, "keep mu2e::GenParticles_*_*_*", "keep *_cosmicFilter_*_*"]
+physics.TriggerPath: [generate, genCounter, g4run, g4status, stepPointMomentumFilter, cosmicFilter, compressPV]
```
brownd1978, line 9: I don't see where the compressPV module is defined.  It's not in this file or any of the #includes.

soleti: It's in JobConfig/primary/prolog.fcl (but defined in JobConfig/cosmic/prolog.fcl) no?

brownd1978:
I must be blind as I can't see it in JobConfig/primary/prolog.fcl  Can you please point me to the line?

JobConfig/primary/prolog.fcl
```diff
@@ -22,6 +22,7 @@ Primary: {
     @table::TrackerMC.StepProducers
     @table::CaloMC.StepProducers
     CrvSteps : @local::CrvSteps
+    compressPV: @local::Cosmic.compressPV
```
soleti, line 25: @brownd1978 here

brownd1978, line 25: I see it's something you added.  primary/prolog should not instantiate a module from Cosmic, that's a reverse-dependency.  Please move this to cosmic/prolog.fcl

JobConfig/cosmic/prolog.fcl
```diff
@@ -103,15 +109,19 @@ Cosmic: {
    }
    dsResample: {
```
brownd1978, line 110: Why does a cosmic job have a module named 'dsResample'?  Was this just a copy from Andrei's example?  I think a name like cosmicResampler would be more intuitive.

brownd1978 (approved):
looks good

brownd1978:
@FNALbuild run build test

oksuzian (approved):
Hi Roberto,
I believe we wanted to structure our fcl file starting with include of "JobConfig/primary/primary.fcl". Other than that I'm fine with this PR

brownd1978:
S2 jobs should inherit from primary.fcl, but S1 jobs are a different type and should not inherit from primary.fcl

PR #438, opened 2021-05-06
Offline compatible with mu2e_artdaq_core v1_05_10

gianipez:
In the setup script,  I used v1_05_09 that has the same functionality as v1_05_10.


kutschke:
This requires a more recent version of mu2e_artdaq_core.  That version of mu2e_artdaq_core is available in Mackenzie's version working towards the art v3_09_00 release.

The current idea is to cherry-pick these commits into Mackenzie's working version - but other options have been discussed.

PR #444, opened 2021-05-10
TrackerAligment moved to its own repo

rlcee:
TrackerAlignment is in its own repo.  I've checked that the git history can be accessed.  The new repo builds under Muse.


sophiemiddleton:
Hi Ray,

Where are the files moved to? I just see the deletes

rlcee:
> Hi Ray,
> 
> Where are the files moved to? I just see the deletes

The code was moved to a new [repo](https://github.com/Mu2e/TrackerAlignment) in the Mu2e organization on github.  The first step was essentially to copy the code there, so there is a no trace of the action here, in Offline.

sophiemiddleton (approved):
Seems fine

kutschke:
Does anyone plan to build both repos (TrackerAlignment and Offline) under muse and verify that everything still works?  Who should do that?

kutschke:
Thanks Richie,

  Unless it’s a large amount of work that would be wasted, it would be good to fix the fcl now and test while the mechanics are fresh in everyone’s mind.

     Rob

From: Richie Bonventre ***@***.***>
Date: Monday, May 10, 2021 at 3:54 PM
To: Mu2e/Offline ***@***.***>
Cc: Rob Kutschke ***@***.***>, Team mention ***@***.***>
Subject: Re: [Mu2e/Offline] TrackerAligment moved to its own repo (#444)


Ray has already tested that it builds. I looked at it briefly, but the alignment fcl is out of date now and needs to be reworked. I'll try to fix it

bonventre (approved):

PR #458, opened 2021-05-24
valCompare one-file mode

rlcee:
Usually valCompare compares two files.  If given only one file, and a request to make a web page or a pdf file, then write output in a format appropriate for displaying one file.  This is just a convenient way to display and share the histograms of one file.


kutschke:
This looks straightforward.  Is there anyone you would like to have a look at it?  if not, I will approve it.

rlcee:
I checked that the comparison we run nightly works OK, I think that's the only concern.  Beyond that, it is just adding a little feature, and probably best to comment on that after using it, so I'd ask you to go ahead and merge.

PR #465, opened 2021-05-28
Only enable ExtMonFNALPixelSD when it is configured to be enabled.

kutschke:
Hot fix to creation of the Sensitive Detector for the ExtMon pixels.  Previously it was always created and registered with G4 even if the configuration said not to.  This led to seg faults because it was not serviced by SenstitiveDetectorHelper, which did obey the configuration.


brownd1978 (approved):
Looks good.

PR #473, opened 2021-06-04
Fixed initial ewm phase always being zero

bonventre:
Fixed a typo that made the initial phase always zero. This PR will change the nightly validation slightly again since the event window times in the 10 validation runs will be different.

### DIFF
--- a/CommonMC/src/EventWindowMarkerProducer_module.cc
+++ b/CommonMC/src/EventWindowMarkerProducer_module.cc
@@ -38,6 +38,8 @@ namespace mu2e {
         fhicl::Atom<unsigned> spillType { Name( "SpillType"), Comment("Whether simulating on or off spill")};
         fhicl::Atom<bool> recoFromMCTruth{ Name("RecoFromMCTruth"), Comment("Create reco PBT from MC truth"),false};
         fhicl::Atom<float> recoFromMCTruthErr{ Name("RecoFromMCTruthErr"), Comment("If creating reco PBT from MC truth, smear by this amount in ns"),0};
+        fhicl::Atom<bool> fixInitialPhase{ Name("FixInitialPhase"), Comment("If false will randomize phase every run"),false};
+        fhicl::Atom<float> initialPhaseShift{ Name("InitialPhaseShift"), Comment("Initial phase shift between beam and detector clocks in fraction of one clock tick."),0};
       };
 
       using Parameters = art::EDProducer::Table<Config>;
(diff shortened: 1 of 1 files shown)


CommonMC/src/EventWindowMarkerProducer_module.cc
```diff
@@ -38,6 +38,8 @@ namespace mu2e {
         fhicl::Atom<unsigned> spillType { Name( "SpillType"), Comment("Whether simulating on or off spill")};
         fhicl::Atom<bool> recoFromMCTruth{ Name("RecoFromMCTruth"), Comment("Create reco PBT from MC truth"),false};
         fhicl::Atom<float> recoFromMCTruthErr{ Name("RecoFromMCTruthErr"), Comment("If creating reco PBT from MC truth, smear by this amount in ns"),0};
+        fhicl::Atom<bool> fixInitialPhase{ Name("FixInitialPhase"), Comment("If false will randomize phase every run"),false};
+        fhicl::Atom<float> initialPhaseShift{ Name("InitialPhaseShift"), Comment("Initial phase shift between beam and detector clocks in fraction of one clock tick."),0};
```
brownd1978, line 42: Andrei has pointed out a better design pattern for this kind of config, namely OptionalAtom.   In this case initialPhaseShift would become an OptionalAtom, and the bool Atom would go away.  The test on the bool would be replaced by a test whether the OptionalAtom is set or not.  This combines the config and prevents ambiguous situations (ie what to do if someone sets fixInitialPhase to false but doesn't provide a value, etc).  An example is

brownd1978 (changes requested):

CommonMC/src/EventWindowMarkerProducer_module.cc
bonventre: That's a lot more convenient, good to know that exists!

brownd1978 (approved):
looks good.  Should there be a production configuration for this for MDC2020.

brownd1978:
@FNALbuild run build test

kutschke:
@FNALbuild run build tests

gaponenko:
Hi,

> Andrei has pointed out a better design pattern for this kind of config, namely OptionalAtom.   In this case initialPhaseShift would become an OptionalAtom, and the bool Atom would go away.  The test on the bool would be replaced by a test whether the OptionalAtom is set or not.  This combines the config and prevents ambiguous situations (ie what to do if someone sets fixInitialPhase to false but doesn't provide a value, etc).  An example is 

This reminded me that there is an esier-to-use interface for Optional
entities.  See https://cdcvs.fnal.gov/redmine/issues/25773
It looks like it should be available in art v3.09.00; I have not tried it yet.

Andrei

brownd1978:
@FNALbuild run build tests

brownd1978:
@FNALbuild run build tests

PR #478, opened 2021-06-07
finishNesting only creates a new LogicalVolume if the VolumeInfo does…

HCasler:
…n't already have one


kutschke (approved):
Thanks Helenka.  This is what we discussed.

PR #483, opened 2021-06-10
deleting G4RunManager explicitly to avoid warnings

resnegfk:


resnegfk:
@FNALbuild run build tests

kutschke:
Please make this a real PR.   

I am happy to get rid of the whole concept of Clean().  We should first check to see if anyone relies on it - please send email to mu2e-hn-sim and ask about it. 

A long time ago I argued that we would never build the world 2 times in one G4 job - if we needed to build 2 different worlds, we would run two jobs.  My reasoning as that I did not trust us to be able to do it correctly - auditing the code to make sure that we did everything correctly would be  a huge job.  I thought this was a widely accepted point of view.

Mu2eG4/src/Mu2eG4_module.cc
```diff
@@ -451,9 +451,9 @@ namespace mu2e {
 
 
   void Mu2eG4::endJob(){
-
     if ( _exportPDTEnd ) exportG4PDT( "End:" );
     _physVolHelper.endRun();
+    delete _runManager;
```
kutschke, line 456: I think you could accomplish the same thing by keeping it a unique_ptr and adding one line here:

_runManager->release(nullptr);

or something like that.  I just don't like leaving bare pointers around as a model for others to follow.

kutschke (commented):
Another question.  Do you understand why this works?  Deleting the run manager in the endJob vs the d'tor seems like no change to me.

Mu2eG4/src/Mu2eG4_module.cc
resnegfk: I am not sure why replacing pointers works. I view it as a temporary solution to remove warnings and I do not like it either.
One thing which it does is defining the exact point where the run manager is deleted. 
The release does not do it: https://www.cplusplus.com/reference/memory/unique_ptr/release/

kutschke:
Ooops.

_runManager->reset();

https://www.cplusplus.com/reference/memory/unique_ptr/reset/



From: Krzysztof Genser ***@***.***>
Date: Thursday, June 10, 2021 at 10:39 AM
To: Mu2e/Offline ***@***.***>
Cc: Rob Kutschke ***@***.***>, Assign ***@***.***>
Subject: Re: [Mu2e/Offline] changed unique ptr pointing to G4RunManager to raw ptr, plus cleanup (#483)


@resnegfk commented on this pull request.

________________________________

In Mu2eG4/src/Mu2eG4_module.cc<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_Mu2e_Offline_pull_483-23discussion-5Fr649301160&d=DwMCaQ&c=gRgGjJ3BkIsb5y6s49QqsA&r=Zi5IH-k2_JiK_Wipiv3pwwSXO9UNaHiLP8sisfPz__k&m=_agSZi0OJPPRGqTufAO_KFhm6p8mCVV6TtxZ8t07UE8&s=DQ7T0USpxlFCf9-atcI_ctIhsPwZOtNrKd5LIVLgN34&e=>:

>      if ( _exportPDTEnd ) exportG4PDT( "End:" );

     _physVolHelper.endRun();

+    delete _runManager;

I am not sure why replacing pointers works. I view it as a temporary solution to remove warnings and I do not like it either.
One thing which it does is defining the exact point where the run manager is deleted.
The release does not do it: https://www.cplusplus.com/reference/memory/unique_ptr/release/<https://urldefense.proofpoint.com/v2/url?u=https-3A__www.cplusplus.com_reference_memory_unique-5Fptr_release_&d=DwMCaQ&c=gRgGjJ3BkIsb5y6s49QqsA&r=Zi5IH-k2_JiK_Wipiv3pwwSXO9UNaHiLP8sisfPz__k&m=_agSZi0OJPPRGqTufAO_KFhm6p8mCVV6TtxZ8t07UE8&s=YsH1cglmtZJqiBb0CNGxnFMrscJkg6--nRz7siOw2JM&e=>

Mu2eG4/src/Mu2eG4_module.cc
kutschke: Adding my previous reply into the thread, where it belongs: Oops, reset() not release.

kutschke: Is the real reason that it works the removal of Clean() and the change of pointer type has nothing to do with it?

resnegfk: Sorry, I missed/misread reset; that should do it. I change it to it and see if it still works. 
The change (warnings going away) comes from that RunMamager deletion, not removal of Clean(); I double check it though.

brownd1978: I agree with Rob that keeping the unique_ptr is preferable to switching to a raw pointer.  They have the same functionality but unique_ptr is less error prone.

kutschke:
I had a conversation with Krzysztof off of this thread.  I created the Clean() function in the very first release of WorldMaker and the comment said that I copied what the G4 docs told me to do.  Now that the advice from G4 has changed, I am even more sure that removing it is the correct thing to do.

resnegfk:
@FNALbuild run build tests

kutschke:
@FNALbuild run build tests

Issue #484, opened 2021-06-10
Bad PhysicsParams defaults


gaponenko:
When GlobalConstantsService loads its configuration from a SimpleConfig file, most parameters are required to be explicitly defined.  But there are a few dangerous lines in the code like

    _captureProtonRate[material] = config.getDouble("physicsParams."+material+".capture.protonRate", 0);
    _captureDeuteronRate[material] = config.getDouble("physicsParams."+material+".capture.deuteronRate", 0);
    _captureNeutronRate[material] = config.getDouble("physicsParams."+material+".capture.neutronRate", 0);
    _capturePhotonRate[material] = config.getDouble("physicsParams."+material+".capture.photonRate", 0);

which default undefined physics rates to zero, potentially leading to jobs happily producing wrong physics output.
I am going to add an explicit exception  throw for non-aluminum material to Pileup_module.cc until this is fixed.

rlcee:
Some information is provided for all materials, and some is not.
I think there are three options here
- do not load anything for a material unless all numbers are available (shorten material list)
- provide all needed numbers for all materials (need a reference)
- load available numbers (create two materials lists for loading purposes, complete materials and incomplete materials)
@kutschke @gaponenko

gaponenko:
I think we ned to allow materials with incomplete information to be used.   For example, if someone studies Mu2e-II physics sensitivity in the mu- to e- channel, they do not need the mu- to e+ channel energy, and if the latter is missing it should not prevent the study in the "information available" channel.   I don't think there should be two lists of materials.  All materials can be treated uniformly, and each accessor must either return valid data or throw.

rlcee:
Here are some ways to implement this, @gaponenko please comment
- attempt to read all values for all materials, use default value of DBL_MAX, then check for this value in the getters.
- call hasName to check the file for each value, keep track of what is available in a PhysicsParams bit flag, check this in the getters
- other?

The first involves a little less code, but uses a flag value in the physics variable.

gaponenko:
Keeping track of what is available and storing values as
std::optional<double> rather than double would be a clean solution,
but the DBL_MAX flag value approach is probably a lot simpler to
implement and looks good enough to me.   I am fine with either solution.

Andrei

rlcee:
One more possible model.  Save the available values as `map<string,double>`, change
[...]

PR #485, opened 2021-06-10
Pileup resampling

gaponenko:
This implements the Pileup generator for art-based stopped muon resampling.


EventGenerator/src/DIOGenerator_tool.cc
```diff
@@ -29,43 +29,54 @@ namespace mu2e {
...
-    {
-
-    }
+    {}
 
+    std::vector<ParticleGeneratorTool::Kinematic> generate() override;
     void generate(std::unique_ptr<GenParticleCollection>& out, const IO::StoppedParticleF& stop) override;
 
-    void setEngine(art::RandomNumberGenerator::base_engine_t& eng) {
+    void finishInitialization(art::RandomNumberGenerator::base_engine_t& eng, const std::string&) override {
```
brownd1978, line 38: Why is the material name unused?  DIO spectrum does depend on material, I guess that's a future development?  If so, please add a comment and a TODO flag

EventGenerator/src/Pileup_module.cc
```diff
@@ -0,0 +1,180 @@
...
+    produces<mu2e::StageParticleCollection>();
+    if(verbosity_ > 0) {
+      mf::LogInfo log("Pileup");
+      log<<"stoppingTargetMaterial = "<<conf().stoppingTargetMaterial()
+         <<", muon lifetime = "<<muonLifeTime_
+         <<", decay fraction = "<<decayFraction_
+         <<std::endl;
+    }
+
+    if(conf().stoppingTargetMaterial() != "Al") {
```
brownd1978, line 107: It is the individual tools that need to be configured for different materials.  Shouldn't the throw be in finishInitialization?  As far as what this module does (lifetime, decay fraction) many materials besides Al are supported.

Mu2eUtilities/src/stoppedMuonList.cc
```diff
@@ -0,0 +1,38 @@
...
+    template<class H>
+    std::vector<art::Ptr<SimParticle> > stoppedMuonListImpl(H simh) {
+
+      std::vector<art::Ptr<SimParticle> > res;
+
+      const auto& sims = *simh;
+
+      for(auto i = sims.begin(); i != sims.end(); ++i) {
+        const auto& inpart = i->second;
+        if((inpart.pdgId() == PDGCode::mu_minus) &&
```
brownd1978, line 17: We will eventually need to process stopped mu+, stopped pi-, etc.  is the idea to have dedicated functions for each?  If so, the name stoppedMuonList is ambiguous.  I think a more extensible design is a generic stoppedParticleList function which takes pdgcode and process code as arguments.  Or make this a functor, and provide the pdgcode and process on construction.

brownd1978 (commented):
This generally looks good, my suggestions are related to making future extension easier.

gaponenko:
Hi Dave,

On Thu, 10 Jun 2021, David Brown wrote:

> @brownd1978 commented on this pull request.
> 
> This generally looks good, my suggestions are related to making future extension easier.
> 
> >      void generate(std::unique_ptr<GenParticleCollection>& out, const IO::StoppedParticleF& stop) override;
>  
> -    void setEngine(art::RandomNumberGenerator::base_engine_t& eng) {
> +    void finishInitialization(art::RandomNumberGenerator::base_engine_t& eng, const std::string&) override {
> 
> Why is the material name unused?  DIO spectrum does depend on material, I guess that's a future development?  If so, please add a comment and a TODO flag

See below.

> 
> > +    , simsToken_{consumes<SimParticleCollection>(conf().inputSimParticles())}
> +    , verbosity_{conf().verbosity()}
> +    , eng_{createEngine(art::ServiceHandle<SeedService>()->getSeed())}
> +    , randFlat_{eng_}
> +    , randExp_{eng_}
> +  {
> +    produces<mu2e::StageParticleCollection>();
> +    if(verbosity_ > 0) {
> +      mf::LogInfo log("Pileup");
> +      log<<"stoppingTargetMaterial = "<<conf().stoppingTargetMaterial()
> +         <<", muon lifetime = "<<muonLifeTime_
> +         <<", decay fraction = "<<decayFraction_
> +         <<std::endl;
> +    }
> +
> +    if(conf().stoppingTargetMaterial() != "Al") {
> 
> It is the individual tools that need to be configured for different materials.  Shouldn't the throw be in finishInitialization?  As far as what this module does (lifetime, decay fraction) many materials besides Al are supported.

That's what I thought, but it does not throw in my testing.  See
https://github.com/Mu2e/Offline/issues/484

In addition to fixing the numbers provided by the
GlobalConstantsService, there is the issue of spectra.  Of course the
DIO spectrum depends on the material, as well as every other spectrum
generated by the Pileup module.  This PR does not touch the existing
implementation of the spectra, which relies on BinnnedSpectrum and:

- assumes that the use provides consistent configuration for individual
  tools to make the spectra consistent

- hardcodes aluminum for some spectra (e.g. MuCapProtonGenerator)
  without saying this explicitly.

IMO the BinnedSpectrum class grew over the years to became one of the
ugliest parts of Mu2e code.  I want to clean this up, which would
involve a significant restructuring.  This would require some not
insignificant amount of effort without changing any of the existing
functionality, so it is not high on my list.  A less error prone
configuration of the Pileup module tools to support non-Al materials
may be an outcome of the code rework.

So yes, future work, but involving more than just the pileup tools.


> 
> > +
> +namespace mu2e {
> +
> +  //================================================================
> +  namespace {
> +
> +    template<class H>
> +    std::vector<art::Ptr<SimParticle> > stoppedMuonListImpl(H simh) {
> +
> +      std::vector<art::Ptr<SimParticle> > res;
> +
> +      const auto& sims = *simh;
> +
> +      for(auto i = sims.begin(); i != sims.end(); ++i) {
> +        const auto& inpart = i->second;
> +        if((inpart.pdgId() == PDGCode::mu_minus) &&
> 
> We will eventually need to process stopped mu+, stopped pi-, etc.  is the idea to have dedicated functions for each?  If so, the name stoppedMuonList is ambiguous.  I think a more extensible design is a generic stoppedParticleList function which takes pdgcode and process code as arguments.  Or make this a functor, and provide the pdgcode and process on construction.


A good point.  I will update this code.

Andrei

gaponenko:
@FNALbuild run build tests

brownd1978 (approved):
Thanks for making the change.

AndrewEdmonds11 (approved):
Very nice. Looks good to me. Thanks, Andrei :)

PR #490, opened 2021-06-12
Remove time maps

brownd1978:
get rid of time maps


kutschke:
The plan is to merge this PR and resolve the failed tests in future PRs so that the mainline work can continue.

PR #492, opened 2021-06-15
Minor updates to the magnetic field configuration

kutschke:
The "reco" configuration is unchanged (reverted earlier changes).

Yuri confirms that he does not want us to use DSSweepExtension; we should use DSExtension.  No changes wrt this map.

Removed the DSExtension map from the "no ds" configuration.  This will save about 30MiB of memory footprint in those jobs.

Removed three more maps from the no_tsu_ps configuration. All three are between the PS and the beam dump.  Total savings of 7.0 MiB of memory.

Removed the three ExtMonUCI maps from the outermaps definition.  These field maps are obsolete.  They describe the magnetic field in the region where the UC Irving design for the Extinction Monitor would have been located.  That region is outside the world maps and is currently filled with dirt and concrete.  This removes about 7.5MiB from the memory footprint of a running job.

To see a figure of where ExtMonUCI maps are located, see pages 22 and 3 of https://mu2e-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=32091  .

### DIFF
--- a/Mu2eG4/geom/bfgeom_reco_v01.txt
+++ b/Mu2eG4/geom/bfgeom_reco_v01.txt
@@ -14,7 +14,8 @@
 #include "Mu2eG4/geom/bfgeom_v01.txt"
 
 vector<string> bfield.innerMaps = {
-  "BFieldMaps/Mau13/DSMap.header"
+  "BFieldMaps/Mau13/DSMap.header",
+  "BFieldMaps/Mau13/DSExtension.header"
 };
 
 vector<string> bfield.outerMaps = {};
--- a/Mu2eG4/geom/bfgeom_reco_v01.txt
+++ b/Mu2eG4/geom/bfgeom_reco_v01.txt
@@ -3,7 +3,7 @@
 // not needed for reconstruction have been removed.
 //
 // Warning:
-//  There are multiple points of maintenance when you change bfield.innerMaps
+//  There are multiple points of maintenance when you change any bfield maps.
 //    - bfgeom_v01.txt           ( the base geometry  )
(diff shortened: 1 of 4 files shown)


brownd1978:
@FNALbuild run build test

Mu2eG4/geom/bfgeom_reco_v01.txt
```diff
@@ -14,7 +14,8 @@
 #include "Mu2eG4/geom/bfgeom_v01.txt"
 
 vector<string> bfield.innerMaps = {
-  "BFieldMaps/Mau13/DSMap.header"
+  "BFieldMaps/Mau13/DSMap.header",
+  "BFieldMaps/Mau13/DSExtension.header"
```
brownd1978, line 18: I am confused about the use case.  CRV digitization and reconstruction don't access the BField map, so why does this matter here?  if the use case is pileup simulation, that job uses  geom_common_no_tsu_ps.txt   which includes DSExtension.

kutschke: You're right.  I will back this one out.  I will also remove these two lines from geom_common_no_tsu_ps.txt

  "BFieldMaps/Mau13/PStoDumpAreaMap.header",
  "BFieldMaps/Mau13/ProtonDumpAreaMap.header",

kutschke:
@FNALbuild run build tests

brownd1978 (approved):
looks good

kutschke:
A comment to see if this triggers the out-of-date check.

kutschke:
A comment to see if this triggers the out-of-date check.

kutschke:
@FNAL build run build tests

PR #494, opened 2021-06-17
Trig flag cleanup

gianipez:
removed un-used data members from the TriggerInfo class


kutschke:
As I read this PR, it won't change nightly validation.   Is that right?  I think it preserves the meaning of trigger bits - is that right?

gianipez:
Hello Rob,

Yeah, the TriggerFlag data-product was something I did when in the very beginning… It’s completely orthogonal w.r.t. the trigger bits the art::TriggerResults provides.

kutschke:
@FNALbuild run build tests

brownd1978 (approved):

PR #499, opened 2021-06-19
updates to the Trigger code + small fix in CaloHitMakerFast

gianipez:
- improvements to the Trigger Analyzer module
- now setting the paths ID in the Menu files
- update the TriggerResultsNavigator code to handle the labels of the `art::TriggerResults` object of the latest art version ("pathName" -> "id:pathName")
- temporary fix in the `CaloHitMakerFast` module (see comments in the code)
- fixed time offset for CalPatRec-trigger_mode (probably needs to be fixed also for the Offline_mode); it moved from `-41` to `-61`
- probably need to adjust the calorimeter-timeOffset also for TPR; I tried to apply the same correction I found for CPR, but it didn't play as expected (aka smaller eff); right now is `-20`, and I tried `-40` and `0`, but both tests were returning worst result than the default.

### DIFF
--- a/CalPatRec/fcl/prolog_trigger.fcl
+++ b/CalPatRec/fcl/prolog_trigger.fcl
@@ -126,7 +126,7 @@ CprTrigger : { @table::CprTrigger
 	    minClusterEnergy                            : 50.    # MeV
 	    minClusterSize                              : 2      # number of crystals
 	    minClusterTime                              : 500.   # ns
-	    dtOffset                                    : -41.8 #FIXME! this is necessary because CaloClusterFast returns an uncalibrated time!
+	    dtOffset                                    : -61. #-41.8 #FIXME! this is necessary because CaloClusterFast returns an uncalibrated time!
 	}
 
 	TTCalHelixFinder             : { @table::CalPatRec.filters.CalHelixFinder
--- a/CaloReco/src/CaloHitMakerFast_module.cc
+++ b/CaloReco/src/CaloHitMakerFast_module.cc
@@ -108,23 +108,25 @@ namespace mu2e {
        for (const auto& caloDigi : caloDigis)
        {
            int crystalID   = cal.caloIDMapper().crystalIDFromSiPMID(caloDigi.SiPMID());
-
-           double baseline = (caloDigi.waveform().at(0)+caloDigi.waveform().at(1)+caloDigi.waveform().at(2)+caloDigi.waveform().at(3))/4.0; 
-           double eDep     = (caloDigi.waveform().at(caloDigi.peakpos())-baseline)*calorimeterCalibrations->ADC2MeV(caloDigi.SiPMID()); 
+	   size_t nSamPed  = caloDigi.peakpos() > 3 ? 4 : std::max(caloDigi.peakpos()-1, 1);
+           double baseline(0);
+	   for (size_t i=0; i<nSamPed; ++i){ baseline += caloDigi.waveform().at(i);}
+	   baseline /= nSamPed;
+           double eDep     = (caloDigi.waveform().at(caloDigi.peakpos())-baseline)*calorimeterCalibrations->ADC2MeV(caloDigi.SiPMID());//FIXME! we should use the function ::Peak2MeV, I also think that we should: (i) discard the hit if eDep is <0 (noise/stange pulse), (ii) require a minimum pulse length. gianipez 
            double time     = caloDigi.t0() + caloDigi.peakpos()*digiSampling_;                      //Giani's definition
            //double time     = caloDigi.t0() + (caloDigi.peakpos()+0.5)*digiSampling_ - shiftTime_; //Bertrand's definition
 
            addPulse(pulseMap, crystalID, time, eDep);
-           if (diagLevel_ > 2) std::cout<<"[FastRecoDigiFromDigi] extracted Digi with crystalID="<<crystalID<<" eDep="<<eDep<<"\t time=" <<time<<std::endl;
+           if (diagLevel_ > 2) std::cout<<"[CaloHitMakerFast] extracted Digi with crystalID="<<crystalID<<" eDep="<<eDep<<"\t time=" <<time<<std::endl;
        }
 
        for (auto& crystal : pulseMap)
        {
-	   int crID = crystal.first;
+	   int  crID     = crystal.first;
            bool isCaphri = std::find(caphriCrystalID_.begin(), caphriCrystalID_.end(), crID) != caphriCrystalID_.end();
 	   for (auto& info : crystal.second)
            {
-               if (diagLevel_ > 1) std::cout<<"[FastRecoDigiFromDigi] extracted Hit with crystalID="<<crID<<" eDep="<<info.eDep_<<"\t time=" <<info.time_<<"\t nSiPM= "<<info.nSiPM_<<std::endl;
+               if (diagLevel_ > 1) std::cout<<"[CaloHitMakerFast] extracted Hit with crystalID="<<crID<<" eDep="<<info.eDep_<<"\t time=" <<info.time_<<"\t nSiPM= "<<info.nSiPM_<<std::endl;
                if (isCaphri) caphriHitsColl.emplace_back(CaloHit(crID, info.nSiPM_, info.time_,info.eDep_));
                else          caloHitsColl.emplace_back(  CaloHit(crID, info.nSiPM_, info.time_,info.eDep_)); 
 	   }
--- a/Mu2eUtilities/src/TriggerResultsNavigator.cc
+++ b/Mu2eUtilities/src/TriggerResultsNavigator.cc
@@ -41,9 +41,24 @@ namespace mu2e {
     }  
 
     //loop over trigResults to fill the map <string, unsigned int)
+    std::string   delimeter=":";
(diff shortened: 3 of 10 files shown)


kutschke:
@FNALbuild run build tests

CaloReco/src/CaloHitMakerFast_module.cc
```diff
@@ -108,23 +108,25 @@ namespace mu2e {
        for (const auto& caloDigi : caloDigis)
        {
            int crystalID   = cal.caloIDMapper().crystalIDFromSiPMID(caloDigi.SiPMID());
-
-           double baseline = (caloDigi.waveform().at(0)+caloDigi.waveform().at(1)+caloDigi.waveform().at(2)+caloDigi.waveform().at(3))/4.0; 
-           double eDep     = (caloDigi.waveform().at(caloDigi.peakpos())-baseline)*calorimeterCalibrations->ADC2MeV(caloDigi.SiPMID()); 
+	   size_t nSamPed  = caloDigi.peakpos() > 3 ? 4 : std::max(caloDigi.peakpos()-1, 1);
+           double baseline(0);
+	   for (size_t i=0; i<nSamPed; ++i){ baseline += caloDigi.waveform().at(i);}
+	   baseline /= nSamPed;
```
kutschke, line 114: Again for future reference.  I like that you are using .at(i) instead of [i] for development.  For time critical code we should switch it back to [i] for production.  I know of several ways to way to write code so that we can flip between the two with a compiler switch.  When you are ready let's talk about it.

Mu2eUtilities/src/TriggerResultsNavigator.cc
```diff
@@ -41,11 +41,26 @@ namespace mu2e {
...
   }
 
+
+  std::string const
+  TriggerResultsNavigator::getTrigPathName(unsigned int const i) const
+  {
+    std::string   delimeter =":";
+    size_t        pos       = _trigPathsNames[i].find(delimeter);
+    if (pos > _trigPathsNames[i].length()) return "TRIG PATH NOT FOUND";
+    return _trigPathsNames[i].substr(pos+1, _trigPathsNames[i].length());
```
kutschke, line 61: How will this function be used?  I generally don't like returning a magic string to denote an error.  But there are probably use cases in which it is the best way - for example if it's only being used in printout and it's normal that it might sometimes be missing.

Trigger/python/genTriggerFcl.py
```diff
@@ -240,7 +240,7 @@ def generate(configFileText="allPaths", verbose=True, doWrite=True):
     #
 
     configFile = open(configFileName, "r")
-
+    pathID     = 100 #we start the Reco paths from 100
```
kutschke, line 243: How many other pieces of code need to know that 100 is a magic number?   I see that this is in the python world.  Will we need a place in the C++ world or the tuple world?  How can we define it once in each world?

Trigger/src/ReadTriggerInfo_module.cc
```diff
@@ -271,13 +285,17 @@ namespace mu2e {
...
+
+    float  _minPOT, _maxPOT;
   };
 
 
   ReadTriggerInfo::ReadTriggerInfo(fhicl::ParameterSet const& pset) :
     art::EDAnalyzer(pset), 
     _diagLevel     (pset.get<int>   ("diagLevel", 0)),
-    _nMaxTrig      (pset.get<size_t>("nFilters", 70)),
+    _nMaxTrig      (pset.get<size_t>("nPathIDs", 200)),
```
kutschke, line 298: We are asking everyone to convert modules to validated fcl.  Please do so.

Trigger/src/ReadTriggerInfo_module.cc
```diff
@@ -286,10 +304,21 @@ namespace mu2e {
...
+    _hsCprTag      (pset.get<art::InputTag>("cprHelixSeedCollection", "CalHelixFinderDe:Positive")), // , "KFFDeMHPar")),
+    _hsTprTag      (pset.get<art::InputTag>("tprHelixSeedCollection", "HelixFinderDe:Positive")), // , "KFFDeMHPar")),
+    _ksTag         (pset.get<art::InputTag>("kalSeedCollection"  , "KFFDeMHPar")),
+    _trkQualTag    (pset.get<art::InputTag>("trackQualCollection", "TrkQualDeMHPar")),
+    _vdTag         (pset.get<art::InputTag>("vdStepPoints","NOTNOW")), // , "compressDigiMCs:virtualdetector")),
     _duty_cycle    (pset.get<float> ("dutyCycle", 1.)),
     _processName   (pset.get<string> ("processName", "globalTrigger")),
-    _nProcess      (pset.get<float> ("nEventsProcessed", 1.))
+    _effBits       (pset.get<std::vector<size_t>>("effBits", std::vector<size_t>{2,8})),
+    _trkMinTanDip  (pset.get<float> ("trkMinTanDip", 0.5)),
```
kutschke, line 316: Please  make a sub-table to hold the cuts.  When you make the change, remove the defaults from the code.

kutschke (changes requested):
I have a few structural change requests and a few suggestions for future improvements.  

Do you expect this to change validation?  As best I can tell the change to dtOffset in CalPatRec/fcl/prolog_trigger.fcl is not seen by validation.   And I don't think that any of the modules are seen by validation.

CalPatRec/fcl/prolog_trigger.fcl
```diff
@@ -126,7 +126,7 @@ CprTrigger : { @table::CprTrigger
 	    minClusterEnergy                            : 50.    # MeV
 	    minClusterSize                              : 2      # number of crystals
 	    minClusterTime                              : 500.   # ns
-	    dtOffset                                    : -41.8 #FIXME! this is necessary because CaloClusterFast returns an uncalibrated time!
+	    dtOffset                                    : -61. #-41.8 #FIXME! this is necessary because CaloClusterFast returns an uncalibrated time!
 	}
```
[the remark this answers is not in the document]
gianipez: I think that once that this offset could be included in the calorimeter channel time calibration, which will differ for Offline and Online reconstruction. Right now, there is no such a prodiction.

CaloReco/src/CaloHitMakerFast_module.cc
gianipez: Interesting. If this is something that affects several parts of the Offline it might be good to identify someone who has time to make this change.

Mu2eUtilities/src/TriggerResultsNavigator.cc
gianipez: I honestly wasn't happy either, but I thought that there might be analysis code that could be ok with having a pathname not found.

Trigger/python/genTriggerFcl.py
gianipez: That is something that was discussed a long time ago in some software meeting; I proposed to leave free the first 100 bits for non-triggger paths. No other code needs to know this number. This script generates only the trigger Menu file. An analyzer will need to know either the list of the "good" bits or the list of the pathnames.

Trigger/src/ReadTriggerInfo_module.cc
gianipez: I can do it, but I'm not super excited about having a table for an analyzer module in the prolog_trigger.fcl file. Are we sure about it?

CalPatRec/fcl/prolog_trigger.fcl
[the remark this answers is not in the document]
kutschke: THanks.  Just keep it on your radar when the proditions object does exist.

CaloReco/src/CaloHitMakerFast_module.cc
kutschke: Do you have any summer students who could do it?

Mu2eUtilities/src/TriggerResultsNavigator.cc
kutschke: The standard answer is to define a small handle-like class and return that.  It would have two data members, the string and bool isValid.  The data should be private and it should throw if you someone tries to access an invalid string.

I have asked Marc and Kyle to see if a standard one exists.  If not I can write one as a template and you can use it.  No need to hold up this PR.

CaloReco/src/CaloHitMakerFast_module.cc
gianipez: No. The summer students working with my group have other projects assigned on the trigger front.

kutschke: I created an issue to followup on this general. 

If we do this in a few places the time cost is small.  If it's ubiquitous then it can add up to a large time cost that is invisible to profilers because it's distributed throughout the code.

kutschke:
@FNALbuild run build tests

kutschke:
@FNALbuild run build tests

kutschke:
@FNALbuild run build tests

Expect ceMix to fail until the input files are updated.

kutschke:
@FNALbuild run build tests

kutschke (approved):
This got lost in the hectic time of the collaboration meeting.  Approving now.  As you move forward please address the issues that I raised.

PR #503, opened 2021-06-25
Primaryfix

brownd1978:
Fix MC primary particle for muon stops with full genealogy.   This includes fixes for Validation.  It requires updates to JobConfig as well.


kutschke:
Off this thread Dave explained that this needed updates to fcl that were submitted to Production after this PR.  I will relaunch the CI to pick up the new fcl.

kutschke:
@FNALbuild run build tests

kutschke:
The issue with ceMix is "No Primary particle found" in event 2

brownd1978:
Yes, the sample used for cemix is now out of date.   Tomorrow I’ll try to rework that so it can follow changes better.   

David Brown

kutschke:
Will merge now because there is value in the nightly tests that we expect to run successfully,

PR #505, opened 2021-06-25
Fix MCPrimary for cosmics

brownd1978:
Bugfix for cosmic generators


kutschke:
@FNAL build run build tests

PR #512, opened 2021-06-29
changed how the digitization window is applied at the CRV

ehrlich-uva:
Changed how the digitization window is applied at the CRV.


CRVResponse/src/CrvWaveformsGenerator_module.cc
```diff
@@ -70,6 +78,10 @@ namespace mu2e
     EDProducer{pset},
     _crvSiPMChargesModuleLabel(pset.get<std::string>("crvSiPMChargesModuleLabel")),
     _singlePEWaveformFileName(pset.get<std::string>("singlePEWaveformFileName")),
+    _eventWindowMarkerLabel(pset.get<std::string>("eventWindowMarker","EWMProducer")),
```
brownd1978, line 81: Do these window times need to be synchronized with the parameters with the same name in CrvResponse?  If so, can you think of a way to coordinate these?  Maybe through DaqConditions?

CRVResponse/src/MakeCrvWaveforms.cc
```diff
@@ -63,14 +63,15 @@ void MakeCrvWaveforms::MakeWaveform(const std::vector<double> &times,
   {
     double timeOfCharge=*iterTime;  //the time when the charge happened
     double charge=*iterCharge/_singlePEReferenceCharge;  //scale it to the 1PE reference charge used for the single PE waveform
-    unsigned int waveformIndex = (timeOfCharge>startTime ? ceil((timeOfCharge-startTime)/digitizationPrecision) : 0);  //waveform index of the first digitization point for this particular charge
+    if(timeOfCharge<startTime) continue; //this shouldn't happen
```
brownd1978, line 66: Does this comment mean that physically it shouldn't happen?  Or that this condition signals an error in the code?

brownd1978 (commented):
I don't see any problems with this as an intermediate PR, but we eventually need a single-source for the digitization timing model, I believe Richie is planning on putting that in DaqConditions.

CRVResponse/src/MakeCrvWaveforms.cc
ehrlich-uva: The function that calls this function makes sure that it doesn't happen. So it means it's an error, if it happens. I will change the code to throw an exception here.

CRVResponse/src/CrvWaveformsGenerator_module.cc
ehrlich-uva: Yes, they need to be the same everywhere. I will change it to get the information from the conditions.

CRVResponse/src/CrvPhotonGenerator_module.cc
```diff
@@ -345,15 +356,16 @@ namespace mu2e
...
                 timeTmp = fmod(timeTmp,_microBunchPeriod);
                 //photons before the digitization start get removed except photons 
-                //in the first 55ns which get moved to the interval between the end of 
-                //the microbunch period and the digitization end
+                //in the first 55ns (digitization end - microbunch period) which get 
+                //moved to the interval between the end of the microbunch period and 
+                //the end of the digitization window.
                 if(timeTmp<digitizationEnd-_microBunchPeriod) timeTmp+=_microBunchPeriod;  
-                if(timeTmp<digitizationStart) continue;
+                if(timeTmp<digitizationStart-_digitizationStartMargin) continue;
```
bonventre, line 363: Does this part still work if digitizationEnd is > _microBunchPeriod?

ehrlich-uva: Yes, I tested it with a digitizationEnd of 1750ns.

PR #513, opened 2021-06-29
fixed calo-only sequence + clean up

gianipez:


Trigger/python/genTriggerFcl.py
```diff
@@ -25,13 +25,13 @@
 
 def appendEpilog(trig_path, relProjectDir, outDir, srcDir, verbose, doWrite, sourceFiles, targetFiles):
 
-    trk_filters      = ['EventPrescale','SDCountFilter','TCFilter', 'HSFilter', 'TSFilter']
-    helix_filters    = ['EventPrescale','SDCountFilter','TCFilter', 'HSFilter']
-    tc_filters       = ['EventPrescale','SDCountFilter','TCFilter']
```
brownd1978, line 30: I have seen a big cosmic ray shower light up an entire detector with many thousands of hits.  Are we sure we want to remove this filter?  How much does it cost to keep it?

brownd1978 (approved):
Thanks for the quick fix.  I will approve as a hot-fix.  We may want to conduct a trigger review at some point of all our trigger lines.

brownd1978:
build failing:
scons: *** [build/sl7-prof-e20-p008/Offline/gen/fcl/Trigger/OffSpillTrigMenu/OffSpillTrigMenu.fcl] Source `Offline/Trigger/data/caloMVANNCE/main_caloMVANNCEEventPrescale.fcl' not found, needed by target `build/sl7-prof-e20-p008/Offline/gen/fcl/Trigger/OffSpillTrigMenu/main.fcl'.

Trigger/python/genTriggerFcl.py
gianipez: keeping them would not fit the model of the Intensity stream I've been developing as it will prevent reco modules run in the early stage to evaluate those quantities. I think that we need to add protections to those modules. The events you are referring to I don't expect to be that frequent... worst case scenario the DEBUG stream implementation will keep us safe.

brownd1978:
@FNALbuild run build test

PR #514, opened 2021-06-30
Revert "changed how the digitization window is applied at the CRV"

brownd1978:


brownd1978:
I mistakenly merged this PR, reverting it now.

PR #518, opened 2021-07-01
changed how the digitization window is applied at the CRV

ehrlich-uva:
digitization window implementation at the CRV


brownd1978:
This is a repeat of the previous PR 512 Ralf submitted.  Due to an error on my part that PR got deleted (it is listed as merged, but it was backed out afterwards) and had to be re-submitted.  The comments made in 512 were previously addressed by Ralf, so we are starting this review from scratch.

bonventre (approved):

ehrlich-uva:
Rebased this PR to resolve conflicts that occurred after the recent merge with PR 523.

kutschke:
@FNALbuild run build tests

CRVResponse/src/MakeCrvWaveforms.cc
```diff
@@ -63,14 +63,15 @@ void MakeCrvWaveforms::MakeWaveform(const std::vector<double> &times,
...
-    size_t waveformIndex = ceil((timeOfCharge-startTime)/digitizationPrecision);  //waveform index of the first digitization point for this particular charge
+    double waveformIndexTmp = ceil((timeOfCharge-startTime)/digitizationPrecision);
+    if(waveformIndexTmp<0) waveformIndexTmp=0;
+    size_t waveformIndex = static_cast<size_t>(lrint(waveformIndexTmp));  //waveform index of the first digitization point for this particular charge
     double waveformTime = waveformIndex*digitizationPrecision + startTime;  //the time for this waveform index
 
     for(; ; waveformIndex++, waveformTime+=digitizationPrecision)
     {
       double singlePEWaveformTime = waveformTime - timeOfCharge;
-      if(singlePEWaveformTime<0) continue;  //this shouldn't happen
```
brownd1978, line 72: IIRC the previous conversation concluded that this should be a throw.   Did the changes above change this conclusion?

ehrlich-uva: The previous conversation was about another line, which I changed over the weekend. This line should not throw: There could be a very rare situation (which I have not observed in my test) that the singlePEWaveformTime is <0 due to precision issues when dealing with floating point numbers (after digitizing the time stamps and converting them back to floating point numbers). If this situation happens, the first point of the single PE waveform will be skipped, which will not result in a noticeable change of the total waveform. That's what the "continue" does.

brownd1978:
@FNALbuild run build test

brownd1978 (approved):
thanks for the explanation

oksuzian (approved):

PR #522, opened 2021-07-06
Fast calo cluster

bechenard:
This is a PR to fix the fast clustering algorithm. I made a few changes to the DAQ code as well, so Giani needs to look at it.

### DIFF
--- a/CaloCluster/src/CaloClusterFast_module.cc
+++ b/CaloCluster/src/CaloClusterFast_module.cc
@@ -92,7 +92,7 @@ namespace mu2e {
 
       std::vector<int> hits;
       hits.reserve(caloHits.size());
-      for (unsigned i=0;i<caloHits.size();++i) if (caloHits[i].energyDep() > EnoiseCut_ ** caloHits[i].nSiPMs() >= minNSiPM_) hits.emplace_back(i);      
+      for (unsigned i=0;i<caloHits.size();++i) if (caloHits[i].energyDep() > EnoiseCut_ && caloHits[i].nSiPMs() >= minNSiPM_) hits.emplace_back(i);      
       auto functorTime = [&caloHits](int a, int b) {return caloHits[a].time() < caloHits[b].time();};
       std::sort(hits.begin(),hits.end(),functorTime);
 
--- a/CaloCluster/src/CaloClusterFast_module.cc
+++ b/CaloCluster/src/CaloClusterFast_module.cc
@@ -33,7 +33,7 @@ namespace mu2e {
             fhicl::Atom<double>         EnoiseCut         { Name("EnoiseCut"),         Comment("Minimum energy for a hit to be in a cluster") }; 
             fhicl::Atom<double>         ExpandCut         { Name("ExpandCut"),         Comment("Minimum energy for a hit to expand cluster") }; 
             fhicl::Atom<double>         deltaTime         { Name("deltaTime"),         Comment("Maximum time difference between seed and hit in cluster") }; 
-            fhicl::Atom<int>            minNSiPM          { Name("minNSiPM"),          Comment("Minimum number of SiPM contributing to the hit") }; 
+            fhicl::Atom<int>            minSiPMPerHit     { Name("minSiPMPerHit"),     Comment("Minimum number of SiPM contributing to the hit") }; 
             fhicl::Atom<bool>           extendSearch      { Name("extendSearch"),      Comment("Search next-next neighbors for clustering") }; 
             fhicl::Atom<int>            diagLevel         { Name("diagLevel"),         Comment("Diag level"),0 }; 
         };
@@ -45,7 +45,7 @@ namespace mu2e {
           EnoiseCut_     (config().EnoiseCut()),
           ExpandCut_     (config().ExpandCut()),
           deltaTime_     (config().deltaTime()),
-          minNSiPM_      (config().minNSiPM()),
+          minSiPMPerHit_ (config().minSiPMPerHit()),
           extendSearch_  (config().extendSearch()),
           diagLevel_     (config().diagLevel())
         {
@@ -61,7 +61,7 @@ namespace mu2e {
         double            EnoiseCut_;
         double            ExpandCut_;
         double            deltaTime_;
-        int               minNSiPM_;
+        int               minSiPMPerHit_;
         bool              extendSearch_;
         int               diagLevel_;
 
@@ -92,7 +92,7 @@ namespace mu2e {
 
       std::vector<int> hits;
       hits.reserve(caloHits.size());
-      for (unsigned i=0;i<caloHits.size();++i) if (caloHits[i].energyDep() > EnoiseCut_ && caloHits[i].nSiPMs() >= minNSiPM_) hits.emplace_back(i);      
+      for (unsigned i=0;i<caloHits.size();++i) if (caloHits[i].energyDep() > EnoiseCut_ && caloHits[i].nSiPMs() >= minSiPMPerHit_) hits.emplace_back(i);      
       auto functorTime = [&caloHits](int a, int b) {return caloHits[a].time() < caloHits[b].time();};
       std::sort(hits.begin(),hits.end(),functorTime);
 
--- a/CaloCluster/src/CaloClusterFast_module.cc
+++ b/CaloCluster/src/CaloClusterFast_module.cc
@@ -33,6 +33,7 @@ namespace mu2e {
             fhicl::Atom<double>         EnoiseCut         { Name("EnoiseCut"),         Comment("Minimum energy for a hit to be in a cluster") }; 
             fhicl::Atom<double>         ExpandCut         { Name("ExpandCut"),         Comment("Minimum energy for a hit to expand cluster") }; 
             fhicl::Atom<double>         deltaTime         { Name("deltaTime"),         Comment("Maximum time difference between seed and hit in cluster") }; 
+            fhicl::Atom<int>            minNSiPM          { Name("minNSiPM"),          Comment("Minimum number of SiPM contributing to the hit") }; 
             fhicl::Atom<bool>           extendSearch      { Name("extendSearch"),      Comment("Search next-next neighbors for clustering") }; 
             fhicl::Atom<int>            diagLevel         { Name("diagLevel"),         Comment("Diag level"),0 }; 
         };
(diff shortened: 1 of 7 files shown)


CaloCluster/src/CaloClusterFast_module.cc
```diff
@@ -33,6 +33,7 @@ namespace mu2e {
             fhicl::Atom<double>         EnoiseCut         { Name("EnoiseCut"),         Comment("Minimum energy for a hit to be in a cluster") }; 
             fhicl::Atom<double>         ExpandCut         { Name("ExpandCut"),         Comment("Minimum energy for a hit to expand cluster") }; 
             fhicl::Atom<double>         deltaTime         { Name("deltaTime"),         Comment("Maximum time difference between seed and hit in cluster") }; 
+            fhicl::Atom<int>            minNSiPM          { Name("minNSiPM"),          Comment("Minimum number of SiPM contributing to the hit") };
```
brownd1978, line 36: Do values other than 1 or 2 make sense for minNSipm?  If not, you might make this a bool (bothSipms) instead.

brownd1978 (commented):
Beyond the diags this looks like a minor change, my only comment is on the parameterization.

bechenard:
Hi Dave,

So far, we have treated the number of SiPMs has an integer, so we might 
want to keep this logic. I'll wait to see what Giani has to say.

Cheers,

Bertrand

CaloCluster/src/CaloClusterFast_module.cc
gianipez: TBH, I think that it's confusing to cut on that variable here. This module will receive the CaloHitCollection already processed and calibrated, so if there are crystals read out only by one SiPM the energy will be calibrated accordingly. @bechenard do you see a specific use case I'm missing?

CaloReco/src/CaloHitMakerFast_module.cc
```diff
@@ -37,9 +37,8 @@ namespace mu2e {
             using Comment = fhicl::Comment;
             fhicl::Atom<art::InputTag> caloDigiCollection { Name("caloDigiCollection"), Comment("CaloDigi collection name") }; 
             fhicl::Atom<double>        digiSampling       { Name("digiSampling"),       Comment("Digitization time sampling") }; 
-            fhicl::Atom<double>        deltaTPulses       { Name("deltaTPulses"),       Comment("Maximum time difference between two signals ") }; 
-            fhicl::Atom<double>        pulseRatioMin      { Name("pulseRatioMin"),      Comment("Min value of energy ratio between pulses ") };
```
gianipez, line 41: mhmhmhmhmhm I much prefer the ratio w.r.t. to the residual for the energy check. This is what has been used so in the test beam analyses. The ratio scales nicely with the energy, while a crude cut on the residual does not.

CaloReco/src/CaloHitMakerFast_module.cc
```diff
@@ -70,8 +68,8 @@ namespace mu2e {
         art::ProductToken<CaloDigiCollection> caloDigisToken_;
         double              digiSampling_;
         double              deltaTPulses_;
-        double              pulseRatioMin_;
-        double              pulseRatioMax_;
+        double              deltaEPulse_;
+        bool                saveOnlyCombined_;
```
gianipez, line 72: this parameter is not used.

gianipez (changes requested):
Hi Bertrand, I made a few requests for changes/justifications. They look pretty straightforward.

CaloCluster/src/CaloClusterFast_module.cc
bechenard: @Giani: This is to filter noise hits only reconstructed by a single SiPM. This is left as an option as I noticed this can improve the resolution a little bit, but more testing is needed to see if that will translate into a gain on the calo trigger. I'll follow Dave's suggestions and change the name to make this more clear.

CaloReco/src/CaloHitMakerFast_module.cc
bechenard: Nope, this is part of what created the mess. The ratio doesn't work at all for low-energy deposits, so you will multiple hits instead of combining them. More precisely, suppose I have two pulses that reconstruct at 2.0 and 2.5 MeV each, then they will fail the ratio test, and the code will reconstruct two hits with 2 MeV each in the same crystal at the same time -> the clustering algorithm overestimates the energy. Using the difference fixes that issue, and 5 MeV is large enough to combine high energy hits in all practical cases. I would strongly suggest using the residual for the trigger (analysis is different)

CaloReco/src/CaloHitMakerFast_module.cc
bechenard: OOOps

CaloCluster/src/CaloClusterFast_module.cc
bechenard: Actually, the situation is more complicated since one SiPM can be dead, in which case it should also be considered as "bothSiPMs". Since we will implement all these subtleties when we'll add DB support, I'll defer the work to that point and keep my lovely integer for now

CaloReco/src/CaloHitMakerFast_module.cc
gianipez: This argument is not convincing me. with a 5 MeV tolerance, you will be combining a 2 MeV signal with a 5 MeV one, which is definitively not what we expect. If there is a problem with the noise simulation it should be addressed in a different way. If a ratio window of +/- 10% is not large enough (that is the value used in the test beam analysis) you should make it wider. In the future maybe we will make it a function/table w.r.t. to the pulse amplitudes, but crude residual sounds not optimal to me.

CaloCluster/src/CaloClusterFast_module.cc
gianipez: I still think that this selection should live in the module that creates the Hits. The clustering should be agnostic to the number of SiPM signals used to create a hit; flagging a hit a noisy is more complicated and I think we should not oversimplify it.

bechenard:
This is getting messy and I won't have much time to work on this soon, so I propose to close this PR and we'll fix it later after more discussions.

brownd1978:
@FNALbuild run build test

brownd1978 (approved):

brownd1978:
Since some of the work on single SiPM is defered to another PR, it might be a good idea to create an issue.

gianipez:
Hello guys,

I really don’t like the use of the “deltaEnergy” parameter in the CaloHitMakerFast. I advocate to go back to the ratio and enlarge the tolerance from 10% to 20%. If the noise model has an issue let’s turn that off up to the moment someone fixes it and let’s create an issue.

bechenard:
Hi Giani,

Even 20% won't cut it for the low energy deposits, and it is not 
entirely due to the noise model. I also think that we need to think 
robustness instead of optimal precision for the trigger. I have a few 
plots that shows the issue but I need to think a bit more about the 
various components. However, the current code fixes the issue and can be 
modified later if needed.

Bertrand

gianipez:
Bertrand, from the cosmic Run analyses performed with the Module-0 it has been shown that this ratio has an RMS of 5% (a cosmic deposits ~20 MeV in a crystal). Now, I can buy the fact that at lower energies we might need to increase this threshold due to neutrons/soft-photons contaminations, but if the sim delivers a difference larger than 20% that means that something is not well understood in the sim. Here is not a matter of robustness in my opinion. We need to understand what the sim is doing before launching the big production I guess.

kutschke:
A comment to see if GitHub will check for conflicts now that PR #523 is merged.

kutschke:
@FNAL build run build tests

ryuwd:
@FNALbuild run build tests

kutschke:
@ryuwd a lowish priority request" do you understand why my launch of the CI tests last night did not produce the log output but your launch this morning did?  When you get a chance, please explain to @HCasler how to diagnose what happened and how to resolve it.  Thanks,

ryuwd:
> @ryuwd a lowish priority request" do you understand why my launch of the CI tests last night did not produce the log output but your launch this morning did? When you get a chance, please explain to @HCasler how to diagnose what happened and how to resolve it. Thanks,

Hi Rob,
There was a misspelling of the command which means FNALbuild didn't trigger the build, so everything seems to be working OK

kutschke:
@ryuwd<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_ryuwd&d=DwMCaQ&c=gRgGjJ3BkIsb5y6s49QqsA&r=Zi5IH-k2_JiK_Wipiv3pwwSXO9UNaHiLP8sisfPz__k&m=CPYGjLr6dVuTnlRId08iH-0p06qhk0MYD5hE7RiEMFo&s=XjA5pj05dxiSRKcP0j9qaTmAxzHajSYWxKTThPlTIlg&e=> a lowish priority request" do you understand why my launch of the CI tests last night did not produce the log output but your launch this morning did? When you get a chance, please explain to @HCasler<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_HCasler&d=DwMCaQ&c=gRgGjJ3BkIsb5y6s49QqsA&r=Zi5IH-k2_JiK_Wipiv3pwwSXO9UNaHiLP8sisfPz__k&m=CPYGjLr6dVuTnlRId08iH-0p06qhk0MYD5hE7RiEMFo&s=-dGuCdOfzOZsV3XeTLQTZt2wlHgF-487vjd5KtgiTBg&e=> how to diagnose what happened and how to resolve it. Thanks,

Hi Rob,
There was a misspelling of the command which means FNALbuild didn't trigger the build, so everything seems to be working OK



Duh.  Thanks.  I see it now.  My apologies for the noise.

gianipez:
Hi @bechenard , I propose to discuss this at the coming calorimeter meeting to understand with the group who can help in getting the code fixed. I'm not very happy about the proposed changes to the hit-maker and the clustering modules, especially because I would like to see first a plot supporting the changes.

bechenard:
Hi Giani,

Ok, I can hopefully produce some more plots tomorrow. The more I think 
abut it, the more I believe we should simply scrape this energy cut...

Cheers,

Bertrand

brownd1978:
@FNALbuild run build tests

kutschke:
@FNALbuild run build tests

kutschke:
@FNALbuild run build tests

kutschke:
The recent build failure is because I retested with muse -1. new new default for CI.

It looks like the only change needed is to add the Offline/ prefix to all of the relevant include directives in CaloDiag/src/CaloNNTest_module.cc

I think that's a new file, which explains why the merge did not fix those lines.

bechenard:
The calorimeter group has suggested additional code developments, so I'm closing this PR and will work on it once I'm back from vacation (I'm tried of getting these emails...)

PR #525, opened 2021-07-07
STM trolley and upstream elements geometry update

michaelmackenzie:
Updates to the STM trolley and sweeper magnet geometry as well as upstream absorbers.
Validation performed:
Geant4 surface checking using old and new geometry configuration
ROOT overlap check using old and new geometry configuration

Summary of the geometry changes with validation images can be found here:
https://mu2e-docdb.fnal.gov/cgi-bin/private/ShowDocument?docid=38804

### DIFF
--- a/GeometryService/inc/STMMaker.hh
+++ b/GeometryService/inc/STMMaker.hh
@@ -52,7 +52,10 @@ namespace mu2e {
     double       _magnetHoleHalfLength;
     double       _magnetHoleHalfWidth;
     double       _magnetHoleHalfHeight;
+    double       _magnetHoleXOffset;
+    double       _magnetHoleYOffset;
     std::string  _magnetMaterial;
+    bool         _magnetHasLiner;
     double       _magnetField;
     bool         _magnetFieldVisible;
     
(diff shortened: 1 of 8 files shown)


dnbrow01:
This looks fine to me.

It's a matter of taste, but I'm not a fan of using the bool function hasLiner a bunch of times as an actual true/false book function and then once or twice as an int.  It works, I just think it makes the code read less cleanly.  I'd prefer something
like  
double zValue = blahblahblah;
if (blah.hasLiner) zValue += blahdeblah;

But I repeat, the code does work as is.

dnbrow01:
That was a comment about constructSTM.cc

dnbrow01 (approved):
Sorry, first comments in wrong place, I guess.

I approve the code as workable.  I do make one comment on style on the pull request page.

oksuzian (approved):
Looks good to CRV-D.

michaelmackenzie:
Hi Dave,

> This looks fine to me.
> 
> It's a matter of taste, but I'm not a fan of using the bool function hasLiner a bunch of times as an actual true/false book function and then once or twice as an int. It works, I just think it makes the code read less cleanly. I'd prefer something
> like
> double zValue = blahblahblah;
> if (blah.hasLiner) zValue += blahdeblah;
> 
> But I repeat, the code does work as is.

I removed this implicit conversion, replacing it with the more readable, multi-line definition.
Thanks for the recommendation!

Best,
Michael

brownd1978:
@FNALbuild run build test

brownd1978:
Thanks everyone for your timely reviews.  This PR is merged, as part of Offline v10_03_00.

PR #528, opened 2021-07-13
IPA resampler

sophiemiddleton:
Code allows use of Pileup_module to be used as a way to produce the IPA stops. I set decay fraction to "1" with Czernecki parameter (B = to that of Carbon). There will be a PR in the Production Repo with scripts to use this code


GlobalConstantsService/data/globalConstants_01.txt
```diff
@@ -40,6 +42,8 @@ double physicsParams.Cu.decayFraction  = 0.07;
 double physicsParams.Se.decayFraction  = 0.07;
 double physicsParams.Sb.decayFraction  = 0.04;
 double physicsParams.Au.decayFraction  = 0.03;
+double physicsParams.C.decayFraction  = 0.86;
+double physicsParams.IPA.decayFraction  = 1.0;
```
brownd1978, line 46: The other entries here are physics numbers, the 1.0 is what you want for a particular application.  Please change this to the physical number, it will just reduce the efficiency of the Michel electron calibration case.

GlobalConstantsService/data/globalConstants_01.txt
```diff
@@ -48,6 +52,8 @@ int    physicsParams.Cu.atomicNumber   = 29;
 int    physicsParams.Se.atomicNumber   = 34;
 int    physicsParams.Sb.atomicNumber   = 51;
 int    physicsParams.Au.atomicNumber   = 79;
+int    physicsParams.C.atomicNumber    = 6;
+int    physicsParams.IPA.atomicNumber  = 6; //Assume Carbon
```
brownd1978, line 56: Please put a 'TODO sign on these approximations (IPA = carbon)

GlobalConstantsService/data/globalConstants_01.txt
sophiemiddleton: Yes, I mean we could just use Carbon if that's the case and remove thje "IPA" instance

GlobalConstantsService/data/globalConstants_01.txt
sophiemiddleton: Sure

GlobalConstantsService/data/globalConstants_01.txt
sophiemiddleton: I guess, maybe not, because at some point we will update this at some point so maybe leave the IPA for now and add the TODO

brownd1978:
@FNALbuild run build test

brownd1978 (approved):

PR #531, opened 2021-07-15
rm duplicated fcl files

rlcee:
all these were moved to Production/Validation


kutschke (approved):
All tests passed and this does not need further review.

PR #538, opened 2021-07-24
fix trigger print

rlcee:
Since art allowed the trigger numbering to be specified, the trigger results print needs to change its access pattern.  It was crashing.


ryuwd:
@FNALbuild run build test

test if successful builds can be retriggered following Mu2e/CI#42

ryuwd:
@FNALbuild run build test

second time lucky

kutschke (approved):
This looks straightforward and does not need additional review.

PR #539, opened 2021-07-26
add calo MC prints

rlcee:
add routines for caloHitMC and caloClusterMC


kutschke:
@FNALbuild run build tests

kutschke (approved):
This looks straightforward and does not need additional review.

I relaunched the CI because of an intervening merge.  Presuming that returns OK I will merge.

PR #540, opened 2021-07-26
remove ReleaseNotes

rlcee:
- notes for Offline v6, 7, 8 are in the [wiki](https://mu2ewiki.fnal.gov/wiki/ReleaseList)
- find old notes by checking out Offline tag v10_05_00, see ReleaseNotes subdir


kutschke (approved):
This does not needed additional review.

PR #547, opened 2021-07-29
do-nothing commit

HCasler:


HCasler:
ok, let's see what happens when I add another commit

HCasler:
Looks like the issue is in the second step of the CI flow. When a commit is added to an existing PR, step 1 (Jenkins receives the webhook) seems to happen fine. In step 2, the CI scripts are supposed to decide whether to run tests or not, and then update the PR by marking the commit as "pending" and adding the "mu2e/buildtest" entry to the set of checks, and then trigger the actual running of the tests (actual tests are step 3). This is happening fine for new PR's, but not when new commits are added.

HCasler:
@FNALbuild run build tests

HCasler:
Weirdly, Jenkins is saying the comment that launched the tests running was the one where I described the Jenkins process, and not the one where I actually said to run the tests. Could be they were just close enough together that when Jenkins pulled down the PR on the former, it picked up the latter

ryuwd:
> This is happening fine for new PR's, but not when new commits are added.

Can you clarify what the undesired behaviour is? We deliberately do not re-trigger tests after new commits are added in order not to spam jenkins with jobs in the case that people `git push` lots of single commits in a short space of time

HCasler:
So as I go through this more, that makes sense. The issue to me was seeing the unfinished checks at the bottom, but maybe that's actually fine, since there are in fact checks that haven't completed (they just also haven't been asked for). It's looking like nothing is actually broken. 

Since we're in the middle of changes to the CI system I wanted to be sure nothing has gone sideways

kutschke:
I confirm that the desired behaviour is that when you push a new commit to an existing PR, the CI build is NOT triggered automatically.  The reason is as Ryun describes.

PR #551, opened 2021-07-30
Import optimized pbar absorber from SU2020

michaelmackenzie:
Importing the pbar window and wedge optimized by Pasha from SU2020. This was validated in SU2020,
and I ran a simple POT job  a Geant4 overlap check without any issues, and by eye it looked correct in
the ROOT geometry viewer. Both geom_common and geom_common_current are updated by updating
the underlying main geometry file.


kutschke:
The error is understood.  It is unrelated to this PR and will be fixed when I merge #548 later this evening.

kutschke:
@FNALbuild run build test

kutschke (approved):
Since this was approved for SU2020 I am approving it without further review.

PR #553, opened 2021-08-02
Remove very first SimParticle when compressing genealogy

AndrewEdmonds11:
In CompressDetStepMCs, we were keeping the very first SimParticle (i.e. the one that points to a GenParticle and has no parent SimParticle) even when we were compressing the genealogy. However, it is not necessary to keep this SimParticle since most of the important information is in the GenParticle. Also, mixing jobs are going over the memory limit in MDC2020, and we can reduce the memory footprint by removing these SimParticles.

With this PR, we now treat the very first SimParticle like all the others. To maintain a link to the GenParticle, the oldest surviving SimParticle now contains the GenParticlePtr.

I have left the code that was treating this very first SimParticle differently commented out. This is just in case we want to either put this back in, or (more likely) add something similar for important SimParticles (e.g. a stopped muon).

I have tested this with 100 MuStopPileup events and checked the verbose output makes sense. I will also see if the CI tests show any issues I may have missed.

### DIFF
--- a/Compression/src/CompressDetStepMCs_module.cc
+++ b/Compression/src/CompressDetStepMCs_module.cc
@@ -16,7 +16,6 @@
 // number of generations back you want to keep. The "oldest" SimParticle remaining
 // has a status code of "truncated" to identify that a truncation has occured
 // - Note 1: N = -1 means keep all generations (i.e. no compression)
-// - Note 2: the very first SimParticle (i.e. the one that has a valid Ptr to a GenParticle) is always kept
 //
 // Dec 2020, Andy Edmonds
 //
@@ -420,8 +419,7 @@ void mu2e::CompressDetStepMCs::compressSimParticles(const art::Event& event) {
     }
   }
 
-  // If we asked for the genealogy to be compressed, we will now end up with some missing links
-  // add them back as truncated SimParticles
+  // If we asked for the genealogy to be compressed, we will now end up with some missing links which we need to fix
   if (_keepNGenerations >= 0) {
     // Go through the particles we are keeping and see if any parents are not there
     for (auto& i_keptSimPart : _simParticlesToKeep[i_product_id]) {
@@ -449,7 +447,7 @@ void mu2e::CompressDetStepMCs::compressSimParticles(const art::Event& event) {
       }
     }
 
-    // We might have turncated some SimParticles so go through the output SimParticleCollection and fix the parent/child links
+    // We might have truncated some SimParticles so go through the output SimParticleCollection and fix the parent/child links
     for (auto& i_simParticle : *_newSimParticles) {
       mu2e::SimParticle& newsim = i_simParticle.second;
       if (newsim.isTruncated()) {
@@ -470,7 +468,14 @@ void mu2e::CompressDetStepMCs::compressSimParticles(const art::Event& event) {
             break; // don't need to go any further
           }
           else {
-            i_ancestorPtr = i_ancestorPtr->parent();
+            // If we have got to the very SimParticle (i.e. the one that points to the GenParticle)
+            if (i_ancestorPtr->isPrimary()) {
+              newsim.genParticle() = i_ancestorPtr->genParticle();// set this particle's GenParticlePtr
+              break; // don't need to go any further
+            }
+            else { // this is just another step in the genealogy
+              i_ancestorPtr = i_ancestorPtr->parent();
+            }
           }
         }
       }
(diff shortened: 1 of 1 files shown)


AndrewEdmonds11:
I can't see the log files but I've managed to recreate the ceMix failure locally. I hope to be able to get to this tomorrow.

kutschke:
There are two ways to see the log files.  The easy one is to log into a Fermilab machine, grab the url to the log file with your mouse and:

curl -O url

You can also do it from your remote machine if you are on VPN.  The other way requires creating a Jenkins account.

AndrewEdmonds11:
So the issue was that the "oldest" SimParticle whose genParticle I was setting, still had a SimParticlePtr that I needed to remove. Although the ceMix job was failing, it was an issue with the muDau job (since that is running genealogy compression). I've tested it locally and it works. Hopefully, the test should succeed now.

AndrewEdmonds11:
run build tests

HCasler:
@FNALbuild run build tests

kutschke:
Andy: Let's make sure that I understand.  Suppose we have

proton (on target) -> pion -> muon (decays) -> electron (beam flash) -> photon (brem in target) -> Compton electron 

and the Compton electron makes a hit.  This has a history of 6 steps.  I think you are saying that the revised history is 5 steps.  Is that right?  Or have I misunderstood?

If the history were:

proton (on target) -> proton ( interact in target) -> pion ....

Then the full history is 7 steps that you reduce to 6.  

Do I have right idea?

kutschke:
@FNAL build run build tests

kutschke:
@FNALbuild run build tests

AndrewEdmonds11:
> 
> 
> Andy: Let's make sure that I understand. Suppose we have
> 
> proton (on target) -> pion -> muon (decays) -> electron (beam flash) -> photon (brem in target) -> Compton electron
> 
> and the Compton electron makes a hit. This has a history of 6 steps. I think you are saying that the revised history is 5 steps. Is that right? Or have I misunderstood?
> 
> If the history were:
> 
> proton (on target) -> proton ( interact in target) -> pion ....
> 
> Then the full history is 7 steps that you reduce to 6.
> 
> Do I have right idea?

Hi Rob,

I'm not really changing the definition of `keepNGenerations` here. To use your example, we have (in terms of GenParticles and SimParticles):

Proton on target (GenParticle) <-- proton-on-target (SimParticle) <-- pion (SimParticle) <-- muon (SimParticle) <-- electron (SimParticle) <-- photon (SimParticle) <-- Compton electron (SimParticle) <-- Compton electron (StrawGasStep)

where the arrow indicate art::Ptrs to previous generations.

Currently, for `keepNGenerations=2`, we are keeping:

Proton on target (GenParticle) <-- proton-on-target (SimParticle) <-- electron(SimParticle) <-- photon (SimParticle) <-- Compton electron (SimParticle) <-- Compton electron (StrawGasStep)

where the Compton electron is generation 0, photon is generation 1, and electron is generation 2. Here we were keeping the proton-on-target SimParticle because it had a GenParticle Ptr rather than a parent SimParticle Ptr.

With this PR, for `keepNGenerations=2`, we would keep:

Proton on target (GenParticle) <-- electron(SimParticle) <-- photon (SimParticle) <-- Compton electron (SimParticle) <-- Compton electron (StrawGasStep)

where we no longer keep that very first proton-on-target SimParticle. Instead we give the electron a GenParticlePtr to point to the GenParticle.

Does that make sense?

kutschke:
Currently, for keepNGenerations=2, we are keeping:

Proton on target (GenParticle) <-- proton-on-target (SimParticle) <-- electron(SimParticle) <-- photon (SimParticle) <-- Compton electron (SimParticle) <-- Compton electron (StrawGasStep)

where the Compton electron is generation 0, photon is generation 1, and electron is generation 2. Here we were keeping the proton-on-target SimParticle because it had a GenParticle Ptr rather than a parent SimParticle Ptr.

With this PR, for keepNGenerations=2, we would keep:

Proton on target (GenParticle) <-- electron(SimParticle) <-- photon (SimParticle) <-- Compton electron (SimParticle) <-- Compton electron (StrawGasStep)

where we no longer keep that very first proton-on-target SimParticle. Instead we give the electron a GenParticlePtr to point to the GenParticle.

Does that make sense?
Thanks Andy,

Yes, that makes sense.  It was a very clear explanation.   I will merge it shortly.

  Rob

kutschke:
@FNALbuild run build tests

kutschke (approved):

PR #554, opened 2021-08-05
dbTool updates

rlcee:
- rename print-calibration to print-content and print-table to print-calibration
- add protection against committing extensions with overlapping IOV
- allow committing IOV with calibration data

### DIFF
--- a/DbService/inc/DbTool.hh
+++ b/DbService/inc/DbTool.hh
@@ -38,8 +38,8 @@ namespace mu2e {
     int run();
     int init();
 
+    int printContent();
     int printCalibration();
-    int printTable();
     int printIov();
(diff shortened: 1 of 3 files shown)


kutschke:
@rlcee as I read this we do not expect it to change the nightly validation so it is safe to merge with other similar PRs.  Is that right?

rlcee:
> @rlcee as I read this we do not expect it to change the nightly validation so it is safe to merge with other similar PRs. Is that right?

Yes, it only effects dbTool prints/commits, nothing in art.

kutschke:
@FNALbuild run build tests

kutschke (approved):

PR #555, opened 2021-08-09
fixed EventDisplay bug

ehrlich-uva:
-Bug fix in EventDisplay for the case of missing protonBunchTimes.
-Use art::Ptr::key() instead of art::Ptr<SimParticle>->id() to identify the correct trajectory in the trajectory collection. This change made it possible to read art files, where access to the art::Ptr<SimParticle> from the trajectory collection was not possible.
-Updated the standalone part of MakeCrvSiPMCharges.cc, which is used for testing purposes. It has no impact on the simulation.

### DIFF
--- a/CRVResponse/src/MakeCrvSiPMCharges.cc
+++ b/CRVResponse/src/MakeCrvSiPMCharges.cc
@@ -247,9 +247,9 @@ int main()
   CLHEP::RandFlat randFlat(engine);
   CLHEP::RandPoissonQ randPoissonQ(engine);
   mu2eCrv::MakeCrvSiPMCharges sim(randFlat,randPoissonQ,"/cvmfs/mu2e.opensciencegrid.org/DataFiles/CRVConditions/v6_0/photonMap.root");
-  sim.SetSiPMConstants(40, 40, 3.0, 500, 1695, 12.0, 8.84e-14, probabilities, inactivePixels);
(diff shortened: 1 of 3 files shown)


ehrlich-uva:
The change in MakeCrvSiPMCharges.cc is for a small section at the end of the code which allows the SiPM response to be compiled as a standalone program for testing purposes. This portion had not been updated in a while, and is not used by the Offline simulation chain. Therefore, this change has no impact on the simulation.

kutschke:
Thanks Ralf,

   Please update the PR comment to indicate this.

  Also I have one question about a hardcoded path.

Rob


The change in MakeCrvSiPMCharges.cc is for a small section at the end of the code which allows the SiPM response to be compiled as a standalone program for testing purposes. This portion had not been updated in a while, and is not used by the Offline simulation chain. Therefore, this change has no impact on the simulation.

kutschke (approved):
Thanks for updating the description text.  I added a suggestion for future reference.

ehrlich-uva:
There is no special environment for the standalone code. It gets compiled as shown in line 16 of the code.
The ConfigFileLookupPolicy gets used for this root file when the SiPM response is run in Offline.

kutschke:
Thanks Ralf.


There is no special environment for the standalone code. It gets compiled as shown in line 16 of the code.
The ConfigFileLookupPolicy gets used for this root file when the SiPM response is run in Offline.

PR #559, opened 2021-08-12
Fixed CaloHitMC bug

bechenard:
Also did a bit of cleaning

### DIFF
--- a/CaloMC/src/CaloHitTruthMatch_module.cc
+++ b/CaloMC/src/CaloHitTruthMatch_module.cc
@@ -141,8 +141,8 @@ namespace mu2e {
       auto pph = event.getValidHandle<PrimaryParticle>(ppToken_);
       auto const& primaryParticles = *pph;
       
-      std::unique_ptr<CaloHitMCCollection>  caloHitMCs(new CaloHitMCCollection);
-      std::unique_ptr<CaloHitMCTruthAssn>   caloHitMCTruth(new CaloHitMCTruthAssn);
+      std::unique_ptr<CaloHitMCCollection>   caloHitMCs(new CaloHitMCCollection);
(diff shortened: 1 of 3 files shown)


kutschke:
Is this the fix to the instability issue that Ray identified shortly before you left for vacation?  Or is it something else?

bechenard:
Hi Rob,

This is a fix for the bug Ray found. He temporarily put a check in the 
code using this data product to avoid the crash (bound checking was done).

Looking back at the discussion, I doubt that this module is the cause of 
the instability. It seems there is an extra SimParticle saved for the 
calorimeter, and that has to be done earlier in the code. What I suspect 
is that CaloHitTruthMatch_module is the first module where this becomes 
apparent.

To further debug, I would need much more information and copy of the 
data products. Is that available somewhere?

Cheers,

Bertrand

kutschke:
HI Bertrand,

   I will merge this PR.

   Ray can give you the information that you need to track down the outstanding instability  in reco.fcl .  It’s back again this morning.

Rob


From: Bertrand Echenard ***@***.***>
Date: Saturday, August 14, 2021 at 10:01 PM
To: Mu2e/Offline ***@***.***>
Cc: Rob Kutschke ***@***.***>, Assign ***@***.***>
Subject: Re: [Mu2e/Offline] Fixed CaloHitMC bug (#559)

Hi Rob,

This is a fix for the bug Ray found. He temporarily put a check in the
code using this data product to avoid the crash (bound checking was done).

Looking back at the discussion, I doubt that this module is the cause of
the instability. It seems there is an extra SimParticle saved for the
calorimeter, and that has to be done earlier in the code. What I suspect
is that CaloHitTruthMatch_module is the first module where this becomes
apparent.

To further debug, I would need much more information and copy of the
data products. Is that available somewhere?

Cheers,

Bertrand

.

PR #562, opened 2021-08-18
add three new dbTool functions

rlcee:
- print-set to print an entire calibration set to a text file
- verify-set to verify that a calibration set is complete for a set of runs
- commit-patch which repairs a calibration set, creating a new version

No effect on any operations or validation


brownd1978:
@FNALbuild run build test

PR #571, opened 2021-09-01
Fixed bug with TrkStrawStatusLong/Short overlaps

bonventre:
Bug fix that allows TrkStrawStatusLong and Short to set flags on the same channel without throwing an exception

### DIFF
--- a/TrackerConditions/src/TrackerStatusMaker.cc
+++ b/TrackerConditions/src/TrackerStatusMaker.cc
@@ -59,11 +59,25 @@ namespace mu2e {
     for (auto const& row : tpls_p->rows()) estatus.insert(TrackerElementStatus(row.id(),tpls_p->sidMask(),row.status()));
     for (auto const& row : tpas_p->rows()) estatus.insert(TrackerElementStatus(row.id(),tpas_p->sidMask(),row.status()));
     for (auto const& row : tssl_p->rows()) estatus.insert(TrackerElementStatus(row.id(),tssl_p->sidMask(),row.status()));
-    for (auto const& row : tsss_p->rows()) estatus.insert(TrackerElementStatus(row.id(),tsss_p->sidMask(),row.status()));
+    unsigned ntotal = tpls_p->rows().size() + tpas_p->rows().size() + tssl_p->rows().size();
+    // tsss can have duplicate entries from tssl
+    for (auto const& row : tsss_p->rows()){
+      auto elem = TrackerElementStatus(row.id(),tsss_p->sidMask(),row.status());
+      auto iter = estatus.find(elem);
+      if (iter == estatus.end()){
+        estatus.insert(elem);
+        ntotal++;
+      }else{
+        TrackerElementStatus temp = *iter;;
+        estatus.erase(iter);
+        temp.status_.merge(row.status());
+        estatus.insert(temp);
+      }
+    }
 
     // check for consistency
     //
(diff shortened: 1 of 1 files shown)


TrackerConditions/src/TrackerStatusMaker.cc
```diff
@@ -59,11 +59,25 @@ namespace mu2e {
     for (auto const& row : tpls_p->rows()) estatus.insert(TrackerElementStatus(row.id(),tpls_p->sidMask(),row.status()));
     for (auto const& row : tpas_p->rows()) estatus.insert(TrackerElementStatus(row.id(),tpas_p->sidMask(),row.status()));
     for (auto const& row : tssl_p->rows()) estatus.insert(TrackerElementStatus(row.id(),tssl_p->sidMask(),row.status()));
-    for (auto const& row : tsss_p->rows()) estatus.insert(TrackerElementStatus(row.id(),tsss_p->sidMask(),row.status()));
+    unsigned ntotal = tpls_p->rows().size() + tpas_p->rows().size() + tssl_p->rows().size();
+    // tsss can have duplicate entries from tssl
+    for (auto const& row : tsss_p->rows()){
+      auto elem = TrackerElementStatus(row.id(),tsss_p->sidMask(),row.status());
+      auto iter = estatus.find(elem);
```
brownd1978, line 66: I think this exposes a problem, that TrackerElementStatus operator < can fail when comparing 2 objects at different levels (say panel and straw).  ie SID 1,1,0 (straw) will match SID 1,1,0 (panel).   We want to allow overlapping Straw and Panel  status objects, so we want operator < to also look at the mask:
 bool operator < (const TrackerElementStatus& other) const { return mask_ < other.mask_ || (mask_ == other.mask_ && sid_ < other.sid_); }

TrackerConditions/src/TrackerStatusMaker.cc
```diff
@@ -59,11 +59,25 @@ namespace mu2e {
...
+      if (iter == estatus.end()){
+        estatus.insert(elem);
+        ntotal++;
+      }else{
+        TrackerElementStatus temp = *iter;;
+        estatus.erase(iter);
```
brownd1978, line 72: Why can't you just do iter->status_.merge(elem.status_)?
I also think the ntotal test is no longer sensible and should be removed.

bonventre: I was having some const issue here. Unless we change the comparison operator and allow all the other tables to merge duplicate rows ntotal still makes sense (checks that you don't set status of a panel/plane twice since that won't work)

TrackerConditions/src/TrackerStatusMaker.cc
bonventre: Probably true - if we merge this PR for now I can continue testing VST in the meantime and we can come back to fix this

TrackerConditions/src/TrackerStatusMaker.cc
brownd1978: Right, I remember now: set won't allow you to modify an element as that might change the sort, invalidating its internal structure.  Remove and replace works.  The alternative is to use something besides set as estat_t.  But for now lets use your solution.

TrackerConditions/src/TrackerStatusMaker.cc
brownd1978: OK.  I'd like to fix this before we forget.  Lets merge this then I'll fix operator <.

brownd1978 (approved):

PR #576, opened 2021-09-08
order on query

rlcee:
The straw alignment table assumed the delivery order of rows, but that is not guaranteed by postgres and postgres recently reminded us of this policy.  I have not been checking for this issue and I can't recall my original thinking or plans on it.  I will have to check the other tables to make sure they are not relying on the same assumption.

### DIFF
--- a/DbTables/inc/DbTable.hh
+++ b/DbTables/inc/DbTable.hh
@@ -38,6 +38,8 @@ namespace mu2e {
     // approx size in bytes - overridden by derived class
     virtual std::size_t size() const =0;
     std::size_t baseSize() const { return _csv.capacity(); }
+    // if table needs to be sorted
+    virtual const std::string orderBy() const {return std::string();}
 
     // take the cvs text from a query and build out the table contents
     int fill(const std::string& csv, bool saveCsv=true);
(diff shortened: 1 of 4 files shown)


rlcee:
Forgot to say, Richie is waiting for this fix in his VST work, so it would be good to get it in quickly.

DbTables/inc/DbTable.hh
```diff
@@ -38,6 +38,8 @@ namespace mu2e {
     // approx size in bytes - overridden by derived class
     virtual std::size_t size() const =0;
     std::size_t baseSize() const { return _csv.capacity(); }
+    // if table needs to be sorted
+    virtual const std::string orderBy() const {return std::string();}
```
kutschke, line 42: Should we make this pure virtual to force derived classes to supply an orderBy function? Or are there cases in which there is no meaningful order so an empty string makes sense as a default?

kutschke (commented):
This looks straightforward.

DbTables/inc/DbTable.hh
rlcee: Most tables do not need to be ordered, so I did not want to force each table to declare an order.  But this a bit of an open question until all tables have been reviewed and I can think about it a bit more.  I will consider this question as part of the process.

kutschke: Thanks.   I will approve and merge now.

kutschke (approved):

PR #577, opened 2021-09-11
Add functionality to dump contents to some analyzers

brownd1978:
dump a text version of the BField map and a TTree of some StepPointMC info.  The new features are used in a toy MC.


Analyses/src/BFieldPlotter_module.cc
```diff
@@ -118,34 +127,46 @@ namespace mu2e {
...
 
     //if a null map, plot the default field from the manager
     const std::string name = (map) ? map->getKey() : "default";
-
     if(_hMap[name]) return; //already filled the maps in a previous event
 
+    // if dumping, create the file
+    std::fstream fs;
+    if(_dump){
+      string dumpfile= name+"Dump.csv";
```
michaelmackenzie, line 152: Should this file name have a configurable tag? To run different configurations without overwriting previous runs.

michaelmackenzie (approved):
Everything looks good to me, and I appreciate the cleanup of the white space I left in the BFieldPlotter module!

Best,
Michael

brownd1978:
@FNALbuild run build test

Analyses/src/BFieldPlotter_module.cc
brownd1978: I made this configurable.  Thanks for your review.

Analyses/src/BFieldPlotter_module.cc
```diff
@@ -44,6 +47,9 @@ namespace mu2e {
       fhicl::Atom<double>      axisTwoMin{Name("axisTwoMin"), Comment("Axis two lower edge (bin centered) for plotting (mm) (plane = x/y/z --> axis two = z/z/y)")};
       fhicl::Atom<double>      axisTwoMax{Name("axisTwoMax"), Comment("Axis two upper edge (bin centered) for plotting (mm) (plane = x/y/z --> axis two = z/z/y)")};
       fhicl::Atom<double>      mapBinSize{Name("mapBinSize"), Comment("Map bin size (mm) (must be a divisor of both axis lengths)"), 10.};
+      fhicl::Atom<bool>	       dump      {Name("dump"),       Comment("Dump map samples to a CSV file"), false};
+      fhicl::Atom<std::string> dumpName  {Name("dumpName"),   Comment("Suffix to the dump file name"),"Dump"};
+      fhicl::Atom<bool>	       detector  {Name("detector"),   Comment("Use Detector coordinate system"), false};
```
gaponenko, line 52: The indentation in this block looks strange in github.  Are there tabs in the new code? If so, please replace with spaces.

Analyses/src/BFieldPlotter_module.cc
```diff
@@ -118,34 +130,46 @@ namespace mu2e {
...
 					<< _axisTwoMin << " < axisTwoValues < " << _axisTwoMax << ", "
 					<< "step size = " << _mapBinSize;
 
     //if a null map, plot the default field from the manager
     const std::string name = (map) ? map->getKey() : "default";
-
     if(_hMap[name]) return; //already filled the maps in a previous event
 
+    // if dumping, create the file
+    std::fstream fs;
```
gaponenko, line 153: Suggest using std::ofstream instead of std::fstream here.

Analyses/src/BFieldPlotter_module.cc
```diff
@@ -159,23 +183,35 @@ namespace mu2e {
...
+	else if(_plane == "z")
 	  point = CLHEP::Hep3Vector(axisOne, axisTwo, _planeValue);
 	if(map)
 	  map->getBFieldWithStatus(point,field);
 	else //if null, get the manager and plot the field returned by it
 	  bf->getBFieldWithStatus(point,field);
 
 	_hMap[name]->Fill(axisOne, axisTwo, field.mag()); //fill with weight of the field magnitude
+
+	if(_detector) point = det->toDetector(point);
```
gaponenko, line 199: Redefining  "point" here may be a landmine for the future.   If one adds new code using "point" above or below this line things will behave differently.  Even worse, the breakage depends on specific job config.
Please use a new variable like
const CLHEP::Hep3Vector dumpPoint = detector_ ? point : det->toDetector(point);

Analyses/src/BFieldPlotter_module.cc
```diff
@@ -118,34 +130,46 @@ namespace mu2e {
...
 
+    // if dumping, create the file
+    std::fstream fs;
+    if(_dump){
+      string dumpfile= name+_dumpname+".csv";
+      fs.open (dumpfile.c_str(), fstream::out);
+      if(fs.is_open()){
+	fs << "# Dump of " << name << endl;
+	fs <<  "#  x   ,  y   ,   z   ,  Bx   ,  By   ,  Bz" << endl;
+      }
```
gaponenko, line 160: Adding an else clause that throws an exception on failure to open the file would be prudent.

Analyses/src/StepPointMCDumper_module.cc
```diff
@@ -189,12 +206,22 @@ namespace mu2e {
...
-
-      if(writeProperTime_) {
-        tau_ = SimParticleGetTau::calculate(i, spMCColls, decayOffCodes_);
+      if(writeVDHit_)hit_ = VDHit(i);
+      if(writeParticleState_) {
+	KinKal::VEC3 pos;
+	if(detectorSystem_)
+	  pos = KinKal::VEC3(det->toDetector(i.position()));
+	else
+	  pos = KinKal::VEC3(i.position());
```
gaponenko, line 215: Suggest to initialize pos in declaration:

const KinKal::VEC3 pos = detectorSystem_ ? 
   KinKal::VEC3(det->toDetector(i.position())) :
   KinKal::VEC3(i.position());

gaponenko (changes requested):
Hello,
I have made a few technical comments on the code.  The one I feel strongly about is the re-definition of the "pos" variable in the middle of code; the others are minor improvements.

Analyses/src/BFieldPlotter_module.cc
brownd1978: Good catch, there was a hidden tab.

Analyses/src/BFieldPlotter_module.cc
brownd1978: OK.

Analyses/src/BFieldPlotter_module.cc
brownd1978: OK

Analyses/src/StepPointMCDumper_module.cc
brownd1978: I have mixed feelings about inline conditionals, I use them, but I think younger coders find them hard to read.  I think there is no performance benefit with modern compilers.

brownd1978:
@FNALbuild run build test

Analyses/src/StepPointMCDumper_module.cc
gaponenko: Hi Dave,
This is not so much about performance (even though it can only help), but about not having "dummy-initialized" or "wrong-initialized" variables.  WIthout inline conditionals one can not make pos here const, and having it const is a good safety and readability thing.  Once I see a variable declared const I don't have to worry about it having a different value couple code pages down; it helps.
About younger coders: the more they see that style the more used to it they will become.
Andrei

Analyses/src/BFieldPlotter_module.cc
```diff
@@ -137,76 +137,76 @@ namespace mu2e {
...
       fs.open (dumpfile.c_str(), fstream::out);
       if(fs.is_open()){
-	fs << "# Dump of " << name << endl;
-	fs <<  "#  x   ,  y   ,   z   ,  Bx   ,  By   ,  Bz" << endl;
-      }
-
+        fs << "# Dump of " << name << endl;
+        fs <<  "#  x   ,  y   ,   z   ,  Bx   ,  By   ,  Bz" << endl;
+      } else
+        throw cet::exception("BADCONFIG") << "Cannot open file " << dumpfile << endl;
```
gaponenko, line 161: Can we please add braces after the "else"?

brownd1978: does this code produce erroneous results?  is it incorrectly indented?  Does it violate Mu2e coding standards?  If not, why are you requesting a change?

gaponenko: I think a Mu2e coding standard doc asked for braces.   I can't locate that docdb at the moment, but I think it makes sense.  Especially in the current case where the "if" has braces -  "else" without them looks weird.

rlcee: Everything we had for coding standards is [here](https://mu2ewiki.fnal.gov/wiki/CodingStandards) and links therein. Maybe you are thinking of [this doc](https://cdcvs.fnal.gov/redmine/attachments/download/19145/CodingStandards.pdf).  After a quick glance through, I didn't find this question addressed.  If there is another doc, let me know so I can link it.

gaponenko:
Hi Ray,

On Thu, 16 Sep 2021, Ray Culbertson wrote:

> Everything we had for coding standards is
> [here](https://mu2ewiki.fnal.gov/wiki/CodingStandards) and links
> therein. Maybe you are thinking of [this
> doc](https://cdcvs.fnal.gov/redmine/attachments/download/19145/CodingStandards.pdf).
> After a quick glance through, I didn't find this question addressed.
> If there is another doc, let me know so I can link it.

Thank you for the links.  Both of these are much newer than what I
remember.  It was something I read soon after first joining Mu2e,
probably in 2011 or 2012.  Rob might know.  I would add a statement on
braces to the wiki, but may be it's just my preference and we should
keep it free-for-all instead.

Andrei

gaponenko (approved):
Dave, thank you for updating the code.
Andrei

Analyses/src/BFieldPlotter_module.cc
kutschke: I have a preference for braces but I don't believe we never formalized it because it was too contentious.

kutschke (approved):

brownd1978:
@FNALbuild run build test

PR #579, opened 2021-09-13
row order review

rlcee:
We were not fully prepared for postgres to return rows in arbitrary order.  This commit is the result of reviewing the tables we have
- val* tables were not vulnerable, but adding these sorts makes the results more readable and prepared for future optimizations
- trk tables were vulnerable and are now prepared
- MVAToolDb was protected, but now won't need to throw
- SimEfficiencies, TrkElelmentStatus do not assume an order
- CalRoIDMap* are vulnerable, but I need to understand better how the indices work before I can fix it.  There are also other aspects of these tables I'd like improve.  I vaguely recall they are not complete. They do not exist in the database yet.

Nothing has to be done by any users of the code - the previously built-in assumptions of the code are now enforced.

### DIFF
--- a/DbService/src/DbReader.cc
+++ b/DbService/src/DbReader.cc
@@ -105,26 +105,47 @@ int mu2e::DbReader::fillValTables(DbValCache& vcache) {
   // load up the queries
   qfv[0].select = tables.query();
   qfv[0].table = tables.dbname();
+  qfv[0].order = tables.orderBy();
+
   qfv[1].select = calibrations.query();
   qfv[1].table = calibrations.dbname();
+  qfv[1].order = calibrations.orderBy();
+
   qfv[2].select = iovs.query();
   qfv[2].table = iovs.dbname();
+  qfv[2].order = iovs.orderBy();
+
   qfv[3].select = groups.query();
   qfv[3].table = groups.dbname();
+  qfv[3].order = groups.orderBy();
+
   qfv[4].select = grouplists.query();
   qfv[4].table = grouplists.dbname();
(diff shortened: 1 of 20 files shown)


brownd1978 (approved):

brownd1978:
Do any of the CI tests invoke the database?  Does the nightly?  If now, how will we know if these commits cause problems?

rlcee:
> Do any of the CI tests invoke the database? Does the nightly? If now, how will we know if these commits cause problems?

No CI or nightly runs the database code.  I successfully ran several simple tests that show the effect of the changes.  A simple calibration set can be loaded in the engine and correctly evaluated.   

We used to run the database in the nightly, but I've forgotten what change/update caused us to drop it temporarily. (BTW, we still do not have reliable service from the query engine.)  A month ago, Richie and I discussed this and we ran in the same problem we have before, that running the database with zeros for alignment produces a slightly different result (probably due to round-off) than not using the database.  The solution here would be for the no-database option to rebuild the geometry the same way that the database with zeros does, then it would get the same round-off and the answer should be identical.  Or we can ignore this effect.  I can turn the database back on at anytime, if you tell me what makes sense to run.

It would make sense to have a database code test suite.  I've thought about this and what I was coming up with seems complicated and maintenance heavy.  It would have to track updates to the code and its capabilities, and add any new Proditions to its tests.  I'll add some thoughts to my ongoing talk at the comp/soft meeting.

Richie is the only one who is just starting to use the database in earnest, so a good check would be for him to run his reco or analysis, whatever depends on the database.

PR #580, opened 2021-09-16
Added Flat Options to the FlatDaughter module

sophiemiddleton:


EventGenerator/src/FlatMuonDaughterGenerator_module.cc
```diff
@@ -112,10 +112,18 @@ namespace mu2e {
     double randomMom = randFlat_.fire(startMom_, endMom_);
     double randomE = sqrt(particleMass_*particleMass_ + randomMom*randomMom);
     double time = mustop->endGlobalTime() + randExp_.fire(muonLifeTime_);
-    
     PDGCode::type pid = static_cast<PDGCode::type>(pdgId_);
+    
+    if (pid == 13) { process = ProcessCode::mu2eFlateMinus; }
```
kutschke, line 117: Please replace all numeric values of PDGCodes with the enum name from PDGCode.hh .   I have asked you to do this before.  Let's fix these as targets of opportunity - so please fix all occurrences in this file.

Also, move the setting of the data member named process from the produce member function to the c'tor.  You only need to do the work once so do it once;  in this case the cost is tiny but just get into the habit of always doing it right so that you do it right when it does matter.  A second reason is that you want to report misconfigurations at the earliest possible opportunity.

kutschke (changes requested):

EventGenerator/src/FlatMuonDaughterGenerator_module.cc
sophiemiddleton: Ok sure, will do

sophiemiddleton: Should be done now, let me know if this is OK

kutschke: It looks good.  Thanks.

kutschke:
@FNALbuild run build tests

kutschke (approved):

PR #584, opened 2021-09-17
2-D Projections of CRV 

NamithaChitrazee:
- 2-D XY and YZ projection tabs of the CRV are added. The hits and the corresponding hit bars are displayed here.
- The "hit" bars are displayed in the 3-D view as well
- Run/Subrun/Event number label added


sophiemiddleton:
@kutschke this is the first stage of a few updates we will put in in the next few weeks. I also updated the PID to use PDGCode (you can see that in this PR since Namitha merged that in from my branch). This PR is mostly to get Namitha's CRV code added in which I think is in good shape, so I'm happy with things for now. I'm going to do a full check through all the TEve code once this is merged and remove any excess and improve a few things which are mostly legacy from previous work, so I'm just reviewing Namitha's new code for now.

sophiemiddleton (approved):
I'm happy with the new additions for the CRV projection. I'll do some more general code upgrades on the TEveMu2e package after this.

kutschke:
@FNALbuild run build test

PR #591, opened 2021-09-21
remove CRLF newlines

rlcee:
These files had CRLF newlines, this changes them to LF.  I'm not sure if we have a policy, but I find this can cause little issues like mysterious diffs, so better to be consistent.


kutschke (approved):
Thanks for doing this. It's been an irritant for a while.  I confirm that the only changes are white space changes.

PR #592, opened 2021-09-21
TEve Code Tidy Ups

sophiemiddleton:
Some basic changes to prevent seg faults and following previous reviews....more to follow


TEveEventDisplay/src/TEveEventDisplay_module.cc
```diff
@@ -100,11 +100,13 @@ namespace mu2e
       application_ = new TApplication( "noapplication", &tmp_argc, tmp_argv );
     } 
     //construct GUI:
-    _frame = new TEveMu2eMainWindow(gClient->GetRoot(), 1000,600, _pset);
-    //build 2D geometries:
-    _frame->CreateCaloProjection();
+    
+    DrawOptions DrawOpts(_filler.addCrvHits_, _filler.addCosmicSeedFit_, _filler.addTracks_, _filler.addClusters_, _filler.addHits_, false, _filler.addMCTraj_); 
+    _frame = new TEveMu2eMainWindow(gClient->GetRoot(), 1000,600, _pset, DrawOpts);
```
kutschke, line 105: Does root/Eve take ownership of this pointer and delete the pointer it at the end of the job?  Or are you responsible for deleting it at the end of the job?

TEveEventDisplay/src/TEveMu2e_base_classes/TEveMu2eMCInterface.h
```diff
@@ -40,6 +40,9 @@ namespace mu2e{
         TEveElementList *fTrackList2DXZ;
         TEveElementList *fTrackList3D;
         std::vector<int> particleIds_;
+        constexpr Int_t linewidth(){ return 3; }
+        constexpr Int_t fontsize(){ return 15; }
+        constexpr double text_z_pos(){ return 3000.; }
```
kutschke, line 45: I recommend doing these as variables not as functions.  For example

constepxpr Int_t linewidth = 3;

TEveEventDisplay/src/TEveMu2e_base_classes/TEveMu2eMainWindow.h
```diff
@@ -72,16 +72,34 @@ class TText;
...
+  struct DrawOptions{
+    // data options
+    bool addCRVInfo = false;
+    bool addCosmicTracks = false;
+    bool addTracks = false;
+    bool addClusters = false; 
+    bool addComboHits = false;
+    bool addCryHits = false;
+    bool addMCTraj = false;
+    DrawOptions(){};
```
kutschke, line 85: The advice we have from the experts is that if the compiler written d'tor will do the right thing, don't write your own d'tor, even an empty one.  This case is an example of that.

TEveEventDisplay/src/TEveMu2e_base_classes/TEveMu2eMainWindow.h
```diff
@@ -109,17 +130,18 @@ namespace mu2e{
       TEveMu2eMCInterface *pass_mc	= new TEveMu2eMCInterface(); 
       TEveMu2eProjectionInterface *pass_proj = new TEveMu2eProjectionInterface();
       
-      double texttime = -1;
+      // data options
+      DrawOptions DrawOpts;
       std::vector<double> *clusterenergy = 0;
```
kutschke, line 135: The preferred style is to use nullptr, not 0. The times that it matters are rare but just always do it correctly and you don't need to think about when it matters.

TEveEventDisplay/src/TEveMu2e_base_classes/TEveMu2eMCInterface.h
kutschke: You might also consider creating a "style" class that has default values - that would allow you to treat the style options as a unit and should reduce repetition and clutter in the code.  Whether or not this really makes sense depends on details of how it will be used.  Please consider this idea as you move forward.

sophiemiddleton: yes, i thought the same thing actually

TEveEventDisplay/src/TEveEventDisplay_module.cc
sophiemiddleton: I think it deletes it

TEveEventDisplay/src/TEveMu2e_base_classes/TEveMu2eMainWindow.h
```diff
@@ -109,17 +130,18 @@ namespace mu2e{
...
       std::vector<double> *hitenergy = 0;
       std::vector<double> times;
 
       TEvePad *fPad;
-	TEvePad	*fPadCRV;
+      TEvePad	*fPadCRV;
```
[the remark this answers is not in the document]
sophiemiddleton: so make them nullptrs?

TEveEventDisplay/src/TEveMu2e_base_classes/TEveMu2eMainWindow.h
sophiemiddleton: sure, I am making that transition now, this =0 thing has caused us a lot of issues with seg faults later on, which is my I am changing to this DrawOpt, I should have updated the header too

TEveEventDisplay/src/TEveEventDisplay_module.cc
kutschke: I think it's a good idea to add a comment about that at the creation of every object that root will eventually delete for you.

A better idea is for us to write a trivial class template that, under the covers, is just a bare pointer.  The idea is that when you create an object that root will eventually take ownership of, then you make it this type of pointer.  This explains to future maintainers who is responsible for deleting pointers.  It also makes it easier to audit code to ensure that there are no memory leaks.

If I write this class template, are you willing to use it?

TEveEventDisplay/src/TEveMu2e_base_classes/TEveMu2eMainWindow.h
kutschke: I don't have time to look at this carefully until next week but here is an initial observation.  I looked at how the variable clusterenergy is used inside TEveMu2eMainWindow.cc .    It looks like you never really use this a a pointer - you always copy into it.  So it should be an object, not a pointer to an object.

TEveEventDisplay/src/TEveMu2e_base_classes/TEveMu2eMainWindow.h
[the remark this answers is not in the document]
kutschke: Yes, initialize them to nullptr here.

TEveEventDisplay/src/TEveMu2e_base_classes/TEveMu2eMainWindow.h
sophiemiddleton: ah. this has to be 0 since it does the whole max/min thing later on. I will make all the others nullptrs though

sophiemiddleton:
Will get back on this tomorrow. Some of the changes I implemented following your suggestions caused a seg fault and I'm trying to chase it while also in a conference

kutschke:
I think that I am still waiting for your to resolve your seg fault and make a new commit.  Is that right?

TEveEventDisplay/src/TEveEventDisplay_module.cc
sophiemiddleton: I can attempt to use it if written, if ROOT/TEve doesnt complain, a lot of this is balancing goof practice to avoiding pitfalls in TEve

kutschke:
@sophiemiddleton is this ready for me to relaunch the tests or is there more to come?

sophiemiddleton:
I will let you know when it's ready, I noticed a small, but non-breaking issue with the event

sophiemiddleton:
Ok @kutschke fixed the event labels - you can launch tests now

kutschke:
@FNALbuild run build tests

HCasler:
Related to the CI and codetools changes that just got merged: I can confirm that this PR was built with the correct branches of Offline and Production -- main, in this case:
[Mon Sep 27 14:09:37 CDT 2021] setup Mu2e/Offline
[2021-09-27 14:09:38,484] Mu2eCI:INFO  |  API Rate Limit: 4785/5000, Reset in 1986 sec i.e. at 2021-09-27 14:42:39
[2021-09-27 14:09:39,165] Mu2eCI:INFO  |  Mu2e/Offline PR 592 base ref is main; the commit sha is 6f0c8d2308718cdca1374d43e2342eaf8a988d19
Cloning into 'Offline'...
Cloning into 'Production'...
From https://github.com/Mu2e/Offline
 * [new ref]         refs/pull/592/head -> pr592

kutschke (approved):

PR #593, opened 2021-09-24
CaloDigi and CaloReco fixes

bechenard:


CaloMC/src/CaloDigiMaker_module.cc
```diff
@@ -136,6 +136,7 @@ namespace mu2e {
       pulseShape_.buildShapes();
             
       noiseGenerator_.initialize(wfExtractor_); 
+noiseGenerator_.plotNoise("noise.pdf");
```
rlcee, line 139: was this a debug line?  It looks like it will always produce a file

CaloReco/src/CaloTemplateWFProcessor.cc
```diff
@@ -258,9 +258,9 @@ namespace mu2e {
 	for (unsigned j=0;j<xvec.size();++j) ywork[j] -= fmutil_.eval_fcn(xvec[j]);
```
rlcee, line 258: ywork is sized by yvec and filled to the limits of xvec.  I would guess that xvec and yvec are always supposed to be the same size, so it doesn't matter, but it is a little harder to follow, and the sane-size requirement is not noted anywhere. Would you mind putting in a check that they are the same size, or at least a note?

rlcee (changes requested):

CaloMC/src/CaloDigiMaker_module.cc
bechenard: OOpsi, this is what happens when you code late being tired... Fixed

CaloReco/src/CaloTemplateWFProcessor.cc
bechenard: Yes, I added a check at the beginning of the function to be sure the rest of the cod works as intended.

I also fixed tabs with spaces (my old, beloved nedit auto-tabbed this file a while ago and I fixed it)

CaloMC/src/CaloDigiMaker_module.cc
kutschke: @rlcee and @bechenard : I believe that the only outstanding work on this PR is for Bertrand to make the production of noise.pdf optional (or remove it).  Then for Ray to approve and then I merge.  Am I missing anything?

bechenard:
No, I already made the changes, didn't it go through?

Bertrand

kutschke:
I just refreshed GitHub and still see only one commit that still  runs plotnoise.  Did you push to the wrong branch?
  
> No, I already made the changes, didn't it go through? Bertrand
> […](#)

bechenard:
Check if making a comment will unstuck the build bot

rlcee (approved):

kutschke:
@FNALbuild run build test

kutschke (approved):

PR #594, opened 2021-09-24
Upstream Trigger Path Mechanics

macndev:
added trigger paths for upstream cprCosmicHelix sequences which previously did not exist.

made upstream tprCosmicHelix paths 'true', they do not just call the parameters for DeM or DeP for the Ue counterparts with same helicity, which was how they previously worked. They now all have Ue* parameters accordingly

added upstream helixFinders, deltaFinders, upstream pieces to make new paths work correctly.


kutschke:
I have verified that all of the changes are either additions or modify a definition with cosmic in its name.  I am asking @gianipez to say whether or not all appropriate conventions are being followed.

gianipez (approved):
Great job, @macndev ! Yiu fixed many typos and introduced quite a few new tracking sequences that will improve significantly the OffSpillTriggerMenu. 
Just as a comment, I'll prepare another PR after this one to get rid of the timing cut in the tracking filters.

PR #595, opened 2021-09-27
Mu2eII_SM21: Add option to not build the IPA

michaelmackenzie:
Added the option to not build the IPA and an example configuration file for this.
I also added a configuration file with the Mu2e era tracker straw parameters for studies
interested in this effect.

The only validation I did was building it and writing out the no IPA gdml file and inspecting it to see the IPA was gone.
I don't expect any overlap issues as an element was simple removed.

Best,
Michael

### DIFF
--- a/Mu2eG4/geom/geom_Mu2eTrackerStraws_common.txt
+++ b/Mu2eG4/geom/geom_Mu2eTrackerStraws_common.txt
@@ -1,25 +1,6 @@
 // Geometry using the Mu2e era tracker straws
 #include "Offline/Mu2eG4/geom/geom_common.txt"
-
-// Mu2e era straw parameters taken from tracker_v6.txt
-double tracker.strawOuterRadius     =     2.5;
-double tracker.strawWallThickness   =     0.015;
-double tracker.strawGap             =     1.25;
-double tracker.wireRadius           =     0.0125;
-string tracker.mat.strawgas  = "StrawGas";
-string tracker.mat.wire      = "G4_W";
-vector<string> tracker.strawMaterials = { "G4_MYLAR", "StrawGas", "G4_W" };
-// define an inactive length at each end of every straw
-double tracker.passivationMargin = -4.;
-// Define the parameters of the metal coatings on the straws and wires.
-double tracker.straw.wallOuterMetal.thickness  = 0.00005;
-string tracker.straw.wallOuterMetal.material   = "G4_Al";
-double tracker.straw.wallInnerMetal1.thickness = 0.00005;
-string tracker.straw.wallInnerMetal1.material  = "G4_Al";
-double tracker.straw.wallInnerMetal2.thickness = 0.00002;
-string tracker.straw.wallInnerMetal2.material  = "G4_Au";
-double tracker.straw.wirePlate.thickness       = 0.00025;
-string tracker.straw.wirePlate.material        = "G4_Au";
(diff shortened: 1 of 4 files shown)


kutschke:
The test failure is expected because the default is still to test against Production/main.  I will resubmit, the test against Production/Mu2eII_SM21 once this one is finished.

There are PRs in to make the proper test the default but they are not yet merged.

Mu2eG4/geom/geom_Mu2eTrackerStraws_common.txt
```diff
@@ -0,0 +1,27 @@
...
+// define an inactive length at each end of every straw
+double tracker.passivationMargin = -4.;
+// Define the parameters of the metal coatings on the straws and wires.
+double tracker.straw.wallOuterMetal.thickness  = 0.00005;
+string tracker.straw.wallOuterMetal.material   = "G4_Al";
+double tracker.straw.wallInnerMetal1.thickness = 0.00005;
+string tracker.straw.wallInnerMetal1.material  = "G4_Al";
+double tracker.straw.wallInnerMetal2.thickness = 0.00002;
+string tracker.straw.wallInnerMetal2.material  = "G4_Au";
+double tracker.straw.wirePlate.thickness       = 0.00025;
```
kutschke, line 21: I suggest that you move this material into it's own file, with a version number in the name.  Then modify the body of this file to be just two #include statements.

kutschke (changes requested):

kutschke:
@FNALbuild run build test with Mu2e/Production#100

Mu2eG4/geom/geom_Mu2eTrackerStraws_common.txt
michaelmackenzie: I moved these parameters to a separate file, with version 1 (even though they start from version 6 of the tracker).

sophiemiddleton:
@kutschke when you are happy with this please merge and then we will create tags and will then merge Bertrand's on top of this one in one of these tags

HCasler:
@FNALbuild run build tests

Mu2eG4/geom/tracker_straw_params_v01.txt
```diff
@@ -0,0 +1,25 @@
+// Mu2e era straw parameters taken from tracker_v6.txt
+
```
kutschke, line 2: These are parameters have the same values as for tracker_v6.txt.  What's the plan - verify with these parameters and then make a new file with updated parameters?

Mu2eG4/geom/geom_Mu2eTrackerStraws_common.txt
kutschke: Thanks.

Mu2eG4/geom/tracker_straw_params_v01.txt
michaelmackenzie: The point of this file is to be able to switch back to the Mu2e era straws while testing the Mu2e-II era versions. I don't believe there will be any updates to this file, unless the Mu2e era straw parameters are updated during this study. A use case would be to run CE reco with and without the updated straw parameters, to see the effect on the track resolution.

kutschke: Thanks.

kutschke (approved):

sophiemiddleton (approved):

PR #597, opened 2021-09-27
protect mixing against unphysical SimEff.

brownd1978:
Also fixes a few whitespace issues.


kutschke (approved):
This looks straightforward.  I will merge it now.

PR #598, opened 2021-09-28
Fixed Calo timing to use proton bunch time and EventTiming prodition

bonventre:
- updated fcl to change blindTime and endTimeBuffer to digitizationStart and digitizationEnd
- digitizationEnd set to 1705 ns (effectively endTimeBuffer = 10 ns)
- CaloTrigger_module.cc not updated, is it used currently?


brownd1978 (approved):
Looks good.

CaloMC/src/CaloDigiMaker_module.cc
```diff
@@ -168,14 +176,20 @@ namespace mu2e {
...
  
-      double mbtime    = (ewMarker.spillType() == EventWindowMarker::SpillType::onspill) ?  accPar->deBuncherPeriod : ewMarker.eventLength();      
+      int waveformSize = (digitizationEnd_ - digitizationStart_) / digiSampling_;
+      if (ewMarker.spillType() != EventWindowMarker::SpillType::onspill){
+        waveformSize = (ewMarker.eventLength() - digitizationStart_) / digiSampling_;
+      }
+
+      // change digitization window to relative to event marker
+      digitizationStart_ -= timeFromProtonsToDRMarker_;
+      digitizationEnd_ -= timeFromProtonsToDRMarker_;
```
bechenard, line 190: Should this be done before calculating the waveformSize (might change the length if the spillType is Offspill)

CaloMC/src/CaloShowerROMaker_module.cc
```diff
@@ -102,7 +104,7 @@ namespace mu2e {
             EDProducer{config},
             ewMarkerTag_      (config().ewMarkerTag()),
             pbtmcTag_         (config().pbtmcTag()),
-            blindTime_        (config().blindTime()),
+            digitizationStart_        (config().digitizationStart()),
```
bechenard, line 107: My indentation......

CaloMC/fcl/common.fcl
```diff
@@ -1,5 +1,6 @@
 BEGIN_PROLOG
-  HitMakerBlindTime      : 500. 
+  HitMakerDigitizationStart    : 500. 
+  HitMakerDigitizationEnd      : 1705.
```
bechenard, line 3: Is there a reason why it is 1705 and not a larger number (1775 if I kept the previous 80ns endBuffer)?

bechenard (commented):
Thanks for doing this, I'm totally swamped these days (but a good swamped).

CaloTrigger_module.cc is an old module, and we're working on a newer version that should be better, so it should be fine if it's left out at this point.

I'm still not sure I understand exactly how timeFromProtonsToDRMarker_ works, it would be useful if you could point me to an explanation. In particular, not sure what are the constraints on the end of digitization (but I don't know the exact algorithm to digitize either, so this should be fine in first approximation)

The rest looks fine, modulo a few spaces...

CaloMC/src/CaloDigiMaker_module.cc
bonventre: I think this should be fine for now, as the specifics depend on the details of how the firmware handles offspill which has probably not been implemented yet. The assumption now is that during offspill there is no jittering in the event window times and the digitization window is exactly from digitizationStart to the end of the window. This is how I've implemented it in the tracker sim for now as well.

bechenard:
If a hit happens at ewMarker.eventLength() - epsilon, then you could 
potentially have an out-of-bound vector access because digitizationStart 
will be smaller later. Am I correct?

CaloMC/fcl/common.fcl
brownd1978: The  StepPoints production job stops at 1705, which keeps the leading edge of the next proton pulses's flash, but cuts out the peak.    Even 50ns beyond that the rate is up by orders of magnitude, if we kept those the payload would be dominated by useless low-energy electron hits.    It doesn't make sense to digitize beyond 1705 as there are no detsteps to digitize.

bechenard:
Ok, that makes sense. The calorimeter DAQ people discussed keeping 
digitizing a pulse even if it is above 1695ns, but I haven't followed 
the code development. We will revisit if need be, but we should be good now.

There is still the (very serious) problem of indentation!

CaloMC/fcl/common.fcl
bonventre: From my understanding of the code, it was never actually at 1775 because all the times folded at 1695 (although maybe I'm missing some details of the digitization process if this is a buffer to keep the end of a waveform or something). We don't actually keep any steps outside 1705 so going later than that doesn't buy you anything as far as getting later hits

bonventre:
Hi Bertrand, the documentation for that parameter is in docdb-37737, you can also look at docdb-19095 slide 8, timeFromProtonsToDRMarker is the parameter "M", and represents the fixed number of clock ticks between the proton pulse and start of the event marker (converted to units of ns). It's currently set to 200 ns, which means the hard bound on the digitization end is 1695+200, although there are sim bounds before that

bechenard:
Thanks, will look at it.

Just the out-of-bound question, and the rest is fine.

bechenard (approved):
Nice

bonventre:
Actually think you were right, if timeFromProtons is 200 ns and digitizationStart is 500 ns, then digitization in offspill should start at 300 ns from marker. Should also fix out of bound issue

brownd1978:
@FNALbuild run build test

brownd1978:
There seems to be a problem with the CaloDigis using this branch.  I checked out your fork/branch, built, then I ran 5K events of ceSimReco and val.fcl on the output.  The CaloShowerStep times look normal:
caloshowerstep_t
 but the digis are mostly missing
calodigi_t
:

brownd1978:
This looks better.  CaloDigis are now being reconstructed, here are the times:
cd_time
So the protonbunchtime offset isn't applied yet, which is fine.  Here are CaloHit times:
ch_time
There the PBT offset is clearly applied, but it looks like we're losing the last 100 ns or so of the MB.  Maybe the digitization window end needs to be longer to allow waveforms that start 1t 1700 to finish?  Bertrand, do you know how that cut is applied?
I'm also worried that the CaloShowerStep has a lot of overflows. Bertrand, do you know what those are?
css_time

brownd1978:
Could the CaloShowerStep overflows be long-lived muons that decay in the _next_ microbunch?  I'll change the plot range and find out.

brownd1978:
The CaloShower overflows are indeed long decay muons:
css_time
If I extend the digitization window 
 HitMakerDigitizationEnd      : 1805
I recover  a significant number of CaloDigis and CaloHits, but the time still doesn't go out to 1705.  Is there anything else that could affect digitization?
ch_extended_time

brownd1978:
Extending the digitization window from 450 (was 500) to 1905 (!) recovered more CaloHits, going out all the way to 1705.  This cut is behaving in a way that's unintuitive to me, I'm hoping Bertrand can explain.  As things stand, I think we want to set this time very high
ch_extended2_time
.

bechenard:
Will look at this tonight

Bertrand

brownd1978:
I am bothered by the gaps in the CaloShowerStep time distribution.  Those look like the effect of beam flash blanking, but that isn't imposed (strictly) by the flash generation, something else must be making that cut.  Whatever is making that cut, it will break OffSpill processing.

brownd1978:
The gaps in the CaloShowerStep validation plot is understood: that's due to the DigiCompression selecting only the steps correlated with digis.  The uncut (DetStepFilter) distribution looks as expected.  There is still long-term work to understand why the upper-end digitization cut needs to be so high, but that is outside the scope of this PR.

brownd1978:
@FNALbuild run build test

PR #600, opened 2021-09-29
complete linking

rlcee:
running with -fsanitize=undefined complains about incomplete, but not fatal, linking such as

build/sl7-prof-e20-p013/Offline/tmp/TrackerConditions/src/StrawElectronicsMaker.os:(.data.rel+0x1698): undefined reference to `typeinfo for mu2e::EventTiming'

these additions stop the complaints and allow testing to continue


kutschke (approved):
@rlcee  I will approve and merge this but I want to be sure that I understand.  Here is what I think - am I right?  We already link using --no-undefined and the link lists were complete enough for that.  I think that -fsanitize=undefined generates more external references and you had to add libraries to the link list to satisfy those requests.    When I look at which libraries you added to which link lists my story seems plausible.

rlcee:
> @rlcee I will approve and merge this but I want to be sure that I understand. Here is what I think - am I right? We already link using --no-undefined and the link lists were complete enough for that. I think that -fsanitize=undefined generates more external references and you had to add libraries to the link list to satisfy those requests. When I look at which libraries you added to which link lists my story seems plausible.

I don't know if sanitize was generating those errors intentionally to enforce a principle, or if it simply accessed more code as part of its checks and therefore generated them incidentally.

kutschke:
> > @rlcee I will approve and merge this but I want to be sure that I understand. Here is what I think - am I right? We already link using --no-undefined and the link lists were complete enough for that. I think that -fsanitize=undefined generates more external references and you had to add libraries to the link list to satisfy those requests. When I look at which libraries you added to which link lists my story seems plausible.
> 
> I don't know if sanitize was generating those errors intentionally to enforce a principle, or if it simply accessed more code as part of its checks and therefore generated them incidentally.

Thanks Ray.  I was guessing the latter.  I had not thought of the other possibility. 

I think we are done with this now.

PR #603, opened 2021-09-29
Add trivial module to create empty MCPrimary object

brownd1978:
This module will be used in the NoPrimary sequence, so the output will be compatible with downstream (reco) processing.


kutschke (approved):
This looks straightforward.

PR #604, opened 2021-09-29
IPA trig prescale + small change in the Trigger analyzer module 

gianipez:


Trigger/src/ReadTriggerInfo_module.cc
```diff
@@ -1161,7 +1161,7 @@ namespace mu2e {
 	  if ( (moduleLabel.find("caloMVACEFilter")!= std::string::npos) || 
 	       (moduleLabel.find("TSFilter")       != std::string::npos) ||
 	       isCosmicHelix ){ 
-	    findTrigIndex(_trigFinal, moduleLabel, index);
+	    //	    findTrigIndex(_trigFinal, moduleLabel, index);
```
brownd1978, line 1164: Why are these lines commented?  Are you intending to restore them in some future commit?  If not, please just delete them.

PR #605, opened 2021-09-29
Updated AlignedTracker to do full construction in fromFcl

bonventre:
should make alignment identical between nominal db and fcl


rlcee (approved):
nice

PR #606, opened 2021-09-30
TEve example fcl file updates

sophiemiddleton:
Mostly includes some new example fcls. Users should try to keep to using these examples and leave the prolog instances alone from now on. I am working on some config file infrastructure, but that will take some time so please add this now and I will put in another PR later with that.

### DIFF
--- a/TEveEventDisplay/src/TEveMu2eTracker.cc
+++ b/TEveEventDisplay/src/TEveMu2eTracker.cc
@@ -26,7 +26,7 @@ namespace mu2e{
         Double_t zpanel{pointmmTocm(2*trkr->g4Tracker()->getPanelEnvelopeParams().zHalfLength())};
         TEveGeoShape *station= new TEveGeoShape();
         CLHEP::Hep3Vector Pos_station(0, 1000, p - dz + zpanel);
-        if(i==0) dp = abs(Pos_station.z() - stationpos [2])/10;
+        if(i==0) dp = pointmmTocm(abs(Pos_station.z() - stationpos [2]));
         stationpos [0] = Pos_station.x();
         stationpos [1] = Pos_station.y();
         stationpos [2] = Pos_station.z();
@@ -70,7 +70,7 @@ namespace mu2e{
         double r = foil.rOut() - foil.rIn();
       
         CLHEP::Hep3Vector center = foil.centerInDetectorSystem();
-        CLHEP::Hep3Vector foilposition(center.x() ,1000+center.y(),startz/10+j/10); // Stopping Target Location FIXME
+        CLHEP::Hep3Vector foilposition(center.x() ,1000+center.y(),pointmmTocm(startz+j)); // Stopping Target Location FIXME
       
         Double_t foilpos[3];
         foilpos [0] = foilposition.x();
--- a/TEveEventDisplay/src/TEveMu2eTracker.cc
+++ b/TEveEventDisplay/src/TEveMu2eTracker.cc
@@ -1,9 +1,6 @@
 #include "Offline/TEveEventDisplay/src/shape_classes/TEveMu2eTracker.h"
-#include "Offline/StoppingTargetGeom/inc/StoppingTarget.hh"
-#include "Offline/StoppingTargetGeom/inc/TargetFoil.hh"
-#include "Offline/CalorimeterGeom/inc/DiskCalorimeter.hh"
-#include "Offline/CalorimeterGeom/inc/Calorimeter.hh"
 
+Int_t transpOpt = 100;
 using namespace mu2e;
 namespace mu2e{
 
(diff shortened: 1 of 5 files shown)


brownd1978 (approved):

TEveEventDisplay/src/dict_classes/StyleOpts.hh
```diff
@@ -0,0 +1,3 @@
+const Int_t transpOpt = 100;
+const Int_t linewidth =  3; 
+const Int_t fontsize = 15;
```
kutschke, line 3: You define these variables at file scope here and also in other files.  That's either illegal in C++ or will produce unexpected behaviour.   I presume that you have not yet had a problem because this file is not yet included anywhere.  

For the record, here are the files that contain definitions of these variables:
TEveEventDisplay/src/TEveMu2eTracker.cc
TEveMu2e_base_classes/TEveMu2eMCInterface.h
TEveEventDisplay/src/TEveMu2eMCInterface.cc

At a minimum, make the content of this file a struct with public data members.  That provides scope protection.  You may also want to make them constexpr, not just const.  That depends on future plans that we can discuss after this PR is in.

After this PR we should talk about future plans.  We talked earlier about being able to save a configuration?  Are these details part of the configuration that a use might want to modify and save?

TEveEventDisplay/src/TEveMu2eTracker.cc
```diff
@@ -19,97 +16,79 @@ namespace mu2e{
       Double_t rmin{pointmmTocm(envelope.innerRadius())};
       Double_t rmax{pointmmTocm(envelope.outerRadius())};
         
-      //Tracker Planes in XZ
+      //Tracker Stations in XZ
       double p = 0.0;
-      for(int i =0;i<20;i++)
-        {
```
kutschke, line 25: What does 20 mean?  It should be a symbolic constant.

TEveEventDisplay/src/TEveMu2eTracker.cc
```diff
@@ -19,97 +16,79 @@ namespace mu2e{
...
-        Double_t panelpos[3];
+      //unsigned int nStations = trkr->nPlanes()/2;
+      unsigned int dp = 0;
+      Double_t stationpos[3];
+      for(unsigned int i =0; i < 20 ;i++){
```
kutschke, line 24: What does 20 mean?  it should be a symbolic constant

kutschke (changes requested):

kutschke:
… I accidentally merged this PR when I merged the two ce+ PRs – too many windows open at the same time.

Please open a new PR soon to address the two issues I brought up: StyleOpts.h will not work if you #include it into a file that already has those quantities defined.  And replace the literal 20 with a symbolic name.

TEveEventDisplay/src/TEveMu2eTracker.cc
sophiemiddleton: yes, dont worry, as I said in my email....Namitha is working on these fixes, we should use nStations which is in fact 18 and she is aware of this...she will put in her PR in the next few days

sophiemiddleton:
> … I accidentally merged this PR when I merged the two ce+ PRs – too many windows open at the same time.
> 
> Please open a new PR soon to address the two issues I brought up: StyleOpts.h will not work if you #include it into a file that already has those quantities defined. And replace the literal 20 with a symbolic name.

Hi Rob, please see my email these hardcoded numbers are being worked on by Namitha, my work here was to provide notes of what needs  improving, please see my email from the other day

TEveEventDisplay/src/dict_classes/StyleOpts.hh
sophiemiddleton: This is a work in progress, as always the TEve code is iterative, this PR is a transitional one, once I have all the numbers I need and understand the overlaps this will be used. Expect the next iteration in the next few days. The purpise of the PR is to get the FCL files in for people to use

kutschke:
Ok on both replies ( this one and the magic number one ).   I will watch for them.

From: Sophie Middleton ***@***.***>
________________________________

In TEveEventDisplay/src/dict_classes/StyleOpts.hh<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_Mu2e_Offline_pull_606-23discussion-5Fr721814207&d=DwMCaQ&c=gRgGjJ3BkIsb5y6s49QqsA&r=Zi5IH-k2_JiK_Wipiv3pwwSXO9UNaHiLP8sisfPz__k&m=rpmSZNIebn87KEohnmvPwjOVpKm03xY8eUR5nWAfRDw&s=8YaEArKP6B0wJSUpNci7NiTv7pjO_osVSCvLxO2KFD8&e=>:

> @@ -0,0 +1,3 @@

+const Int_t transpOpt = 100;

+const Int_t linewidth =  3;

+const Int_t fontsize = 15;

This is a work in progress, as always the TEve code is iterative, this PR is a transitional one, once I have all the numbers I need and understand the overlaps this will be used. Expect the next iteration in the next few days. The purpise of the PR is to get the FCL files in for people to use

PR #610, opened 2021-10-04
removed cut on t0 in the TimeClusterFilter module

gianipez:
removed a useless cut in the TimeClusterFilter module. 
- it was not needed as the DAQ acceptance window will be encoded in the firmware of the ROCs
- it was creating configuration issues for running the OnSpillMenu track sequences also in the OffSpill events

### DIFF
--- a/Trigger/data/cprCosmicHelixDeM/main_cprCosmicHelixDeMTCFilter.fcl
+++ b/Trigger/data/cprCosmicHelixDeM/main_cprCosmicHelixDeMTCFilter.fcl
@@ -1,4 +1,2 @@
 physics.filters.cprCosmicHelixDeMTCFilter.requireCaloCluster : false
 physics.filters.cprCosmicHelixDeMTCFilter.minNHits : 10.0
-physics.filters.cprCosmicHelixDeMTCFilter.minTime : 0.0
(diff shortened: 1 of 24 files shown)


kutschke (commented):
You say that for real data the cut is not needed here because there is already an upstream cut, in the ROCs.   Please add something to the PR comments that explain why it is not needed for simulated events - my guess is that there is already an upstream cut.  If I am right, where is the upstream cut?

brownd1978 (approved):
I believe Giani is right, we can create whatever time cut we want through the configuration of the time cluster finders, and a cut here is unnecessary.

brownd1978:
I apparently hit 'merge' instead of 'comment' by mistake when I reviewed this PR.   Unless a problem shows up in the rest of the review I suggest we leave it.

kutschke:
> I apparently hit 'merge' instead of 'comment' by mistake when I reviewed this PR. Unless a problem shows up in the rest of the review I suggest we leave it.

That's Ok with me.  My only comment was  a request for more complete documentation.  Your followup comment answers it.  I still think that information should be part of the PR description and would appreciate it if Giani looks after it when he has a chance.

PR #611, opened 2021-10-05
WIP: Lsqsum class clean up + added check on the number of point used to evaluate the chi2 in CalPatRec

gianipez:
- Lsqsum class clean up
- added check on the number of point used to evaluate the chi2 in CalPatRec

### DIFF
--- a/CalPatRec/src/CalHelixFinderAlg.cc
+++ b/CalPatRec/src/CalHelixFinderAlg.cc
@@ -1278,9 +1278,9 @@ namespace mu2e {
 // gianipez: procedure for aligning the phi vector
 //-----------------------------------------------------------------------------
     ::LsqSums4        szphi;
-    int               count(0);
-    float            chi2min, deltaPhi, dphi_max(0);
-
+    int               count(0), minNFitHits(5);
+    float             chi2min, deltaPhi, dphi_max(0);
+    
     HitInfo_t         iworst;//(-1,-1);
 
     const char        banner[200] = "doLinearFitPhiZ";
@@ -1340,7 +1340,7 @@ namespace mu2e {
       }
 	
       //update the dfdz and phi0 if...	
-      if ( (count>=5) &&      //FIXME!
+      if ( (count>=minNFitHits) &&      //FIXME!
 	   (faceHitChi2 < 2.) &&
 	   ( (fabs(phiZInfo.dfdz - Helix._szphi.dfdz()) < 8.e-4) ) &&//  || //require that the new value of dfdz is
 				 //close to the starting one. update dfdz only if:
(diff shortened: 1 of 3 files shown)


CalPatRec/src/CalHelixFinderAlg.cc
```diff
@@ -1278,9 +1278,9 @@ namespace mu2e {
 // gianipez: procedure for aligning the phi vector
 //-----------------------------------------------------------------------------
     ::LsqSums4        szphi;
```
kutschke, line 1280: Why is this class in the global namespace?

CalPatRec/src/CalHelixFinderAlg.cc
```diff
@@ -1278,9 +1278,9 @@ namespace mu2e {
...
-    int               count(0);
-    float            chi2min, deltaPhi, dphi_max(0);
-
+    int               count(0), minNFitHits(5);
+    float             chi2min, deltaPhi, dphi_max(0);
```
kutschke, line 1282: Please make minNFitHits constexpr.

This file already has another variable, _minHHits.  Is there a good reason to make a new one?

kutschke (commented):
Please audit your code to look for other places where you need this protection

CalPatRec/src/CalHelixFinderAlg.cc
gianipez: No problem.
the data member is used for a different purpose; `_minHits` is used to select at the end of the search if a helix has enough hits, while the new I introduced is just a threshold for updating the fit results.

CalPatRec/src/CalHelixFinderAlg.cc
gianipez: I don't recall the reason TBH.

kutschke: Please make a separate PR that makes only the change of moving functions from the global to the mu2e namespace and updating the code where needed.

CalPatRec/src/CalHelixFinderAlg.cc
kutschke: When you make the change, please add a comment to distinguish it from _minHits.

kutschke (commented):
Please inspect other files that use the two LsqSum classes to be sure that the same protection is in place.   In addition to the files in this PR, the strings LsqSum appears in the following files:
TrkReco/inc/RobustHelixFinderData.hh
CalPatRec/inc/CalHelixFinderData.hh
CalPatRec/src/MergeHelixFinder_module.cc

kutschke:
Comment to toggle the status.

PR #612, opened 2021-10-05
Added digi diagnostics for VST data analysis

bonventre:
SHD and CHD now can show digi parameters as well. Also minor changes to help VST and MDC analysis with database


brownd1978 (approved):
The only thing that worries me is adding the digi index.  In the context of this PR it is a bug fix, the bigger issues with that are way outside the scope of this PR and require a redesign of ComboHit to address.

PR #613, opened 2021-10-06
Addition of livetime record to CRY stage-1 simulaiton

oksuzian:


MCDataProducts/inc/CosmicLivetime.hh
```diff
@@ -42,6 +42,10 @@ namespace mu2e {
         _livetime = _primaries / (M_PI * _area * _fluxConstant * (EfToOneMinusGamma - EiToOneMinusGamma) / (1. + eslope));
     }
 
+    CosmicLivetime( float livetime ){
```
kutschke, line 45: Please use initializer syntax and initialize all variables.

kutschke (changes requested):

MCDataProducts/inc/CosmicLivetime.hh
kutschke: I understand they are meaningless in this case but they should be initialized and not have values of random uninitialized memory.

Use the same values as are used in the default c'tor.  You should also initialize lifetime with initializer syntax.

kutschke (approved):
I still prefer that you use intializer syntax to initialize _liveitme in the new c'tor.

MCDataProducts/inc/CosmicLivetime.hh
```diff
@@ -42,6 +43,15 @@ namespace mu2e {
...
 
+    CosmicLivetime( float livetime ):
+      _primaries(0),
+      _area(0),
+      _lowE(0),
+      _highE(0),
+      _fluxConstant(0) {
+      _livetime = livetime;
+    }
+
```
soleti, line 54: Here can't you initialize `_livetime` as you do for the other parameters? However, I am not sure that in CRY you don't have those numbers. It must be using a formula internally to calculate the livetime no? Probably primaries could be misleading since you don't generate protons in CRY, but it must assume an area, an energy interval and a flux constant to calculate the livetime no?

oksuzian: I don't see a big point in storing these variables.
We won't be able to reconstruct the livetime based on these quantities. 
However, I can add them for completeness.

soleti: Ok, your call. I personally think it would be nice to store them somewhere for reference, but if you think it's misleading just keep 0 (or -1)

kutschke:
comment to force update of status

kutschke:
@FNALbuild run build test

MCDataProducts/inc/CosmicLivetime.hh
brownd1978: My personal take is that any information which is available and correct should be loaded, even if we don't know of a use case today.  If accessing that info is a lot of work, then that work can be deferred to a future PR, but it should not be lost (ie add FIXME and an issue).

EventGenerator/src/CRYEventGenerator_module.cc
```diff
@@ -72,6 +76,20 @@ namespace mu2e {
     mf::LogInfo("CRYEventGenerator") << oss.str();
   }
 
+  void CryEventGenerator::endSubRun(art::SubRun &subrun)
+  {
+    // All inputs (except getLiveTime) in CosmicLivetime are not used in the livetime calculation
+    // Livetime is calculated internally by cry: cryGen->getLiveTime()
+    std::unique_ptr<CosmicLivetime> livetime(new CosmicLivetime(1,
+                                                                cryGen->getSubboxLength(),
```
brownd1978, line 84: I am confused by the units: this accessor claims to be a length, but the constructor needs an area.  Is the accessor incorrectly named, or should this value be squared?

EventGenerator/src/CRYEventGenerator_module.cc
```diff
@@ -72,6 +76,20 @@ namespace mu2e {
...
+                                                                cryGen->getMinShowerEn()/1000.,
+                                                                cryGen->getMaxShowerEn()/1000.,
+                                                                1.8e4,
```
brownd1978, line 87: How is this constant calculated?  is it merely a conversion factor?  Or does it have physical units that are somehow constant for this generator?  Either way, could you please add a comment?

oksuzian:
@FNALbuild run build test

brownd1978 (approved):

soleti (approved):

PR #615, opened 2021-10-08
start TriggerInfo print and val

rlcee:


Validation/src/ValTriggerInfo.cc
```diff
@@ -0,0 +1,32 @@
...
+  // histogram contents change, and will not match previous versions
+  _hVer->Fill(0.0);
+
+  _hccs->Fill(obj.caloClusters().size());
+  _htrk->Fill(obj.tracks().size());
+  _hhel->Fill(obj.helixes().size());
+  _hhit->Fill(obj.hitClusters().size());
+  _hcts->Fill(obj.caloTrigSeeds().size());
+  _hcos->Fill(obj.cosmics().size());
+
```
brownd1978, line 30: Giani should comment, but my impression is this object describes what detector reco object caused the trigger to pass.  As such, a bit map of whether the trigger was generated by a track, or cluster, or helix, or ... might be interesting.  Also, it might be interesting to record the energy/momentum and time of the reco object(s) that triggered.

gianipez:
Hi @rlcee , I just realized one thing that I think is important for you to know. When we run the trigger Menu, EACH sequence generates N>=1 TriggerInfo objects. For example, a typical track trigger generates 3 TriggerInfo objects: one from the TimeClusterFilter, one from the HelixSeedFilter and lastly one from the TrackSeedFilter. This is to underline the fact that the number of entries in the Trigger-related histograms will be different from the number of triggered events. 
Other two reasons causing this effect are:
1. the fact that a given triggered events could have fired multiple triggers
2. the fact that there might be more tracks/clusters that passed the trigger selections in a given event

rlcee:
I'll assume Dave's comment about a bitmap was answered by Giani.

The question of whether this code should also follow the Ptrs and histogram the pointee objects, I think falls into what I've been calling "validation level 2".  Level 1 is what we have now (with, I think, one exception) where only the content of each product is histogrammed.  Not introducing evolving/custom cuts, links to other products, or derived calculations is intentional so that level 1 ~never fails, never has to be configured, and remains ~fast.  Confirming the intended E cuts in the trigger would fall into level 2, where we relax all those requirements and you can analyze the trigger, or do other data validation, for example,  across the whole event.  We can also allow histograms to be defined based on the trigger table, onspill/offspill, etc.  We will want this eventually.  I can also imagine a level 3 where we rerun parts of the production, or do even more expansive connections like comparing STM database to Calo and track hits, or "analysis" or "calibration".

Anyway, the point is that this is level 1 so just basic alerts to product characteristics and changes.  I am happy to start a level 2 concept for the trigger, if you would like.  It could demo the concept. Or maybe a data validation module for the VST events?

gianipez (approved):
Approving this review as the suggestions made I think will be part of the discussion of the "level 2" validation concept. In the `Trigger/src` there is an analyzer that I've been using with a similar intent, aka keep track of the physical observables that characterize each trigger chain (p, E, nHits, etc).

PR #617, opened 2021-10-10
Stopped Muon Generators

sophiemiddleton:
Adds in generators for:

* Leading Log conversion
* DIO (from new table)
* a few corrections to others


sophiemiddleton:
@FNALbuild run build tests with Mu2e/Production#114

sophiemiddleton:
@FNALbuild run build tests Mu2e/Production#114

MCDataProducts/inc/ProcessCode.hh
```diff
@@ -87,6 +87,7 @@ namespace mu2e {
       mu2eExternalRMC,         mu2eFlateMinus,      mu2eFlatePlus, mu2eFlatPhoton, // 175
       mu2eunused1, mu2eunused2, mu2eunused3, mu2eunused4, 
       mu2eunused5, mu2eunused6, mu2eunused7, mu2eunused8, 
+      mu2eCePlusLeadingLog,
```
gaponenko, line 90: Questions on the addition of the mu2eCePlusLeadingLog process code:

1) the SU2020 sims did generate mu- to e+ at leading log already, what code did they use?  Do we want to change that?

2) Should new process codes replace mu2eunused1 etc, or should they be addded after as is done here?

I do not have a strong opinion on either of these questions, but others might.

brownd1978: SU2020 used tuple-based stops, it is not reusable.  That code is in Offline/main anyways, since years.  The point of this PR is to replace it.
I prefer to infill the 'unused' codes before extending the enum.

kutschke: I prefer that we define new enum values rather than re-use retired ones.  Also, I propose the following policy.  If you want to extend an enum on Mu2eII_SM21 you must first extend it on main and then either merge or cherry-pick the change into Mu2eII_SM21.   If not we can get into the position that the same value has different meanings on different branches.   This particular PR is aimed at main, so my comment is not important here.

EventGenerator/src/DecayInOrbitGenerator_module.cc
```diff
@@ -0,0 +1,143 @@
+// This module will be used to make DIO electrons from mu- stopped in any target material
```
kutschke, line 1: I don't think we want to have two DIO modules.  Yes it's more work now to add these features to the existing module but it will be less work in the long run.

MCDataProducts/inc/ProcessCode.hh
sophiemiddleton: @brownd1978 OK., do I will replace the "unused1" with my new code

ConditionsService/data/heeck_et_al_2021_aluminum.tbl
```diff
@@ -0,0 +1,1162 @@
+1	0.0004364202
```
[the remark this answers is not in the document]
sophiemiddleton: Should I rename the file?

gaponenko:
Hi,

The new module is called DecayInOrbitGenerator, but there is nothing
DIO-specific in the code.   It is a simplified version of the Pileup
module: one can set decayFraction=1 in Pileup and get the same spectrum
(produced by the same DIOGenerator tool).

We can add a "SingleProcessGenerator" module to call just one tool.
But then that module should not have DIO in its name.
Or we can use the existing PileUp module to do the job.
In either case adding DecayInOrbitGenerator is not justified.

Andrei

kutschke:
> There aren't 2 DIO modules that work from art-based muon stops. The tuple-based stops generators are a dead end: there's no plan to revise or update them, we will remove them when we are sure we no longer need them. The new modules are needed to support physics channels from art-based stops in MDC2020 and going forwards. Sophie will discuss this more in her talk tomorrow.
> […](#)

Thanks for the clarification.

EventGenerator/src/DecayInOrbitGenerator_module.cc
```diff
@@ -0,0 +1,143 @@
...
+      addParticles(output.get(), mustop, time, Generator_.get());
+
+    }
+
+    event.put(std::move(output));
+  }
+
+  //================================================================
+  void DecayInOrbitGenerator::addParticles(StageParticleCollection* output,
+                            art::Ptr<SimParticle> mustop,
```
kutschke, line 122: Should this be passed by const& not by copy?

EventGenerator/src/GeneratorPlots_module.cc
```diff
@@ -81,12 +81,10 @@ namespace mu2e {
 
     if(!findData(event)) 
       throw cet::exception("RECO")<<"No data in  event"<< endl; 
-
-    GetGenPartInfo(event);
+    else GetGenPartInfo(event);
 }
```
kutschke, line 85: I have a strong style preference for always using {}.  I have seen too many cases where future maintenance failed due to not paying attention to the implied {}.

EventGenerator/src/LeadingLogGenerator_tool.cc
```diff
@@ -0,0 +1,89 @@
...
+  };
+
+
+  std::vector<ParticleGeneratorTool::Kinematic> LeadingLogGenerator::generate() {
+    std::vector<ParticleGeneratorTool::Kinematic>  res;
+
+    double energy = _spectrum.sample(_randSpectrum->fire());
+
+    const double p = energy * sqrt(1 - std::pow(_mass/energy,2));
+    CLHEP::Hep3Vector p3 = _randomUnitSphere->fire(p);
```
kutschke, line 65: I know that everyone does this but it's dangerous.  gcc happens to know that you you mean x*x not exp(2.0*log(x)), which is much, much slower to evaluate.  I don't know about clang or Intel or other compilers we might use down the road.  Look at cetlib/pow.h which lets you write  cetlib::square(x); for an arbitrary integer power there is cetlib::pow<N>(x).  Also, evaluating the difference of squares the way you do it is susceptible to precision issues.  In real Mu2e cases, we won't have precision issues in this line.  The better way to write a difference of squares is (a-b)*(a+b).    And pow.h has a helper function template for that too.

EventGenerator/src/LeadingLogGenerator_tool.cc
```diff
@@ -0,0 +1,89 @@
...
+
+    if(pid == PDGCode::e_minus){ _process = ProcessCode::mu2eCeMinusLeadingLog; }
+    if(pid == PDGCode::e_plus){ _process = ProcessCode::mu2eCePlusLeadingLog; }
+    ParticleGeneratorTool::Kinematic k{pid, _process, fourmom};
+    res.emplace_back(k);
+
+    return res;
+  }
+
+  void LeadingLogGenerator::generate(std::unique_ptr<GenParticleCollection>& out, const IO::StoppedParticleF& stop) {
```
kutschke, line 76: This fragment is identical to DecayInOrbitGenerator.  Is there a good way to consolidate it or are their future plans that require it to be duplicate now?

gaponenko:
More on the modules: LeadingLog_module is also redundant.
Superficially there is an if statement that sets the "process"
variable to mu2eCeMinusLeadingLog or mu2eCePlusLeadingLog, but then
"process" is not used for anything.  So again, either the Pileup
module or a generic "single process" module would do; no need for
LeadingLog_module.

Andrei

gaponenko:
Hi,

> There is no perfect answer, it is always a judgement call.  I am
> surprised Andrei is so strongly in favor of combining in this case:

Let's talk in the meeting, I think there are nuances here.

Andrei

EventGenerator/src/LeadingLogGenerator_tool.cc
gaponenko: Yes, (a-b)*(a+b) is the best here for numeric precision.  For the record, in places where one needs to square a value, there is std::pow(x,2).  I would not use third-party libs since there  is an std equvalent.

EventGenerator/src/SingleProcessGenerator_module.cc
```diff
@@ -0,0 +1,143 @@
...
+  class SingleProcessGenerator : public art::EDProducer {
+  public:
+    struct Config {
+      using Name=fhicl::Name;
+      using Comment=fhicl::Comment;
+      fhicl::Atom<art::InputTag> inputSimParticles{Name("inputSimParticles"),
+          Comment("A SimParticleCollection with muons stopping in the target.")};
+      fhicl::Atom<std::string> stoppingTargetMaterial{
+        Name("stoppingTargetMaterial"),
+          Comment("Material determines muon life time, capture fraction, and particle spectra.\n"
```
brownd1978, line 49: AFAIK this restriction applies only to LL DIO, as that has only been calculated for Al.  If other spectra are used this is (possibly) not a restriction, or a different restriction.  This is another situation where we need to tie BinnedSpectrum more tightly to the configuration.  That is outside the scope of this PR, I will add an issue.

kutschke:
@sophiemiddleton  I requested changes to pass arguments by const& and not by value?  You replied that it broke the build.  I replied with a suggestion that should fix that problem.  Did you try it?

sophiemiddleton:
What is the status of this PR?

PR #618, opened 2021-10-11
TEve 2-D Tracker view tidy up

NamithaChitrazee:
Some minor corrections made to the 2-D projection of the Tracker and the hits.


TEveEventDisplay/src/TEveMu2eTracker.cc
```diff
@@ -130,7 +113,7 @@ namespace mu2e{
           CLHEP::Hep3Vector diskPos = calo->disk(idisk).geomInfo().origin() - _detSysOrigin;
           diskPos += CLHEP::Hep3Vector(0.0, 10000.0, (-holeDZ+frontPanelHalfThick));
           double k = diskOuterRailOut + diskInnerRingIn;
-          for(int ic=0; ic<30; ic++){
+          for(int ic=0; ic<calo->nCrystal()/45; ic++){
```
brownd1978, line 116: Where does the number 45 come from?  This is integer arithmetic, by default it will round down, is that what you want?

brownd1978 (commented):
This mostly looks good, just the question about the hard-coded integer.

TEveEventDisplay/src/TEveMu2eTracker.cc
```diff
@@ -130,7 +113,7 @@ namespace mu2e{
...
             CLHEP::Hep3Vector crystalposition(diskPos.x(), (diskPos.y() + k), diskPos.z());
```
brownd1978, line 117: I must be missing something; it doesn't seem as if the loop variable 'ic' is used inside the loop.

brownd1978:
I confirm that something went wrong merging this PR, and the commits are
not actually in the repo.  I've tested that later and earlier merged PRs
are correctly in the repo.

I have no idea what happened here.  There is a comment about the PR being
accidentally merged, but whether accidental or not GitHub clearly states it
was merged 7 days ago,
yet the git history of the relevant files do not show the changes, or even
any activity at the relevant times.

The CrvUpdates branch of your repo is still there.  Please check that out,
and start a new PR based on that.  We can continue to  diagnose the history
after the commits are back in.

Dave

TEveEventDisplay/src/TEveMu2eTracker.cc
brownd1978: OK I see, you are decrementing 'k' each iteration.
This loop needs to be refactored to make the intent more clear.  For instance, the Hep3Vector is never used as such, it would be clearer to increment crystalpos[1] directly, without using 'k' (BTW, i,j, and k are usually used to name integers, not floats).  The crystalpos array should be defined outside the loop, then the x and z components don't need to be reset each iteration as they don't change.   Some comments about what is being done and why n/45 is the correct number are also needed.  Making this a 'while' loop, stopping when crystalpos[1] reaches some physical value, would be less obscure than an integer count.  Please contact Sophie for guidance.

sophiemiddleton:
Yes, I agree. If the "45" is not easy to find. In this view, an outline of the calorimeter (no crystals) could be fine. I sense the only way the code can do this would be to plot all crystals which are really not needed in this view. Since we have all the crystals in the XY view this longitudinal view is fine without crystal outlines to save plotting time.

TEveEventDisplay/src/TEveMu2eTracker.cc
```diff
@@ -105,25 +106,25 @@ namespace mu2e{
...
+          while(diskXZwidth > -(diskOuterRailOut + diskInnerRingIn))
+            {
+            crystalpos [1] = pointmmTocm(diskPos.y() + diskXZwidth);
+
             TEveGeoShape *crystalXZ = new TEveGeoShape();
             crystalXZ->SetShape(new TGeoBBox("Crystal",pointmmTocm(wrapperDXY),pointmmTocm(wrapperDXY),pointmmTocm(crystalDZ), crystalpos));
             crystalXZ->SetMainTransparency(transpOpt);   
             orthodetXZ->AddElement(crystalXZ);
-            k = k - wrapperDXY;
+            diskXZwidth = diskXZwidth - wrapperDXY;
```
brownd1978, line 124: Standard syntax for this line is  diskXZwidth -= wrapperDXY;

PR #619, opened 2021-10-11
make valCompare more sensitive

rlcee:
- if both histograms are empty, was special category, no alarm, now perfect agreement
- if one histogram empty and the other not, was special category, no alarm, now use Poisson prob in KS, and fraction tests fails
- if the two histograms have inconsistent binning, was special category, no alarm,  now an alarm
- in addition, still note if one or both histograms are empty in the summary

This pattern is better suited for how we use valCompare today, when we are asking "are these files the same?"  In the future, if we use it for asking "are the files consistent?", we will probably want to add tuning for that.  In any case, it is now set to alarm on special cases of diffs, which is the most conservative.

### DIFF
--- a/Validation/root/TValCompare.cc
+++ b/Validation/root/TValCompare.cc
@@ -249,7 +249,7 @@ void TValCompare::Report(Option_t* Opt) {
 //_____________________________________________________________________________
 void TValCompare::Summary(Option_t* Opt) {
 
-  int n0=0,ns=0,n1=0,n2=0,n3=0,n10=0,n11=0,n100=0;
+  int n0=0,ns=0,ne=0,n1=0,n2=0,n3=0,n11=0,n100=0;
   TIter it(&fList);
   TValHist* hh;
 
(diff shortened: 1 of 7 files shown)


Validation/root/TValCompare.cc
```diff
@@ -266,9 +266,9 @@ void TValCompare::Summary(Option_t* Opt) {
       else if(hh->GetStatus()==1 ) n1++;
       else if(hh->GetStatus()==2 ) n2++;
       else if(hh->GetStatus()==3 ) n3++;
-      else if(hh->GetStatus()==10) n10++;
       else if(hh->GetStatus()==11) n11++;
```
brownd1978, line 269: Please make these status values enums, to improve readability.

brownd1978 (changes requested):
Thanks for filling the holes in the validation comparison.  I just have a minor request to improve readability.

Validation/root/TValCompare.cc
```diff
@@ -249,7 +249,7 @@ void TValCompare::Report(Option_t* Opt) {
 //_____________________________________________________________________________
 void TValCompare::Summary(Option_t* Opt) {
 
-  int n0=0,ns=0,n1=0,n2=0,n3=0,n10=0,n11=0,n100=0;
+  int n0=0,ns=0,ne=0,n1=0,n2=0,n3=0,n11=0,n100=0;
   TIter it(&fList);
```
brownd1978, line 253: Thanks for introducing the enum.  I suggest following-through with n0 -> nPerfect, etc.

PR #620, opened 2021-10-11
added protection to dfdz calculation with lsqsum2

gianipez:


CalPatRec/src/CalHelixFinderAlg.cc
```diff
@@ -1441,7 +1441,9 @@ namespace mu2e {
       success = true;
     }
     //----------------------------------------------------------------------//
-   if ((Helix._szphi.dfdz()*_dfdzsign) < 0.) { 
+    if (Helix._szphi.qn() < minNFitHits) {
+      success = false;
+    }else if ((Helix._szphi.dfdz()*_dfdzsign) < 0.) { 
       success = false;
```
brownd1978, line 1447: It looks odd to have 2 logic cases result in the same action (success=false).  Maybe combine the clauses?  Or do you forsee treating these cases differently in future?

gianipez: Thanks @rlcee ! I just pushed an improved logic, which should make the code safer. 
@brownd1978 @kutschke I agree that the code should be revisited, but I think this is out of the scope of this PR and it's not really urgent as of today the performance of the code didn't show any alarming behavior (in terms of physics performance). I propose to create an issue in order to keep track of this and address it in a later stage when we will have datasets that will allow a deep investigation.

rlcee (approved):
Thanks

PR #621, opened 2021-10-12
Event Display fixes

sophiemiddleton:
To fix that Mu2e/Offline#606 didnt merge correctly


brownd1978 (changes requested):
I would rather cherry-pick the original commits.  Also this is only fixing 3 of the 5 files, it is better if everything were restored to its intended state before making any changes.  Please review 622, if that correctly includes these changes I will merge it and close 621.

PR #625, opened 2021-10-14
art Source for processing tracker VST data

bonventre:


kutschke:
The build error is a missing Offline/  in several of the #include directives in Offline/Sources/src/FromTrackerPrototypeData_source.cc .

The CI builds with setup muse -1 .  The default remains -2 .

kutschke:
@bonventre What testing have you done?  What is the priority for this merge?  If you have tested the relevant near-term use cases then I am happy to merge now and have a second PR to address Kyle's recommendations, should there be any.

kutschke:
@FNALbuild run build test

kutschke:
Kyle reports on the art redmine ticket that our code is OK as is.   We do not need to add protection within our code, either by getting a ServiceHandle to TFileService or in any other way.

PR #632, opened 2021-10-21
Fixes for Offpill digitization

brownd1978:
Loosen or remove cuts on the digitization window.  Extend TrkPatRec (time cluster finder) in trigger and reco to support up to  100usec long events.  Remove some spurious, misleading parameters set in CalPatRec config.  Extend validation plots to cover OffSpill timing window.  Cleanup of HelixSeed dependencies got mixed in.


rlcee (approved):
All fine for print and val

PR #633, opened 2021-10-21
Mixing CosmicLivetime at event level and store it at SubRun level

soleti:
This PR fixes the mixing of the `CosmicLivetime` data products. We store them at `art::Event` level in the Source module `FromCorsikaBinary`, which is then passed to the `CORSIKAEventGenerator` (our S1 stage). We resample these files with a `ResamplingMixer` (our S2 stage), which now stores the sum of the livetime of the mixed events as a `art::SubRun` data product.

### DIFF
--- a/EventMixing/src/Mu2eProductMixer.cc
+++ b/EventMixing/src/Mu2eProductMixer.cc
@@ -136,6 +136,7 @@ namespace mu2e {
 
   }
 
+  //================================================================
   void Mu2eProductMixer::startEvent(art::Event const& e) {
     if(applyTimeOffset_){
     // find the time offset in the event, and copy it locally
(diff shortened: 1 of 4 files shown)


EventMixing/src/Mu2eProductMixer.cc
```diff
@@ -350,10 +360,15 @@ namespace mu2e {
...
-      throw cet::exception("BADINPUT")<<"Mu2eProductMixer/evt: can't mix CosmicLiveTime" << std::endl;
-    for(const auto& x: in) {
-      out = *x;
+
+    if(!in.empty()) {
+      area_ = in[0]->area();
+      lowE_ = in[0]->lowE();
+      highE_ = in[0]->highE();
+      fluxConstant_ = in[0]->fluxConstant();
+      for(const auto& x: in) {
```
brownd1978, line 369: Is there a use case for mixing >1 cosmic events where this calculation makes sense?  If not, it would be better to throw, as before.

PR #641, opened 2021-10-29
organize DbService verbose settings

rlcee:
Organize debug prints a bit more:
1 will be on by default (when db is on by default)
  only echo the conditions set and override tables
2 good for validation and production, log total time and memory
5 log max about configuration and startup
6 start logging every table request and db read
9 print head and tail of each request/read
10 dump full content of each request/read


kutschke (approved):
Looks good.

PR #644, opened 2021-11-02
Updated default tracker digitization window

bonventre:


brownd1978 (approved):
IIRC this has no impact on OffSpill.  The earlier time is fine, the G4 cutoff is 250 ns.  The G4 end time is 1705,  but there's substantial delay after that so we could go out longer if you want.

PR #645, opened 2021-11-04
Used geant4 10.7 hadronic verbosity control

resnegfk:
Replaced older hadronic verbosity control with the newer one (in the remaining file used in the MT mode)


brownd1978 (approved):

kutschke (commented):
Will this change reduce printout?  Increase it?  Provide new ways to let us adjust the verbosity at run time?

resnegfk:
It will decrease some printout (which Ray noted). 
We have been using this approach in the sequential mode (and master thread) for a while now.
The PR makes a change missed earlier.

kutschke (approved):

kutschke:
Thanks for the explanation.

PR #648, opened 2021-11-04
Update PS endcap model

michaelmackenzie:
Update the PS endcap model for radiation and background studies. This includes work done by Dave Brown (WKU) and
myself. Some relevant documentation of these updates is given in docdb 39987. A new top-level geometry file was created
for this update, `geom_2021_PhaseI_v02.txt`, and `geom_common_current.txt` was advanced to use this geometry.

Validation run:
ROOT overlap check
Geant4 overlap check
No issues running with `geom_common.txt` which uses the previous geometry, and by eye the geometry looks unchanged

Best,
Michael

### DIFF
--- a/GeometryService/inc/PSEnclosureMaker.hh
+++ b/GeometryService/inc/PSEnclosureMaker.hh
@@ -24,8 +24,10 @@ namespace mu2e {
 
                                             // The center of the downstream surface of the PS
                                             const CLHEP::Hep3Vector& psEndRefPoint);
-  };
+  private:
+    static void readWindow(int i, const SimpleConfig& c, const CLHEP::Hep3Vector& winRefPoint, std::unique_ptr<PSEnclosure>& res);
 
+  };
 }  //namespace mu2e
 
 #endif /* ProductionSolenoidGeom_PSEnclosureMaker_hh */
--- a/GeometryService/src/PSEnclosureMaker.cc
+++ b/GeometryService/src/PSEnclosureMaker.cc
@@ -109,18 +109,17 @@ namespace mu2e {
                      CLHEP::HepRotation(CLHEP::HepRotation::IDENTITY),
                      shellMaterialName);
       if(vers == 2) {
-        res = std::unique_ptr<PSEnclosure>
-          (new PSEnclosure(
-                           shellCone,
-                           // end plate
-                           Tube(shellMaterialName,
-                                // The vacuum volume is flush to the PS surface
-                                psEndRefPoint + CLHEP::Hep3Vector(0,0, -shellLength - 0.5*endPlateThickness),
-                                0.,
-                                0.5*shellODWest,
-                                0.5*endPlateThickness
-                                )
-                           )
+        res = std::make_unique<PSEnclosure>
+          (
+           shellCone,
+           // end plate
+           Tube(shellMaterialName,
+                // The vacuum volume is flush to the PS surface
+                psEndRefPoint + CLHEP::Hep3Vector(0,0, -shellLength - 0.5*endPlateThickness),
+                0.,
+                0.5*shellODWest,
+                0.5*endPlateThickness
+                )
            );
       } else {
         std::vector<double> zPlanes; c.getVectorDouble("PSEnclosure.endPlate.zPlanes", zPlanes);
@@ -130,17 +129,16 @@ namespace mu2e {
         // if(zPlanes.size() > 0 ) {
         //   zlength = abs(zPlanes[0] - zPlanes[zPlanes.size() - 1]); //planes should be ordered
         // }
-        res = std::unique_ptr<PSEnclosure>
-          (new PSEnclosure(shellCone,
-                           // end plate
-                           Polycone(zPlanes,
-                                    rIns,
-                                    rOuts,
-                                    // The vacuum volume is flush to the PS surface
-                                    psEndRefPoint + CLHEP::Hep3Vector(0,0, -shellLength),
-                                    shellMaterialName
-                                    )
-                           )
+        res = std::make_unique<PSEnclosure>
+          (shellCone,
+           // end plate
+           Polycone(zPlanes,
+                    rIns,
+                    rOuts,
+                    // The vacuum volume is flush to the PS surface
+                    psEndRefPoint + CLHEP::Hep3Vector(0,0, -shellLength),
+                    shellMaterialName
+                    )
            );
       }
 
@@ -159,24 +157,24 @@ namespace mu2e {
 
       const CLHEP::Hep3Vector shellOriginInMu2e(psEndRefPoint + CLHEP::Hep3Vector(0,0, -0.5*shellLength));
 
-      res = std::unique_ptr<PSEnclosure> (new PSEnclosure(
-                            // cylindrical shell
-                            Tube(shellMaterialName,
-                                 shellOriginInMu2e,
-                                 0.5*(shellOD - 2*shellThickness),
-                                 0.5*shellOD,
-                                 0.5*shellLength),
+      res = std::make_unique<PSEnclosure>
+        (
+         // cylindrical shell
+         Tube(shellMaterialName,
+              shellOriginInMu2e,
+              0.5*(shellOD - 2*shellThickness),
+              0.5*shellOD,
+              0.5*shellLength),
 
-                            // end plate
-                            Tube(shellMaterialName,
-                                 // The vacuum volume is flush to the PS surface
-                                 psEndRefPoint + CLHEP::Hep3Vector(0,0, -shellLength - 0.5*endPlateThickness),
-                                 0.,
-                                 0.5*shellOD,
-                                 0.5*endPlateThickness
-                                 )
-                                                          )
-                            );
+         // end plate
+         Tube(shellMaterialName,
+              // The vacuum volume is flush to the PS surface
+              psEndRefPoint + CLHEP::Hep3Vector(0,0, -shellLength - 0.5*endPlateThickness),
+              0.,
+              0.5*shellOD,
+              0.5*endPlateThickness
+              )
+         );
     }
 
     //----------------------------------------------------------------
--- a/GeometryService/src/PSEnclosureMaker.cc
+++ b/GeometryService/src/PSEnclosureMaker.cc
@@ -22,24 +22,27 @@ namespace mu2e {
       const double yoff = c.getDouble(prefix+"y"   )*CLHEP::mm;
       const double zoff = c.getDouble(prefix+"z",0.)*CLHEP::mm;
       const double halfThick = 0.5*c.getDouble(prefix+"thickness")*CLHEP::mm;
+      const CLHEP::Hep3Vector windowCenterInMu2e = winRefPoint + CLHEP::Hep3Vector(xoff, yoff, zoff-halfThick);
 
       const bool windHasFrame = c.getBool(prefix+"hasFrame",false);
       res->hasFrames_.push_back(windHasFrame);
 
-      const double fWid = c.getDouble(prefix+"frameRadialWidth")*CLHEP::mm;
-      const double fHalfThick = 0.5*c.getDouble(prefix+"frameThickness")*CLHEP::mm;
-      const double rFin = c.getDouble(prefix+"r")*CLHEP::mm;//Same as rad of window
-      const double rFout = rFin+fWid;
+      if(windHasFrame) {
+        const double fWid = c.getDouble(prefix+"frameRadialWidth")*CLHEP::mm;
+        const double fHalfThick = 0.5*c.getDouble(prefix+"frameThickness")*CLHEP::mm;
+        const double rFin = c.getDouble(prefix+"r")*CLHEP::mm;//Same as rad of window
+        const double rFout = rFin+fWid;
 
-      const CLHEP::Hep3Vector windowCenterInMu2e = winRefPoint + CLHEP::Hep3Vector(xoff, yoff, zoff-fHalfThick);
+        //center the frame on the window, but pushed in z to avoid overlaps with the pipe
+        const CLHEP::Hep3Vector frameCenterInMu2e = winRefPoint + CLHEP::Hep3Vector(xoff, yoff, zoff-2.*fHalfThick);
 
-      res->wFrames_.push_back( Tube(c.getString(prefix+"frameMaterialName"),
-                                    windowCenterInMu2e,
-                                    rFin, // rIn
-                                    rFout,
-                                    fHalfThick
-                                    ));
-
+        res->wFrames_.push_back( Tube(c.getString(prefix+"frameMaterialName"),
+                                      frameCenterInMu2e,
+                                      rFin, // rIn
+                                      rFout,
+                                      fHalfThick
+                                      ));
+      }
 
       //retrieve the window pipe information
       if(res->version() > 2) {
@@ -48,9 +51,9 @@ namespace mu2e {
         const double thetax = c.getDouble(prefix+"pipe.thetaX")*CLHEP::degree;
         const double thetay = c.getDouble(prefix+"pipe.thetaY")*CLHEP::degree;
         const double zplate = abs(res->endPlatePolycone().zPlanes()[0] - res->endPlatePolycone().zPlanes().back());
-        const CLHEP::Hep3Vector origin = windowCenterInMu2e; //set at the same 
+        const CLHEP::Hep3Vector origin = windowCenterInMu2e; //set at the same
         //origin to ensure it intersects it.
-        //Set the pipe to be long enough to intersect both the 
+        //Set the pipe to be long enough to intersect both the
         //plate and the window.
         double zlength = abs(res->endPlatePolycone().originInMu2e().z() - windowCenterInMu2e.z()) + zplate;
         zlength /= abs(std::cos(thetax)*std::cos(thetay));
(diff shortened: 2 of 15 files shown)


kutschke:
@michaelmackenzie Is this PR targeted at a particular piece of work?  Is there a goal for getting that work started?

michaelmackenzie:
Hi Rob,

> @michaelmackenzie Is this PR targeted at a particular piece of work? Is there a goal for getting that work started?

The goal is to improve the model so Vitaly can perform a radiation estimate outside the PS and in the remote handling
room. There was also concern about backgrounds from proton interactions in this endcap, which could then be delayed
with respect to the rest of the beam (though this is just my understanding). Andy is planning on checking the high energy
electron background after this update I believe.

Once people are happy enough with the geometry, I can send a gdml file to Vitaly to begin the radiation estimates.

Best,
Michael

kutschke:
Thanks Michael. In interpret this to mean that the reviewers should treat this as a high-is priority item.  Let's set a goal of being done by Monday EOB unless serious issue are discovered.

Mu2eG4/src/constructPSEnclosure.cc
```diff
@@ -91,101 +103,275 @@ namespace mu2e {
...
+      for(unsigned iwindow = 0; iwindow<pse->nWindows(); ++iwindow) {
+        Tube pipe = pse->windowPipes()[iwindow];
+        std::ostringstream pipename, solidname, windowname, boxname;
+        pipename << "PSEnclosurePipeHole_" << iwindow+1;
+        solidname << "PSEnclosurePlateTmp_" << iwindow+1;
+        boxname   << "PSEnclosurePlateBox_" << iwindow+1;
+        windowname   << "PSEnclosureWindow_" << iwindow+1;
+        //make a solid pipe that is the hole in the pipe, to make a hole in the endplate
+        G4Tubs* pipetube = new G4Tubs(pipename.str(), 0., pipe.innerRadius(), pipe.halfLength(), 0., CLHEP::twopi);
+        G4RotationMatrix* matrix = new G4RotationMatrix(pipe.rotation());
```
resnegfk, line 272: new G4RotationMatrix(pipe.rotation()) is a memory leak.
Please use AntiLeakRegistry registry as e.g., in constructPSShield and many other places 
(and perhaps use CLHEP::HepRotation instead of the typedefed G4RotationMatrix for consistency with the majority of the Mu2eG4 code)

Mu2eG4/geom/psEnclosure_v05.txt
```diff
@@ -0,0 +1,102 @@
+// The configuration to be used with the as-built design, reflected
+// The configuration to be used with the as-built design, reflected
```
resnegfk, line 2: Perhaps remove the duplicated line and mention the new note in the docdb?

resnegfk (changes requested):
An important improvement and a loot of good work.  

I feel that we need to decide which code pattern (signature) of the nest and
finishNesting functions should be used when writing new code.

At some point the, in principle more general, "lookup token" approach
was introduced by Kyle, but it has not always been used since.

Consider merging this PR after making the change related to using
the AntiLeakRegistry for the rotation matrix and responding to other comments.

Afterwards consider making another PR related to the signatures of the
nest functions in the just written/updated code.

Mu2eG4/src/constructPS.cc
```diff
@@ -292,17 +295,67 @@ namespace mu2e {
...
+      auto polycone = pse->endPlatePolycone();
+      std::vector<double> rInners, rOuters(polycone.rOuter()), zPlanes(polycone.zPlanes());
+      if ( verbosityLevel > 1) {
+        cout << __func__ << " printing PS enclosure planes to subtract from PS vacuum:\n";
+      }
+      for(unsigned iplane = 0; iplane <= polycone.numZPlanes(); ++iplane) {
+        rInners.push_back(0.); //clear from center to outer radius
+        if(iplane < polycone.numZPlanes()) {
+          rOuters[iplane] = rOuters[iplane] + 0.01; //add an extra 10 um gap
+          zPlanes[iplane] = zPlanes[iplane] + 0.01; //add an extra 10 um gap
```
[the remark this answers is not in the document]
michaelmackenzie: This doesn't change the dimensions of the endcap, just creating a 10 um gap between the endcap and the
PS vacuum as this is a subtraction volume from the PS vacuum, not the endcap itself. I can remove this gap
though if it's best to not have it, I added it earlier on when I was having some overlap issues I was testing.

Mu2eG4/src/constructPSEnclosure.cc
```diff
@@ -53,12 +66,12 @@ namespace mu2e {
...
-			       pse->shellCone().phi0(),
-			       pse->shellCone().deltaPhi() };
+                               pse->shellCone().outerRadius1(),
+                               pse->shellCone().innerRadius2(),
+                               pse->shellCone().outerRadius2(),
+                               pse->shellCone().halfLength(),
+                               pse->shellCone().phi0(),
+                               pse->shellCone().deltaPhi() };
       nestCons( sName,
 		consParams,
```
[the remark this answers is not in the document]
michaelmackenzie: This makes sense, I'll try to separate whitespace changes from code changes in the future.

GeometryService/src/PSEnclosureMaker.cc
```diff
@@ -55,62 +98,93 @@ namespace mu2e {
...
+      } else {
+        std::vector<double> zPlanes; c.getVectorDouble("PSEnclosure.endPlate.zPlanes", zPlanes);
+        std::vector<double> rIns   ; c.getVectorDouble("PSEnclosure.endPlate.rIns"   , rIns   );
+        std::vector<double> rOuts  ; c.getVectorDouble("PSEnclosure.endPlate.rOuts"  , rOuts  );
+        // double zlength = 0.;
+        // if(zPlanes.size() > 0 ) {
+        //   zlength = abs(zPlanes[0] - zPlanes[zPlanes.size() - 1]); //planes should be ordered
+        // }
+        res = std::unique_ptr<PSEnclosure>
+          (new PSEnclosure(shellCone,
```
gaponenko, line 134: Suggest

res = std::make_unique<PSEnclosure>(shellCone, ...);

Mu2eG4/geom/geom_2021_PhaseI_v02.txt
```diff
@@ -0,0 +1,121 @@
+// Updated to be used in the MDC2020 simulation campaign, based off Mu2eG4/geom/geom_2019_PhaseI_hayman_v2.txt
+// updates the OPA support structures to remove overlaps
+
```
gaponenko, line 3: Why is OPA update a part of the PS endcap geometry PR?  Or is the comment here wrong?

ProductionSolenoidGeom/src/PSEnclosure.cc
```diff
@@ -28,7 +28,7 @@ namespace mu2e {
 
       }
 
-    } else {
+    } else if(pse.version() == 2){
```
gaponenko, line 31: What happens if pse.version() is neither 1 nor 2?   Please add an "else" clause at the end, e.g. throw an exception with a message about lacking support for other enclosure versions.

gaponenko (approved):
Hello,

The result of this work as shown in docdb-39987 is very impressive!  Wow!
I took a quick look at the code and have some minor suggestions, below.
I think Krzysztof did a more thorough code review.

Thank you,
Andrei

kutschke:
@michaelmackenzie No pressure - I am just checking in.   I understand that you may have commitments this week due to the collaboration meeting.  It looks to me like the ball is in your court on this PR.  Do you agree?

michaelmackenzie:
@kutschke yes, I believe the next steps are for me to address the points made above, which I will hopefully do tomorrow

Mu2eG4/src/constructPS.cc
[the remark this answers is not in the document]
michaelmackenzie: I removed this buffer, and didn't run into additional overlaps.

Mu2eG4/src/constructPSEnclosure.cc
michaelmackenzie: I've added this to the AntiLeakRegistry, thank you for noticing this!

Mu2eG4/geom/psEnclosure_v05.txt
michaelmackenzie: I've updated the comment here, and removed the duplicate.

GeometryService/src/PSEnclosureMaker.cc
michaelmackenzie: I'm not sure how to make this change, simply replacing `std::unique_ptr` with `std::make_unique` at this line for example doesn't compile. Do I have to make additional changes elsewhere?

Mu2eG4/geom/geom_2021_PhaseI_v02.txt
michaelmackenzie: I've updated this comment to reflect this update, instead of the previous one.

ProductionSolenoidGeom/src/PSEnclosure.cc
michaelmackenzie: I've added a print option for the new version of the PS enclosure, version 3.

GeometryService/src/PSEnclosureMaker.cc
michaelmackenzie: Here's an example of a change I tried to make:
```
        res = std::unique_ptr<PSEnclosure>
          (new PSEnclosure(
                           shellCone,
                           // end plate
                           Tube(shellMaterialName,
                                // The vacuum volume is flush to the PS surface
                                psEndRefPoint + CLHEP::Hep3Vector(0,0, -shellLength - 0.5*endPlateThickness),
                                0.,
                                0.5*shellODWest,
                                0.5*endPlateThickness
                                )
                           )
           );
```
converted to:
```
        res = std::make_unique<PSEnclosure>
          (
           shellCone,
           // end plate
           Tube(shellMaterialName,
                // The vacuum volume is flush to the PS surface
                psEndRefPoint + CLHEP::Hep3Vector(0,0, -shellLength - 0.5*endPlateThickness),
                0.,
                0.5*shellODWest,
                0.5*endPlateThickness
                )
           );

```
runs into this error:
```
/cvmfs/mu2e.opensciencegrid.org/artexternals/gcc/v9_3_0/Linux64bit+3.10-2.17/include/c++/9.3.0/bits/unique_ptr.h: In instantiation of 'typename std::_MakeUniq<_Tp>::__single_object std::make_unique(_Args&& ...) [with _Tp = mu2e::PSEnclosure; _Args = {mu2e::Cone&, mu2e::Tube}; typename std::_MakeUniq<_Tp>::__single_object = std::unique_ptr<mu2e::PSEnclosure>]':
Offline/GeometryService/src/PSEnclosureMaker.cc:123:12:   required from here
/cvmfs/mu2e.opensciencegrid.org/artexternals/gcc/v9_3_0/Linux64bit+3.10-2.17/include/c++/9.3.0/bits/unique_ptr.h:857:30: error: 'mu2e::PSEnclosure::PSEnclosure(const mu2e::Cone&, const mu2e::Tube&, int)' is private within this context
  857 |     { return unique_ptr<_Tp>(new _Tp(std::forward<_Args>(__args)...)); }
      |                              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from Offline/GeometryService/src/PSEnclosureMaker.cc:10:
./Offline/ProductionSolenoidGeom/inc/PSEnclosure.hh:70:5: note: declared private here
   70 |     PSEnclosure (const Cone& shellCone, const Tube& ep, int vers=2 )
      |     ^~~~~~~~~~~
```

dnbrow01: Well, if the constructor was made public in PSEnclosure.hh, this particular line should be happy.
I don't know if this breaks anything else, though.   I've not done a lot with make_unique.

michaelmackenzie: I agree, this seems to be the issue. I suppose the question is then is it better to make the constructor public and use make_unique, or leave it private and use unique_ptr? Perhaps Andrei has a preference?

kutschke: For now just make the c'tor public.  Whoever created this did not follow the pattern.  The design is that a Geometry object is meant to be a simple container of information that does not know how to construct itself.  The matching maker class was meant to have all of the smarts about how to build it.  The idea is that the maker is allowed to depend on arbitrarily complex external things, like conditions, databases, other parts of the but the geometry object class should be simple.  If you do this right then the users of the geometry object do not have any compile time or run time dependence on the infrastructure that knows how to make the geometry object.  

Part of that pattern was the at the c'tor of the geometry object was private and friended to the maker.  

For now just make it public.

kutschke:
@FNALbuild run build test

GeometryService/src/PSEnclosureMaker.cc
gaponenko: Strange.   There is no requirement that the template arg of make_unique is default constructable.   Nothing in the c++20 standard draft, nothing  in my internet searches.   And a simple standalone test happily builds a test case below,  make_unique without default ctr.

The reason I would like to avoid default constructors is the same as not having uninitialized variables in general.
As a silly example,  it would allow one to compile something like

PSEnclosure pe;
function_that_needs_enclosure_geometry(pe);

without remembering to use the geometry service.
And what is the benefit of having default constructor for such an object?

Here is the test case

//================================================================
// compile with
// g++ -std=c++17

#include <memory>
#include <iostream>

class C {
  int x_;
  int y_;
  C() {}
public:
  C(int x,int y) : x_{x}, y_{y} {}
  int sum() { return x_ + y_; }
};

int main() {
  auto res = std::make_unique<C>(1,2);
  std::cout<<"OK"<<std::endl;
  return 0;
}
//================================================================

gaponenko: Ah, not the default constructor but all the normal constructors.   Yes, that makes sense.   I just tried with the actual Offline code, and the default constructor can be kept private.  If the real constructors are public then everything compiles.

The (private) default constructor declaration can probably be removed from the code; the reason I added it in the first place was to do I/O of geometry objects into art files, but Rob discourages that.

resnegfk (approved):

On a slightly different topic: At some point we should globally replace cout/cerr/endl with the G4 counterparts in the construct....cc and other Geant4 related files, but since this was not the focus of this PR let's leave them in the currently modified files as they are for now.

GeometryService/src/PSEnclosureMaker.cc
kutschke: One the reason behind the design is exactly the case that Andrei illustrated. Has anyone looked to see how widely this pattern has been broken?

michaelmackenzie: I made the non-default constructors public and switched to `make_unique`. I tried removing the default constructor and ran into some issues doing this. Perhaps this is best done on a separate PR?

michaelmackenzie:
@FNALbuild run build test

GeometryService/src/PSEnclosureMaker.cc
gaponenko: Sure, please leave the private default constructor alone if its removal causes any problems.

GeometryService/inc/PSEnclosureMaker.hh
```diff
@@ -24,8 +24,10 @@ namespace mu2e {
 
                                             // The center of the downstream surface of the PS
                                             const CLHEP::Hep3Vector& psEndRefPoint);
-  };
+  private:
+    static void readWindow(int i, const SimpleConfig& c, const CLHEP::Hep3Vector& winRefPoint, std::unique_ptr<PSEnclosure>& res);
```
kutschke, line 28: I just noticed this now.  There is no reason for this to be  static member function.  If you can make it a free function in an anonymous namespace in the same file, that is the preferred solution.  A regular member function is also OK.

kutschke: In the interest of moving this forward, please make it just a regular member function and add a comment to suggest a future improvement.  Then I think we are ready to merge.

michaelmackenzie: This was originally an anonymous namespace free function, as you suggested, but then I wanted to access/alter private data members of PSEnclosure and this didn't work when I first tried. If I try to make just this function non-static, it fails to compile since the `make` function is static and calls this function. Would you suggest I add public functions that alter the private data of the PSEnclosure object to return this function to being an anonymous namespace free function?

kutschke: You are right that if it modifies private member data of the PSEnclosure object then a free function won't work.  You are right that a static member function may not call a non-static one.  The problem is that the function make should not have been static. Let's leave it as it is since I don't know where the tentacles lead and there are many similar instances to clean up.   

So I think we are done with this review.  Thanks for your work on this and for your patience.

kutschke: One last question.  The comment on the PR says that you updated geom_common_current.txt but I don't see it among the modified files.  Which did you intend?

michaelmackenzie: Strange! I had changed this file, but I must have accidentally reverted it at some point before committing it. I'll update this again, thank you for catching this!

michaelmackenzie:
@FNALbuild run build test

PR #649, opened 2021-11-04
increased speed of Crv waveform generator

ehrlich-uva:
increased speed of Crv waveform generator (issue #630)
bug fixes in CrvReco and CrvCoincidenceClusterFinder modules

### DIFF
--- a/CRVResponse/src/CrvCoincidenceClusterFinder_module.cc
+++ b/CRVResponse/src/CrvCoincidenceClusterFinder_module.cc
@@ -217,6 +217,7 @@ namespace mu2e
             sumXY+=ht->_x*ht->_y;
           }
           //filled one full position&time cluster
+          if(layerSet.size()<2) continue;
 
           //average counter position (PE weighted), slope, layers
           avgCounterPos/=PEs;
--- a/CRVResponse/src/CrvWaveformsGenerator_module.cc
+++ b/CRVResponse/src/CrvWaveformsGenerator_module.cc
@@ -70,7 +70,14 @@ namespace mu2e
     std::vector<double> _digitizationPointShiftFEBsSide0, _digitizationPointShiftFEBsSide1;
     std::vector<double> _timeShiftFEBsSide0, _timeShiftFEBsSide1;
 
+    struct ChargeCluster
+    {
+      std::vector<double> times, charges;
+    };
 
+    void FindChargeClusters(const std::vector<CrvSiPMCharges::SingleCharge> &timesAndCharges,
+                            std::vector<ChargeCluster> &chargeClusters,
+                            double timeShiftFEB);
     bool SingleWaveformStart(std::vector<double> &fullWaveform, size_t i);
   };
 
@@ -169,16 +176,8 @@ namespace mu2e
       if(SiPM%2==1 && FEB<_timeShiftFEBsSide1.size()) timeShiftFEB=_timeShiftFEBsSide1[FEB];
 
       const std::vector<CrvSiPMCharges::SingleCharge> &timesAndCharges = iter->GetCharges();
-      std::vector<double> times, charges;
-      double firstChargeTime = NAN;
-      for(size_t i=0; i<timesAndCharges.size(); ++i)
-      {
-        //No check whether times are within digitizationStart-_digitizationMargin and digitizationEnd
-        double timeTmp=timesAndCharges[i]._time+timeShiftFEB;  //apply timeShift to account for inaccuraries in the FEB time calibration
-        if(isnan(firstChargeTime) || firstChargeTime>timeTmp) firstChargeTime=timeTmp;
-        times.push_back(timeTmp);
-        charges.push_back(timesAndCharges[i]._charge);
-      }
+      std::vector<ChargeCluster> chargeClusters;
+      FindChargeClusters(timesAndCharges, chargeClusters, timeShiftFEB);
 
       //need to find where this FEB's TDC=0 is located with respect to the global time
       //can be anywhere within the digitization period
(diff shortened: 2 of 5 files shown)


CRVResponse/src/CrvCoincidenceClusterFinder_module.cc
```diff
@@ -217,6 +217,7 @@ namespace mu2e
             sumXY+=ht->_x*ht->_y;
           }
           //filled one full position&time cluster
+          if(layerSet.size()<2) continue;
```
brownd1978, line 220: does 2 represent a coincidence count?  Could it ever change? ie could you imagine wanting to run with 2-bar coincidences, maybe in commissioning or OffSpill?  If so, please make this a configuration parameter.

CRVResponse/src/CrvWaveformsGenerator_module.cc
```diff
@@ -70,7 +70,14 @@ namespace mu2e
     std::vector<double> _digitizationPointShiftFEBsSide0, _digitizationPointShiftFEBsSide1;
     std::vector<double> _timeShiftFEBsSide0, _timeShiftFEBsSide1;
 
+    struct ChargeCluster
+    {
+      std::vector<double> times, charges;
```
brownd1978, line 75: are times and charges are correlated vectors?  If so I suggest instead a vector  structs, where the struct is time, charge.  That is more efficient in terms of memory caching.

CRVResponse/src/CrvWaveformsGenerator_module.cc
```diff
@@ -187,76 +186,104 @@ namespace mu2e
...
-          std::array<double,CrvDigiMC::NSamples> voltages;
-          for(size_t singleWaveformIndex=0; singleWaveformIndex<CrvDigiMC::NSamples; i++, singleWaveformIndex++)
+          if(SingleWaveformStart(fullWaveform, i)) //acts as a zero suppression
           {
-            if(i<fullWaveform.size() && TDCstartTimeAdjusted+i*_digitizationPeriod<=digitizationEnd) voltages[singleWaveformIndex]=fullWaveform[i];
-            else voltages[singleWaveformIndex]=0.0;  //so that all unused single waveform samples are set to zero
-          }
+            //start new single waveform
+            double digiStartTime=TDCstartTimeAdjusted+i*_digitizationPeriod;
+            if(digiStartTime<digitizationStart) continue; //digis cannot start before the digitization interval
```
brownd1978, line 213: This would be clearer if if these tests were combined into a single statement.

CRVResponse/src/CrvWaveformsGenerator_module.cc
```diff
@@ -187,76 +186,104 @@ namespace mu2e
...
+            if(digiStartTime>digitizationEnd) continue; //digis cannot start after the digitization interval
+//            if(digiStartTime+(CrvDigiMC::NSamples-1)*_digitizationPeriod>digitizationEnd) continue; //digis cannot end after the digitization interval
+
+            //collect voltages
+            std::array<double,CrvDigiMC::NSamples> voltages;
+            for(size_t singleWaveformIndex=0; singleWaveformIndex<CrvDigiMC::NSamples; ++i, ++singleWaveformIndex)
+            {
+              if(i<fullWaveform.size() && TDCstartTimeAdjusted+i*_digitizationPeriod<=digitizationEnd) voltages[singleWaveformIndex]=fullWaveform[i];
```
brownd1978, line 221: isn't this test (<=digitizationEnd) redundant with the above line 214
is it intentional that you are incrementing the outer loop variable (i++)?  That is also getting incremented in the loop.  Did you mean instead (line 207) while(i < fullWaveform.size())?

CRVResponse/src/CrvWaveformsGenerator_module.cc
```diff
@@ -187,76 +186,104 @@ namespace mu2e
...
     event.put(std::move(crvDigiMCCollection));
   } // end produce
 
+  void CrvWaveformsGenerator::FindChargeClusters(const std::vector<CrvSiPMCharges::SingleCharge> &timesAndCharges,
+                                                   std::vector<ChargeCluster> &chargeClusters,
+                                                   double timeShiftFEB)
+  {
+    for(size_t i=0; i<timesAndCharges.size(); ++i)
+    {
+      //No check whether times are within digitizationStart-_digitizationMargin and digitizationEnd
```
brownd1978, line 274: it will be more efficient to allocate the size of chargeClusters: chargeClusters.reserve(timesAndCharges.size());

CRVResponse/src/CrvWaveformsGenerator_module.cc
brownd1978: Also, if there is a known maximum size for this vector consider using an array, or at least reserving the vector to that size.

CRVResponse/src/MakeCrvWaveforms.cc
```diff
@@ -56,6 +58,7 @@ void MakeCrvWaveforms::MakeWaveform(const std::vector<double> &times,
   waveform.clear();
 
   if(times.size()==0) return;
+  waveform.resize(8);
```
brownd1978, line 61: is 8 a hardware or firmware number?  Can it ever change?  If so it should be a configuration parameter.  If not, consider making waveform an std::array<double,8>

brownd1978 (commented):
Suggestions for clarifying and simplifying the code

CRVResponse/src/MakeCrvWaveforms.cc
ehrlich-uva: This waveform vector is not the final waveform. It is the full waveform from which relevant pieces will later be extracted to form the final waveforms. The final waveforms are stored in the CrvDigis as std::array<int16_t,8>.
The full waveform vector has a variable length depending on the distribution of the charges and is being constructed by summing up single PE waveforms. In order to avoid resizing this vector too often, I started the process with a reasonable length of 8. However, I have already made a change that replaces the number 8 with an estimate of the expected vector length.

CRVResponse/src/CrvCoincidenceClusterFinder_module.cc
ehrlich-uva: No, it represents the number of CRV layers. One needs to have at least 2 layers to form coincidences. We form coincidences based on 2, 3, or 4 layers.

CRVResponse/src/CrvWaveformsGenerator_module.cc
ehrlich-uva: Line 221 is not redundant. Line 214 checks the beginning of the 8-sample CrvDigi to see whether it should even exist. Line 221 checks individual samples of the CrvDigis, and fills the remaining samples past digitizationEnd and/or past the end of the full waveform (that was filled based on the arriving charges) with zeros. Of course the question remains how the real electronics responds at the end of the digitization window.
The outer loop (line 207) scans the full waveform until it finds relevant pieces to store in the CrvDigis. The inner loop (line 219) continues the loop through this waveform and fills the CrvDigis. Both loops advance the vector index i of the full waveform.

CRVResponse/src/CrvWaveformsGenerator_module.cc
ehrlich-uva: I can add a reserve().

CRVResponse/src/CrvWaveformsGenerator_module.cc
ehrlich-uva: Yes, they are correlated. I can combine them into a struct or a pair. I could reserve the vector to the number of charges.

CRVResponse/src/CrvCoincidenceClusterFinder_module.cc
kutschke: It's best to make this a variable declared constexpr.  When constexpr  it has the same run time speed as a literal value but the name of the variable lets you tell the story better.

brownd1978:
Hi Ralf, are you still working on any of review comments for this PR, or are we waiting for reviewer responses?

brownd1978:
@FNALbuild run build test

kutschke (approved):
I only looked at a one magic number issue. It can be another PR.

oksuzian (approved):

brownd1978 (approved):
Thanks for making the changes

brownd1978:
@FNALbuild run build test

PR #650, opened 2021-11-04
Changed Crv digi/binary/fragment converters and CrvDigi generator

ehrlich-uva:
Changed Crv digi-to-binary and fragment-to-digi converters, the CrvDigi generator, and the CrvDigi class to work with the current Crv data format. 
Bug fix in EventDisplay.

This change is not backward compatible, because the data format of the CrvDigis changed.
-ADCs: from unsigned int to int16_t
-SiPM#: from int to uint8_t
The default setting for the pedestal was changed from 100 to 0 to use the same value as the current real Crv data. This needs to be kept in mind, when older data (where the CrvDigis use a different pedestal) is reconstructed.


CRVResponse/src/MakeCrvDigis.cc
```diff
@@ -8,13 +8,12 @@ void MakeCrvDigis::SetWaveform(const std::vector<double> &waveform, double ADCco
   _ADCs.clear();
```
brownd1978, line 8:
```suggestion
  _ADCs.clear();
  _ADCs.reserve(waveform.size();
```

CRVResponse/src/MakeCrvDigis.cc
```diff
@@ -8,13 +8,12 @@ void MakeCrvDigis::SetWaveform(const std::vector<double> &waveform, double ADCco
...
   {
-    if(waveform[i]*ADCconversionFactor+pedestal>0) _ADCs.push_back(static_cast<unsigned int>(waveform[i]*ADCconversionFactor+pedestal+0.5));
-    else _ADCs.push_back(0);
+    _ADCs.push_back(static_cast<int16_t>(waveform[i]*ADCconversionFactor+pedestal+0.5));
   }
 
   int TDCtmp=lrint(startTime/digitizationPrecision);
   if(TDCtmp<0) throw std::logic_error("ERROR: found a waveform start time (relative to the event marker) < 0"); //this shouldn't happen
-  _TDC=static_cast<unsigned int>(TDCtmp);
+  _TDC=static_cast<uint16_t>(TDCtmp);
```
brownd1978, line 16: This will truncate in case TDCtmp > 65535.  Is that what the hardware does?  Or does it saturate?  If it saturates, you should add an std::max(TDCtmp,std::numeric_limits<uint16_t>.max()));

ConditionsService/data/conditions_01.txt
```diff
@@ -64,7 +64,7 @@ double crv.digitizationPeriod = 12.55; //ns
 
 //CRV calibration parameters 
 //This is only a placeholder. In the future, these values need to be a function of time and SiPM number.
-double crv.pedestal           = 100;    //ADC
+double crv.pedestal           = 0;      //ADC
```
brownd1978, line 67: Please talk to Ray about designing a calibration condition to describe this pedestal.

brownd1978 (commented):
I leave it to others to comment on the artbinary code

eflumerf:
I think it would be best to do the changes in mu2e_artdaq_core first instead of copying files into Offline

eflumerf:
Also, are these data format changes documented in DocDB 4914?

kutschke:
> I think it would be best to do the changes in mu2e_artdaq_core first instead of copying files into Offline

@eflumerf I am not sure that I follow your comment about a new mu2e_artdaq_core.  Are you suggesting that one of the changes he is making belongs in mu2e_artdaq_core, not here?  Or is it something else?

eflumerf:
Yes, I am referring to DAQ/inc/CRVFragmentTmp.hh

eflumerf:
@ehrlich-uva I just checked, and you are listed as a developer on the mu2e_artdaq Redmine project. Please make your changes to CRVFragment in mu2e_artdaq_core (`git clone ssh://p-mu2e_artdaq@cdcvs.fnal.gov/cvs/projects/mu2e_artdaq-core`), and let me know, and I'll make a tag for you.

kutschke:
> Yes, I am referring to DAQ/inc/CRVFragmentTmp.hh

Thanks Eric.  I think that the right path is:

1. Eric/Ralf makes the change in mu2e-artdaq-core/Overlays/CRVFragment.hh
2. Mackenzie will build and deploy on cvmfs.  Make a new envset on cvmfs to choose this mu2e_artdaq-core
3. Ralf will edit this PR to use the code from mu2e_artdaq-core and choose the new envset.  And delete the Tmp file.
4. Re-run the CI.

@eflumerf Do you agree?

@eflumerf   We currently use mu2e-artdaq-core version v1_06_04.  The head is at v1_06_07.  Two questions:

1. Is this change included in the head of mu2e-artdaq_core? Or do we need a new PR and a new tag?
2. Will v1_06_07 work with the current head of Offline (ie before Ralf's PR) or does Offline need changes?

@ehrlich-uva After this change will we be able to read CRV information from files made with the current head of Offline?

kutschke:
> Sounds like a good plan. I do not believe there are any breaking changes from the Offline perspective with v1_06_07 (or v1_06_08 after Ralf makes his changes).

There is a step missing from the plan.  I see that the head of mu2e_artdaq_core depends on versions of artdaq_core and  pcie_linux_kernel_module that are not installed in our cvmfs.  Please provide Mackenzie with installable builds of the required versions of these UPS products so that she can install them on cvmfs.  What about TRACE?  Am I forgetting any  UPS products?

eflumerf:
`https://scisoft.fnal.gov/scisoft/packages/artdaq_core/v3_08_00/artdaq_core-3.08.00-slf7-x86_64-e20-s112-prof.tar.bz2`
`https://scisoft.fnal.gov/scisoft/packages/TRACE/v3_17_03/TRACE-3.17.03-noarch.tar.bz2`

There are some changes in `pcie_linux_kernel_module` that we discovered while doing Trigger HST work. They only affect the DAQ side of things, but it would be nice to get Offline using the same version when we have the opportunity, so I'll cut a new tag of pcie as well.

ehrlich-uva:
I can make the changes in mu2e-artdaq-core. However, this is only a temporary solution to read Simon's files which he gets, if only one ROC FPGA is used. It is not clear yet, how the data format will look like if more than one FPGA is used. Specifically, the presence of the microbunch number is unknown at the moment, i.e. whether it will show up for every FPGA (which would cause a problem and may require more changes) or only once after the ROC status packet (current implementation in CRVFragmentTmp).

ehrlich-uva:
I pushed the changes to mu2e_artdaq_core. However, we may have to change it again soon depending on how the format will look like, if more than one ROC FPGAs are used.

CRVResponse/src/MakeCrvDigis.cc
ehrlich-uva: the current implementation for the timestamp has only 12 bit (--> 2^12*12.55ns=0.05ms) which is insufficient for the interspill microbunch. i think the hardware starts over at 0 for timestamps larger than 2^12.
one proposal is to change it to 16 bit (--> 2^16*12.55ns=0.8ms).

CRVResponse/src/MakeCrvDigis.cc
ehrlich-uva: I will add a reserve().

kutschke: As you revisit code, please always look for cases where you know the size of the vector and use reserve.

CRVResponse/src/MakeCrvDigis.cc
kutschke: Just checking: I believe that the code you are committing now will support the 100 micro-second Offspill event windows.  Is that right?

ehrlich-uva: Yes, the uint16_t supports the 100us OffSpill windows. However, the firmware still requires some changes for that (currently it only uses 12 bits / up to 50us). Once this is done, we will need another change in the CRVFragment Overlay class in mu2e_artdaq_core.

kutschke:
@ehrlich-uva There have been some developments regarding envsets.  The envset p015, which uses the new mu2e_artdaq_core, has been available for a few days now; when you get a chance you can remove the temporary hack in this PR and use the code from mu2e_artdaq_core.  

The second part is that there will soon be a p017.  Once that exists I will merge a PR #657 into main and you will need to update your .muse file to use p017, not p015.    For the record, PR #657 updates .muse to use p016 (directly from p014) to support other new features.  The envset p017 contains both that change and the new mu2e_artdaq_core that you need.

I will a comment to this PR once that happens.  In the mean time you can continue to test using p015.

brownd1978:
Hi Ralf, are you still working on any of review comments for this PR, or are we waiting for reviewer responses?

brownd1978:
@FNALbuild run build test

kutschke:
p017 is available.  I will go ahead and merge in #657 now.  Then you can change p015 to p017 in your .muse file and test again interactively.  This may generate a merge conflict but we can fix it on the GitHub side.

This evening I have a commitment starting form about 4:15 to about 8:00 PM.  I will check in then.

kutschke:
As I expected, when I merged #657 it created a conflict.  I resolved the conflict using the in-GitHub editor.  The one thing that I am not sure about is what happens if Ralf makes addition commits to this PR - that will be breaking new ground.

kutschke:
@FNALbuild run build test

kutschke (approved):
I have only seriously looked at the envset issue and one or two others that caught my eye based on initial comments from others.   Those are all resolved.

CRVResponse/src/MakeCrvDigis.cc
eflumerf: Ping for this comment.

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
```diff
@@ -1114,6 +1116,15 @@ uint8_t ArtBinaryPacketsFromDigis::compressCrvDigi(int adc) {
     toReturn = 255;
   return toReturn;
 }
+*/
```
eflumerf, line 1119: I must say I'm not a fan of leaving commented code around. If it's no longer useful, delete it.

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
```diff
@@ -1155,21 +1158,18 @@ void ArtBinaryPacketsFromDigis::fillCrvHeaderPacket(CrvDataPacket& crvData, uint
   // Word 0
   adc_t nBytes =
       sizeof(DataBlockHeader) + sizeof(CRVROCStatusPacket) + sizeof(CRVHitReadoutPacket) * nHits;
-  while (nBytes % 16 != 0)
-    nBytes++;
+  while (nBytes % 16 != 0) nBytes++;
   crvData.header.s.TransferByteCount = nBytes;
   // Word 1
-  crvData.header.s.PacketType = 5; // PacketType::Dataheader;
+  crvData.header.s.PacketType = 0x05; // PacketType::Dataheader;
```
eflumerf, line 1164:
```suggestion
  crvData.header.s.PacketType = DTCLib::DTC_PacketType_DataHeader;
```

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
ehrlich-uva: This is code that was supposed to be used for the compression algorithm. However, Simon's data doesn't use compression, yet. I can delete this code section until the compression algorithm is implemented.

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
ehrlich-uva: I will change this part.

eflumerf (approved):

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
```diff
@@ -1114,6 +1116,15 @@ uint8_t ArtBinaryPacketsFromDigis::compressCrvDigi(int adc) {
     toReturn = 255;
   return toReturn;
 }
+*/
+int16_t ArtBinaryPacketsFromDigis::compressCrvDigi(int16_t adc) 
+{
+  // TODO: Temporary implementation until we have the real compression used at the FEBs
+  // FEBs use only 12 bits out of the 16 bits
+  if(adc>2047) adc=2047;
```
oksuzian, line 1104: Can you avoid using hardcoded values?

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
```diff
@@ -1131,15 +1142,7 @@ void ArtBinaryPacketsFromDigis::fillCrvDataPacket(const CrvDigi& digi, CRVHitRea
   hit.SiPMID = SiPMID;
   hit.HitTime = digi.GetStartTDC();
   hit.NumSamples = 8;
```
oksuzian, line 1124: same here

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
```diff
@@ -1131,15 +1142,7 @@ void ArtBinaryPacketsFromDigis::fillCrvDataPacket(const CrvDigi& digi, CRVHitRea
...
-  hit.WaveformSample0 = compressCrvDigi(
-      digi.GetADCs().at(0)); // TODO: There should be a better way of filling the waveform
-  hit.WaveformSample1 = compressCrvDigi(digi.GetADCs().at(1));
-  hit.WaveformSample2 = compressCrvDigi(digi.GetADCs().at(2));
-  hit.WaveformSample3 = compressCrvDigi(digi.GetADCs().at(3));
-  hit.WaveformSample4 = compressCrvDigi(digi.GetADCs().at(4));
-  hit.WaveformSample5 = compressCrvDigi(digi.GetADCs().at(5));
-  hit.WaveformSample6 = compressCrvDigi(digi.GetADCs().at(6));
-  hit.WaveformSample7 = compressCrvDigi(digi.GetADCs().at(7));
+  for(int i=0; i<8; ++i) hit.WaveformSamples[i].ADC=compressCrvDigi(digi.GetADCs().at(i));  //TODO: needs to be changed
```
oksuzian, line 1125: and here

oksuzian (approved):

brownd1978:
@FNALbuild run build test

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
```diff
@@ -1096,23 +1097,13 @@ void ArtBinaryPacketsFromDigis::processCrvData(art::Event& evt, uint64_t& eventN
 //--------------------------------------------------------------------------------
 // crate a crvPacket from the digi
 //--------------------------------------------------------------------------------
-uint8_t ArtBinaryPacketsFromDigis::compressCrvDigi(int adc) {
+int16_t ArtBinaryPacketsFromDigis::compressCrvDigi(int16_t adc) 
+{
```
brownd1978, line 1101: Is there a reason this is int16_t and not uint16_t?  Can the ADC really take on negative values?  If so, what does that mean?  Will downstream code produce a correct interpretation if it is negative?  If not, I suggest changing to uint16_t.

brownd1978 (commented):
Ralf, have you tried reading existing digi files with these changes?  Does the rootIO fail?  Or does automatic schema evolution handle older data correctly?  A good test would be to run the reco validation script: 
mu2e -c Production/Validation/reco.fcl /pnfs/mu2e/tape/phy-sim/dig/mu2e/CeEndpointMixTriggered/MDC2020km/art/d0/03/dig.mu2e.CeEndpointMixTriggered.MDC2020km.001210_00000001.art

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
ehrlich-uva: Yes, the current firmware can (and does) return negative ADC values. That's why I had to change it to the signed int. The downstream code has no problem with negative ADC values; it only cares about the difference to the pedestal (which can also be negative).

brownd1978:
@FNALbuild run build test

brownd1978 (approved):

PR #651, opened 2021-11-05
reduce default printing

rlcee:


fcl/standardMessageDestinations.fcl
```diff
@@ -138,6 +138,8 @@ mf_coutInfo:
    Configuration : { limit : 0 }
    HITS : { limit : 0 }
    COSMIC_STEPPOINTS : { limit : 0 }
+   Summary : { limit : 0 }
+   INFO : { limit : 0 }
  } # end categories
```
gaponenko, line 143: Does limit:0 disable the printing?

rlcee: > Does limit:0 disable the printing?

Yes, the number is how many prints of this message are allowed,
so zero means no printing

kutschke: I believe that if you ask for the message facility end of job summary report you get the number of messages in each category, but only for categories with > 0 messages.  So there is value in setting a limit of 0 as a default.

Ray: do I have that right?  Do we have that summary on by default?

gaponenko: Hi Ray,

> Yes, the number is how many prints of this message are allowed, so zero means no printing

What is not clear to me is the role of all those INFO printouts to message facility in the code.

I agree that we want to minimize routing "everything is OK" printouts in production jobs, but there is always a tension between production and code development.   If a new developer starts working on  an Offline module, ideally things would be setup for all the info printouts to show up in their tests, I think.   But the same printouts would be off for production running.

When I first started working on Offline,  I realized that not all my prints to message_facility showed up (I do not remember the details).   I guessed there were some obscure settings somewhere to make them work, but that complexity immediately disqualified the use of message facility for code development/debugging in my mind.   I just started printing  all the stuff I needed to see to std::cout.    I am not saying that this is the right thing to do, but this is how it worked in practice.

I think the "out-of-the-box" setting  of Offline should be for production running.   If we can have a simple instruction for code developers, "after git clone and muse setup do this to enable INFO printouts in all your interactive jobs",  then having all those message facility prints in our C++ code makes sense.    Otherwise they are basically dead code, not used in production, not used for code development.   May be just for experts to localize a problem without re-compiling the code.

I think addressing this question is outside of the scope of the current PR, which I am approving, but we should keep thinking about this.
Andrei

gaponenko (approved):

rlcee:
Andrei, you raise good points.  In my talk, I offered to provide other settings. Right now all our fcl defaults to mf_interactive.  The original plan, which you can see in the files is a different set of settings for production and for debug etc.  If you set
services.message : @local::mf_debugging
you will get the message back, but in error.log
There are already mf_warning, presumably only warning or higher, etc
I'm happy to provide any preference with any name, I just need to know what those suggestions are.  I'll make sure there is one which is unsuppressed and writes to cout.

Another point I am concerned about, which everyone should be aware of, is that you can only suppress individual error  messages by their short name, "Summary" for the filters.  These strings really should really be unique to the world, like "GenFilterSummary".   This way, we can control each individually, for example if there are occasional corrupt data, we might want to print the first 3 occurrences of corruption X, but all occurrences of corruption Y, if you see what I mean.  You can't suppress them by code source.  You can suppress all of the same level - info, warning, error.

kutschke:
@FNALbuild run build test

fcl/standardMessageDestinations.fcl
kutschke: I agree that this is out of scope.  I have relaunched the tests and merged.  I will create an issue to follow up.

kutschke:
I created two new issues to followup on the two topics that came up during this discussion.

kutschke (approved):

PR #655, opened 2021-11-10
more prints behind verbose flags

rlcee:


kutschke (approved):

resnegfk (changes requested):
Thanks for working on this. 
The Geant4 related modules should use G4cout/G4cerr, esp. in the new/modified code.
I see that in one case G4out was replaced with cout and a cerr was kept in another.
And a more philosophical comment: perhaps some of those printouts should be printed even at level 0, and be suppressed at -1 level, but I let Ray to decide that.

resnegfk (approved):

rlcee:
I fixed the G4cout, but I'm not sure what you are thinking about for the flag > - 1.  In previous discussions, we said that items that should always be logged should not go through the message system or verbose flags.  For example, recording the geometry file. I thought one line noting that this was an MT job was reasonable to put in this same category.  I didn't see anything else I thought should be printed with flags set to 0, but I don't pretend to understand geant philosophy or the potential value of some of the prints.  Did you want to print some of these at 0, then set the flags to -1 or?

kutschke:
@FNALbuild run build test

kutschke:
@resnegfk it looks like you approved  just the last commit.  Did you mean to approve the full PR?

resnegfk:
On my page it shows that I approved the entire PR (which I intended)
Answering Ray's question: I think that some of the printout could stay when the the print level is set to 0, and suppressed when it is set to -1; but I do not feel strongly about it and leave to Ray to decide.

kutschke:
I also see check beside your name at the top right.  But in the summary it does not show your approval.  No idea why those two things disagree.  I will merge now.

PR #656, opened 2021-11-11
new CRV light default print

rlcee:
After consulting with Ralf, reduce default Crv light file prints from several stanzas of

Reading CRV lookup tables /cvmfs/mu2e.opensciencegrid.org/DataFiles/CRVConditions/v6_0/LookupTable_4550_0 ... Done.
CRV sector 0 (R1) uses /cvmfs/mu2e.opensciencegrid.org/DataFiles/CRVConditions/v6_0/LookupTable_4550_0 with scintillation yield of 16548 photons/MeV

to
CRV light files: /cvmfs/mu2e.opensciencegrid.org/DataFiles/CRVConditions/v6_0
CRV light sectors: 23 hash:10539762652061218902
mirroring the default prints for other input files.

The previous format is still available with a debug flag.

In other default print cases, the hashes are over the entire input file content, but here the hash is just on the file names.  Hashing the content would require intrusive code, or re-reading the files.

### DIFF
--- a/CRVResponse/src/CrvPhotonGenerator_module.cc
+++ b/CRVResponse/src/CrvPhotonGenerator_module.cc
@@ -222,6 +222,7 @@ namespace mu2e
       filedirs.insert( std::filesystem::path(filespec).parent_path() );
       photonMaker->LoadLookupTable(filespec,_debug);
       photonMaker->SetScintillationYield(_scintillationYields[i]);
+      if(_debug>0) std::cout<<"CRV sector "<<i<<" ("<<_CRVSectors[i]<<") uses "<<_makeCrvPhotons.back()->GetFileName()<<" with scintillation yield of "<<_scintillationYields[i]<<" photons/MeV"<<std::endl;
     }
 
     std::cout << "CRV light files:";
--- a/CRVResponse/src/CrvPhotonGenerator_module.cc
+++ b/CRVResponse/src/CrvPhotonGenerator_module.cc
@@ -210,7 +210,7 @@ namespace mu2e
         {
            tableLoaded=true;
            _makeCrvPhotons.emplace_back(_makeCrvPhotons[j]);
-           if(_debug>0) std::cout<<"CRV sector "<<i<<" ("<<_CRVSectors[i]<<") uses "<<_makeCrvPhotons.back()->GetFileName()<<std::endl;
+	   if(_debug>0) std::cout<<"CRV sector "<<i<<" ("<<_CRVSectors[i]<<") uses "<<_makeCrvPhotons.back()->GetFileName()<<" with scintillation yield of "<<_scintillationYields[i]<<" photons/MeV"<<std::endl;
            break;
         }
       }
@@ -222,7 +222,6 @@ namespace mu2e
       filedirs.insert( std::filesystem::path(filespec).parent_path() );
       photonMaker->LoadLookupTable(filespec,_debug);
       photonMaker->SetScintillationYield(_scintillationYields[i]);
-      if(_debug>0) std::cout<<"CRV sector "<<i<<" ("<<_CRVSectors[i]<<") uses "<<_makeCrvPhotons.back()->GetFileName()<<" with scintillation yield of "<<_scintillationYields[i]<<" photons/MeV"<<std::endl;
     }
 
     std::cout << "CRV light files:";
--- a/CRVResponse/src/CrvPhotonGenerator_module.cc
+++ b/CRVResponse/src/CrvPhotonGenerator_module.cc
@@ -41,6 +41,9 @@
 #include "CLHEP/Random/Randomize.h"
 
 #include <string>
+#include <filesystem>
+#include <set>
+#include <boost/functional/hash.hpp>
 
 #include <TDirectory.h>
 #include <TFile.h>
@@ -58,6 +61,7 @@ namespace mu2e
     using Comment=fhicl::Comment;
     struct Config 
     {
+      fhicl::Atom<int> debug{ Name("debugLevel"),Comment("Debug Level"), 0};
       fhicl::Sequence<std::string> moduleLabels{ Name("crvStepModuleLabels"), Comment("CrvStepModule labels")};
       fhicl::Sequence<std::string> processNames{ Name("crvStepProcessNames"), Comment("process names of CrvSteps")};
       fhicl::Sequence<std::string> CRVSectors{ Name("CRVSectors"), Comment("Crv sectors")};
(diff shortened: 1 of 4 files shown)


CRVResponse/src/CrvPhotonGenerator_module.cc
```diff
@@ -202,18 +210,31 @@ namespace mu2e
...
+    std::cout << "CRV light files:";
+    for(auto const& dir : filedirs ){
+      std::cout << " " << dir;
+    }
+    std::cout << std::endl;
+    size_t hash = 0;
+    for(auto const& mcp: _makeCrvPhotons) {
+      boost::hash_combine<std::string>(hash,mcp->GetFileName());
     }
+    std::cout << "CRV light sectors: " << _makeCrvPhotons.size() << " hash:" << hash <<std::endl;
```
kutschke, line 237: The printout between lines 225 an 235 and on line 237 is not protected by a debug level. I am checking that this is intended.

rlcee: Yes, intended.  Recall when we discussed the print for the geometry and conditions files.  We decided that they should not be able to be turned off, since we wanted to force the logging of the critical inputs, including the hash.  This line follows that pattern.

CRVResponse/src/CrvPhotonGenerator_module.cc
```diff
@@ -202,18 +210,31 @@ namespace mu2e
...
 
       _makeCrvPhotons.emplace_back(boost::shared_ptr<mu2eCrv::MakeCrvPhotons>(new mu2eCrv::MakeCrvPhotons(_randFlat, _randGaussQ, _randPoissonQ)));
       boost::shared_ptr<mu2eCrv::MakeCrvPhotons> &photonMaker=_makeCrvPhotons.back();
-      photonMaker->LoadLookupTable(_resolveFullPath(_lookupTableFileNames[i]));
+      std::string filespec = _resolveFullPath(_lookupTableFileNames[i]);
+      filedirs.insert( std::filesystem::path(filespec).parent_path() );
+      photonMaker->LoadLookupTable(filespec,_debug);
       photonMaker->SetScintillationYield(_scintillationYields[i]);
-      std::cout<<"CRV sector "<<i<<" ("<<_CRVSectors[i]<<") uses "<<_makeCrvPhotons.back()->GetFileName()<<" with scintillation yield of "<<_scintillationYields[i]<<" photons/MeV"<<std::endl;
+      if(_debug>0) std::cout<<"CRV sector "<<i<<" ("<<_CRVSectors[i]<<") uses "<<_makeCrvPhotons.back()->GetFileName()<<" with scintillation yield of "<<_scintillationYields[i]<<" photons/MeV"<<std::endl;
```
ehrlich-uva, line 225: The sector-by-sector print outs of the scintillation yields is of some importance, because we are running simulations with different scintillation yields where the log files may be useful to recover the values that were used. What is the requirement to print out information at debug level 0?

rlcee: My goal is to keep asking if prints are necessary and apply some pressure to keep it down.  I'll leave it 
to the group to determine "necessary".   For this study, you could turn the debug to 1 in your fcl.
Or will the changes be made with different light files in different directories, in which 
case the directory path print captures it?  We also have the choice to set the debug level in 
different ways, for example, leave it off by default but turn it on by default for all Production fcl.  
We already do that for other flags.  If the decision is to restore the print, we could try to reduce its
visual impact a bit.

kutschke: @ehrlich-uva  are you happy with Ray's reply or do you still want a change?  If you are happy, please approve the PR.

ehrlich-uva: The scintillation yields of the individual CRV sectors is important information. That's why I was wondering whether it could be kept in the log file regardless of the debug level. 
As an alternative, we could set the default debug level to 1 for the CrvPhotonGenerator so that all relevant information gets printed out by default. In this case, debug>0 in MakeCrvPhotons.cc could be changed to debug>1, since this part is less relevant.
@rlcee The lookup-table files and directories don't change when different scintillation yields are used.

CRVResponse/src/CrvPhotonGenerator_module.cc
```diff
@@ -202,18 +210,31 @@ namespace mu2e
         {
            tableLoaded=true;
            _makeCrvPhotons.emplace_back(_makeCrvPhotons[j]);
-           std::cout<<"CRV sector "<<i<<" ("<<_CRVSectors[i]<<") uses "<<_makeCrvPhotons.back()->GetFileName()<<std::endl;
+           if(_debug>0) std::cout<<"CRV sector "<<i<<" ("<<_CRVSectors[i]<<") uses "<<_makeCrvPhotons.back()->GetFileName()<<std::endl;
```
ehrlich-uva, line 213: Could you please replace this line by
if(_debug>0) std::cout<<"CRV sector "<<i<<" ("<<_CRVSectors[i]<<") uses "<<_makeCrvPhotons.back()->GetFileName()<<" with scintillation yield of "<<_scintillationYields[i]<<" photons/MeV"<<std::endl;

CRVResponse/src/CrvPhotonGenerator_module.cc
```diff
@@ -222,7 +222,6 @@ namespace mu2e
       filedirs.insert( std::filesystem::path(filespec).parent_path() );
       photonMaker->LoadLookupTable(filespec,_debug);
       photonMaker->SetScintillationYield(_scintillationYields[i]);
-      if(_debug>0) std::cout<<"CRV sector "<<i<<" ("<<_CRVSectors[i]<<") uses "<<_makeCrvPhotons.back()->GetFileName()<<" with scintillation yield of "<<_scintillationYields[i]<<" photons/MeV"<<std::endl;
```
ehrlich-uva: This line should have stayed.

kutschke: It looks like Ray has addressed this now.  When you agree that this is ready, please approve the PR.

kutschke:
@FNALbuild run build test

ehrlich-uva:
Approved.

kutschke:
Thanks @ehrlich-uva   .   When I said, please approve, I meant to go to the "Files change" link at the top right of this page.  In the upper right of that page, there is a green "Review changes" box.  Click on that which will open a popup window.  I there click the "Approve" radio button.  If you have want to enter a comment you can do so.  Then click on the green "Submit review" button.   Please do so when you have a chance.

ehrlich-uva (approved):

PR #657, opened 2021-11-13
Start pywrap

rlcee:
This should all be turned off and harmless by default.  To build wrappers, use Muse v2_10_00 or higher and add the switch "--mu2ePyWrap" to the muse build command.  Muse has added the output directories to the PTYTHONPATH, so the wrappers are ready to use.  See [the wiki](https://mu2ewiki.fnal.gov/wiki/PyWrap) for examples.  I had to ifdef out some code in the enum classes. The swig interface files have lots of features, so this code could probably accommodated with some more work on the interface specification.  In any case, the code works now for translating the enums.

A note to Mackenzie and Rob.  I went from p014 to p016, which is necessary to setup swig.  I see p015 is in development now, so when that is ready, please add the swig setup before committing it.  Since p015 is already out there, I'd make your p015 plus swig into p017.  Please commit them to MuseConfig.

### DIFF
--- a/DbService/src/SConscript
+++ b/DbService/src/SConscript
@@ -46,6 +46,8 @@ helper.make_plugins( [ mainlib,
 BINLIBS   = [ mainlib, 'mu2e_DbTables' , 'cetlib', 'cetlib_except', "pq" ]
 helper.make_bin("dbTool",BINLIBS,[])
 
+# turn pywrap.i into a python interface
+helper.make_pywrap (['mu2e_DbTables'])
 
 # This tells emacs to view this file in python mode.
 # Local Variables:
--- a/.muse
+++ b/.muse
@@ -1,7 +1,6 @@
 # prefer to build with this environment
-ENVSET p014
+ENVSET p016
 # add these to python path
-PYTHONPATH TrackerAlignment/scripts
 PYTHONPATH Trigger/python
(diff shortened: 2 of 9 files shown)


DbService/src/SConscript
```diff
@@ -46,6 +46,8 @@ helper.make_plugins( [ mainlib,
 BINLIBS   = [ mainlib, 'mu2e_DbTables' , 'cetlib', 'cetlib_except', "pq" ]
 helper.make_bin("dbTool",BINLIBS,[])
 
+# turn pywrap.i into a python interface
+helper.make_pywrap (['mu2e_DbTables'])
```
kutschke, line 50: Where is the source for make_pywrap?  I don't see it either in Offline or in Muse.

rlcee: [here](https://github.com/Mu2e/Muse/blob/91127678ce9ae6600fa435408548cdd4221099d0/python/mu2e_helper.py#L125) in Muse.  You must have just overlooked it.

kutschke: Thanks.  Master/main confusion in my clone.

kutschke (approved):
This is ready to merge.  But the merge needs to wait until the envset p017 is available - see the comments on PR #650 .

PR #658, opened 2021-11-15
Move stable geometry forward to current version

michaelmackenzie:
Update `geom_common` to use the latest geometry version, with the updated PS endcap model, from `geom_common_current`.
The only validation done is that a job writing out the gdml file for `geom_common` properly runs, as the underlying model was
already validated for `geom_common_current`.


kutschke (approved):
This was discussed in an email thread.

PR #659, opened 2021-11-17
Add SimEfficiencies2 table, change SimBookkeeper prodition to use it

rlcee:
The SimEfficiencies table used int to store the numerator and denominator of the efficiency ratio, which is now overflowing.  The table SimEfficiencies2 is the same, but using long long for these numbers (BIGINT in the database).  The prodition in this commit now can only use SimEfficiencies2 and doesn't know about the original table.  

This is our first table version migration.  The old code can continue to use old conditions sets but the new code will only run on new conditions sets with the new table.  I'll add the topic of table migration to a slide to my discussion talk with some details.

### DIFF
--- a/DbService/data/create.sql
+++ b/DbService/data/create.sql
@@ -292,3 +292,10 @@ CREATE TABLE sim.efficiencies
    CONSTRAINT sim_efficiencies_pk PRIMARY KEY (cid,tag) );
 GRANT SELECT ON sim.efficiencies TO PUBLIC;
 GRANT INSERT ON sim.efficiencies TO sim_role;
+
+CREATE TABLE sim.efficiencies2
+  (cid INTEGER, 
+   tag TEXT, numerator BIGINT, denominator BIGINT, eff NUMERIC,  
+   CONSTRAINT sim_efficiencies2_pk PRIMARY KEY (cid,tag) );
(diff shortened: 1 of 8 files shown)


AndrewEdmonds11 (approved):
Looks good to me :)

kutschke:
Andy and Ray:  do you understand how this will impact MDC2020? We already have files that we plan to keep.  Does this change invalidate them?

kutschke:
@soleti and @brownd1978 .  I added you as reviewers only to make sure that you are on board with me merging this in the middle of MDC2020.  I suspect that it's needed for MDC2020 but I don't know that for sure.

AndrewEdmonds11:
On the mixing side, I think it should just work. Ray's changed the SimBookkeeper class to use the new SimEfficiencies2 table, and the MixBackgroundFrame module uses the SimBookkeeper interface instead of accessing the table directly. I'm not sure if there will be something to change on the database side so I will leave Ray to comment on that.

soleti (approved):
Thank @kutschke, we had an email discussion with Ray and we are onboard with these changes.

kutschke:
> Thank @kutschke, we had an email discussion with Ray and we are onboard with these changes.

I will take this as approval from both Roberto and Dave.  I will merge now.

PR #660, opened 2021-11-18
FlagBkg clustering hit fix

bechenard:
A fix for the bug Richie encountered with the FlagBkgHit module, making the module independent of the "microbunch length". I took the opportunity to do some cleaning as well. Check that the track reco performance is essentially unchanged (there is a small change in the hit flagging algo but the impact is very limited)

### DIFF
--- a/TrkHitReco/fcl/prolog.fcl
+++ b/TrkHitReco/fcl/prolog.fcl
@@ -59,13 +59,7 @@ TNTClusterer : {
     MaxDistance      : 100.0
     MinHitError      : 5.0
     TimeRMS          : 2.0
-    DeltaTimeBinMin  : 10
     MedianCentroid   : false
-    preFilter        : false
-    pfTimeBin        : 20
-    pfPhiBin         : 0.1
-    pfMinHit         : 10
-    pfMinSumHit      : 20
     ComboInit        : true
     TestFlag         : true
     BackgroundMask   : []
--- a/TrkHitReco/src/FlagBkgHits_module.cc
+++ b/TrkHitReco/src/FlagBkgHits_module.cc
@@ -9,9 +9,7 @@
 
 #include "Offline/ConditionsService/inc/ConditionsHandle.hh"
 #include "Offline/ConfigTools/inc/ConfigFileLookupPolicy.hh"
-#include "Offline/ConditionsService/inc/AcceleratorParams.hh"
 
-#include "Offline/DataProducts/inc/EventWindowMarker.hh"
 #include "Offline/MCDataProducts/inc/StrawDigiMC.hh"
 #include "Offline/RecoDataProducts/inc/StrawHit.hh"
 #include "Offline/RecoDataProducts/inc/StrawHitFlag.hh"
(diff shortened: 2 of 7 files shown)


kutschke:
Bertrand: Any suggestions  for appropriate reviewers?

kutschke:
@FNALbuild run build test

bechenard:
I did a lot of cleanup, and the real changes are minimal. Anyone who can read code could do it.

kutschke:
> I did a lot of cleanup, and the real changes are minimal. Anyone who can read code could do it.
> […](#)

I had a quick look.  My impression is that "cleanup" means just removal of obsolete code.  Is that right?

Do you expect changes to nightly validation.  I don't think that we make any Offspill digs during nightly validation.

kutschke:
What tests have you done to check the behaviour of this change?   I can see that it should work well for reco.fcl since the events are mixed and I bet that maxtime is stable to within a few ns.   I can see possible pathologies with both ceSimReco and with reco of offline simulated offline events; when the only hits in the event are from track of interest, track will be pegged to maxtime - is that an issue?

bechenard:
Not sure I understand the question. How get a bunch of hits with 
different times, and I just need to find the max time to build the 
proper data structure for the algorithm. It is on an event by event 
basis, so this shouldn't impact things downstream. The only difference 
is that the definition of the bin width in that structure is now allowed 
to be a bit smaller than it was before, which means a few edge effects 
for the clustering algorithm.

I tested the old/new version with reprocessing Validation/reco.fcl on a 
digi file provided from Dave Brown (latest version). I also tested with 
the VST file that Ray provided me.

Not sure I understand your last sentence :-)

kutschke:
Thanks Bertrand.  Please also try ceSimReco - probably 1k or 2k events is fine.

bechenard:
Could you send me the command, I forgot what it is...

Bertrand

kutschke:
> Could you send me the command, I forgot what it is... Bertrand
> […](#)

mu2e -c Production/Validation/ceSimReco.fcl -n 2000

bonventre (approved):
This should fix the VST data crashes too - worst case when there's a bad hit time now it will just create spurious clusters

brownd1978 (approved):
This looks good to me.  It makes sense to use the hits themselves to define the time window, that minimizes the places where we need to explicitly keep track of on/offspill and such.  I'm glad that you're pruning out obsolete methods.  We will need to generate some OffSpill MDC data before this can be fully tested.

kutschke:
Hi Bertand,


  *   Not sure I understand your last sentence :-)

Taking this Offline.  First checking something – this is the code that looks for clusters of hits in time, azimuth and radius, right?

Normally a time cluster has a low level of hits on both the low time side and the high time side.  If the cluster ends at the edge of the binning, there is no low level of hits on the high side.  Is the algorithm robust against that?  I bet that in ceSimReco that will happen a lot.

Rob

bechenard:
I'm glad you are glad!

kutschke (approved):
I accidentally got Bertrand and myself into an offline thread.  He ran valCompare on the output of ceSimReco and reports:

"Slight changes in all plots but the results are globally compatible. this is what I would expect if i tweaked the hit clustering algo parameters."

Thanks for checking this for me.

brownd1978:
@FNALbuild run build test

PR #661, opened 2021-11-20
fix some references to null pointers

rlcee:
fsanitize test finds pointers that were not created were dereferenced.  The pointers were not created because they were inside version checks, but this use was outside the version checks.  This commit just moves the dereference to within version checks.


kutschke (approved):
This is straightforward.  Thanks for looking after it.

PR #663, opened 2021-11-28
start support for TH2

rlcee:


kutschke (approved):
This looks straightforward and it will not change nightly validation.  So we don't need to think about it interfering with other merges.  Merging now.

PR #664, opened 2021-11-29
Added cstTimeCluster to Offspill and ExtrPos menu

bonventre:
Tracker only trigger requiring 8 digis in the event and 4 combohits in the time cluster

### DIFF
--- a/Trigger/data/cstTimeCluster/main_cstTimeClusterEventPrescale.fcl
+++ b/Trigger/data/cstTimeCluster/main_cstTimeClusterEventPrescale.fcl
@@ -1,2 +1,1 @@
 physics.filters.cstTimeClusterEventPrescale.nPrescale    : 1
-physics.filters.cstTimeClusterEventPrescale.triggerFlag  : ["PrescaleRandom"]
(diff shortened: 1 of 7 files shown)


Trigger/data/cstTimeCluster/main_cstTimeClusterEventPrescale.fcl
```diff
@@ -0,0 +1,2 @@
+physics.filters.cstTimeClusterEventPrescale.nPrescale    : 1
+physics.filters.cstTimeClusterEventPrescale.triggerFlag  : ["PrescaleRandom"]
```
gianipez, line 2: this line is not needed anymore. Please, remove it

gianipez (changes requested):
Hi @bonventre , 

Thanks for integrating this sequence in the various Menu. I posted one comment AND I also noticed another thing I would like you to change; in CosmicRec/fcl/prolog_trigger.fcl, the definition of the sequence `cstTimeCluster` uses the tracker-recohit sequence `TTprepareHits`, while I think it's enough to use the `TTmakefastHits` (which doesn't run the ComptonTagger module (FlagBkgHits...). Can you commit also this changes?

Cheers,

Giani

gianipez (approved):

PR #666, opened 2021-11-30
Addition of Time Cluster and Colour coding the hits

NamithaChitrazee:
- Two more functions "Add Time Cluster" and "Add Trk Hits" have been added to the Data Interface.
- The "Add Trk Hits" option shows all the hits in a colour coded manner where the hits part of the track are displayed in green and the rest are red. 
- The selection options for the user to view these features is provided in the prolog.


kutschke:
I am looking at this now.  Two comments for future reference.

First, please make a freshly named working branch for each new feature.  This PR (666) can proceed with the to-be-merged branch named main but I feel like it's a landmine if this becomes regular practice.  If you have several feature branches in parallel development, please develop them on separate branches.  You can have one more branch in which the features are merged together for testing together.

Second we should revisit the design of the class Data_Collections.  I know that you inherited this design so I am not blaming you.  I don't like the data members that point to data products being bare pointers.  They should be an appropriate type of safe pointer, probably an art::Handle ( not an art::ValidHandle).    The only way that it's Ok for them to be bare pointers is if all are required to be present at all times.  That's not a good idea because we want to be able to look at events in which some data products are missing.

TEveEventDisplay/src/TEveMu2eMainWindow.cc
```diff
@@ -862,6 +863,7 @@ namespace mu2e{
       _firstLoop = firstLoop;
       _accumulate = accumulate;
       _data.chcol = data.chcol;
+      _data.tccol = data.tccol;
```
kutschke, line 866: Please make a new PR that addresses the comments below.  

Are you copying the entire content of data to _data?  If so, the right way is to use the assignment operator of the class Data_Collections .  The one that is automatically generated by the compiler will do the right thing.  Again, I now that this is code that you inherited.  I see that the assignment operator is declared in this code but it I don't see where it is defined.  Does it actually have a definition?  If not, I *think* that the declaration blocks the compiler from providing the default one.

And a second comment. In the signature of TEveMu2eMainWindow::setEvent, the argument data is passed by reference, which is good.  But it is not const reference.  Is it actually modified by this code?  If not please make it const reference.  If you can't, because lower level code is not const correct, please fix that too.

kutschke (approved):

kutschke:
@NamithaChitrazee I have approved this PR and will merge it now.  In my comments I ask you to make one new PR soon to address a few issues; please do that soon.  I have a few other comments about organizing future work.

TEveEventDisplay/src/TEveMu2eMainWindow.cc
NamithaChitrazee: The variable _data is defined in the header file "TEveMu2eMainWindow.h". 

 I understand your comment but I am sorry since I followed the pattern that was set previously for functions like AddComboHits() or AddCrystalHits() while developing these new functions, I think @sophiemiddleton would be able to answer/explain this much better.

NamithaChitrazee:
> 

Thank you. I have forwarded these issues to @sophiemiddleton as well. We shall take care of them soon.

TEveEventDisplay/src/TEveMu2eMainWindow.cc
sophiemiddleton: I will check. There is a work around in that code to allow for add/remove tick box on the GUI and that data_ variable is part of that. It's a bit "hacky" but getting TEve to do what you want is a bit that way in general

kutschke: > The variable _data is defined in the header file "TEveMu2eMainWindow.h".
> 
> I understand your comment but I am sorry since I followed the pattern that was set previously for functions like AddComboHits() or AddCrystalHits() while developing these new functions, I think @sophiemiddleton would be able to answer/explain this much better.

I understand that you followed an existing pattern.  That's OK, even expected.  I would like to evolve towards better patterns as we move ahead, particularly was we move into REve.

It does look to me as if you are copying the entire content of _data.  Am I right?   If so you should do that in one line by implementing the correct member function in the class.  Do you know how to do that?

Also fix const correctness when you can - doing it properly needs to be done from the bottom up so it's tedious.

kutschke: > I will check. There is a work around in that code to allow for add/remove tick box on the GUI and that data_ variable is part of that. It's a bit "hacky" but getting TEve to do what you want is a bit that way in general.

I am talking about the data members defined here:
https://github.com/NamithaChitrazee/Offline-1/blob/42715230c5fff46f969b7658f457b79cdac601a8/TEveEventDisplay/src/dict_classes/Data_Collections.h#L46

All are bare pointers to data products. It's possible to read a file that does not contain one of these data products, so the pointers can stay zero.  You should use a type of pointer that will throw an exception if you forget to check for validity and attempt to dereference an invalid pointer.

sophiemiddleton: Yes, I know we need to rewrite this. I have rewritten it a bit for REve but further advice is welcome (please email my with suggestions so I dont lose track)

PR #671, opened 2021-12-09
remove unused dictionary builds

rlcee:


kutschke (approved):
This matches what we discussed.

PR #673, opened 2021-12-09
Adding time offset to GenParticles

soleti:
This PR adds the possibility to add a random time offset to the input GenParticles.

### DIFF
--- a/CommonMC/src/CosmicTimeOffset_module.cc
+++ b/CommonMC/src/CosmicTimeOffset_module.cc
@@ -70,9 +70,9 @@ namespace mu2e {
     produces<SimTimeOffset>();
 
     addTimeOffset_ = conf().cosmicModuleLabel(cosmicModuleLabel_);
-    mayConsume<GenParticleCollection>(cosmicModuleLabel_);
 
     if (addTimeOffset_) {
+      consumes<GenParticleCollection>(cosmicModuleLabel_);
       produces<GenParticleCollection>();
     }
   }
--- a/CommonMC/src/CosmicTimeOffset_module.cc
+++ b/CommonMC/src/CosmicTimeOffset_module.cc
@@ -90,7 +90,8 @@ namespace mu2e {
   //================================================================
   void CosmicTimeOffset::produce(art::Event& event) {
     // Generate and record offset
-    std::unique_ptr<SimTimeOffset> toff(new SimTimeOffset(flatTime_.fire()));
+    const float timeOffset = flatTime_.fire();
+    std::unique_ptr<SimTimeOffset> toff(new SimTimeOffset(timeOffset));
     if( verbosityLevel_ > 1) std::cout << "CosmicTimeOffset " << toff->timeOffset_ << std::endl;
     event.put(std::move(toff));
 
@@ -100,7 +101,6 @@ namespace mu2e {
       event.getByLabel(cosmicModuleLabel_, cosmicParticles);
       const GenParticleCollection &particles(*cosmicParticles);
       std::unique_ptr<GenParticleCollection> offsetParticles(new GenParticleCollection);
-      const float timeOffset = flatTime_.fire();
 
       for (GenParticleCollection::const_iterator i = particles.begin(); i != particles.end(); ++i) {
         GenParticle const &particle = *i;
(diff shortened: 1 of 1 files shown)


CommonMC/src/CosmicTimeOffset_module.cc
```diff
@@ -31,6 +32,8 @@ namespace mu2e {
       using Name=fhicl::Name;
       using Comment=fhicl::Comment;
 
+      fhicl::Atom<std::string> cosmicModuleLabel{Name("cosmicModuleLabel"), Comment("Name of CORSIKA module label"), "generate"};
```
brownd1978, line 35: Isn't this more general than CORSIKA?  Also, shouldn't this be an optionalAtom, as we don't need it except for Extracted.  If you make it optional then you don't need the explicit boolean parameter.

CommonMC/src/CosmicTimeOffset_module.cc
```diff
@@ -55,13 +60,16 @@ namespace mu2e {
...
+    , engine_(createEngine(art::ServiceHandle<SeedService>()->getSeed()))
+    , cosmicModuleLabel_(conf().cosmicModuleLabel())
+    , addTimeOffset_(conf().addTimeOffset())
     , verbosityLevel_(conf().verbosityLevel())
     , intervalStart_(conf().intervalStart())
     , intervalEnd_(conf().intervalEnd())
     , flatTime_(engine_, intervalStart_, intervalEnd_)
   {
     produces<SimTimeOffset>();
+    produces<GenParticleCollection>();
```
brownd1978, line 72: This produces should be conditional on addTimeOffset_.  Also you need a mayConsume on the input GenParticle collection.

CommonMC/src/CosmicTimeOffset_module.cc
```diff
@@ -82,6 +90,23 @@ namespace mu2e {
...
+      art::Handle<mu2e::GenParticleCollection> cosmicParticles;
+      event.getByLabel(cosmicModuleLabel_, cosmicParticles);
+      const GenParticleCollection &particles(*cosmicParticles);
+      std::unique_ptr<GenParticleCollection> offsetParticles(new GenParticleCollection);
+
+      for (GenParticleCollection::const_iterator i = particles.begin(); i != particles.end(); ++i) {
+        GenParticle const &particle = *i;
+        offsetParticles->push_back(GenParticle(particle.pdgId(), particle.generatorId(),
+                                               particle.position(), particle.momentum(),
+                                               particle.time() + flatTime_.fire()));
```
brownd1978, line 105: This will give a different random time offset to every particle.  You want the same random offset to all particles in the same event.

brownd1978 (changes requested):

CommonMC/src/CosmicTimeOffset_module.cc
soleti: Yes, sorry, the comment is wrong and yes, it should be Optional

CommonMC/src/CosmicTimeOffset_module.cc
soleti: Ok

CommonMC/src/CosmicTimeOffset_module.cc
soleti: True, fixed.

CommonMC/src/CosmicTimeOffset_module.cc
```diff
@@ -82,6 +93,24 @@ namespace mu2e {
...
     if( verbosityLevel_ > 1) std::cout << "CosmicTimeOffset " << toff->timeOffset_ << std::endl;
     event.put(std::move(toff));
+
+    if (addTimeOffset_ ) {
+      // Add offset to all particles
+      art::Handle<mu2e::GenParticleCollection> cosmicParticles;
+      event.getByLabel(cosmicModuleLabel_, cosmicParticles);
+      const GenParticleCollection &particles(*cosmicParticles);
+      std::unique_ptr<GenParticleCollection> offsetParticles(new GenParticleCollection);
+      const float timeOffset = flatTime_.fire();
```
brownd1978, line 103: Actually this needs to be the same offset value is stored in the SimTimeOffset, for consistency.

soleti: Done

brownd1978 (approved):

brownd1978:
@FNALbuild run build test

PR #677, opened 2021-12-18
start alternate particle list service

rlcee:
- The previous code read from 3 tables, combined and modfied data.  This code simply captured the output of that process as the basis of the new table.
- mass is in MeV, lifetime is in ns (was MeV, ks)
- Rob needs to update doc 1120
- there were HepPDT names, geant names and PDGCode.hh names, this list uses primarily geant names and adds the concept of "code names" which don't contain any special characters so they can be used in code
- once this service passes review, we will migrate the use cases from the old service to the new service and then remove the old

### DIFF
--- a/GlobalConstantsService/inc/ParticleData.hh
+++ b/GlobalConstantsService/inc/ParticleData.hh
@@ -38,12 +38,7 @@ namespace mu2e {
     float mass() const { return _mass; }
     float lifetime() const { return _lifetime; }
 
-    void print( std::ostream& ostr=std::cout) const {
-      ostr << *this;
-    }
-
-    friend std::ostream& operator<<( std::ostream& output, 
-                                     const ParticleData& pd );
+    void print( std::ostream& ostr=std::cout) const;
 
   private:
     int _id;
@@ -54,6 +49,11 @@ namespace mu2e {
     float _lifetime;
 
   };  // ParticleData
+
+
+  std::ostream& operator<<( std::ostream& output, 
+                            const ParticleData& pd );
+
 
 } //end namespace mu2e
 
--- a/GlobalConstantsService/inc/ParticleData.hh
+++ b/GlobalConstantsService/inc/ParticleData.hh
@@ -1,0 +1,63 @@
+#ifndef GlobalConstantsService_ParticleData_hh
+#define GlobalConstantsService_ParticleData_hh
+//
+// Partcile data loaded from ParticleDataList
+//
+// the content is
+//  - PDG ID
+//  - displayname (includes special chars like +-*)
+//  - progamming name (includes only alphanumeric and _)
+//        this should be the same as  DataProducts/inc/PDGCode.hh
+//  - charge (units of electron charge)
+//  - mass (MeV)
+//  - lifetime (ns)
+//
+
+#include <iostream>
+#include <iomanip>
+
+#include "Offline/GlobalConstantsService/inc/ParticleData.hh"
+
+namespace mu2e {
+
+
+  class ParticleData {
+
+  public:
+
+    ParticleData(int id, const std::string& name, const std::string& codeName,
+                 float charge, float mass, float lifetime): 
+      _id(id),_name(name),_codeName(codeName),_charge(charge),
+      _mass(mass),_lifetime(lifetime) {}
+
+    int id() const { return _id; }
+    const std::string& name() const { return _name; }
+    const std::string& displayName() const { return _name; }
+    const std::string& codeName() const { return _codeName; }
+    float charge() const { return _charge; }
+    float mass() const { return _mass; }
+    float lifetime() const { return _lifetime; }
+
+    void print( std::ostream& ostr=std::cout) const {
+      ostr << std::setw(11) << _id;
+      ostr << std::setw(22) << _name;
+      ostr << std::setw(35) << _codeName;
+      ostr << std::setw(10) << _charge;
+      ostr << std::setw(12) << _mass;
+      ostr << std::setw(12) << _lifetime;
+      ostr << std::endl;
+    }
+
+  private:
+    int _id;
+    std::string _name;
+    std::string _codeName;
+    float _charge;
+    float _mass;
+    float _lifetime;
+
+  };  // ParticleData
+
+} //end namespace mu2e
+
+#endif /* GlobalConstantsService_ParticleData_hh */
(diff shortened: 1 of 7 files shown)


GlobalConstantsService/inc/ParticleDataList.hh
```diff
@@ -0,0 +1,69 @@
...
+// A simple particle data list
+//
+// this list was created from 
+// - HepPDT list
+// - mass_width_2008.mc
+// - g4nuclei.tbl
+// - aspects of the geant particle list
+// - aspects of DataProducts/inc/PDGCode.hh
+//
+// Units are standard units: electron_charge, MeV, and ns
```
kutschke, line 13: proton charge, not election charge, to get the sign right.

GlobalConstantsService/src/ParticleDataList.cc
```diff
@@ -0,0 +1,121 @@
...
+    float mass = (pA+exc)*protonMass;
+
+    std::ostringstream pName;
+    pName << "Mu2e_" << id;
+    std::string name = pName.str();
+
+    _list.try_emplace(id,id, name, name, float(pZ), mass, 0.0);
+
+    return _list.at(id);
+
```
kutschke, line 94: Do we want a smarter error message in the case that try_emplace fails?  As written you will just get the out of bounds message from _list.

GlobalConstantsService/src/ParticleDataList.cc
```diff
@@ -0,0 +1,121 @@
...
+  // *************************************************************
+
+  ParticleDataList::ParticleDataList( SimpleConfig const& config ) {
+
+    ConfigFileLookupPolicy findConfig;
+    
+    std::string tableFilename = findConfig(
+                    config.getString("particleDataList.filename",
+                   "Offline/GlobalConstantsService/data/ParticleList.txt") );
```
kutschke, line 24: I recommend no default on the grounds that if "particleDataList.filename" is intended to be in the fcl file, but is misspelled, I would like it to be an error.

GlobalConstantsService/src/ParticleDataList.cc
```diff
@@ -0,0 +1,121 @@
...
+    std::ifstream in(tableFilename.c_str());
+    if ( !in ) {
+      throw cet::exception("PARTICLELIST_OPEN_FAIL")
+          << "Unable to open particle data file: "
+          << tableFilename << "\n";
+    }
+
+    std::string line;
+    std::string str;
+    std::vector<std::string> words;
```
kutschke, line 39: I suggest

constexpr int exepectedWords = 7;
words.reserve(expectedWords);

and replace the literal 7 everywhere with this constexpr

GlobalConstantsService/src/ParticleDataList.cc
```diff
@@ -0,0 +1,121 @@
...
+          << "ParticleList: Not 7 words on line: "
+          << line << "\n";
+      }
+      int id = std::stoi(words[0]);
+      _list.try_emplace(id, id, words[1], words[2], 
+         std::stof(words[4]), std::stof(words[5]), std::stof(words[6]) );
+      _names.try_emplace(words[1],id);
+      if( words[2] != "none" ) _names.emplace(words[2],id);
+      if( words[3] != "none" ) _names.emplace(words[3],id);
+      words.clear();
```
kutschke, line 56: Do you also want to add the aliases here too?

GlobalConstantsService/inc/ParticleData.hh
```diff
@@ -0,0 +1,63 @@
...
+
+    void print( std::ostream& ostr=std::cout) const {
+      ostr << std::setw(11) << _id;
+      ostr << std::setw(22) << _name;
+      ostr << std::setw(35) << _codeName;
+      ostr << std::setw(10) << _charge;
+      ostr << std::setw(12) << _mass;
+      ostr << std::setw(12) << _lifetime;
+      ostr << std::endl;
+    }
```
kutschke, line 49: I suggest writing operator << and then implementing the print function using that.   Also I would put the implementation of operator << in the .cc file since I think that the code ends up being biggish to inline.

I know that I started this pattern but I would like to modernize it.

kutschke (commented):
I did not look at the content of the particle list carefully but I have looked at everything else.

GlobalConstantsService/src/ParticleDataList.cc
rlcee: if it finds the key, it will not change anything. If it does not find the key, it will construct the pair.  Either way, the key will be there.   The only failure might be something like running out of memory.  What error are you thinking we need to catch?

GlobalConstantsService/src/ParticleDataList.cc
rlcee: words[3] is the alias, so I'm not sure what you mean

GlobalConstantsService/src/ParticleDataList.cc
kutschke: You are right - there is no real use case.  If it runs out of memory it will core dump and we won't be able to issue a well phrased warning.

GlobalConstantsService/src/ParticleDataList.cc
kutschke: Duh.  So the one that's missing is lookup by displayname.  What's the argument for not supporting that?

rlcee: That's line 53, words[1]

kutschke: Thanks.  I was reading the shape of the code, not what it said.

GlobalConstantsService/inc/ParticleDataList.hh
kutschke: Thanks for fixing this.l

GlobalConstantsService/src/ParticleDataList.cc
kutschke: thanks.

GlobalConstantsService/src/ParticleDataList.cc
kutschke: Thanks.

GlobalConstantsService/inc/ParticleData.hh
kutschke: Thanks for doing this.  One last request.  If operator << is implemented using the public interface, then you do not need to friend it.  I believe everything that you do print is in the public interface.

kutschke: Thanks.

kutschke:
@FNALbuild run build test

kutschke (approved):
Thanks for addressing my comments.

PR #678, opened 2021-12-19
Poly cone fix

kutschke:
In the following I ask the reviewers to approve if the fixes to the code are good enough for now.  If they prefer a better solution for the longer term, I prefer that they submit a separate PR to address it.

For some time now the gdml file generated by the head of Offline has broken both the TEve and REve based event displays.  The issue is the following: we have several polycones in our geometry in which two planes have identical values for all three of (z, rmin and rmax).   The G4 experts tell me that this silently produces undefined behaviour in G4.  The ROOT graphics experts tell me that having two identical planes in a poly cone is illegal.

For the event displays we produce a gdml file from our G4 geometry and then read it into a TGeometry object.  There are errors when reading the gdml file into the TGeometry.   In a TEve or REve based event display the job crashes.  If I read the gdml file in a standalone module that does nothing else, I see printed error messages about the identical planes.  The error message mentions 4 polycones, given below along with the person I believe is responsible for that section of the geometry
 
 collimator1channel   ( Andrei )
 BSTS                         ( Dave (Kentucky) )
 BSTC                         ( Dave (Kentucky) )
 PSShieldEndRing1   ( Dave (Kentucky) )

I wrote code to walk the G4SolidStore and identify all G4PolyCones that have this problem.  There are only the four illegal polycones, the four mentioned above.  In a separate PR I will add this code to Offline .

The Root and G4 people say that it is OK if two planes have the same value of z so long as at least one of rmin or rmax changes.

I have verified that the standalone test works with a gdml file made from this geometry (no  printed warnings) and that the TEve event display works with this geometry.  The TEve event display seg faulted on exit; I don't know if that is related to this and I have informed Sophie.

After my changes, the root overlap check passes and the G4 surface check passed with 6 different random numbers.

### DIFF
--- a/Mu2eG4/src/constructExtMonFNALBuilding.cc
+++ b/Mu2eG4/src/constructExtMonFNALBuilding.cc
@@ -100,7 +100,10 @@ namespace mu2e {
                                            std::pow(boxdz * abs(tan(collimator.angleH())) + dr/abs(cos(collimator.angleH())), 2)
                                            );
 
-    double zPlane[] = {-cylHalfLength, -0.5*collimator.radiusTransitiondZ(), +0.5*collimator.radiusTransitiondZ(), +cylHalfLength };
+    // Fixme: protect against two planes at the same value of z.
+    double const epsilon= ( collimator.radiusTransitiondZ() == 0. ) ? 0.1 : 0.;
+
+    double zPlane[] = {-cylHalfLength, -0.5*collimator.radiusTransitiondZ()-epsilon, +0.5*collimator.radiusTransitiondZ()+epsilon, +cylHalfLength };
     double rzero[] = {0,0,0,0,0,0,0,0,0,0,0,0,0,0};
 
     //----------------------------------------------------------------
@@ -205,6 +208,7 @@ namespace mu2e {
     VolumeInfo channel(collimator.name()+"Channel",
                        CLHEP::Hep3Vector(0,0,0),
                        alignmentHole.centerInWorld);
+
 
     channel.solid = new G4IntersectionSolid(channel.name,
                                             parent.solid,
--- a/GeometryService/src/MBSMaker.cc
+++ b/GeometryService/src/MBSMaker.cc
@@ -262,7 +262,7 @@ namespace mu2e {
                       _CLV2MaterialName));
       
     } else if ( _MBSVersion >= 2 ) {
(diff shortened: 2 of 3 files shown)


kutschke:
@gaponenko and @dnbrow01  you as reviewers.  Please read the description of the PR to get the big picture and then look at the code that has your name on it in the note.  You are welcome to look at the other person's code but need not do so.

@michaelmackenzie I added you to keep you in the geometry loop.  You and Dave can sort out who does what.

@resnegfk I added you so that you are aware of this. 

I have also asked Sophie to check that this suits her needs.  I only did a minimal test of the event display.

kutschke:
Thanks Dave.

> But there are legitimate cases for having the z-values the same. So I'd say this is not the long-term solution.

Yes.  I did not touch the cases with the same z value but different radial values.  There is one other rule that I forgot to mention: if two planes have the same z, the radial extents must "overlap" - I don't have a precise definition of "overlap" so I don't know if touching but not overlapping is OK or not.  

> There is no reason I can see to have z, rin, and rout all the same for two or more planes. I don't remember why we left the code this way - perhaps the fact that G4 worked around it allowed this issue to not stay on our radars.

I suspect that a particular version had a radial step and then parameters evolved to remove the step but the code was not revisited.  Or perhaps that a step was foreseen as an option but we did not choose that option.

> To get something working quickly, I think it is ok to commit your workarounds, but I'll definitely have to revisit this and figure out what we were doing with this code and creating a robust solution.

If the design of the MBS is stable enough, there are a few polycones that could be turned into G4Tubs.

resnegfk (approved):

Mu2eG4/src/constructExtMonFNALBuilding.cc
```diff
@@ -206,6 +209,7 @@ namespace mu2e {
                        CLHEP::Hep3Vector(0,0,0),
                        alignmentHole.centerInWorld);
 
+
```
gaponenko, line 212: extraneous new line?

gaponenko (approved):
Hi Rob,  the fix to ExtMon geometry is good.   Thank you,
Andrei

michaelmackenzie (approved):
This all makes sense to me, thank you for addressing this!

brownd1978 (approved):

PR #680, opened 2021-12-22
move ParticleDataTable uses to new ParticleDataList

rlcee:
Switch to new, simpler ParticleDataList. Validation tests
- 1000 ceSimReco
- 500 potSim
- 300 reco
show exact matches.  I know this is a precise test because a few small diffs in precision with the old system did show differences in the ceSimReco and reco validation.

I also checked uses in TrackerAlignment, REve and TrkAna, and there will be one follow-on commit in TrkAna.


kutschke:
> 
> I also checked uses in TrackerAlignment, REve and TrkAna, and there will be one follow-on commit in TrkAna.

WHen you say that you checked  TrackerAlignment and REve, do you mean that they do not currently use the old system and therefore need no maintenance?  Is that correct? 

I see the PR in TrkAna.

Have you looked at  TrackerMCTune?

Outside of this PR the appropriate people will need to look at Tutorial and Stntuple.

rlcee:
> > I also checked uses in TrackerAlignment, REve and TrkAna, and there will be one follow-on commit in TrkAna.
> 
> WHen you say that you checked TrackerAlignment and REve, do you mean that they do not currently use the old system and therefore need no maintenance? Is that correct?

Yes

> I see the PR in TrkAna.
> 
> Have you looked at TrackerMCTune?

TrackerMCTune doesn't use ParticleDataTable

> 
> Outside of this PR the appropriate people will need to look at Tutorial and Stntuple.

Tutorial doesn't use ParticleDataTable

kutschke:
Thanks Ray.  I am surprised that Tutorial does not use ParticleDataTable but it's certainly simpler that it does not.

rlcee:
#672 references the ParticleDataTable include, which should be removed since it is not used.  This can be done independently from this PR.

If I run the geant particle print routine, which prints 6 sig fig, and compare to the text in the new particle table, all the printed digits agree exactly for the mass of the ele, mu, pi0 and pi+.  This is not definitive since the variables are doubles, but I'd have to figure out some code to go further in sig fig.  Let me know if I have to do that.

kutschke:
@FNALbuild run build test

kutschke (approved):
Ray's tests show exact matches and the changes are straightforward.  I am approve this and will merge.

PR #681, opened 2022-01-05
DbService update for DQM

rlcee:
- add DQM database to connection file which allows the Db code to be used for DQM database
- always add database name to queries. Needed for DQM, safer for future
- allow multiple where clauses in reader query


kutschke:
A comment to force an update of the status.

kutschke:
@FNALbuild run build test

kutschke (approved):
This looks straight forward and I don't expect it to change validation.

PR #676, opened 2021-12-13
constant reference edits for TFileDirectory objects

macndev:
This work is intermediate work being completed between iterations of art, v3.9 and v3.10, where TFileDirectory objects now need to be passed by constant reference. The modifications here were validated with 1k ceSimReco events against a reference branch, returning perfect matches:

macndev@mu2ebuild01.fnal.gov:/mu2e/app/users/macndev/OfflineArt310_ref:valCompare -s val_ref.root ../OfflineArt310/val_working.root
TValCompare Status Summary:
  370 Compared
    6 marked to skip
    0 had unknown status
    0 could not be compared
   24 had at least one histogram empty
    0 failed loose comparison
    0 passed loose comparison, failed tight
    0 passed tight comparison, not perfect match
  364 had perfect match
  364 passed loose or better
  364 passed tight or better

PR #682, opened 2022-01-07
Art v310 new

macndev:
compatibility fixes for art v3.10 - Many of these fixes call TFileDirectory objects by reference or by constant reference, like in PR #676. Other included fixes were to prefer forward declaration for the TFileDirectory namespace rather than using the TFileDirectory.h include file and use different syntax for checking for empty paths when defining TFileDirectory objects.


kutschke:
@FNALbuild run build test

kutschke:
@FNALbuild run build test

kutschke (approved):
After the fix to GeometryService/src/PSEnclosureMaker.cc (added to this PR),  Mackenzie reran the validation tests.   There are 3 exact matches:

https://mu2e.fnal.gov/atwork/computing/ops/val/plots/art_v3_10/ce/result_v2.html
https://mu2e.fnal.gov/atwork/computing/ops/val/plots/art_v3_10/pot/result_v2.html
https://mu2e.fnal.gov/atwork/computing/ops/val/plots/art_v3_10/reco/result.html

These are for 2k each ceSimReco and potSim;  and for 725 reco events.

I will approve and merge this and get higher stats validation on the overnight validation.

PR #685, opened 2022-01-13
Modify variant geometry files to be based on geom_common.txt

kutschke:
These geometry files are now deltas on geom_common.txt, not on a named geometry file.

### DIFF
--- a/Mu2eG4/geom/geom_common_no_tsu_ps.txt
+++ b/Mu2eG4/geom/geom_common_no_tsu_ps.txt
@@ -1,7 +1,7 @@
 //
 // Variant on geom_common.txt for production running of G4 jobs
-// in which the DS is hidden by stopping rules in G4.
-// The DS field map is removed to save space
+// in which the TSu and PS field maps are not needed.  These
+// field maps are removed to save space.
 //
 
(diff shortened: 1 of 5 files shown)


Mu2eG4/geom/geom_common_no_tsu_ps.txt
```diff
@@ -4,7 +4,7 @@
 // The DS field map is removed to save space
```
michaelmackenzie, line 4: This should say the TS upstream and PS fields are removed, and the DS is not hidden

kutschke: Thanks.  Fixed and pushed.

michaelmackenzie (approved):

kutschke:
I have verified that g4test_03.fcl ( which looks at single conversion electrons ) gives sensible output using geom_common_no_tsu_ps.txt and geom_common_reco.txt .  I have checked that it runs a few events with geom_common_no_ds.txt - but there is no expectation that it would give sensible results.  Doing a real test for no_ds would take a long run of potSim and I don't think that we need to do it.

brownd1978 (approved):

PR #686, opened 2022-01-14
Mu2eII_SM21: Merge head of Offline including new physics generators

sophiemiddleton:
Last big PR for Mu2eII_SM21 to include the new physics generators


kutschke (approved):
After an off-github discussion  with Sophie, we decided that threshold for merging is that it passes the CI.  Any additional testing is the responsibility of the Mu2e-II co-convenors.

Issue #690, opened 2022-01-16
Refactor Mu2eG4/src/SensitiveDetectorHelper.cc 


kutschke:
SensitiveDetectorHelper (SDHelper) was developed prior to our use of G4MT.  In sequential mode there is a single SensitiveDetectorHelper (SDHelper) object that is a data member of the Mu2eG4 module class.  It is passed by reference/pointer into the various classes that use it.

In MT mode, with N worker threads, there are N+2 SDHelper objects.  There is one object in each worker thread and one in the master thread, each held as a data member of the corresponding RunManager class.   The last instance is a temporary object in the c'tor of Mu2eG4/src/Mu2eG4IOConfigHelper.cc.

Some of the functions of SensitiveDetectorHelper must only be called from the master thread and others must only be called from a worker thread.  The class should be refactored to separate these functions so that we cannot call a function from the wrong type of thread.   Calling functions from the wrong thread was the source of the bug fixed in PR 

Some notes towards the design discussion.   A good target would be to separate the class into 3 classes:

1. One class should be a class that can be used in any thread; most of the static state should be in this class.
2. A second class should contain only the functions that may only be called in the master thread
3. The third class should contain only the functions that may only be called in a worker thread; most of the time varying state should be in this class.

Issue #601, opened 2021-09-29
reference binding to null pointers


rlcee:
references to null pointers, found with -fsanitize=undefined

Offline/Mu2eG4/src/constructMBS.cc:77:57: runtime error: reference binding to null pointer of type 'const struct Tube'
Offline/Mu2eG4/src/constructMBS.cc:78:57: runtime error: reference binding to null pointer of type 'const struct Tube'
Offline/Mu2eG4/src/constructMBS.cc:79:54: runtime error: reference binding to null pointer of type 'const struct Tube'
Offline/Mu2eG4/src/constructMBS.cc:80:54: runtime error: reference binding to null pointer of type 'const struct Tube'
Offline/Mu2eG4/src/constructMBS.cc:82:68: runtime error: reference binding to null pointer of type 'const struct Tube'

Offline/Mu2eG4/src/Mu2eG4EventAction.cc:111:34: runtime error: reference binding to null pointer of type 'struct StepPointMCCollection'
Offline/CaloMC/src/CaloShowerStepMaker_module.cc:249:42: runtime error: reference binding to null pointer of type 'const struct PhysicalVolumeInfoMultiCollection'

++++ source event
Begin processing the 5th record. run: 1 subRun: 0 event: 5 at 28-Sep-2021 20:11:19 CDT
/cvmfs/mu2e.opensciencegrid.org/artexternals/gcc/v9_3_0/Linux64bit+3.10-2.17/include/c++/9.3.0/bits/stl_iterator.h:820:17: runtime error: reference binding to null pointer of type 'const struct CaloCluster'
/cvmfs/mu2e.opensciencegrid.org/artexternals/gcc/v9_3_0/Linux64bit+3.10-2.17/include/c++/9.3.0/bits/stl_vector.h:1132:16: runtime error: reference binding to null pointer of type 'const struct value_type'
++++ finished source event

rlcee:
1) constructMBS is fixed in #661 
2) Offline/Mu2eG4/src/Mu2eG4EventAction.cc:111:34: runtime error: reference binding to null pointer of type 'struct StepPointMCCollection' should be tackled by one of the mu2e geant experts
3) CaloShowerStepMaker_module.cc is obvious what's wrong, but not obvious how to fix it efficiently and cleanly
4) in source event - I think these are in art since source is EmptyEvent .  After the above get fixed, I'll try to set up an art ticket

rlcee:
both ceSimReco and reco paths no longer have this problem.

PR #691, opened 2022-01-16
Fix bug in the registration of Logical Volume Sensitive Detectors

kutschke:
Logical Volume Sensitive Detectors (LVSD) have never worked correctly in MT mode.  This commit fixes the immediate problem but the code remains fragile.  Issue #690 in the Offline GitHub issue tracker has been created to remove the fragility.

The immediate problem is that in MT mode there are many SensitiveDetectorHelper objects and the member function:

   void SensitiveDetectorHelper::instantiateLVSDs(const SimpleConfig& config);

correctly instantiates the objects but stores the pointers in the wrong SensitiveDetectorHelper object.  They are stored in the instance that lives in the Mu2eG4MTRunManager object, not in the object that lives in the appropriate Mu2eG4WorkerRunManager instance.

The solution is to leave the instantation as is but to move the storage of pointers into:

  void SensitiveDetectorHelper::registerSensitiveDetectors();

which is called at the right time using the right SensitiveDetectorHelper instance.

As part of the fix I made it manifest that instantiateLVSDs does not register any pointers; I also  moved an invariant computation out of a loop.

### DIFF
--- a/Mu2eG4/src/SensitiveDetectorHelper.cc
+++ b/Mu2eG4/src/SensitiveDetectorHelper.cc
@@ -137,9 +137,9 @@ namespace mu2e {
     art::ServiceHandle<Mu2eG4Helper> helper;
 
     for(auto& iter : lvsd_) {
-      iter.second.sensitiveDetector = new Mu2eG4SensitiveDetector(iter.first, config);
-      SDman->AddNewDetector(iter.second.sensitiveDetector);
-      helper->locateVolInfo(iter.first).logical->SetSensitiveDetector(iter.second.sensitiveDetector);
+      auto sd = new Mu2eG4SensitiveDetector(iter.first, config);
+      SDman->AddNewDetector(sd);
+      helper->locateVolInfo(iter.first).logical->SetSensitiveDetector(sd);
     }
   }
 
@@ -149,6 +149,8 @@ namespace mu2e {
   void SensitiveDetectorHelper::registerSensitiveDetectors(){
     G4SDManager* sdManager = G4SDManager::GetSDMpointer();
 
+    const bool printWarnings = (verbosityLevel_ > -1) ? true : false;
+
     for ( InstanceMap::iterator i=stepInstances_.begin();
           i != stepInstances_.end(); ++i ) {
       StepInstance& step(i->second);
@@ -156,7 +158,6 @@ namespace mu2e {
         std::cout << __func__ << " looking for sd with name: "
                   << step.stepName << std::endl;
             }
-      bool printWarnings = (verbosityLevel_ > -1) ? true : false;
       step.sensitiveDetector =
         dynamic_cast<Mu2eG4SensitiveDetector*>(sdManager->FindSensitiveDetector(step.stepName.c_str(),printWarnings));
     }
@@ -165,6 +166,17 @@ namespace mu2e {
       dynamic_cast<ExtMonFNALPixelSD*>(sdManager->
                                        FindSensitiveDetector(SensitiveDetectorName::ExtMonFNAL()))
       : nullptr;
+
+    // Logical volume SDs, it any.
+    for ( auto& i : lvsd_ ){
+      StepInstance& step(i.second);
+      if (verbosityLevel_ > 0 ) {
+        std::cout << __func__ << " looking for lvsd with name: "
+                  << step.stepName << "   " << this << std::endl;
+      }
+       step.sensitiveDetector =
+         dynamic_cast<Mu2eG4SensitiveDetector*>(sdManager->FindSensitiveDetector(step.stepName.c_str(),printWarnings));
+    }
 
   }
 


kutschke:
@oksuzia  Yuri: please test that this solves your problem with logical volume sensitive detectors.  Please check in both MT and sequential modes; there should not no change in sequential mode.

@goodenou Please check my understanding of the MT-ology.  I have some notes I can send you if that would help.

@resnegfk I included you mostly to keep you in the loop. Review what you feel is appropriate.

rlcee:
While you are in this code, would you consider looking at issue #601 item 2?

kutschke:
> While you are in this code, would you consider looking at issue #601 item 2?

I will have a look and if I figure it out I will submit as a separate PR.

I believe that to reproduce, I need to rebuild Offline/Mu2eG4/src/constructMBS.os with -fsanitize=undefined and then re-link  Mu2eG4.so.   Is that right?  Do I need the flag on the link?

rlcee:
> > While you are in this code, would you consider looking at issue #601 item 2?
> 
> I will have a look and if I figure it out I will submit as a separate PR.
> 
> I believe that to reproduce, I need to rebuild Offline/Mu2eG4/src/constructMBS.os with -fsanitize=undefined and then re-link Mu2eG4.so. Is that right? Do I need the flag on the link?

https://mu2ewiki.fnal.gov/wiki/CodeDebugging#fsanitize_bounds_check
Set a build flag and an environmental.  I believe the flag does need to 
be set on compile and link.  I haven't tried to build only one lib with the flag,
only the whole repo, so I can't say if that works as expected.

kutschke:
Ooops. When I added Yuri as a reviewer two days ago, I meant to add Andy.  It was his bug report that triggered this work.  So I just added Andy.

@oksuzian :  let me know if you plan to look at this.  if not, I will remove you from the reviewer list.

Sorry for the confusion.

AndrewEdmonds11 (commented):
Thanks, Rob. I can confirm that the segfault I was seeing no longer occurs in multi-threaded. However, when you say there should be no change between multi-threaded and single-threaded mode, do you mean that they should produce identical outputs? Because I don't see that...

kutschke:
> Thanks, Rob. I can confirm that the segfault I was seeing no longer occurs in multi-threaded. However, when you say there should be no change between multi-threaded and single-threaded mode, do you mean that they should produce identical outputs? Because I don't see that...

Thanks Andy.  Sorry.  I meant no change on a statistical basis - the random number seed management is different in the two cases so you won't get an exact match.  I think it's good enough if the high stats plots, which are also immune from single event artifacts, compare well.

AndrewEdmonds11:
Thanks, Rob. The tests I did were just 100 events, so I'll run some higher-stats ones tonight and let you know how the comparison looks.

kutschke:
Thanks Andy.

From: Andrew Edmonds ***@***.***>
Date: Tuesday, January 18, 2022 at 4:16 PM
To: Mu2e/Offline ***@***.***>
Cc: Rob Kutschke ***@***.***>, Mention ***@***.***>
Subject: Re: [Mu2e/Offline] Fix bug in the registration of Logical Volume Sensitive Detectors (PR #691)


Thanks, Rob. The tests I did were just 100 events, so I'll run some higher-stats ones tonight and let you know how the comparison looks.

—
Reply to this email directly, view it on GitHub<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_Mu2e_Offline_pull_691-23issuecomment-2D1015883258&d=DwMCaQ&c=gRgGjJ3BkIsb5y6s49QqsA&r=Zi5IH-k2_JiK_Wipiv3pwwSXO9UNaHiLP8sisfPz__k&m=Jk9y7M9xt0E4eQAbD2NbHJqJL_j-UIuiO4ZZHVDRCgUO_R-v2wOOuaPOw6iQKBk9&s=ll-Ly0fY0NrK0BYxwe1aoyd3y_PBuOsxoAlIPiRuWlg&e=>, or unsubscribe<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_notifications_unsubscribe-2Dauth_ABHY5ZTJSXCGBZXDREVW4TTUWXRB7ANCNFSM5MDJ3HLA&d=DwMCaQ&c=gRgGjJ3BkIsb5y6s49QqsA&r=Zi5IH-k2_JiK_Wipiv3pwwSXO9UNaHiLP8sisfPz__k&m=Jk9y7M9xt0E4eQAbD2NbHJqJL_j-UIuiO4ZZHVDRCgUO_R-v2wOOuaPOw6iQKBk9&s=jnVSGFmt7WWqCo4LA-o06bbLbU5ewwGlgTPj5JIlv5c&e=>.
Triage notifications on the go with GitHub Mobile for iOS<https://urldefense.proofpoint.com/v2/url?u=https-3A__apps.apple.com_app_apple-2Dstore_id1477376905-3Fct-3Dnotification-2Demail-26mt-3D8-26pt-3D524675&d=DwMCaQ&c=gRgGjJ3BkIsb5y6s49QqsA&r=Zi5IH-k2_JiK_Wipiv3pwwSXO9UNaHiLP8sisfPz__k&m=Jk9y7M9xt0E4eQAbD2NbHJqJL_j-UIuiO4ZZHVDRCgUO_R-v2wOOuaPOw6iQKBk9&s=q6R8kxHAELIxGbHW7rlgr6mpbfeihi07Sw3izT5a68M&e=> or Android<https://urldefense.proofpoint.com/v2/url?u=https-3A__play.google.com_store_apps_details-3Fid-3Dcom.github.android-26referrer-3Dutm-5Fcampaign-253Dnotification-2Demail-2526utm-5Fmedium-253Demail-2526utm-5Fsource-253Dgithub&d=DwMCaQ&c=gRgGjJ3BkIsb5y6s49QqsA&r=Zi5IH-k2_JiK_Wipiv3pwwSXO9UNaHiLP8sisfPz__k&m=Jk9y7M9xt0E4eQAbD2NbHJqJL_j-UIuiO4ZZHVDRCgUO_R-v2wOOuaPOw6iQKBk9&s=TMkCrByT8xuTfnjeWNLeZgSTTAqA5G59-7kGXiEGodI&e=>.

AndrewEdmonds11 (approved):
Hi Rob,

I ran 10k POT overnight in single-threaded and multi-threaded mode and the results are statistically consistent. This plot shows the energy deposit of StepPointMCs in the primary beam window from the single-threaded (red) and multi-threaded (blue) jobs

Mu2eSim_PSEndCap_BugFixCheck

Thanks,
Andy

goodenou (commented):
Hi Rob. The code changes you had added look fine. I am curious about the addition of lines 170-180. That loop over the logical volumes has not been in any of the code going back to pre-MT days. Why was the lack of it causing a seg fault?

kutschke:
Hi Andy,

  Thanks for doing the tests - the plot looks good.  I am happy with the results and am ready to move ahead when the reviewers are.

Rob

kutschke:
> Hi Rob. The code changes you had added look fine. I am curious about the addition of lines 170-180. That loop over the logical volumes has not been in any of the code going back to pre-MT days. Why was the lack of it causing a seg fault?

Hi Lisa.  

  First a bit of github-ology - remember to click the approve button when you are read to approve. It's well hidden - let me know if you don't find it.

Next, be sure to read the information at the start of issue #690.  

Here we go: for N worker threads there are N+2 SDHelper objects, one in each worker thread, one in the master thread and one that is outside of the scope of this discussion.   The SDHelper instances are owned by the appropriate run manager object and are passed into other objects by ref/pointer.   Recall that we often pass helpers into the c'tor of a callback object and hold it by pointer as a data member of the callback object; then we use the helper in some of the member functions of the callback object.  

The short version is this: the function SensitiveDetectorHelper::instantiateLVSDs should have been named instantiateAndRegisterLVSDs because it does both jobs.  The instantiate part is OK.  It is called within a worker thread so it correctly creates the objects and registers them with worker thread G4SDManager object.  The problem is that it is called on the SDHelper instance that lives in the master thread - we get away with this because the function only uses the stack and heap, never thead local memory.   So all of that is OK.  But when we stash the pointer to the newly created LVSD object into the SDHelper object, we store it in the master thread SDHelper object.  So the worker thread SDHelper objects never know about their LVSD objects and when we try to use them they fail. 

 On the other hand SensitiveDetectorHelper::registerSensitiveDetectors is called in each worker thread using the SDHelper instance from that thread.  So now the pointers to the SD objects made in instantiateLVSDs are stored in the right place and the code works.

  The hack fix is just to add the LVSDs to the work done in registerSensitiveDetectors, which is called from the work thread instances of SDHelper. 

The better solution is outlined in issue #690: split SDHelper into pieces so that it's manifest what is called from which thread.  This is a lower priority than the work you are doing now to fix the problems at static d'tor time.

In sequential mode there is a single SDHelper object - so there is no ambigutity about which instance to use in which case.  So, back in the day, it was OK to have all of these functions bundled into one class; and that's the way it was designed.

I don't believe that we ever gave you  fcl file that exercised this feature.  So there is no reason for you to have encountered this issue before.

goodenou (approved):
Thank you for the explanation Rob. I approve this PR.

kutschke:
I had added Yuri by accident.  In a private email thread he says that it's OK to remove hime as a reviewer.  So I did.

resnegfk (approved):

PR #694, opened 2022-01-25
update to p019 to remove hepPDT

rlcee:
the only change is to remove setup of hepPDT, which is now unused


kutschke (approved):
Ray and I had talked about this earlier. I checked the p019 is indeed installed on /cvmfs.

PR #696, opened 2022-01-27
protect against empty select statement

rlcee:
present, but empty, select clause leads to incorrect results.  Recent dqm work was the first test of this case..


kutschke (approved):
Thanks for finding and fixing this.

PR #697, opened 2022-01-28
added mother volumes around CRV sectors

ehrlich-uva:


kutschke:
For the record, this reduces the number of daughter volumes in the HallAir volume from 1700 to 949; that's good.   A few questions for Ralf:

Have you run both the root and G4 overlap checkers?

What tests have you done to verify that the new code is working correctly?

Please pick a suitable test job and report on the CPU time and memory used by the job.  I am looking for the numbers from the art log file:

```
 TimeReport ---------- Time summary [sec] -------
 TimeReport CPU = 70.728307 Real = 73.627714

 MemReport  ---------- Memory summary [base-10 MB] ------ 
 MemReport  VmPeak = 2771.57 VmHWM = 2045.25

```

  Please report on two tests: first, do just one event, to measure any changes to the startup time and memory use.  The second is to run for at least 30 minutes of CPU time.  Be sure to pick a job that is dominated by the G4 time and memory, not by other modules - maybe you should only run the generator and G4.

ehrlich-uva:
I did the following overlap checks:
- Mu2eG4/fcl/surfaceCheck.fcl
- Mu2eG4/fcl/gdmldump.fcl with bin/overlapCheck.sh

My tests were done with Production/Validation/cosmicSimReco.fcl (which includes the entire CRV sim/reco package and not just the Mu2eG4 part) with 20,000 events and 5 different seeds for the original geometry and new geometry with mother volumes. The spreadsheet shows the following comparisons: 
- the total time, the time for just the Mu2eG4 module, and the memory usage
- a comparison for the CrvSteps (number and average start_z) and CrvRecoPulses (number and average PEs)

motherVolumes

I will make more tests with just the Mu2eG4 module and for 1 event and many events (>30 minutes).

ehrlich-uva:
I cut the cosmicSimReco.fcl to only include the generate and Mu2eG4 modules. I run it for the original and new geometry with 5 different seeds. Here is the comparison for 50,000 events:
motherVolumes50000
And here for 1 event:
motherVolumes1

resnegfk:
A lot of important, and at the same tedious/hard work.  

I would suggest adding comments explaining what is happening in the added code, 
like what is placed in what, shifted by what and why, how the unions are formed, common surfaces avoided etc.

Other comments:
Change NULL to nullptr
I find the dependence on ROOT through the Form function somewhat surprising.
Perhaps use std::ostringstream from sstream in future?

Perhaps use
services.TimeTracker.printSummary: true
for time reports in future?

ehrlich-uva:
Thank you, Krzysztof. I will make the changes.
services.TimeTracker.printSummary was set to true. That's how I got the timing information in the spreadsheets.

resnegfk:
> Thank you, Krzysztof. I will make the changes. services.TimeTracker.printSummary was set to true. That's how I got the timing information in the spreadsheets.

Thank you Ralf. To clarify what I meant: I am only suggesting to add comments at this time (while you still remember all the details :) ) to preserve that should the code need to be modified in the future, and use nullptr where appropriate. 

The Form replacement should probably be another PR if we decide that the ROOT dependence should be avoided

kutschke (approved):

kutschke:
@ehrlich-uva I am checking in to see where we are on this.  I think that I am waiting for you to respond to come comments from Krzysztof.  Is that also your understanding.

ehrlich-uva:
I'm going to finish in the next couple of days.

ehrlich-uva:
I added comments to the code and changed NULL to nullptr

Mu2eG4/src/constructCRV.cc
```diff
@@ -76,23 +83,216 @@ namespace mu2e
...
+      {
+        //strong back dimensions
+        if(l==0)
+        {
+          const CRSAluminumSheet &aluminumSheet0 = module0.getAluminumSheets().at(0);
+          const CRSAluminumSheet &aluminumSheet1 = module1.getAluminumSheets().at(0);
+          const CLHEP::Hep3Vector &centerInMu2e0=aluminumSheet0.getPosition();
+          const CLHEP::Hep3Vector &centerInMu2e1=aluminumSheet1.getPosition();
+          motherCenterInMu2e.emplace_back(0.5*(centerInMu2e0+centerInMu2e1));
+          std::vector<double> halfLengths;
```
resnegfk, line 111: I believe these (and similarly used) halfLengths could be just C-style arrays and could be declared outside the loops they are in (and if the vectors or std::arrays were to be used std::vector<double> halfLengths(3) would be better)

Mu2eG4/src/constructCRV.cc
```diff
@@ -76,23 +83,216 @@ namespace mu2e
...
+      G4VSolid *motherSolid=nullptr;
+      for(size_t l=0; l<motherCenterInMu2e.size(); ++l)
+      {
+        //solid of one l-th layers of the CRV sector
+        G4VSolid *motherSolidPart=new G4Box(Form("%s_%s_%lu","CRSmotherSolidPart",CRVsectorName.c_str(),l),
+                                            motherHalfLengths[l][0],
+                                            motherHalfLengths[l][1],
+                                            motherHalfLengths[l][2]);
+        //need an additional solid that overlaps the solids of two adjacent layers,
+        //because G4UnionSolids require fully overlapping solids
```
resnegfk, line 182: I think
// //because G4UnionSolids require constituent solids to have a fully overlapping region
would be more clear

kutschke: I agree with Krzysztof - describing this well will be the most valuable to a future maintainer.

kutschke: A technique that I find useful is to put numbered notes at the top of the file; then the comments in the code can refer to them.  That way a long comment does not disrupt the layout of the code and you can write as long a comment as is needed to tell the story well.  This is also useful if you need to make the same comment in several places.  I don't know if that will help in this cause - just be aware it's an option.

resnegfk (approved):
As far as I can tell from reading the code, based on my quick tests, the integration tests and tests described in Mu2e doc 40541 v3 it all seems to be good. 
One would need to redo the work to be really sure :)
I made some minor comments regarding constructCRV.cc which can be addressed later.
When doing it one can also consider removing trailing white space from the files touched by this PR.
I created https://github.com/Mu2e/Offline/issues/709 to consider replacing TString::Form

kutschke:
@FNALbuild run build test

PR #698, opened 2022-01-29
Making the code compatible with Geant4 v11

resnegfk:
This pull request makes changes enabling the use of Geant4 v11.0 while preserving compatibility with the current,
10.7, version of Geant4

Changes in the first commit add a newly introduced process name "NoProcess" which was added for accounting 
purposes and is not assigned to any of Geant4 particle process managers, but is used to acknowledge, that
in certain situations, no process has acted, therefore we need to add it to all particles here.
While adding it in PhysicsProcessInfo the class was refactored, which was long overdue.
While refactoring a small logic error was corrected. It will result in more particles to be added to the ProcInfo in it,
but it seems that that information was not relied upon, so it will likely not change any of the current results.
Two of the following commits remove obsolete entities, and another one adapts the code to the Geant4 code evolution, 
neither of which should result in any change of functionality.

### DIFF
--- a/MCDataProducts/inc/ProcessCode.hh
+++ b/MCDataProducts/inc/ProcessCode.hh
@@ -87,7 +87,7 @@ namespace mu2e {
       mu2eExternalRMC,         mu2eFlateMinus,      mu2eFlatePlus, mu2eFlatPhoton, // 175
       mu2eCePlusLeadingLog, mu2ePionCaptureAtRest, mu2eExternalRPC, mu2eInternalRPC, 
       mu2eunused5, mu2eunused6, mu2eunused7, mu2eunused8, 
-      uninitialized,
+      uninitialized, NoProcess,
       lastEnum,
       // An alias for backward compatibility
       mu2eHallAir = mu2eKillerVolume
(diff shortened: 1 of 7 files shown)


brownd1978 (approved):

kutschke (approved):
Do you expect this PR to produce an exact match in the nightly validation or small changes?

resnegfk:
I am expecting an exact match (this is what I got with g4test_03.fcl)

kutschke:
> I am expecting an exact match (this is what I got with g4test_03.fcl)

Thanks.  I will check with Dave and, unless he has conflicting requirements, I will merge today.

resnegfk:
> > I am expecting an exact match (this is what I got with g4test_03.fcl)
> 
> Thanks. I will check with Dave and, unless he has conflicting requirements, I will merge today.

It would still be prudent to make this PR the only one which needs to be validated that it indeed does not result in any changes, due to the PhysicsProcesInfo refactoring

kutschke:
> It would still be prudent to make this PR the only one which needs to be validated that it indeed does not result in any changes, due to the PhysicsProcesInfo refactoring

Yes. That's the plan.  Merging now.

kutschke:
> > It would still be prudent to make this PR the only one which needs to be validated that it indeed does not result in any changes, due to the PhysicsProcesInfo refactoring
> 
> Yes. That's the plan. Merging now.

The tests ran with an exact match last night.  Thanks for looking after this.

PR #699, opened 2022-02-04
Improved 2D Projections for TEve and GDML added as an input parameter

sophiemiddleton:
Includes two small but significant changes:
1) The new YZ projection is used instead of the hack we had before
2) GDML is now taken as an input parameter allowing this to be used with Mu2e-II (or other) geometry versions
3) Option to look at the production target and see Proton trajectory (and others) inside the PS

### DIFF
--- a/TEveEventDisplay/fcl/prolog.fcl
+++ b/TEveEventDisplay/fcl/prolog.fcl
@@ -38,7 +38,8 @@ TEveEventDisplayBase : {
       addMCTraj				          : true 
 
     }
-  particles :   [11,13,2212,2112,211,22,212]
+    gdmlname :    "Offline/TEveEventDisplay/src/mu2e.gdml"
+    particles :   [11,13,2212,2112,211,22,212]
   }  
   
 TEveEventDisplayUpstream : {
@@ -66,6 +67,7 @@ TEveEventDisplayUpstream : {
       addCosmicSeedFit	          : false
       addMCTraj				            : true
    }
+    gdmlname :    "Offline/TEveEventDisplay/src/mu2e.gdml"
     particles :   [11,13,2212,2112,211,22,212]
     
   }  
@@ -80,7 +82,7 @@ TEveEventDisplayHelix : {
       isMCOnly : false
       filler : {
         ComboHitCollection 	  : "makePH"
-	TimeClusterCollection                  : "MHDeM"
+	      TimeClusterCollection                  : "MHDeM"
         CrvRecoPulseCollection	  :  "SelectRecoMC"
         CosmicTrackSeedCollection :  NULL
         CaloClusterCollection     :  "CaloClusterMaker"
@@ -90,8 +92,8 @@ TEveEventDisplayHelix : {
         TrkExtTrajCollection      :  NULL
         MCTrajectoryCollection    :  "compressRecoMCs"
         addHits 		  : true
-	addTrkHits 		  : true
-	addTimeClusters           : true
+	      addTrkHits 		  : true
+	      addTimeClusters           : true
         addTracks 		  : true
         addCrvHits		  : false
         addCosmicSeedFit	  : false
@@ -99,6 +101,7 @@ TEveEventDisplayHelix : {
         addTrkExtTrajs		  : false
         addMCTraj		  : true
       }
+      gdmlname :    "Offline/TEveEventDisplay/src/mu2e.gdml"
       particles :   [-11,11,13,-13,2212,2112,211,22,212]
   }   
 
@@ -126,7 +129,7 @@ TEveEventDisplayCosmics : {
       addTrkExtTrajs		        : false
       addMCTraj				          : false 
    }
-   gdmlname :         "Offline/TEveEventDisplay/src/mu2e.gdml"
+   gdmlname :    "Offline/TEveEventDisplay/src/mu2e.gdml"
    particles :   [11,13,2212,2112,211,22,212]
   }   
 
--- a/TEveEventDisplay/fcl/prolog.fcl
+++ b/TEveEventDisplay/fcl/prolog.fcl
@@ -11,10 +11,7 @@ TEveNonGDMLTest : {
 TEveEventDisplayBase : {
     module_type : TEveEventDisplay
     accumulate : false
-    showCRV : false
-    showBuilding : false  
-    showDSOnly : true 
-    showEvent : true
+
     isMCOnly : false
     filler : {
       ComboHitCollection 	  : "makePH"
@@ -39,6 +36,12 @@ TEveEventDisplayBase : {
     }
     gdmlname :    "Offline/TEveEventDisplay/src/mu2e.gdml"
     particles :   [11,13,2212,2112,211,22,212]
+    show : {
+      showCRV : false
+      showBuilding : false  
+      showDSOnly : true 
+      showInsidePS : false
+    }
   }  
   
 TEveEventDisplayHelix : {
@@ -47,8 +50,7 @@ TEveEventDisplayHelix : {
   
 TEveEventDisplayUpstream : {
     @table::TEveEventDisplayBase
-    showDSOnly : false
-    showInsidePS : true
+    
     isMCOnly : true
     filler : {
       ComboHitCollection 		      : NULL
@@ -69,13 +71,17 @@ TEveEventDisplayUpstream : {
       addCosmicSeedFit	          : false
       addMCTraj				            : true
    }    
+   show : {
+      showDSOnly : false
+      showInsidePS : true
+    }
   }  
   
  
 
 TEveEventDisplayCosmics : { # for stright cosmic for alignment (ordinary cosmics can use the base)
     @table::TEveEventDisplayBase
-    showCRV : true
+
     filler : {
       ComboHitCollection 		    : "makeSH"
       TimeClusterCollection                  : "MHDeM"
@@ -97,6 +103,9 @@ TEveEventDisplayCosmics : { # for stright cosmic for alignment (ordinary cosmics
       addTrkExtTrajs		        : false
       addMCTraj				          : false 
    }
+    show :{
+      showCRV : true
+    }
   }   
 
 
--- a/TEveEventDisplay/fcl/prolog.fcl
+++ b/TEveEventDisplay/fcl/prolog.fcl
@@ -126,6 +126,7 @@ TEveEventDisplayCosmics : {
       addTrkExtTrajs		        : false
       addMCTraj				          : false 
    }
+   gdmlname :         "Offline/TEveEventDisplay/src/mu2e.gdml"
    particles :   [11,13,2212,2112,211,22,212]
   }   
 
--- a/TEveEventDisplay/fcl/prolog.fcl
+++ b/TEveEventDisplay/fcl/prolog.fcl
@@ -46,16 +46,17 @@ TEveEventDisplayUpstream : {
     module_type : TEveEventDisplay
     showCRV : false
     showDSOnly : false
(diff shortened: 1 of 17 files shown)


sophiemiddleton:
@FNALbuild run build test

kutschke:
I just added Ray as a reviewer to address one specific question.  If you look at line 41 of TEveEventDisplay/fcl/prolog.fcl there is a hardcoded name of a gdml file.  I don't really like the idea of the EventDisplay team maintaining a special gdml file that they need to maintain in their file space.  I prefer that people get in the habit of using the current gdml file; by "current" I mean the version created by the code that the person is running, whether it's from a Musing or from their own build.  So a few questions for Ray and Sophie:

1. Sophie: is there a compelling reason NOT to use the current gdml file?
2. Ray.  What is the best way to access the current gdml file? I see two use cases.  One is when the users is working from a Musing and the other is when they are working from a local build.

TEveEventDisplay/fcl/prolog.fcl
```diff
@@ -38,23 +38,25 @@ TEveEventDisplayBase : {
...
       ComboHitCollection 		      : NULL
-      TimeClusterCollection                  : NULL
+      TimeClusterCollection       : NULL
       CosmicTrackSeedCollection 	: NULL
       CrvRecoPulseCollection		  : NULL
       CaloClusterCollection		    : NULL
-      CaloHitCollection    : NULL
-      MCTrajectoryCollection      : "BeamFilter" #"g4run"
+      CaloHitCollection           : NULL
+      MCTrajectoryCollection      : "g4run" #"BeamFilter"
```
kutschke, line 59: What's the intention here?  Is this something that end users might normally edit?  If so I would like to change the design so that end users only make changes using overrides in their top level fcl.    Perhaps it would make more sense to define two configurations, by defining the second as a delta on the first.  Then the thing you edit is the name of the configuration in your top level fcl file.

TEveEventDisplay/fcl/prolog.fcl
```diff
@@ -126,6 +130,7 @@ TEveEventDisplayCosmics : {
       addTrkExtTrajs		        : false
       addMCTraj				          : false 
    }
+   gdmlname :    "Offline/TEveEventDisplay/src/mu2e.gdml"
```
kutschke, line 133: This comment started out narrow: please refactor the code so that the default gdml file name appears in exactly one place and is propagated to the others by the use of @local::

But I see that the file has many configurations repeated wholesale with only minor changes.  So I am making a broader request.  Please refactor this so that there is a base configuration plus additional configurations that override  selected definitions.  This will be much easier to maintain.

Please also consider making a sub table to hold all of the "show" definitions.  The purpose is to reduce the number of items at the top level of configuration.

sophiemiddleton:
Hi Rob, 

I mean yes in theory there's no reason not to use the current. Would advise that the users generates the GDML everytime they run the event display, or that this is not in the prolog, forcing them to make it and add it in the "driver fcl" level?

TEveEventDisplay/src/TEveEventDisplay_module.cc
```diff
@@ -56,13 +58,15 @@ namespace mu2e
       int _diagLevel;
       bool _showBuilding;
       bool _showDSOnly;
+      bool _showInsidePS;
       bool _showCRV;
       bool _showEvent;
```
kutschke, line 63: In parallel to the change I requested in fcl, please put  all of the "_show*" member data into a struct.  This will reduce maintenance at the function calls as this struct becomes more complex.  Remember to pass the struct by const reference.  You should nest the configuration code for the struct; ie move it out of the top level and into the code for the struct.

sophiemiddleton:
Oh, I see you are thinking of a musing way to do this, @rlcee let me know what you think and what suggestions you have

kutschke (changes requested):

TEveEventDisplay/src/TEveEventDisplay_module.cc
sophiemiddleton: Sure, I can do that, I have draw options in a struct now too

TEveEventDisplay/fcl/prolog.fcl
sophiemiddleton: No the user would not edit the prolog. The driver fcls are already in the CallerExampleFcls directory

sophiemiddleton:
Hi @kutschke - I have implemented many of your suggestions. I need to look into Ray's comments about adding in the path to the GDML

sophiemiddleton:
@FNALbuild run build test

TEveEventDisplay/fcl/prolog.fcl
kutschke: If that's the case, why do you need to retain commented out code?

kutschke:
@sophiemiddleton We should discuss how to close this out.  Issue #703 is outside of the scope of this PR.  

This leaves two things that we can include in this PR or manage in separate PRs.

1. Thanks for the fix in the last commit.  But the magic number 10 should be a constexpr - the best choice would be to use   conversions supplied in CLHEP::SystemOfUnits.  There are many other places that have this so doing them all can be one PR by itself.  I prefer that you do it as a separate PR, not packaged with other changes.
2.  When I asked you to collect all of the show* variables into a struct, I meant for the change to propagate back to the fcl file so that all of the show variables would be in their own subtable.  Let's get the current PR done and do this as a PR on it's own.

Is there anything else that you want in this PR?

kutschke:
@FNALbuild run build test

kutschke:
About subtable.  The configuration now is:

```
TEveEventDisplayBase : {
    module_type : TEveEventDisplay
    showCRV : false
    showBuilding : false  
    showDSOnly : true 
    showEvent : true 
   // more stuff
}
```

My suggestion is to write it as:
```
TEveEventDisplayBase : {
    module_type : TEveEventDisplay
    show : {
        CRV : false
        Building : false  
        DSOnly : true 
        Event : true 
    }
   // more stuff
}
```

Then update to the code to create the new struct.  This will simplify future  maintenance.  In my experience heirarchical configurations are much easier for to understand than are flat ones.

sophiemiddleton:
Ah ok, but then what form is "show" in the fcl parameter list in the module?

kutschke:
In the Config in the module, the parameter “show” would be a Table or an OptionalTable. And the other parameters are Atoms or OptionalAtoms inside the table.  If you look at Mu2eG4/inc/Mu2eG4Config.hh  you can see an example of a complex configuration with nested tables.    This configuration is used in both Mu2eG4_module and Mu2eG4MT_module.


From: Sophie Middleton ***@***.***>
Date: Thursday, February 10, 2022 at 9:48 PM



Ah ok, but then what form is "show" in the fcl parameter list in the ***@***.***>

sophiemiddleton:
Hi @kutschke I attempted the table. I already do that for the "filler" anyway. The issue I now have is were to initialize the values of the struct.

For example - the module takes in this "show" table and passes that to a GeomOption which is a struct which has the fcl table and is stored within the TEveMu2eMainWindow.hh.....but of course it has to then turn the fcl input to the raw bool object so it can use it in the MainWindow. Usually the transformation is done in an initializer but since the struct doesnt have a constructor in the same way a full class would I dont see how we can translate for example:

  struct GeomOptions{
    using Name = fhicl::Name;
    using Comment = fhicl::Comment;
    fhicl::Atom<bool> showCRV {Name("showCRV"),false};
    fhicl::Atom<bool> showBuilding {Name("showBuilding"),false};
    fhicl::Atom<bool> showDSOnly {Name("showDSOnly"),true};
    fhicl::Atom<bool> showInsidePS {Name("showInsidePS"),false};
    fhicl::Atom<bool> showEvent {Name("showEvent"),false};
   };

such that in the code I can access in a way like this: is GeomOptionObject.showBuilding == true....I'm sure its doable I just dont know how to pass the input FCL to an object and how that works with a struct?

sophiemiddleton:
Hi Rob, did you have any tips here - I can move to a full class if it has to be?

kutschke:
> Hi Rob, did you have any tips here - I can move to a full class if it has to be?

I will get back to you after dinner this evening.

kutschke:
> Hi Rob, did you have any tips here - I can move to a full class if it has to be?

On mu2egpvm* have a look at  /mu2e/app/users/kutschke/Development/NestedConfig

Note that you need to define 2 structs: Show is the one you use in your code and ShowConfig is the one that is used to define the table in the configuration.  Then you build the first from the second.  

If you want to write things in separate files I would add a few more tweaks so the code that only needs to see the bar Show struct does not need to be exposed to ShowConfig.

Let me know if I answered the question you needed answered.

sophiemiddleton:
Hi Rob, its not quite the same as what I had in mind but I think it works OK like that (previously GeomOptions was a struct inside the Window class, not the module, but its fine)

kutschke:
> 



> Hi Rob, its not quite the same as what I had in mind but I think it works OK like that (previously GeomOptions was a struct inside the Window class, not the module, but its fine)

I can modify this so that the struct is inside another class, not a data member of the module.  I might not get it done until tomorrow.

sophiemiddleton:
That would be good because I can also do the same for the DrawOptions too as they are also in the MainWindow class and a  struct ( you can see my partial attempt for the GeomOptions already)

kutschke:
Have a look again at  /mu2e/app/users/kutschke/Development/NestedConfig .

This time I made a mockup of a Show struct that lives inside a GeomUtiils struct.  This is exercised in NestedConfig/src/NestedConfig2_module.cc and executed with 

`mu2e -c NestedConfig/fcl/test2.fcl`

Is this closer to what you are looking for.  Note that I chose to keep the knowledge of configuration system out of GeomUtils.

sophiemiddleton:
Hi Rob, Thank you for the examples, I have it working now

sophiemiddleton:
@FNALbuild run build test

sophiemiddleton:
@FNALbuild run build test

kutschke:
Thanks Sophie.  I showed my Nested configuration example to Kyle and got some feedback from him.  I think his ideas are a good improvement.  So I am asking you to tweak your code to follow this example.  Then we are done with this PR.

The new example is in:
/mu2e/app/users/kutschke/Development/NestedConfig/NestedConfig/inc/GeomUtilsConfig.hh
/mu2e/app/users/kutschke/Development/NestedConfig/NestedConfig/src/NestedConfig4_module.c
 /mu2e/app/users/kutschke/Development/NestedConfig/NestedConfig/fcl/test4.fcl 

I moved the definition of ShowConfig into GeomUtilsConfig.hh and also put the parser function in the same file.  I renamed the parser function to "convert" - this is a special name that art knows about and if the function exists, it will automatically call it at the right time.

If you do

`  diff NestedConfig2_module.cc NestedConfig4_module.cc `

you can see a few other minor edits.  The key piece of this is using the TableAs template, instead of the Table template. Because we used TableAs, when we get to the line:

`_geomUtils(conf().show())`

art automatically knows to call the convert function.

The other thing that I like about this pattern is that there are multiple points of maintenance when you modify GeomUtils::Show - you modify the struct, the ShowConfig struct and the convert function.  This pattern does a better job of keeping these 3 things close together.  I chose not to put them all in one file because I don't want GeomUtils to see the validated fhicl infrastructure.

sophiemiddleton:
@kutschke - I have updated it to correspond to what Kyle suggested

kutschke (approved):
Thanks.  That's what I wanted.  And you can add more tables inside DisplayUntilsConfig when you discover that you need them.  As targets of opportunity, please refactor the configuration to introduce hierarchy where it makes sense to do so.

PR #701, opened 2022-02-08
IOV updates

rlcee:
- disallow spaces in IOVs
- fix print of IOV with only one subrun=0


kutschke (approved):
This looks straightforward.

PR #702, opened 2022-02-08
DbService tweaks

rlcee:
- fix prints
- add retry timeout parameter

### DIFF
--- a/DbService/src/DbService_service.cc
+++ b/DbService/src/DbService_service.cc
@@ -61,21 +61,28 @@ namespace mu2e {
     } // end loop over files
 
     int cacheLifetime = 0;
-    _config.cacheLifetime(cacheLifetime);
-    _engine.reader().setCacheLifetime(cacheLifetime);
+    if(_config.cacheLifetime(cacheLifetime)) {
+      _engine.reader().setCacheLifetime(cacheLifetime);
+    }
+
+    int retryTimeout = 0;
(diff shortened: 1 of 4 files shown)


DbService/src/DbService_service.cc
```diff
@@ -61,22 +61,29 @@ namespace mu2e {
...
     // fastStart overrides this and starts reading the DB imediately.
     bool fastStart = false;
-    _config.fastStart(fastStart);
-    if (fastStart) {
-      // this prepares IOV infrastructure for efficient queries
-      if(_verbose>2) std::cout << "DbService::beginJob initializing engine "
-			       <<std::endl;
-      _engine.beginJob();
+    if(_config.fastStart(fastStart)) {
+      if (fastStart) {
```
kutschke, line 79: I think that this test is redundant.  The call to _config.fastStart(argument) both returns true/false and writes the return value into the argument.

rlcee: I think the expression in the if tests if the optional parameter was present.
It is a little redundant in that if the parameter was not present, then the 
fastStart variable will not be set in the conf call, and it will have the value I initialized it
to, which I could use predictably.  I wrote this way to be parallel other optimal parameters that are not
bool's.  For example, a present, but empty sequence might be treated differently
than a missing optional parameter.  If is just a little more logically organized
to test if the parameter was input before using it.

PR #707, opened 2022-02-14
add time selection options

rlcee:


kutschke:
In this PR does "time selection"  refer to IOVs based on time, not subrun range?  Does it refer to the time at which data was commited to the DB?   Something else?

In any case, I think it makes sense to have a presentation about this at a Wednesday meeting.

rlcee:
> In this PR does "time selection" refer to IOVs based on time, not subrun range? Does it refer to the time at which data was commited to the DB? Something else?

It refers to the time the values were committed to the database, mostly so you can print the content that you recently committed, to check or remind yourself.  There are other features I want to add eventually.  This one is called
"ctime" for 'create time" in anticipation of also providing "dtime" for "data time" and run number
selection, and possibly other selection options based on needs that can only come from 
the experience of thinking "I wish I could X".  It is hard to know when to stop anticipating and wait for 
the feedback.  Right now I don't know if I should spend my time on a web database browser
or this tool, or both, for example.

> In any case, I think it makes sense to have a presentation about this at a Wednesday meeting.

I think there have only been maybe 4 people who have ever run dbTool, maybe not even that,
and we are far from the complex content that these options will eventually help you
pick through.  At some point, I will have to give some tutorials, but for now this is a minor
feature addition so I'm not sure it warrants wide notice.  Another point is that I'll be making
other upgrades in this development cycle, and I could summarize at the end, though even that 
will be obscure to most users.  Even during the run, I only expect a handful of people will be 
in a position to need to use dbTool - basically only the calibrators for each subdet and the calibration
manger.

kutschke:
Thanks Ray.  That works for me.  As you move forward, please add comments where appropriate to make the meaning of "time selection" clear.  I will approve and merge now.

PR #708, opened 2022-02-14
switched from comboHit::time to comboHit::correctedTime in Cpr

gianipez:


gianipez:
tagging @brownd1978 @pavel1murat 
I'm checking the final rates and post some slides on docdb soon.

Cheers,

Giani

CalPatRec/src/CalHelixFinder_module.cc
```diff
@@ -338,6 +338,9 @@ namespace mu2e {
 
 	int loc = _data.nseeds[0];
 	if (loc < _data.maxSeeds()) {
+	  if (index_best == 2){
```
brownd1978, line 341: I don't understand this logic.  What does index_best==2 correspond to physically?

CalPatRec/src/CalTimePeakFinder_module.cc
```diff
@@ -200,7 +200,7 @@ namespace mu2e {
     static const double twopi(2*pi);
 
     double              mphi(-9999.);
-    double              meanDriftTime = 1.25/0.06;// half straw tube radius / drift velocity
+    double              meanDriftTime = -3.3;//1.25/0.06;// half straw tube radius / drift velocity
```
brownd1978, line 203: What does this number correspond to?  If it depends on gas properties it should come from StrawResponse.

CalPatRec/src/CalTimePeakFinder_module.cc
```diff
@@ -243,7 +243,7 @@ namespace mu2e {
           for(int istr=0; istr<nch;++istr) {
 
             hit    = &_data.chcol->at(istr);
-            time   = hit->time();
+            time   = hit->correctedTime();//time();
```
brownd1978, line 246: Please remove replaced code

brownd1978 (commented):
Can you explain how the DtMin,Max were tuned?  I guess that will be in your talk tomorrow.

CalPatRec/src/CalTimePeakFinder_module.cc
gianipez: Waiting for @bonventre to comment on this

CalPatRec/src/CalTimePeakFinder_module.cc
gianipez: done

CalPatRec/src/CalHelixFinder_module.cc
gianipez: the case where two helices will be stored. The current diagnostic doesn't handle such a case, that's why I'm selecting the first one. This is OK as I usually run the diagnostic only with the single-particle datasets

CalPatRec/src/CalTimePeakFinder_module.cc
bonventre: The TOT calibration currently has a 3 ns bias (needs to be recalibrated based on MDC2020 samples), so this is expected right now but should be closer to zero once that's fixed. After recalibrating should check this against https://github.com/Mu2e/Offline/blob/main/TrkReco/inc/TrkTimeCalculator.hh#L37, maybe should be the same fcl parameter if there still ends up being an offset

gianipez: good point. I use now an fcl parameter that was already there.

gianipez:
Hi there,
I got rid of the hardcoded value and instead incorporated this new offset in the existing one.

Please, let me know if further actions are needed.

Cheers,

Giani

gianipez:
@FNALbuild run build test

gianipez:
@FNALbuild run build test

brownd1978 (approved):
I note that there are a number of hard-coded time offsets, these need to be addressed in a more systematic way, though that's outside the scope of this PR

bonventre (approved):

PR #710, opened 2022-02-15
proper IOV subtraction

rlcee:
Since I suspect someone will ask, here is the explanation.  Suppose a db IOV is run 100-200 and it is overridden by a text file with IOV runs 110-120.  If the requested R/S is in the database IOV, but needs to be shortened due to the presence of the text IOV, then it should be either 100-109 or 121-200.  To know which should be returned, you need to pass in the current R/S.

### DIFF
--- a/DbService/src/DbEngine.cc
+++ b/DbService/src/DbEngine.cc
@@ -380,7 +380,7 @@ mu2e::DbLiveTable mu2e::DbEngine::update(int tid, uint32_t run,
   // table interval from the database table's interval
   for(auto& oltab : _override) { // loop over override tables
     if(oltab.tid()==tid) { // if override is the right type
-      iov.subtract(oltab.iov());
+      iov.subtract(oltab.iov(),run,subrun);
     }
   }
   


kutschke:
I thought that we had a requirement of non-overlapping IOVs?

rlcee:
> I thought that we had a requirement of non-overlapping IOVs?

That rule applies to a database calibration set.  In this case, we have a text file, which can overlap, and override, an existing database IOV.

kutschke:
> > I thought that we had a requirement of non-overlapping IOVs?
> 
> That rule applies to a database calibration set. In this case, we have a text file, which can overlap, and override, an existing database IOV.

Using your example from above, does the code as is do the right thing for the use case that I run a job on a file that has events over the full range of runs 100 to 200?

rlcee:
> > > I thought that we had a requirement of non-overlapping IOVs?
> > 
> > 
> > That rule applies to a database calibration set. In this case, we have a text file, which can overlap, and override, an existing database IOV.
> 
> Using your example from above, does the code as is do the right thing for the use case that I run a job on a file that has events over the full range of runs 100 to 200?

Yes.  When you come in with run 100, it checks the text file IOV, and it is not in that IOV, so looks up db IOV and it is in that IOV, so then it subtracts the text file IOV and delivers 100-109 to the handle.  When the event hits 110, it matches the text file IOV and the handle gets 110-120 for it's IOV.  When it hits 121, it doesn't match text file IOV and does match db IOV again, and the subtract line delivers 121-200 to the handle.

kutschke:
Thanks.  Next suppose I have two different database objects in the text file, with different ranges, will that also work correctly?

rlcee:
> Thanks. Next suppose I have two different database objects in the text file, with different ranges, will that also work correctly?

Yes, whether those two objects are two different tables, or two IOVs of the same table.

kutschke:
Thanks.

kutschke (approved):

PR #714, opened 2022-02-17
Update UPS product version to use new 'd' version.

eflumerf:
Make installTableFile get more things from MUSE environment, and make it
add to the table file instead of overwriting.


eflumerf:
I applied tag `d10_10_01` to this commit in my fork.

kutschke:
Thanks Eric.  The CI has no coverage for this change so I will merge now.  So far as I know the only way to get the tag into the Offline repo in the Mu2e GitHub organization is for me to copy it by hand from your fork.  Do you know of another way?

kutschke (approved):
The CI has on coverage for this change - and he changes are straightforward.  Approving and merging.

PR #716, opened 2022-02-17
Update installTableFile.sh

eflumerf:
Put the Offline bin directory in PATH

### DIFF
--- a/ups_install/installTableFile.sh
+++ b/ups_install/installTableFile.sh
@@ -62,7 +62,8 @@ Common:
 
     exeActionRequired(GetFQDir)
     envSet (OFFLINE_LIB, \${OFFLINE_DIR}/\${OFFLINE_FQ}/lib )
-    pathAppend( CET_PLUGIN_PATH, \${OFFLINE_LIB} )
+    pathAppend( CET_PLUGIN_PATH, \${OFFLINE_LIB} )    
+    pathPrepend(PATH, "\${OFFLINE_DIR}/\${OFFLINE_FQ}/bin")
 
     if ( test \`uname\` = "Darwin" )
       pathPrepend(DYLD_LIBRARY_PATH, \${\${UPS_PROD_NAME_UC}_LIB})


ups_install/installTableFile.sh
```diff
@@ -62,7 +62,8 @@ Common:
 
     exeActionRequired(GetFQDir)
     envSet (OFFLINE_LIB, \${OFFLINE_DIR}/\${OFFLINE_FQ}/lib )
-    pathAppend( CET_PLUGIN_PATH, \${OFFLINE_LIB} )
+    pathAppend( CET_PLUGIN_PATH, \${OFFLINE_LIB} )    
+    pathPrepend(PATH, "\${OFFLINE_DIR}/\${OFFLINE_FQ}/bin")
```
kutschke, line 66: I just setup the product and tried

ls ${OFFLINE_DIR}/${OFFLINE_FQ_DIR}/bin

It looks like Offline_FQ is not defined correctly - it's missing the "trigger" piece.

kutschke (changes requested):

eflumerf:
`ls ${OFFLINE_DIR}/${OFFLINE_FQ}/bin` works for me using the product on CVMFS...note that it's `$OFFLINE_FQ` not `$OFFLINE_FQ_DIR`

kutschke:
> `ls ${OFFLINE_DIR}/${OFFLINE_FQ}/bin` works for me using the product on CVMFS...note that it's `$OFFLINE_FQ` not `$OFFLINE_FQ_DIR`

Ooops.  My apoligies

Three mistakes on my part.  First a typo: I did use OFFLINE_FQ.  Second when I did the setup I forgot trig in qualifiers.  Third I did not use setup in paranoid mode with -B and "+" on each qualifier.

ups_install/installTableFile.sh
kutschke: This is ok it was a mistake on my part - see the conversation.  Resolving this now.

kutschke (approved):

kutschke:
> I didn’t touch the script that makes those files, so I guess it’s historical and should probably be fixed at some point… Eric

I agree that this is outside the scope of this PR.  And the PR is merged.

Now that I have dredged up memories from the distant past, I think that this was done on purpose.  My instinct now is that we will never want non-trig builds in UPS.  @eflumerf and @rlcee do you agree?  If so, I will put in a PR to take out the ones we don't want.

PR #717, opened 2022-02-18
add option to match to the nearest calibration

rlcee:
From time to time a user might want to run reco on a run before the formal calibration set is prepared, or perhaps a bad run or a test run where the formal calibration will never be prepared.  Turning this option on will cause the code to accept the calibration from the nearest available earlier run.  A messagelogger warning saying the result may be unreliable and unreproducible, is printed.

### DIFF
--- a/DbService/inc/DbEngine.hh
+++ b/DbService/inc/DbEngine.hh
@@ -18,7 +18,8 @@ namespace mu2e {
   class DbEngine {
   public:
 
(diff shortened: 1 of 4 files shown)


kutschke (approved):
This is a good addition.

kutschke:
Comment to check if the CI is marked as out of date - I think it should be.

kutschke:
@FNALbuild run build test

PR #720, opened 2022-02-25
Refactor IoV code

rlcee:
Refactoring code to make it more logical, maintainable and reusable for tools and utilities.
- create DbSet to hold a flattened IoV set (only lists of iovs)
- create DbValTool to hold code to operate on the IoV heirarchy content
- simplify DbEngine by using these new classes
Log file prints will change a little, but no results will change.  Tested with validation on reco job and various test vectors.  It should not change CI or nightly result.


DbTables/inc/DbSet.hh
```diff
@@ -0,0 +1,60 @@
...
+
+    // IoV extended with its related cid
+    class EIoV {
+    public:
+      EIoV():_cid(-1) {}
+      EIoV(int cid, DbIoV const& iov):_cid(cid),_iov(iov) {}
+      int cid() const {return _cid;}
+      DbIoV const& iov() const {return _iov;}
+      void setCid(int cid) {_cid = cid;}
+      void setIov(DbIoV const& iov) {_iov = iov;}
```
kutschke, line 29: Are there good use cases for the two setter functions?  Unless there are good use, cases I prefer to leave them out in order to force single phase construction.

DbTables/src/DbSet.cc
```diff
@@ -0,0 +1,83 @@
...
+#include "cetlib_except/exception.h"
+#include <iomanip>
+#include <iostream>
+
+namespace mu2e {
+
+  //***********************************************
+
+void DbSet::add(int tid, int cid, DbIoV const& iov) { 
+  _emap[tid].emplace_back(cid, iov);
```
kutschke, line 12: Is it OK to add a duplicate?  If not, should it trigger a warning or error?

DbTables/src/DbSet.cc
```diff
@@ -0,0 +1,83 @@
...
+  _emap[tid].emplace_back(cid, iov); 
+}
+
+  //***********************************************
+
+DbSet::EIoV DbSet::find(int tid, uint32_t run, uint32_t subrun) const {
+
+  auto iter = _emap.find(tid);
+  if (iter == _emap.end()) {
+    return EIoV();
```
kutschke, line 21: What is the behaviour if there is no match, the function returns a default constructed EIoV, the caller never checks validity and uses it anyway?

If it's required that the caller check validity, and if this class is intended only to be used inside DbService, we can leave it as is. But if it could be used outside of the DbService we should think about protection for the caller not checking.

DbTables/src/DbSet.cc
```diff
@@ -0,0 +1,83 @@
...
+
+  return EIoV();
+
+}
+
+  //***********************************************
+
+void DbSet::clear() {
+  for (auto& p : _emap) p.second.clear();
+}
```
kutschke, line 57: I am curious why you don't need to clear _emap too?

DbService/src/DbValTool.cc
```diff
@@ -0,0 +1,242 @@
...
+
+int DbValTool::findPid(std::string purpose) const {
+
+  if (purpose.empty()) return -1;
+
+  for (auto const& pp : _valcache.valPurposes().rows()) {
+    if (pp.name() == purpose) return pp.pid();
+  }
+  return -1;
+}
```
kutschke, line 17: Same question as for DbSet.cc  It is safe to count on the caller to test return values for validity?  Is there well defined and safe behaviour if they do not?

Again, if this is for internal use only we can decide to keep it as is.

kutschke (commented):
This is a nice refactoring.  I have a handful of questions.

DbTables/inc/DbSet.hh
rlcee: one deprecated use case, I'll remove them

DbTables/src/DbSet.cc
rlcee: Since the lists can get long and the input should have already been checked for duplicates, I don't think it makes sense to check here.  If it is true duplicate, then it doesn't affect the existing main algorithm.  The same cid and different IoV is allowed.  Trying to imagine a use case where it is more wild-west and needs to be checked.  I think I'd add a separate duplicate check.  But I don't imagine any of this will be used by any sort of non-expert.

DbTables/src/DbSet.cc
rlcee: It is required that the code does not throw, but the code can be arranged like "bool find(...EIoV& eiov)" if you prefer.  The default EIoV has cid=-1 and IoV is null (matches no range) so should not give false results.  The existing caller does check. This pattern just reflected what I had in Engine previously..  Let me know if you prefer the other pattern.

DbTables/src/DbSet.cc
rlcee: I will add

DbService/src/DbValTool.cc
rlcee: The code should not throw, but I can change to "bool findByName(int& tid)" if you prefer.  I don't imagine non-expert users, but you never know for sure.

DbTables/src/DbSet.cc
kutschke: Thanks for the explanation.

DbTables/src/DbSet.cc
kutschke: Thanks.  We can leave it as is.  The behaviour sounds like it's well defined.  So long as this is for internal use and the callers check, then its OK.

DbService/src/DbValTool.cc
kutschke: Please do this.

kutschke:
@FNALbuild run build test

kutschke (approved):

PR #721, opened 2022-02-25
install database table cache purge

rlcee:


rlcee:
I forgot to consider thread safety, I'll have to make one more commit..

rlcee:
> I forgot to consider thread safety, I'll have to make one more commit..

I think it is OK now, please proceed

DbTables/src/DbCache.cc
```diff
@@ -1,8 +1,18 @@
...
 
 void mu2e::DbCache::add(int cid, mu2e::DbTable::cptr_t const& ptr) { 
+
   _tables[cid] = ptr;
+
+  _cidFifo.push(cid);
+  _nAdded++;
+  _size += ptr->size();
+  if(_size>_hwmSize) _hwmSize = _size;
+  if(_nAdded%20==0) purge();
```
kutschke, line 14: Do we want to make "20" run-time configurable?

DbTables/src/DbCache.cc
```diff
@@ -14,18 +24,23 @@ mu2e::DbTable::cptr_t mu2e::DbCache::get(int cid) {
...
-}
+  if(_size < _limitSize) return;
 
-size_t mu2e::DbCache::size() {
-  size_t s=0;
-  for(auto const& t : _tables) s += t.second->size();
-  return s;
+  _nPurges++;
+  while(_size>0.9*_limitSize) {
+    int cid = _cidFifo.front();
```
kutschke, line 33: I would like to understand the following scenario.  There are a lot of things about this system that I have not yet understood in detail so I guess it's time to learn.   Suppose that my module always gets one table and conditionally gets a second table.   I am following the rules: I hold the DbHandles as data members and call get on the first handle in every event and I call get on the second handle on some but not all events.  Suppose that a lot of other modules also ask for tables, directly or indirectly, so that the table I always get is the oldest table in the delete queue.  Finally suppose that I have never called get on the second table on the first few events; when I finally do call get on it, it's the 20th table added to cache and triggers a purge; if the cache is more than 90% full then my first table will be deleted.  What happens to the  T const& that I have to the first table?  Have I just invalidated it?

DbTables/src/DbCache.cc
```diff
@@ -14,18 +24,23 @@ mu2e::DbTable::cptr_t mu2e::DbCache::get(int cid) {
...
-  return 0;
-}
+  if(_size < _limitSize) return;
 
-size_t mu2e::DbCache::size() {
-  size_t s=0;
-  for(auto const& t : _tables) s += t.second->size();
-  return s;
+  _nPurges++;
+  while(_size>0.9*_limitSize) {
```
kutschke, line 32: Do we want to make "0.9" run time configurable.

kutschke (commented):
This looks like a good step forward - I have one big question and two minor suggestions.

DbTables/src/DbCache.cc
rlcee: I can make the 20 and the 0.9 configurable.

DbTables/src/DbCache.cc
rlcee: A few more rules which may guide your thinking to new questions.  Once a table is loaded into memory, it will always occupy that memory (it doesn't move).  The cache passes out shared pointers to the table data, so a handle points to a table data with a reference count, so it will always be valid, even after the cache deletes its pointer.   Tables being passed out to handles and the purge can't interfere.  Tables can always be reloaded.  Even if there were 10 large tables being asked for by handles and the cache could only hold one table, it will still work (probably inefficiently).

DbTables/src/DbCache.cc
kutschke: Thanks.  I will watch for the commit.

DbTables/src/DbCache.cc
kutschke: Thanks.  I think this means that, in the scenario I described, it's possible that no memory will actually be recovered by the purge, which is a good thing since nothing is really purgeable at that time.   Does this invalidate your bookkeeping of the HWM since you have no idea when the memory will actually be released?

rlcee: Yes, there may always be some memory held exclusively by a handle and it will not be counted in the DbService totals or subject to purging.   If you dig into it, the memory accounting is approximate from the start.  I expect the largest table will be order 1MB, so a few even a few copies of that unaccounted for should be in the noise.

kutschke:
@FNALbuild run build test

kutschke (approved):
Thanks for the changes and the explanations.  If we discover that we need better accounting of the memory use we can do it at that time.

PR #724, opened 2022-02-28
write multiple IoV on a TABLE line

rlcee:
One calibration entry might be used in multiple IoV.  This is especially true in new data where we expect to advance the IoV one or a few runs at a time.   If we need to represent this calibration set in a text file, then we want to write the calibration content once and then write all the IoV where this content applies, so as to not repeat the content.


kutschke:
@FNALbuild run build test

kutschke (approved):
This looks straightforward.

PR #725, opened 2022-03-01
make constants static

rlcee:
a bit more appropriate design

### DIFF
--- a/DbTables/inc/DbIoV.hh
+++ b/DbTables/inc/DbIoV.hh
@@ -81,8 +81,8 @@ namespace mu2e {
     uint32_t endRun() const {return _endRun;}
     uint32_t endSubrun() const {return _endSubrun;}
 
-    uint32_t maxRun() const { return maxr; }
-    uint32_t maxSubrun() const { return maxsr; }
+    static uint32_t maxRun() { return maxr; }
+    static uint32_t maxSubrun() { return maxsr; }
 
     // in format: run subrun run subrun
     std::string simpleString() const;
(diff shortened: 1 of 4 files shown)


DbTables/inc/DbIoV.hh
```diff
@@ -81,8 +81,8 @@ namespace mu2e {
     uint32_t endRun() const {return _endRun;}
     uint32_t endSubrun() const {return _endSubrun;}
 
-    uint32_t maxRun() const { return maxr; }
-    uint32_t maxSubrun() const { return maxsr; }
+    static uint32_t maxRun() { return maxr; }
+    static uint32_t maxSubrun() { return maxsr; }
```
kutschke, line 85: I suggest making the definitions of max4 and maxsr, static constexpr.   The benefit is minimizing cache misses. You can also make them public and get rid of the accessor functions.  Or you can keep them private and keep the accessor - it's a tiny bit more work for the compiler but I think it will produce the same compiled code.

kutschke (commented):
One Mom and apple pie suggestion.

DbTables/inc/DbIoV.hh
rlcee: I had actually written as public constants, but then switched to accessors to make it feel like numeric_limits::max() and give it some encapsulation.  I did miss one trick, to call the methods constexpr.  The POD variable const or constexpr is the same, according to the web but perhaps constexpr shows intent.  Is public member contsexpr variables your first choice?  I don't feel strongly.

kutschke: I had not thought about making the member functions constexpr - they can just return the value written as a literal and you can get rid of the two static data members.  I like the idea of having the look and fell of numeric_limits::max().
A great idea.

kutschke: I see that the commit doing this is in.  Thanks.

kutschke (approved):
Thanks for the improvements.

PR #729, opened 2022-03-07
Updated the clustering process of the Crv coincidence finder

ehrlich-uva:
- added different cuts for the initial clustering (to keep the number of hits combinations down that need to be checked for coincidences) and the final clustering (based only on coincidence hits).
- used a better solution to keep hits from opposite counter ends in one coincidence cluster.
- added a protection to prevent hits from coincidence groups to be broken apart at the final clustering.

### DIFF
--- a/CRVResponse/fcl/prolog_v10.fcl
+++ b/CRVResponse/fcl/prolog_v10.fcl
@@ -158,8 +158,6 @@ BEGIN_PROLOG
       crvRecoPulsesModuleLabel         : "CrvRecoPulses"
       #cluster settings
       clusterMaxDistance               : 250 //mm
-      clusterMaxTimeDifference         : 20  //20ns (same as max time difference in coincidence setting) //when overlap option is not used
-      clusterMinOverlapTime            : 30  //30ns (=at least 3 ADC samples, same as coincidence setting) //when overlap option is used
       #sector-specific coincidence settings
       sectorConfig :
       [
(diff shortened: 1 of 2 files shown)


kutschke:
Ralf, please describe briefly the tests you have done.  Does this change the execution time significantly?

I will not merge this today because we would like another PR to establish a new validation baseline tonight.

ehrlich-uva:
The execution time (when running the slow reco validation script) increased slightly from an average of 0.16 s / event to 0.19 s / event when running at mu2egpvm0x. As a comparison, KFFDeM:KalFinalFit takes 0.37 s / event at mu2egpvm0x.

I tested the following things with the cosmicSimReco validation script:
- It still finds the same number of events with coincidences: 5323. That's the most important parameter.
- The peak of the PE distribution stayed at around 350. 
  The low PE portion looks again more like before the changes of 2/21. 
  Many clusters with a low number of PEs disappeared, because the coincidences are not broken apart anymore, 
  and the cluster finder tries to keep hits from opposite counter ends together.
- The number of crv layers used for each cluster was 3 or 4 as expected.

I tested the following things with the cosmicSimReco validation script:
- The distribution of the reco bar IDs showed no significant changes.
- The number of crv layers used for each cluster was 3 or 4 as expected.

For both scripts, I scanned the cluster times (start and end) for the first ~100 events to see whether they look normal.

kutschke:
@FNALbuild run build test

kutschke (approved):

PR #731, opened 2022-03-10
Fix santize errors: binding a nullptr to a reference. 

kutschke:
The sanitizer found two places in which a pointer with a value of nullptr was bound to a reference.  In both cases the caller held an object by pointer and dereferenced the pointer to pass the object by reference into a function;  the function immediately took the address of the argument and held it by pointer internally.

In some cases the original pointer is allowed to have a value of nullptr and the code handles this correctly.  However the sanitizer objected to binding a nullptr to a reference at the function call.  So I changed the argument type to pointer and made all of the needed follow-on changes in the callers.  This includes locations that had not yet had coverage by the sanitizer.


brownd1978 (approved):
Looks good.

PR #732, opened 2022-03-15
Improve calo digi timing sim

bechenard:
I improve the internal model of the calo digi timing ad the boundary effects. I have tested with a small CEsimreco.fcl production, and it works as intended. The digitization time window are my guess. Richie can review...


CaloMC/fcl/common.fcl
```diff
@@ -1,6 +1,6 @@
 BEGIN_PROLOG
-  HitMakerDigitizationStart    : 450.
-  HitMakerDigitizationEnd      : 2005.0
+  HitMakerDigitizationStart    : 250.
```
brownd1978, line 2: This seems to include the event window marker offset, which should come from DaqConditions I believe.

brownd1978 (commented):
Having a common digitization start time setting is a good idea.  It needs to coordinate with the DAQ timing, so the parameters always refer to proton time.

CaloMC/fcl/common.fcl
bonventre: The way we've been encoding it for the other systems is that the fcl parameterization for the digitization is given in terms of nominal time from protons, and everywhere it is actually used it subtracts out timeFromProtonsToDRMarker from DAQConditions. We can eventually move all of this to DAQConditions and provide a method for getting these times, but for this PR I think it should be done here.

bechenard:
It doesn't really make sense - the FPGA will only know that it needs to 
wait X ns after receiving the start signal from the CFO, it doesn't know 
about the  timeFromProtonsToDRMarker. I believe this should be changed, 
but I'll make this PR compatible with the rest of the code...

- Bertrand

bonventre:
Yeah... the thought I think was that in some sense the literal FPGA setting is some other variable in a run config database in some other units probably, and then somewhere it has to be propagated to proditions, and a conversion is happening anyway. Regardless, I think if we get all the subsystems working, then we can consolidate these to DAQConditions and then any change of that sort will be simpler

bechenard:
Fixed

CaloMC/fcl/common.fcl
```diff
@@ -1,6 +1,6 @@
 BEGIN_PROLOG
   HitMakerDigitizationStart    : 450.
-  HitMakerDigitizationEnd      : 2005.0
+  HitMakerDigitizationEnd      : 1795
```
brownd1978, line 3: Is it understood why this number extends beyond the microbunch?  Is it to include tails of showers that start near the end of the microbunch period?  Note that, in the MDC simulations, we only accumulate pileup 10ns past the end of the microbunch, so simulations of the times past that are not necessarily physically accurate.

bechenard: It needs to go over the microbunch length if you want to properly digitize a signal at the boundary of the microbunch. The exact number is a guess on my side, I'll fix it once the FPGA code is finalized. At this point, I'll take anything past the microbunch

brownd1978: OK.  We will need to understand what this implies for simulated calo signals that extend past the microbunch end, my guess is the pileup for those will be underestimated, perhaps by a lot.  That issue is outside the scope of this PR.

brownd1978 (approved):
These changes are now technically correct.  The question of the accuracy of the simulation for signals past or even near the microbunch end is still open, that's outside the scope of this PR.

bonventre (approved):

brownd1978:
@FNALbuild run build test

PR #733, opened 2022-03-16
Strawman header classes for discussion.

kutschke:
Strawman versions of header classes for events, subruns and runs.  The plan is for these to be added to events by the TDAQ system.  This PR is discussed in the talk:

      https://mu2e-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=41418


rlcee:
A few thoughts, just for consideration.

DataProducts/inc/EventWindowMarker.hh
has spilltype, which now overlaps this product - should there be a transition?  Also it has event length, which I wonder whether that should be in here?

Event number is event EWT-beginRun EWT, true?  Redundant?  Checks?  Slightly wrong? Confusing?

If you have processing node, do you want datalogger node?

Is the STM and ExtMon, 100% sure, going to be coordinated by EWT and never time? (If time might be used, it is a reason to require more accuracy).

HCasler:
Test comment: @FNALbuild you should not understand this

gianipez:
tagging @eflumerf

kutschke:
@FNALbuild run build test

kutschke:
Following an out-of-thread discussion with Eric and Giani, we have decided that the right place for these classes is in mu2e-artdaq-core.    They propose to create the header objects on the artdaq side and to pass them to the art side using the input module that copies the artdaq::Fragments into the art processes.   This answers the question of how these objects will be populated.   

I will close this PR and contribute the code to the mu2e_artdaq-core repo.

PR #738, opened 2022-03-22
Added extra buffer in caloShowerRO

bechenard:


HCasler:
FYI Jenkins is saying all the tests succeeded, but is reporting errors updating the test report on the PR and adding comments. It had no problem with PR#739, so not sure what the problem is yet.

kutschke:
> FYI Jenkins is saying all the tests succeeded, but is reporting errors updating the test report on the PR and adding comments. It had no problem with PR#739, so not sure what the problem is yet.

Thanks for catching this and point it out.

This morning GitHub was flakey.  Around 11 AM I  tried to make a release and it failed with an error message, something like "Something's wrong.  Try again."  A few minutes ago I was able to to make the tag.  Very likely the same glitch caused this problem.  

If I remember correctly, it won't let me re-run the CI until either make a commit to this PR or the head changes.  Is that right?

HCasler:
Based on the logs, it looks like we got a couple of 500 responses from GitHub when trying to post the updates, so that makes sense.
`If I remember correctly, it won't let me re-run the CI until either make a commit to this PR or the head changes. Is that right?`
Without looking at the code, I seem to recall us changing that because of issues like this. Might as well try running it again.

kutschke:
Let's see if this actually runs the tests or if it says that nothing has changed so there is nothing new to test:

@FNALbulid run build test

kutschke:
Done

From: Helenka Casler ***@***.***>
Date: Tuesday, March 22, 2022 at 1:35 PM
To: Mu2e/Offline ***@***.***>
Cc: Robert K Kutschke ***@***.***>, Team mention ***@***.***>
Subject: Re: [Mu2e/Offline] Added extra buffer in caloShowerRO (PR #738)

Based on the logs, it looks like we got a couple of 500 responses from GitHub when trying to post the updates, so that makes sense.
If I remember correctly, it won't let me re-run the CI until either make a commit to this PR or the head changes. Is that right?
Without looking at the code, I seem to recall us changing that because of issues like this. Might as well try running it again.

kutschke:
Ooops. Thanks for catching my typo.

kutschke:
This looks simple enough that I don't think that we need another reviewer.   I expect to see differences in ceSimReco but not reco - is that correct?

bechenard:
I guess that’s right. If you see differences, this would be right at the beginning of the call digitization time

Cheers
Bertrand

kutschke:
@FNALbuild run build test

kutschke (approved):
This PR has not yet been properly reviewed but the code that it adds is disabled by default.  Bertrand is unavailable to present this work for a few weeks.  I will merge this now to minimize downstream maintenance of the PR with the understanding that there will be a proper review at the time of the PR that enables this code.

PR #740, opened 2022-03-23
Enabling optional use of density effect in ionization loss calculations

resnegfk:
Added code, to be used with Geant4 10.7.p03 and above, to enable a future use of more accurate energy loss calculations,
esp. in metals, which may affect, e.g., muons stops. TBD.


kutschke:
@FNALbuild run build test

kutschke (approved):
Thanks for getting this in ahead of time.

PR #743, opened 2022-03-24
remove reference to null pointer

rlcee:
This should remove the last "reference to null pointer" warning in Offline sanitize.  The functionality is identical, just without going through the intermediate step of making a null reference.  Verified and validated with 500 events of ceSimReco (where the warning originally appeared).


kutschke (approved):
Thanks for tracking this down.

kutschke (approved):
I reviewed this myself.

PR #746, opened 2022-03-26
Fix some remaining Muse one-path issues.

kutschke:
There were a few fcl files in the Print area that were still configured for Muse 2-path.  I changed these to 1-path.

There were a few places in comments that still used 2-path syntax.  I changed these to 1-path too.


kutschke:
@FNALbuild run build test

rlcee (approved):

Issue #745, opened 2022-03-26
Repeatability of PrimaryProtonGun_module.cc


kutschke:
EventGenerator/src/PrimaryProtonGun_module.cc is an art::ReplicatedProducer

When run in MT mode, this does not produce repeatable results because of how environmental effects interact with the scheduler; that is, Event N in one job might land on instance 1 of the module and in a different job it might land on instance 2.  These will have different random number state and produce different events.

For physics purposes this is OK.  But we need repeatability for debugging purposes.

See also #744 - these may need a common solution.

rlcee:
Wasn't this module made a ReplicatedProducer for an early version of the geant MT mode, and now it would fine to downgrade to legacy module, and only run one copy?  Isn't this sort of state problem the main reason for not making a module MT?

kutschke:
Reverting to legacy feels like a step backwards but I think you are right that it will solve both this issue and and #744.  I will test; I don't think that it will have serious execution time consequences but I will measure to be sure.  If it's good enough I will revert it.

rlcee:
@kutschke I propose closing this and #744 and let this module be addressed when/if we review modules generally for promoting to thread safe.  This one only sticks out since it was used in a the early, now obsolete, form of G4 threading.

rlcee:
Since there is no need for this module to run multi-threaded, we will leave it as a legacy module and this issue is solved.

Issue #744, opened 2022-03-26
PrimaryProtonGun_module.cc uses duplicate seeds in MT mode


kutschke:
EventGenerator/src/PrimaryProtonGun_module.cc is an art::ReplicatedProducer.  The issue is that all instances are seeded identically.  This needs to be fixed.

I believe that the MDC2020 POT samples are OK even though they were run with 2 threads.  The reason is that the random engine used in each G4 thread was uniquely re-seeded at the start of each event.  So there is enough downstream randomness.

This is the only replicated producer in our code base.  We should develop a solution that transfers easily to future replicated producers.

See also issue #745 - these may need a common solution.

rlcee:
Since there is no need for this module to run multi-threaded, we will leave it as a legacy module and this issue is solved.

Issue #482, opened 2021-06-09
Investigate event reproducibility in jobs using Mu2eG4MT_module


resnegfk:

kutschke:
Krzysztof reported that the reproducer is:

> mu2e -n 250000 -c Production/JobConfig/validation/stoppedMuonsSingleStageMT.fcl >& stoppedMuonsSingleStageMT.log
> This takes about 40min.

This runs 5 threads and 5 schedules.  Two consecutive runs of this job produce a different number of output events.  I added some printout to see which events were being written out and discovered that there is very little overlap in the events that are written out, maybe 10%.

The issue is not in G4MT but in the interaction of the event generator EventGenerator/src/PrimaryProtonGun_module.cc with the MT environment.  When you run the same job 2 times,  the match of event number to the instance of PrimaryProtonGun is random - it depends on the thread scheduler and environmental effects.  So in two separate jobs the match of generated event to G4 worker threads is broken.

This is not a problem for physics purposes but it is a problem because we do not have the repeatability needed for debugging purposes.  Issue #745 has been created to address this.

This investigation also uncovered an unrelated problem in EventGenerator/src/PrimaryProtonGun_module.cc:  all instances get the same seed, which is reported in issue  #744 .

kutschke:
The immediate problems have been fixed with PR #747.  The new issues #744 and #745 have been created to create a better solution.

PR #747, opened 2022-03-27
Revert from a ReplicatedProducer to a legacy EDProducer

kutschke:
This is a, hopefully, temporary solution to Offline issues #744 and #745  that were uncovered when working on issue #482.

Using Production/JobConfig/validation/stoppedMuonsSingleStageMT.fcl  I ran 4 test jobs, 2 with the original code and with this PR.  All were 250000 events.  The two with this PR produced identical results while the two originals differed as described in issue #482 .

This PR could slow down the real time performance of the two jobs because of less parallelism.  So I did a measurement and the change is not large enough to be important.    The execution times for the 4 jobs were

Before this PR:
TimeReport CPU = 9158.873317 Real = 1878.142641
TimeReport CPU = 9288.815459 Real = 1904.679477

 Mean and standard deviation:  1891 +/- 23

After this PR:
TimeReport CPU = 9301.399397 Real = 1906.883257
TimeReport CPU = 9358.623558 Real = 1916.216917

 Mean and standard deviation:   1911 +/- 7

The difference between the two means is about 1%, with a significance of about 1 sigma.  I believe this is acceptable.

### DIFF
--- a/EventGenerator/src/PrimaryProtonGun_module.cc
+++ b/EventGenerator/src/PrimaryProtonGun_module.cc
@@ -1,6 +1,8 @@
 
 /*
-  This is a Replicated Module.
+  This was a Replicated Module and was reverted to Legacy on Mar 27, 2002
+  until GitHub issues #744 and #745 can be resolved.
+
   A plug-in for running PrimaryProtonGun-based event generator for running in MT art.
   It produces a GenParticleCollection of primary protons using the PrimaryProtonGun.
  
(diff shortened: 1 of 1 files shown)


resnegfk (approved):
In my opinion, the very low additional CPU cost is fully acceptable to regain the reproducibility.

EventGenerator/src/PrimaryProtonGun_module.cc
```diff
@@ -1,6 +1,8 @@
 
 /*
-  This is a Replicated Module.
+  This was a Replicated Module and was reverted to Legacy on Mar 27, 2002
```
goodenou, line 3: Typo here. You probably want this to say Mar 27, 2022.

goodenou (approved):
I recommend a fix to the date in line 3, but otherwise this looks good.

EventGenerator/src/PrimaryProtonGun_module.cc
kutschke: Thanks.  It's fixed.

brownd1978 (approved):

PR #748, opened 2022-03-28
Fixed bug in TEve 

sophiemiddleton:
* A leftover from when we couldnt use the TEve projections


kutschke (approved):
An evergreen comment: please do not write code with magic numbers  - give every numeric literal a constexpr name and use it.   Please do this as targets of opportunity as you move forward.

This PR does not have test coverage beyond the CI tests, which have passed.  So I will approve this and merge.

PR #753, opened 2022-04-05
enabling forcing specific muon decay proper time

resnegfk:
The PR enables forcing specific muon decay proper time 
(if the decay process is the one which ends the track, and not, e.g., 
a capture at rest often seen in case of low energy muons).

The setting is inactive by default, so no changes in behavior are expected when merging this PR.

There is a negligible CPU penalty to check if a double is greater or equal zero in the tracking action.

The PR is a proof of concept, which can form a basis for a more elaborate solution.

### DIFF
--- a/Mu2eG4/src/Mu2eG4TrackingAction.cc
+++ b/Mu2eG4/src/Mu2eG4TrackingAction.cc
@@ -191,7 +191,7 @@ namespace mu2e {
 
     // Need to cast away const-ness to do this.
     const_cast<G4Track*>(trk)->SetUserInformation(ti);
-    if (_muonPreAssignedDecayProperTime>0) {
+    if (_muonPreAssignedDecayProperTime>=0.0) {
       const G4DynamicParticle* dynPart = trk->GetDynamicParticle();
       if (dynPart->GetPDGcode() == 13) {
         // if track is a muon and preassigned proper time is set, assing it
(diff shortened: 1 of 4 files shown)


kutschke:
@FNALbuild run build test

Mu2eG4/src/Mu2eG4TrackingAction.cc
```diff
@@ -184,6 +191,19 @@ namespace mu2e {
 
     // Need to cast away const-ness to do this.
     const_cast<G4Track*>(trk)->SetUserInformation(ti);
+    if (_muonPreAssignedDecayProperTime>=0.0) {
+      const G4DynamicParticle* dynPart = trk->GetDynamicParticle();
+      if (dynPart->GetPDGcode() == 13) {
```
kutschke, line 196: Please use the PDGCode enum rather than the literal 13.

kutschke (changes requested):

resnegfk:
@FNALbuild run build test

Mu2eG4/src/Mu2eG4TrackingAction.cc
resnegfk: Done (Also in two other cases in this file); Thanks for reminding me about it

kutschke: Thanks.

kutschke (approved):
Thanks for rerunning the tests.

PR #754, opened 2022-04-06
Update the main program per the upstream art changes.

gaponenko:
See https://github.com/art-framework-suite/art/pull/122

This does not affect any successful jobs, but removes an extraneous "successful completion" printout for some jobs that crash.


kutschke (approved):
Thanks for updating this.

PR #756, opened 2022-04-06
Fix whitespace in a subset of directories

rlcee:
In subdirectories I'm largely responsible for,
- removed tabs
- remove trailing whitespace
- install .clang-format files in subdirectories
- reformat by format files 

While the diff's are extensive, the overall appearance of the code changes very little

### DIFF
--- a/DbService/.clang-format
+++ b/DbService/.clang-format
@@ -1,0 +1,18 @@
+---
+Language:        Cpp
+BasedOnStyle:  google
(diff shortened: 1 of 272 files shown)


kutschke (approved):
I don't know a good way to meaningfully review this PR.  I will wait for a day when we can merge it on it's own so that the nightly validation can confirm no changes.  This cannot be today since there is already a PR merged that will produce changes.

Issue #755, opened 2022-04-06
Follow up to PR#706


kutschke:
The geometry changes to PR #706 have been vetted but the review has outstanding requests for changes to EventGenerator/src/CaloCalibGun_module.cc .  The plan is to merge the PR with the understanding that the outstanding requests must be resolved in a separate PR before that code is used.

sophiemiddleton:
I am working on this today. I have done most of the last few "coding" points you made on the previous PR. For the G4 naming, we only have one calibration system. So is there somewhere there could be ambiguity, I dont see it at a glance, did you have something in mind?

sophiemiddleton:
I put in a PR: https://github.com/Mu2e/Offline/pull/758 which should answer most of the issues you had

rlcee:
@kutschke @sophiemiddleton Please evaluate and describe what still needs to be done, or close.

sophiemiddleton:
I think the follow up request (758) fixed it, right?

sophiemiddleton:
Yes

PR #758, opened 2022-04-07
Calo Photon Generator Edits

sophiemiddleton:
Hope to address issues in: https://github.com/Mu2e/Offline/issues/755

### DIFF
--- a/EventGenerator/src/CaloCalibGun_module.cc
+++ b/EventGenerator/src/CaloCalibGun_module.cc
@@ -155,7 +155,8 @@ namespace mu2e {
     std::unique_ptr<GenParticleCollection> output(new GenParticleCollection);
     PrimaryParticle primaryParticles;
 
-    for (unsigned int ig = 0; ig < _mean; ++ig) 
+    double nPhotons = randPoissonQ.fire();
+    for (unsigned int ig = 0; ig < _nPhotons; ++ig) 
     {
       double xpipe, ypipe, zpipe;
       //Pick position - find either 0,1 - these are indices of the sign list (so 0=-1, 1=+1)
--- a/EventGenerator/src/CaloCalibGun_module.cc
+++ b/EventGenerator/src/CaloCalibGun_module.cc
@@ -98,6 +98,7 @@ namespace mu2e {
     std::vector<double>    _pipeTorRadius;
     std::vector<double>    _randomRad;
     CLHEP::Hep3Vector      _zPipeCenter;
+    unsigned int _nPipes;
 
   };
 
@@ -141,6 +142,7 @@ namespace mu2e {
       xsmall = _cal->caloInfo().getDouble("radSmTor");
       xdistance = _cal->caloInfo().getDouble("xdistance");
       rInnerManifold = _cal->caloInfo().getDouble("rInnerManifold");
+      _nPipes = _cal->caloInfo().getInt("nPipes");
 
   }
   //================================================================
@@ -161,9 +163,9 @@ namespace mu2e {
     // find the z position based on above:
     zpipe = pipeR*sin(theta);
 
-    double rtest = _randFlat.fire();
     // select an index from list of pipes:
-    int idx = int( std::lower_bound(_randomRad.begin(), _randomRad.end(), rtest) - _randomRad.begin());
+    int idx = rand() % ( _nPipes ) + 1 ;
+
     // select the LgTor from list extracted from geom service above:
     double radLgTor = _pipeTorRadius[idx];
 
--- a/EventGenerator/src/CaloCalibGun_module.cc
+++ b/EventGenerator/src/CaloCalibGun_module.cc
@@ -164,7 +164,7 @@ namespace mu2e {
     zpipe = pipeR*sin(theta);
 
     // select an index from list of pipes:
-    unsigned int idx = rand() % ( _nPipes ) + 1 ;
+    unsigned int idx = int(5*_randFlat.fire())+1;
 
     // select the LgTor from list extracted from geom service above:
     double radLgTor = _pipeTorRadius[idx];
--- a/EventGenerator/src/CaloCalibGun_module.cc
+++ b/EventGenerator/src/CaloCalibGun_module.cc
@@ -1,6 +1,11 @@
-// Simulate the photons coming from the pipe calibration source
-// based on CaloCalibGun orginally written by Bertrand Echenard (2014)
-// Current module author: Sophie Middleton (2022)
+/* Simulate the photons coming from the pipe calibration source
+ based on CaloCalibGun orginally written by Bertrand Echenard (2014)
+ Current module author: Sophie Middleton (2022)
+ 
+ Assumptions: 
+ * We treat all 5 pipes as equal volume when we pick a pipe. 
+ * We treat each torus is if it were a right circular cylinder with a length equal to the arc length of the centerline of the torus
+*/
 
 // Framework includes
 #include "messagefacility/MessageLogger/MessageLogger.h"
@@ -96,7 +101,6 @@ namespace mu2e {
 
     double                 _pipeRadius;
     std::vector<double>    _pipeTorRadius;
-    std::vector<double>    _randomRad;
     CLHEP::Hep3Vector      _zPipeCenter;
     unsigned int _nPipes;
 
@@ -126,13 +130,8 @@ namespace mu2e {
 
       _pipeRadius      = _cal->caloInfo().getDouble("pipeRadius");
       _pipeTorRadius   = _cal->caloInfo().getVDouble("pipeTorRadius");
-      _randomRad       = _cal->caloInfo().getVDouble("pipeTorRadius");
       _zPipeCenter     = _cal->disk(_nDisk).geomInfo().origin()-CLHEP::Hep3Vector(0,0,_cal->disk(_nDisk).geomInfo().size().z()/2.0-_pipeRadius);
       _nPipes = _cal->caloInfo().getInt("nPipes");
-      // we normalize to the volume of the pipe (proportional to 2*pi*R if they have all the same radius) to draw a random number from which to generate the photons
-      double sumR(0);
-      std::for_each(_randomRad.begin(), _randomRad.end(), [&](double& d) {sumR+=d; d = sumR; });
-      std::for_each(_randomRad.begin(), _randomRad.end(), [&](double& d) {d /= sumR;});
    
       //Define the parameters of the pipes:
       phi_lbd = _cal->caloInfo().getVDouble("largeTorPhi");
--- a/EventGenerator/src/CaloCalibGun_module.cc
+++ b/EventGenerator/src/CaloCalibGun_module.cc
@@ -46,7 +46,7 @@ namespace mu2e {
     struct Config {
       using Name=fhicl::Name;
       using Comment=fhicl::Comment;
-      fhicl::Atom<unsigned> mean{Name("mean"),1};
+      //fhicl::Atom<unsigned> mean{Name("mean"),1};
       fhicl::Atom<double> energy{Name("PhotonEnergy"),6.13};//MeV
       fhicl::Atom<double> cosmin{Name("cosmin"),-1.};
       fhicl::Atom<double> cosmax{Name("cosmax"),1.};
@@ -65,7 +65,7 @@ namespace mu2e {
 
   private:
 
-    unsigned _mean;
+    ///unsigned _mean;
     double _energy;
     double _cosmin;
     double _cosmax;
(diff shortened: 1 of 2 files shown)


kutschke:
Thanks for making the requested changes - it's a big improvement.

When you are done with the work below, please remake the generator plot in Fig 2 of your writeup.  The jobs should be fast since you don't need to rerun G4.   No need to update the note - this is just to verify that we did not break something in the course of this work.

There are 4 outstanding issues and I will add @frankcporter and @bechenard as reviewers.   Please approve the PR when the issue I mention you in is resolved to your satisfaction.

1. This one is for Frank.  In the email thread that said that you were comfortable with the approximations.  I want to double check one of the approximations.  The choice of pipe 1 to 5 is made using only the radius of the large torus section not on the full length of each pipe; the generation point is then uniformly distributed throughout the pipe.  This under samples the inner pipes. These end effects are fairly large for the innermost pipe.  You have the information to do this correctly and you use it when choosing which of the 3 pipe sections to use.  Frank:  is this OK as is or should it be fixed?
2.  In the thread on PR #706, Dave said that you should not create the PrimaryParticle data product in this module.  You should created it later in your workflow using an appropriately configured instance of FindMCPrimary.  See https://github.com/Mu2e/Offline/pull/706#issuecomment-10904901423.   So please remove all references to that product.
3. Something has been bothering me since I started looking at this. Bertrand and I have talked about it.  The Jacobean to transform from (x,y,z) to your 3 spatial coordinates describing a torus is not 1.  In general you need to include the Jacobean somewhere in this generator.  However I think I know the answer.  It turns out that the expression for the volume of a torus is;  (2 pi R) * (pi r^2), where r is the pipe radius and R is the radius of the big circle through the center of the pipe.  So it's the same as the volume of a right circular cylinder with the same radius as the pipe and a length equal to the circumference of the centerline of the torus.  From this I conclude that there must be an exact cancelation of the effect of the Jacobean when you integrate a constant function over the volume of a torus.  The Jacobean would be important if the efficiency were a function of location within the torus or if you sampled over a subset of the volume.   I am confident enough in this that I am not asking for any more work on this but if anyone knows the answer with certainty I would appreciate hearing it.  Bertrand are you OK with this?
4. The last thing is another computing issue.  After you have chosen a pipe, you randomly chose a point in each of the 3 pipe sections and then pick a section.  It's better to pick a section first and then compute just one point. I would also encapsulate the calculation in a function - you only need two functions since you can use the same function for the two torus sections; the second function is for the straight section.  I recommend declaring the functions in an anonymous namespace within this file.

bechenard:
Hi Rob,

Regarding point 3, it's not really the Jacobian but the sampling 
distribution. For a cylinder, selecting a point uniformly in the 
cross-section disk is given by r*sqrt(random). I believe you have a 
non-trivial transformation as soon as the metric of your space isn't 
constant (i.e. a n-dimensional box), but I don't have any proof.

This said, it should be fine to generate the photons from the center of 
the torus. As I mentioned earlier, we just want to have a good idea of 
what the calibration spectrum will look like, but we won't need the MC 
to extract the calibration constant.

Cheers,

Bertrand

kutschke:
> Hi Rob, Regarding point 3, it's not really the Jacobian but the sampling distribution. For a cylinder, selecting a point uniformly in the cross-section disk is given by r*sqrt(random). I believe you have a non-trivial transformation as soon as the metric of your space isn't constant (i.e. a n-dimensional box), but I don't have any proof. This said, it should be fine to generate the photons from the center of the torus. As I mentioned earlier, we just want to have a good idea of what the calibration spectrum will look like, but we won't need the MC to extract the calibration constant. Cheers, Bertrand
> […](#)

She fixed the code to correctly generate photons over a uniform disk.  We caught that early in the review and it is now R*sqrt(_randflat).  What I was not sure about is if that's the whole story.  Last night I thought there was a magical cancellation that saves us but this morning I think that's wrong.

Here is my current thought experiment.  Imagine a torus lying flat on a horizontal surface.  Make a vertical circular cut around the centerline to define  an inner section and an outer section.  I am pretty sure that the two sections have different volumes but that her algorithm will generate equal populations in the two sections.  Her 2D differential volume element should be  circular wedge, not a disk.  I have not yet figured out the mathematical form - that's connected to the Jacobian.

sophiemiddleton:
@kutschke, @dnbrow01 can you provide an example of your point in (2)? I already use FindMCPrimary in the FCL file, but the reco code will throw errors without a primary particle being created in the module, is there a way around?

sophiemiddleton:
@kutschke for point (4) you mean where I find the "sample" and then ask if x,y are within certain sections? But to get the information about those sections you have to do all the calculations preceding that "if(sample.....)elseif..." statement. The sections are not all the same size, so it would be complicated to select a section first, although I agree the way it seems weird. Did I misunderstand your comment?

dnbrow01:
> @kutschke, @dnbrow01 can you provide an example of your point in (2)? I already use FindMCPrimary in the FCL file, but the reco code will throw errors without a primary particle being created in the module, is there a way around?

Sorry, I think you want @brownd1978 ?  re:  PR 758.

frankcporter:
Hi Rob –

I hope I am not missing the point.
Yes, approximating the torus as a cylinder is not quite right.
But in this case the cylinder is “long and thin”, so the approximation
being made should be rather good, and good enough for the purposes
we need it for. Am I missing something?

cheers,
frank

Frank Porter
Professor of Physics
Caltech
***@***.******@***.***>
(626) 395-6777

frankcporter:
On this point:

  1.  This one is for Frank. In the email thread that said that you were comfortable with the approximations. I want to double check one of the approximations. The choice of pipe 1 to 5 is made using only the radius of the large torus section not on the full length of each pipe; the generation point is then uniformly distributed throughout the pipe. This under samples the inner pipes. These end effects are fairly large for the innermost pipe. You have the information to do this correctly and you use it when choosing which of the 3 pipe sections to use. Frank: is this OK as is or should it be fixed?
Thanks, Rob. Yes, I think this should be fixed. The longest pipes are actually the ones with the smallest radii, so it is even a (somewhat) bigger effect than at first sight. I would like to request:


  1.  Add to the documentation the manner of choosing the pipe (idx) [I had assumed it was just done randomly on the index, but Rob is correct that in the code it is based on the large bend radius].
  2.  Change the code (line 166 of EventGenerator/src/CaloCalibGun_module.cc) to base the idx selection on the total pipe length. An alternative, which is good to +- 5%, would be to select randomly on (1 to 5) [that is, as I had thought was being done].

cheers,
frank

sophiemiddleton:
thanks @frankcporter for spotting that. Yes, it should be a random pipe (I have changed it). I think at some point it was 1 - 5 i.e. pipe number and it got changed (I cant remember why).

kutschke:
> > @kutschke, @dnbrow01 can you provide an example of your point in (2)? I already use FindMCPrimary in the FCL file, but the reco code will throw errors without a primary particle being created in the module, is there a way around?
> 
> Sorry, I think you want @brownd1978 ? re: PR 758.

Sophie, please be more specific about what errors are thrown.

kutschke:
> @kutschke for point (4) you mean where I find the "sample" and then ask if x,y are within certain sections? But to get the information about those sections you have to do all the calculations preceding that "if(sample.....)elseif..." statement. The sections are not all the same size, so it would be complicated to select a section first, although I agree the way it seems weird. Did I misunderstand your comment?

In order to pick a section, you only need to compute, three things: circLgTor + circSmTor + lenStrait and then choose the random number named sample.  None of these 3 quantities depend on the random numbers drawn to compute the positions in the 3 sections - they are static properties of the geometry.  In fact you can move most of this computation to beginRun and you should.

kutschke:
Thanks for updating the way that you choose the pipe.  I have several followup comments.

1. Ideally you would generate decays uniformly throughout the volume of the pipes; you make two approximations that deviate from this.   Please document these approximations as comments at the top of the code.  One is that you treat all 5 pipes as equal volume when you pick a pipe.  The second is that you treat each torus is if it were a right circular cylinder with a length equal to the arc length of the centerline of the torus.  I am OK with these approximations but be clear that you are making them - otherwise the code just looks wrong.
2. At the line: https://github.com/sophiemiddleton/Offline/blob/c3ca78a1eaf64607c06fe9d3dbdb22aca08bd383/EventGenerator/src/CaloCalibGun_module.cc#L132  That comment was  and is wrong.  Also the next 3 lines are no longer needed.   Remove all 4 lines and remove the declaration of _randomRad in the header.
3. At line https://github.com/sophiemiddleton/Offline/blob/c3ca78a1eaf64607c06fe9d3dbdb22aca08bd383/EventGenerator/src/CaloCalibGun_module.cc#L168  Use _randFlat not rand.  We require that all random engines be managed by the RandomNumberService with seeds managed by SeedService.  This is to ensure that we can control the uniqueness and repeatability of how random variates are chosen.  You will need to update the algorithm since the range of values returned by rand and _randFlat are different.

kutschke:
> I will be on vacation from tomorrow for a week. I will try to get to this today, but it may not be done til after my vacation

Ok.  I will watch for it either way.  Enjoy your vacation.

sophiemiddleton:
@kutschke I have completeed 1 and 2 of your additional comments, can you advise how to get a random number of 1-5 from randFlat. Is it not just 0 or 1?

kutschke:
> @kutschke I have completeed 1 and 2 of your additional comments, can you advise how to get a random number of 1-5 from randFlat. Is it not just 0 or 1?

This will give 0 to 4:
`int i = std::round( 5.*flat1_.fire());`

And this will give 1 to 5:
`int i = std::round( 5.*flat1_.fire())+1;`

I believe that RandFlat returns values on the semi-open interval [0,1).  So you will never get 5 in the first case or 6 in the second.

bechenard:
Hi Sophie,

Something like "int(5*random.flat())+1" should do the trick.

Bertrand

sophiemiddleton:
OK, thanks for the advice, I'll try that

kutschke:
OOps - blundered by comment.  Editted it together it right.

bechenard:
I think we just crossed emails. On a purely mathematical (and pedantic 
note), the end of the interval has measure null, so it doesn't rally 
matter if it is open or closed. Ok, I shut up...

kutschke:
> I think we just crossed emails. On a purely mathematical (and pedantic note), the end of the interval has measure null, so it doesn't rally matter if it is open or closed. Ok, I shut up...

Your answer was might and mine was wrong - int not round, is right.

sophiemiddleton:
Ok, I made the changes and checked it still works

EventGenerator/src/CaloCalibGun_module.cc
```diff
@@ -101,138 +118,120 @@ namespace mu2e {
...
+
+    // pick a random theta, between 0 and 2*pi:
+    double theta = _randFlat.fire() * 2.0 * CLHEP::pi;
+    // pick a random point on the radius
+    double pipeR = _pipeRadius * sqrt(_randFlat.fire());
+    // find the z position based on above:
+    zpipe = pipeR*sin(theta);
+
+    // select an index from list of pipes:
+    unsigned int idx = int(5*_randFlat.fire())+1;
```
kutschke, line 167: Don't you want an index in the range 0 to 4, not 1 to 5?  You are indexing into std::vector with dimension 5.

kutschke (changes requested):

frankcporter:
Just to remark that you are using the wrong measure on
the (discrete, finite) set of computer numbers – the end of an interval
is not of measure zero, so if the interval is closed, the probability
is not zero to sample the end.

cheers,
frank

Frank Porter
Professor of Physics
Caltech
***@***.******@***.***>
(626) 395-6777

bechenard:
Correct, I forgot the discretness of the computer numbers....

sophiemiddleton:
Sorry, I edited the code to check and forgot to change back (Rob is right it should be 0 - 4, that is what I had checked and it worked)

gaponenko:
On Fri, 22 Apr 2022, Bertrand Echenard wrote:

> I think we just crossed emails. On a purely mathematical (and pedantic 
> note), the end of the interval has measure null, so it doesn't rally 
> matter if it is open or closed. 

Yes.  On the other hand on a real computer with a finite number of bits 
used to represent a floating point number there is a finite probability
to get a sample exactly at the end of an interval unless it is excluded
by the API.

Andrei

bechenard:
Yes, Frank already made that comment. I adopted a theorist point of 
view, glossing over the limitations of this finite world!

kutschke:
@FNALbuild run build test

EventGenerator/src/CaloCalibGun_module.cc
kutschke: Thanks for fixing this.

kutschke (approved):

PR #763, opened 2022-04-14
Update GeometryService to use validated fhicl.

kutschke:
Update plus one clean up.

### DIFF
--- a/GeometryService/inc/GeometryService.hh
+++ b/GeometryService/inc/GeometryService.hh
@@ -15,7 +15,10 @@
 #include "fhiclcpp/ParameterSet.h"
 #include "art/Framework/Services/Registry/ActivityRegistry.h"
 #include "art/Framework/Services/Registry/ServiceHandle.h"
+#include "art/Framework/Services/Registry/ServiceTable.h"
 #include "art/Framework/Services/Registry/ServiceDeclarationMacros.h"
+#include "fhiclcpp/types/Atom.h"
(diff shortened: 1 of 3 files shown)


brownd1978 (approved):
I'm happy you standardized the _ on variable names as well.

GeometryService/inc/GeometryService.hh
```diff
@@ -69,7 +95,7 @@ public:
 
     }
 
-    bool isStandardMu2eDetector() const { return standardMu2eDetector_; }
+    bool isStandardMu2eDetector() const { return _standardMu2eDetector; }
```
gaponenko, line 98: Hi Rob,   technically (some?) names starting with an underscore are reserved for C++ implementers, so it is a bad practice to introduce them into user code.    Trailing underscores should be preferred.
Andrei

brownd1978: While what you say is true, the majority of the Mu2e code base has prepended underscores.  Changing pre to post underscores was not part of the code consistency discussion.  A mix of pre and post-pended underscores is clearly an inconsistency, which was agreed by all as something to avoid.   Rob's commit made the file more self-consistent with the minimum of change, so satisfies these goals.

kutschke: These days I use trailing on new code.  For old code I value consistency within the file over the small risk.

PR #765, opened 2022-04-15
Separate B field config

kutschke:
This PR starts the removal of the BField configuration from the geometry file.  The automatic CI jobs will break until a matching PR is made in Production;  I will then resubmit the CI and correct any issues that appear in the second run.

I believe that this is enough to evaluate the changes.  There remain a number geometry and fcl files to update.

One question.   The standard configuration of the GeometryService now looks like:

 ```
GeometryService         : {
        inputFile           : "Offline/Mu2eG4/geom/geom_common.txt"
        bFieldFile          : "Offline/Mu2eG4/geom/bfgeom_v01.txt"
        simulatedDetector   : { tool_type: "Mu2e" }
  }
```
Another option would be:

```
GeometryService         : {
        inputFiles : {
            basegeom  : "Offline/Mu2eG4/geom/geom_common.txt"
            bfield    : "Offline/Mu2eG4/geom/bfgeom_v01.txt"
       }
        simulatedDetector   : { tool_type: "Mu2e" }
  }
```

The latter makes sense if we intend to do additional spliting down the road.  I don't know if we want to do that.

### CODE REVIEW

**Comment by kutschke** [comment]
The matching Production PR is 170: https://github.com/Mu2e/Production/pull/170

**Comment by kutschke** [comment]
To all of the reviewers: There are two PRs that need to be reviewed together, this one and Production #170.   The initial CI will fail and I will resubmit it using the test with feature; if any issues arise I will track them down.  There remains some cleanup work to do but I want to converge on this work before I do the cleanup of many files.

**Comment by kutschke** [comment]
@FNALbuild test with Mu2e/Production#170

File: Mu2eG4/geom/geom_2021_PhaseI_v02.txt [hunk 1]
```diff
@@ -9,7 +9,6 @@ bool hasProtonAbsorber  = true;
 bool hasTSdA           = true;
 bool hasExternalShielding = true;
 bool hasDiskCalorimeter = true;
-bool hasBFieldManager   = true;
```
AndrewEdmonds11, line 12: One minor consideration: Is it better to keep this parameter and set it to false explicitly? I see that it defaults to false in the code so it doesn't really matter

AndrewEdmonds11 (approved):
This all looks good to me. One very minor comment that isn't a showstopper for me.

brownd1978 (approved):
Thanks for making this split.  Looking at the content of the BField SimpleConfig files, I see they consist of a list of field headers for inner and outer maps, scale factor (currently global) and interpolation config.  Some features (toyDS, G4BL?) strike me as obsolete.  I suggest a further refactoring eventually to provide this info directly to GeometryService via fcl parameters (map headers and interpolation), with individual scale factors coming from the database (based on real current values), providing more control and eliminating the SimpleConfig indirection. That refactoring is beyond the scope of this PR.

kutschke:
> Thanks for making this split. Looking at the content of the BField SimpleConfig files, I see they consist of a list of field headers for inner and outer maps, scale factor (currently global) and interpolation config. Some features (toyDS, G4BL?) strike me as obsolete. I suggest a further refactoring eventually to provide this info directly to GeometryService via fcl parameters (map headers and interpolation), with individual scale factors coming from the database (based on real current values), providing more control and eliminating the SimpleConfig indirection. That refactoring is beyond the scope of this PR.

Some of these are indeed obsolete for Mu2e and I agree that it's out of scope for this PR.  I have created issue #766 to address this.

Analyses/fcl/DSBFieldDump.fcl
```diff
@@ -11,6 +11,7 @@ services : {
   GlobalConstantsService  : { inputFile      : "Offline/GlobalConstantsService/data/globalConstants_01.txt" }
     GeometryService	    : {
       inputFile      : "Offline/Mu2eG4/geom/geom_common.txt"
+      bFieldFile          : "Offline/Mu2eG4/geom/bfgeom_v01.txt"
```
rlcee, line 14: Given the override below, shouldn't this be
Offline/Mu2eG4/geom/bfgeom_reco_v01.txt

Mu2eG4/fcl/surfaceCheck.fcl
```diff
@@ -36,5 +36,6 @@ services.SeedService.maxUniqueEngines :  20
 physics.producers.generate.inputfile : "Offline/EventGenerator/defaultConfigs/surfaceCheck.txt"
 
 services.GeometryService.inputFile : "Offline/Mu2eG4/geom/geom_SurfaceCheck.txt"
+services.GeometryService.bFieldFile : "Offline/Mu2eG4/geom/bfgeom_reco_v01.txt"
```
rlcee, line 39: Doesn't thin effective add a field to this job when there wasn't one before?  Why not no_field?

rlcee (approved):
I'll approve with two minor inline comments - I just don't understand the intent.  The questions seem obvious so it makes me think I am missing something...

soleti (approved):

kutschke:
I have made all of the main line changes to this PR and to Production 170.  I am still looking at a few files that might be stale or might still be useful

@FNALbuild test with Mu2e/Production#170

kutschke:
It looks like the tests ran without merging the requested PR from Production.  I probably copied and pasted, creating an parsable command.

@FNALbuild run build test with Mu2e/Production/#170

kutschke:
This time for sure.  The first time had the embedded url bug and the second time had a spurious extra character

@FNALbuild run build test with Mu2e/Production#170

PR #778, opened 2022-04-26
Updated TrkPreampStraw script

bonventre:
Updated prodition generation script, also added flag that allows overriding database for strawElectronics time offsets that will be used in production digitization

### DIFF
--- a/TrackerConditions/test/TrkPreampStrawRandomize.py
+++ b/TrackerConditions/test/TrkPreampStrawRandomize.py
@@ -3,34 +3,54 @@
 import sys
 import numpy as np
 
+NPlanes = 36
+NPanels = 6
+NStraws = 96
+NUStraws = NPlanes*NPanels*NStraws
+
 if len(sys.argv) < 7:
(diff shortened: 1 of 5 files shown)


TrackerConditions/test/TrkPreampStrawRandomize.py
```diff
@@ -3,34 +3,54 @@
 import sys
 import numpy as np
 
+NPlanes = 36
```
brownd1978, line 6: These numbers can be obtained as StrawId::_nplanes;, etc, please use those instead.

brownd1978 (changes requested):

TrackerConditions/test/TrkPreampStrawRandomize.py
brownd1978: Otherwise the PR looks good.

brownd1978: Since this is python using StrawId.hh is problematic, outside the scope of this PR.  I'll accept it for now.

brownd1978 (approved):
We should understand how to better integrate python and c++ to avoid code duplication, which is outside the scope of this PR.

PR #780, opened 2022-04-27
Fix ExtMon whitespace.

gaponenko:
Whitespace only fixes.


kutschke (approved):
I confirmed that the only changes are to whitespace.

PR #783, opened 2022-04-29
Python Strawid

rlcee:
Here is the python using the PyWrap interface to the c++ StrawId class.  I don't feel strongly if you use it or not.  One might say that mixing code is an unnecessary complication, or one might say it is our future..  If merged, should have no effect on validation.

### DIFF
--- a/DataProducts/src/SConscript
+++ b/DataProducts/src/SConscript
@@ -24,6 +24,8 @@ helper.make_dict_and_map( [ mainlib,
                             'cetlib_except',
                             'Core'
                           ] )
+# turn pywrap.i into a python interface
+helper.make_pywrap ()
 
 # This tells emacs to view this file in python mode.
 # Local Variables:
--- a/DataProducts/src/pywrap.i
+++ b/DataProducts/src/pywrap.i
@@ -1,0 +1,22 @@
+//
+// swig interface file to wrap c++ code for python
+//
+
+%module DataProducts
+
+%include "std_string.i"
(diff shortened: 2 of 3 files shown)


kutschke:
> Here is the python using the PyWrap interface to the c++ StrawId class. I don't feel strongly if you use it or not. One might say that mixing code is an unnecessary complication, or one might say it is our future.. If merged, should have no effect on validation.

I think that there is value in this approach - it reduces points of maintenance, although this example is very unlikely to ever change.   The other option would be to code-gen both python and c++ but that would require an investment in new infrastructure.  I think that this is right way to go unless we discover issues.

brownd1978 (approved):

kutschke:
@bonventre a gentle reminder that this is in your queue

kutschke:
@FNALbuild run build test

kutschke:
@bonventre a gentle reminder that this is in your queue

bonventre:
Hey sorry for the delay, I've been out on vacation. We may want to add StrawStatus as well

rlcee:
> Hey sorry for the delay, I've been out on vacation. We may want to add StrawStatus as well

The interface to StrawStatus is fairly complicated (c++ vs python-wise) and it take a bit of work to get some methods working. Are you planning to use it in a particular way?  Like input a int bit word and test by string bit by name?

bonventre:
I think then lets wait until a specific use case comes up

bonventre (approved):

PR #784, opened 2022-04-29
Using physics constructors in the Minimal physics list etc.

resnegfk:
This pull request rewrites the Minimal physics list using a more current approach using physics constructors,
making it easier to expand it by adding other constructors if desired, e.g. for tests.
It also makes physics constructors consistent with more recent versions of Geant4.
No change in the results is expected.

### DIFF
--- a/Mu2eG4/src/Mu2eG4CustomizationPhysicsConstructor.cc
+++ b/Mu2eG4/src/Mu2eG4CustomizationPhysicsConstructor.cc
@@ -21,8 +21,6 @@ namespace mu2e {
 
   G4_DECLARE_PHYSCONSTR_FACTORY(Mu2eG4CustomizationPhysicsConstructor);
 
-  G4ThreadLocal G4bool Mu2eG4CustomizationPhysicsConstructor::wasActivated = false;
-
   Mu2eG4CustomizationPhysicsConstructor::
   Mu2eG4CustomizationPhysicsConstructor(const Mu2eG4Config::Physics* phys
                                         , const Mu2eG4Config::Debug* debug
@@ -44,9 +42,6 @@ namespace mu2e {
     , mu2elimits_(nullptr)
   {}
 
-  Mu2eG4CustomizationPhysicsConstructor::~Mu2eG4CustomizationPhysicsConstructor()
-  {}
-
   void Mu2eG4CustomizationPhysicsConstructor::ConstructParticle()
   {
     // Empty on purpose, for now
@@ -54,8 +49,6 @@ namespace mu2e {
 
   void Mu2eG4CustomizationPhysicsConstructor::ConstructProcess()
   {
-    if(wasActivated) { return; }
-    wasActivated = true;
 
     if (debug_->diagLevel()>0) {
       G4cout << "Mu2eG4CustomizationPhysicsConstructor::"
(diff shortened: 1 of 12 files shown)


Mu2eG4/src/Mu2eG4CustomizationPhysicsConstructor.cc
```diff
@@ -44,18 +42,13 @@ namespace mu2e {
...
-  {}
-
   void Mu2eG4CustomizationPhysicsConstructor::ConstructParticle()
   {
     // Empty on purpose, for now
   }
 
   void Mu2eG4CustomizationPhysicsConstructor::ConstructProcess()
   {
-    if(wasActivated) { return; }
```
kutschke, line 57: Please explain why this is not needed?  Is it because this code is only every called once, in the master thread?

kutschke (commented):
This looks good.  I have two questions in-line.

kutschke:
@goodenou Lisa, I added you as a reviewer to this PR to look at one thing so I don't think this will only take you a few minutes.  In two classes this PR removes protection against a member function being called multiple times; the classes are  Mu2eG4CustomizationPhysicsConstructor and MuonsWithSpinPhysicsConstructor.  Were these lines simply leftovers from development/debugging or is there more to the story?

Mu2eG4/src/Mu2eG4StepLimiterPhysicsConstructor.cc
```diff
@@ -14,17 +14,17 @@
 #include "Offline/Mu2eG4/inc/Mu2eG4StepLimiterPhysicsConstructor.hh"
 #include "Offline/Mu2eG4/inc/addStepLimiter.hh"
 
-using namespace std;
+// G4 includes
+#include "Geant4/G4PhysicsConstructorFactory.hh"
 
 namespace mu2e {
 
+  G4_DECLARE_PHYSCONSTR_FACTORY(Mu2eG4StepLimiterPhysicsConstructor);
```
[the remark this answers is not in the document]
resnegfk: It is for dynamic selection of physics constructor at the run (vs. compile) time which we have not been and are not using
I made it available for a potential future use.

Mu2eG4/src/Mu2eG4CustomizationPhysicsConstructor.cc
resnegfk: That approach was abandoned circa version 10.6+ and the static thread local variable was removed from all the physics constructors.
I still need to check what exactly replaced it.

kutschke: Thanks. That's all I need.

kutschke (approved):

kutschke:
Krzysztof's answers to my questions have satisfied me enough to merge now.  @goodenou: when you have a chance please have a look at the changes I asked you to look at and let me know if you agree that they are OK/

goodenou:
Hi Rob. Are you referring to the bool 'wasActivated'? I did not do any development work on the two files you mention, so I do not know the reason behind the existence of 'wasActivated'.

kutschke:
> Hi Rob. Are you referring to the bool 'wasActivated'? I did not do any development work on the two files you mention, so I do not know the reason behind the existence of 'wasActivated'.

Thanks Lisa.  As soon as I saw "G4ThreadLocal" I stopped thinking and decided you needed to be involved.  Using git blame, I see that this Krzysztof was the last person to touch this line.  Sorry for the noise.

PR #786, opened 2022-05-06
global whitespace cleanup

rlcee:


kutschke (approved):
I made a throw away clone and  merged into a  test branch.  I diffed this using 

`diff -w -b main`

This did not find any differences.  Merging now.

PR #789, opened 2022-05-07
White space 220507

kutschke:
This is another round of white space fixes.  I looked at the files that Ray did not feel comfortable changing and I decided that I felt comfortable changing more of them.  I changed some files using Ray’s bash one-liners and changed some using emacs commands.  I diff'ed all of them after the change and everything looks as expected.   Note that git diff -w -b main shows a few files in which trailing blank lines were removed – I am not sure how to suppress those.

   I ran ceSimReco 2k events before and after  and there is an exact match.

   With these changes the only files with hard tabs are *.png, *.root, *.dedx, and files that are computer generated spectra. There is also one Makefile, a few script files ( bash or .pl ) and valgrind configuration file.

  There are still 523 files with trailing white space.  Almost all of them (471) are SimpleConfig files in Mu2eG4/geom and its subdirectories .  The remaining 52 files are of the types that also have hard tabs plus *.xml files.

   This brings all of the following file name patterns into compliance with the white space rules: *.C, *.h, *.mac, *.icc, SConscript and and SConstruct.  I suggest that we add these to the checks in the pre-commit hook. 

   I think it is safe to run the trim trailing white space on all of the files rooted at Mu2eG4/geom.  Then the pattern Mu2eG4/geom/*.txt can be added to the pre-commit checks.


brownd1978 (approved):
Only whitespace changes in diff.

PR #792, opened 2022-05-10
Channel level delta t calibration

bonventre:


TrackerConditions/src/StrawResponseMaker.cc
```diff
@@ -98,19 +98,36 @@ namespace mu2e {
...
+    if ((int) _config.ehalfPVScale().size() != _config.eBins() ||
         (int) _config.tdCentralRes().size() != _config.eBins() ||
         (int) _config.tdResSlope().size() != _config.eBins() ||
         (int) _config.totDriftTime().size() != _config.totTBins()*_config.totEBins()){
       throw cet::exception("BADCONFIG")
         << "StrawResponse calibration vector lengths incorrect" << "\n";
     }
 
+    std::array<double, StrawId::_nustraws> strawHalfPropVelocity;
+    if (_config.strawHalfPropVelocity().size() == 0){
```
brownd1978, line 110: This logic is very flexible.  Do we want that also in the eventual database table?  Or should we just pick per-straw, per ustraw eventually?  Also, have you thought about how this looks with asymmetric thresholds?  Both questions are outside the scope of the PR.

TrackerConditions/fcl/prolog.fcl
```diff
@@ -156,19 +156,18 @@ StrawResponse : {
   totTBinWidth : 5
   totEBins : 10
   totEBinWidth : 0.00025
-
   // KeV
```
brownd1978, line 160: This table looks like a candidate for a future prodition.

TrackerConditions/src/StrawResponseMaker.cc
bonventre: I think we can leave this on the fcl side later to make overriding the db easier. For asymmetric thresholds, the error is larger on one side which would be easy to add, but also there can be energy dependent changes (like the effective time offset for the center of the straw), it's not clear to me yet how much of that we need to model here

brownd1978: Given it's unclear whether straw or ustraw is the right table segmentation I agree leaving this flexible is good.  I'm not sure about the empty table being interpreted as the default value.  Is that to support a bootstrap?  If so, it might be better as an explicit table override, to avoid accidentally omitting the table resulting in a silent error.

bonventre: This is just to avoid a huge default fcl array, when we add to the database this override will be made explicit

PR #795, opened 2022-05-11
fix units in TEve following ROOT changes

sophiemiddleton:
Units in 2D and 3D now the same thanks to ROOT fix


sophiemiddleton:
@FNALbuild run build test

sophiemiddleton:
Looks fine now, not sure what caused earlier issue

TEveEventDisplay/src/TEveMu2eDataInterface.cc
```diff
@@ -571,8 +571,8 @@ namespace mu2e{
               segment.helix().position(fltL,pos);
               CLHEP::Hep3Vector p = GenVector::Hep3Vec(pos);
               CLHEP::Hep3Vector InMu2e = det->toMu2e(p);
-              line->SetPostionAndDirectionFromKalRep((InMu2e.z()));
-              line->SetNextPoint((InMu2e.x()), (InMu2e.y()), (InMu2e.z()));
+              line->SetPostionAndDirectionFromKalRep((InMu2e.z()));              
+              line->SetNextPoint(pointmmTocm(InMu2e.x()), pointmmTocm(InMu2e.y()), pointmmTocm(InMu2e.z()));
```
kutschke, line 575: I think that I mentioned this before.  It's out of scope for this review but it will make the code base much easier to maintain and I strongly recommend that you do it at your earliest convenience.   Define a set of  inline free functions, to minimize the repeated work.  One example would be:

SetNextPoint ( TEvePointSet* t, CLHEP::Hep3Vector const& p, double factor=1){
  t->SetNextPoint ( factor*p.x(), factor*p.y(), factor*p.z());
}

There are 14 calls to SetNextPoint in this file alone.  Even better, template it on the first argument.  That way it will work with any class that has a member function with the signature SetNextPoint(double, double,double) .

kutschke (commented):
Please bring your code into compliance with the new whitespace rules:  the following files have lines with trailing whitespace:'

TEveEventDisplay/src/TEveMu2eDataInterface.cc
TEveEventDisplay/src/TEveMu2eMCInterface.cc
TEveEventDisplay/src/TEveMu2e_base_classes/TEveMu2eMCInterface.h

kutschke (changes requested):
The previous comment (about white space) was supposed to be flagged with "Request Changes".

kutschke:
@sophiemiddleton Did root change the units in it's public interface or is this a leftover from earlier work?

sophiemiddleton:
Hi, It comes from a previous bug in ROOT which is now fixed

TEveEventDisplay/src/TEveMu2eDataInterface.cc
sophiemiddleton: True, I will get to this. At the moment I'm focussing on getting REve upto standard, but TEve isnt perfect, I know, due to the way it has evolved

kutschke: > True, I will get to this. At the moment I'm focussing on getting REve upto standard, but TEve isnt perfect, I know, due to the way it has evolved

We will require this sort of change in the REve display so please do it proactively.

kutschke:
I will let the suggested improvements slide to another time but I will not merge the PR until it is compliant with the whitespace policy.

sophiemiddleton:
How do I see the WhiteSpace errors?

kutschke:
> How do I see the WhiteSpace errors?

https://mu2ewiki.fnal.gov/wiki/Editors

sophiemiddleton:
Yes, think I found them

kutschke:
@FNALbuild run build test

kutschke (approved):
Thanks for fixing the whitespace.  Presuming the CI test complete cleanly, I will merge.

PR #803, opened 2022-05-20
use splitString 

rlcee:
to remove duplicate code and simplify


kutschke:
@FNALbuild run build test

kutschke (approved):

PR #805, opened 2022-05-23
test PR - has whitespace problems

HCasler:


HCasler:
I left the "found errors" symbol as a wavy dash, based on the code for the  FIXME/TODO check

HCasler:
Do you mean the background of just the dash symbol or the table cell/row? I don't think I can do the former, unless markdown emojis have features I wasn't aware of. It might be possible to highlight the table row in some way.

I can also change the symbol to something else. We could use something like :heavy_exclamation_mark: or :point_right:. The list of markdown emojis is here: https://gist.github.com/rxaviers/7360908

kutschke:
> I can also change the symbol to something else. We could use something like ❗ or 👉. The list of markdown emojis is here: https://gist.github.com/rxaviers/7360908

I did mean the former but I was exploring.   Thanks for the reference to the markdown emojis.  I am looking for orange or yellow.  What about ⚠️ or 🔶  ?  I think that ⚠️ might be lost against the red and green symbols.  So I am leaning towards 🔶 .

HCasler:
I think :large_orange_diamond: will show up very clearly compared to :white_check_mark: and :x:. This will be an easy change.

kutschke:
> I think 🔶 will show up very clearly compared to ✅ and ❌. This will be an easy change.

Please go ahead.

HCasler:
It's in Mu2e/codetools#90

HCasler:
@FNALbuild run build tests

kutschke:
It looks great.  Thanks.

PR #807, opened 2022-06-01
Fix loop dependencies, part 1.

kutschke:
This PR fixes two compile time loop dependencies in our libraries.  It also fixes two other issues that were encountered while fixing these.  The two loops were:

RecoDataProducts/inc/CosmicTrack.hh used code from Mu2eUtilities which creates a loop.  There was no error before because code that should have been in the .cc file but was in the .hh file and therefore was never exposed to our link time loop prevention.  

The second loop was spurious and was in RecoDataProducts/inc/CosmicTrackSeed.hh .  I removed the spurious #include  and the spurious library in the link list.  No other code changes were needed.

The drive-by changes were in RecoDataProducts/inc/CosmicTrack.hh.  I removed
```using namespace std;```
which is not allowed in header files.  There were also helper structs in the global name space.  I move them to the mu2e namespace.   These changes required touching many other files.

I have verified that ceSimReco.fcl -n 2000 produces identical output before and after.

I have verified that the head of REve builds against this code.


kutschke:
@sophiemiddleton I made the changes that we discussed to two of cosmic data products.  I have added you as a reviewer.

brownd1978 (approved):

PR #808, opened 2022-06-01
Rmc trigger path

gianipez:
added the RMC trigger chain to the menus. The logic of this trigger was discussed during the Trigger meeting we had on May 16 2022 (https://mu2e-docdb.fnal.gov/cgi-bin/sso/DisplayMeeting?sessionid=12828)


gianipez:
I had to use clang to get a format that was allowing me to commit. `clang-format -i` didn't provide the best formatting result IMHO, and I also had to do several manual fixes. In the twiki I didn't find any specific info for helping format the FHICL files in the required style. Any idea how to overcome this? I also tried the `.clang-format` style in the `DbService` area, but it didn't help much with the FHICL format. 

I think that among the users, there are at least three different styles used to visualize and format the FHICL files: the formatting style from Kyle, the other one from Roberto and `tcl`. I honestly find it hard to pick one of these and enforce it on everybody. 

It would be simpler maybe to run routinely run a given script in the repo to standardize the formatting of the code, or at least it would save quite some time on the users' side.

Cheers,

Giani

gianipez:
@FNALbuild run build test

Trigger/data/OnSpillTrigMenu.config
```diff
@@ -8,3 +8,4 @@ cprDeM_highP_stopTarg:150
 cprDeP_highP_stopTarg:151
 cprDeM_lowP_stopTarg:160
 cprDeP_lowP_stopTarg:161
+caloFast_rmc:220
```
brownd1978, line 11: Please add the missing newline

CaloFilters/src/CaloClusterCounterFilter_module.cc
```diff
@@ -17,134 +17,164 @@
...
+    auto const& cluster = *icl;
+    CLHEP::Hep3Vector pos = cluster.cog3Vector();
+    float clRadius = std::sqrt(std::pow(pos.x(), 2) + std::pow(pos.y(), 2));
+    if (_diagLevel > 0) {
+      std::cout << "[CaloClusterCounter::filter] cluster "
+                << std::distance(caloClusters->begin(), icl) << " diskID = " << cluster.diskID()
+                << " E = " << cluster.energyDep() << " radial distance = " << clRadius
+                << " nCels = " << cluster.size() << std::endl;
+    }
+    if ((std::find(_diskID.begin(), _diskID.end(), cluster.diskID()) != _diskID.end()) &&
```
brownd1978, line 160: I'm confused why you use std::find here, since the conditional you apply seems trivial.  Or is it possible for clusters to have a non-physical diskID?  If so, how are such clusters created?

brownd1978 (changes requested):

CaloFilters/src/CaloClusterCounterFilter_module.cc
gianipez: I wanted to make the module more flexible for other possible use. I wanted to enable the possibility of using both disks; _diskID is the data-member with the allowed diskIDs and in the RMC specific config is set to `[0]`, while to use both disks it should be set to `[0, 1]`

brownd1978:
@FNALbuild run build test

gianipez:
@FNALbuild run build test

HCasler:
Testing a CI change:

@FNALbuild run build tests

brownd1978 (approved):

PR #812, opened 2022-06-09
Merge helices improvement

NamithaChitrazee:
- A uniform chi-squared calculation for all helices, irrespective of the source (TrkPatRe or CalPatRec) is implemented and used for the comparison of the helices. 
- The "greater the number of active StrawHits, better the helix" criteria relaxed to if "the difference in the number of active StrawHits between the two helix candidates < 5, select the helix with the better chi-squared value".

### DIFF
--- a/TrkReco/src/MergeHelices_module.cc
+++ b/TrkReco/src/MergeHelices_module.cc
@@ -10,11 +10,26 @@
 #include "art/Framework/Principal/Selector.h"
 #include "art/Framework/Principal/Event.h"
 #include "art/Framework/Principal/Handle.h"
+
+#include "Offline/GeometryService/inc/GeometryService.hh"
+#include "Offline/GeometryService/inc/GeomHandle.hh"
+#include "Offline/GeometryService/inc/DetectorSystem.hh"
+#include "Offline/CalorimeterGeom/inc/DiskCalorimeter.hh"
+#include "Offline/CalorimeterGeom/inc/Calorimeter.hh"
+#include "Offline/CalPatRec/inc/CalTimePeakFinder_types.hh"
+
 // mu2e data products
 #include "Offline/RecoDataProducts/inc/HelixSeed.hh"
 #include "Offline/RecoDataProducts/inc/TimeCluster.hh"
+
+#include "Offline/RecoDataProducts/inc/CaloCluster.hh"
+
 // utilities
 #include "Offline/TrkReco/inc/TrkUtilities.hh"
+
+#include "Offline/Mu2eUtilities/inc/LsqSums2.hh"
+#include "Offline/Mu2eUtilities/inc/LsqSums4.hh"
+#include "Offline/Mu2eUtilities/inc/polyAtan2.hh"
 // C++
 #include <vector>
 #include <memory>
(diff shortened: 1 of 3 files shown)


TrkReco/src/MergeHelices_module.cc
```diff
@@ -141,25 +151,28 @@ namespace mu2e {
     unsigned nh1, nh2, nover;
     countHits(evt,h1,h2, nh1, nh2, nover);
     unsigned minh = std::min(nh1, nh2);
+    float chih1xy(0),chih1zphi(0),chih2xy(0),chih2zphi(0);
+    unsigned deltanh = 5;
```
brownd1978, line 155: This number needs to be fhicl-configurable.  It should also be a class member.  Please  add an appropriate atom to the Config struct, the set the value using that in the constructor.

TrkReco/src/MergeHelices_module.cc
```diff
@@ -60,12 +61,14 @@ namespace mu2e {
       HelixComp compareHelices(art::Event const& evt,
           HelixSeed const& h1, HelixSeed const& h2);
       void countHits(art::Event const& evt,
-          HelixSeed const& h1, HelixSeed const& h2,
-          unsigned& nh1, unsigned& nh2, unsigned& nover);
+	  HelixSeed const& h1, HelixSeed const& h2,
+	  unsigned& nh1, unsigned& nh2, unsigned& nover);
       unsigned countOverlaps(SHIV const& s1, SHIV const& s2);
-  };
+      void findchisq(art::Event const& evt,
```
brownd1978, line 67: This function should be declared 'const'

TrkReco/src/MergeHelices_module.cc
```diff
@@ -183,13 +196,46 @@ namespace mu2e {
...
   }
-
+	
+  // The chisquared calculation method used here is 
+  // taken from the LsqSums approach followed in CalPatRec
+  void MergeHelices::findchisq(art::Event const& evt,
+      HelixSeed const& h1,float& chixy, float& chizphi) {
+    ::LsqSums4 sxy;
+    ::LsqSums4 szphi;
+    for(size_t ihit=0;ihit < h1.hits().size(); ihit++) {
```
brownd1978, line 206: Since you never use the 'ihit' variable except to dereference the hit, you can simplify the code by using a range loop:
for (auto const& hh : h1.hits()) {

TrkReco/src/MergeHelices_module.cc
```diff
@@ -183,13 +196,46 @@ namespace mu2e {
...
+      auto const& hh = h1.hits()[ihit];
+      float transErr = 5./sqrt(12.);
```
brownd1978, line 208: This number doesn't change, so calculate it in the constructor.  The value 5 here (and 1.1 and 0.75 below) are algorithmically tuned values, they should be made into class variables and initialized via fhicl parameter.  If there's a physical explanation what they mean or why they are set to these values (maybe Pasha knows) that shoudl be explained.  Note that the code should not take default values; they should be set in the appropriate prolog.fcl file.

brownd1978 (changes requested):

TrkReco/src/MergeHelices_module.cc
NamithaChitrazee: Hi Dave, so I should set the deltaNhits, XY and Phi weights (1.1 and 0.75) in the prolog.fcl right ? Shall make the changes ASAP

NamithaChitrazee:
Hi Dave, I think I have made all the necessary changes.

brownd1978 (approved):

brownd1978:
@FNALbuild run build test

NamithaChitrazee:
Sorry, I had not updated prolog_trigger.fcl. Hopefully, it should build OK now.

brownd1978:
@FNALbuild run build test

PR #814, opened 2022-06-10
Remaining components for VST calibration

bonventre:
Small changes to straight line fit, flags to enable straight line type longitudinal reco, and temporary dbTool changes for tracker calibration scripts


TrackerConditions/fcl/prolog.fcl
```diff
@@ -190,6 +190,8 @@ StrawResponse : {
     0.0705138 , 0.0657311 , 0.0559622 , 0.0685536 , 0.0604229 , 0.0574693 ,
```
brownd1978, line 190: Please add a comment documenting the provenance of these numbers

brownd1978 (changes requested):

TrackerConditions/fcl/prolog.fcl
bonventre: These are pretty old, I think from sims a few years ago, the new values are in the Production PR

rlcee (approved):
This all looks OK.  For the dbTool changes, I suggest we merge this version, then I will rewrite print-cid into print-run which will provide info for a given run, including CID and content.  I will also modify all parse-able output into returnable strings instead of cout output for use when DbTool class is called from other code.  

I did not review any algorithm changes.

brownd1978:
@FNALbuild run build test

brownd1978 (approved):

rlcee:
I don't mean to be a pain, but I noticed that the latest commit (Mon at 12:51) brought in two whitespace warnings.  Over the last week, it seems the majority of PR's have been putting in extra whitespace.  Since our intent was that these would be blocked at the "git commit" command by Muse, and this git CI check would never be triggered, I am really wondering how that check is being avoided.  @bonventre , do you know?  Thanks

bonventre:
Well, in my case I cheated by creating my commit by editing in the github UI... guess we should avoid that. Not sure how other PRs have been bypassing it

PR #815, opened 2022-06-16
add print-run

rlcee:
- remove "print-cid"
- add "print-run" which provides info about what is in a calibration set for a given run.
- options for print-run include drilling down to cid or table content
- all parse-able output is now available by string via getResult(), this enables imbedding DbTool in other code

### DIFF
--- a/DbService/inc/DbTool.hh
+++ b/DbService/inc/DbTool.hh
@@ -11,6 +11,15 @@
 //   DbTool tool;
 //   tool.init();
 //   tool.commitCalibrationSql(coll);
+//       or
+//    DbTool:vec_str args;
+//    args.push_back("print-content");
+//    args.push_back("--cid");
+//    args.push_back("67");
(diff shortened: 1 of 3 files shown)


rlcee:
> Any suggestions for reviewers for this PR?

Changes will force minor updates to Richie's code

kutschke:
@bonventre I added you as a reviewer because Ray says that this will require updates to your code.  Please review this.  Also, I will merge soon after you approve - so pick a time when you are ready to deal with the fallout.

bonventre (approved):

rlcee:
> Hi Ray, looks like if I now do print-run with the content flag , getResult() will then give me the same thing as calling findCid and returnContent(cid) would? Should be easy to edit my side. Feel free to merge this whenever, the only dependence is TrackerCalibration which I can fix later.

Correct.

PR #821, opened 2022-06-24
Recording the times before interaction/decay

resnegfk:
This pull requests adds the particle times just before the interaction/decay to SimParticle,
enabling the choice of using, e.g., the times before or after the decays.

In addition, some of the diagnostic printouts are enhanced/regularized.

### DIFF
--- a/MCDataProducts/inc/SimParticle.hh
+++ b/MCDataProducts/inc/SimParticle.hh
@@ -68,8 +68,6 @@ namespace mu2e {
       _endMomentum(),
       _endGlobalTime(0.),
       _endProperTime(0.),
-      // _preLastStepGlobalTime(0.),
-      // _preLastStepProperTime(0.),
       _endVolumeIndex(0),
       _endG4Status(),
       _stoppingCode(),
@@ -87,8 +85,8 @@ namespace mu2e {
                  art::Ptr<GenParticle> const&   agenParticle,
                  const CLHEP::Hep3Vector&       aposition,
                  const CLHEP::HepLorentzVector& amomentum,
-                 float                         astartGlobalTime,
-                 float                         astartProperTime,
+                 float                          astartGlobalTime,
+                 float                          astartProperTime,
                  unsigned                       astartVolumeIndex,
                  unsigned                       astartG4Status,
                  ProcessCode                    acreationCode,
@@ -111,8 +109,6 @@ namespace mu2e {
       _endMomentum(),
       _endGlobalTime(0.),
       _endProperTime(0.),
-      // _preLastStepGlobalTime(0.),
-      // _preLastStepProperTime(0.),
       _endVolumeIndex(0),
       _endG4Status(),
       _stoppingCode(),
@@ -127,22 +123,18 @@ namespace mu2e {
 
     void addEndInfo( CLHEP::Hep3Vector       aendPosition,
                      CLHEP::HepLorentzVector aendMomentum,
-                     float                  aendGlobalTime,
-                     float                  aendProperTime,
-                     // float                  apreLastStepGlobalTime,
-                     // float                  apreLastStepProperTime,
+                     float                   aendGlobalTime,
+                     float                   aendProperTime,
                      unsigned                aendVolumeIndex,
                      unsigned                aendG4Status,
                      ProcessCode             astoppingCode,
                      float                   endKE,
                      int                     nSteps,
-                     float                  trackLength){
+                     float                   trackLength){
       _endPosition     = aendPosition;
       _endMomentum     = aendMomentum;
       _endGlobalTime   = aendGlobalTime;
       _endProperTime   = aendProperTime;
-      // _preLastStepGlobalTime = apreLastStepGlobalTime;
-      // _preLastStepProperTime = apreLastStepProperTime;
       _endVolumeIndex  = aendVolumeIndex;
       _endG4Status     = aendG4Status;
       _stoppingCode    = astoppingCode;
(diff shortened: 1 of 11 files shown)


Mu2eG4/src/Mu2eSpecialCutsProcess.cc
```diff
@@ -65,15 +65,17 @@ namespace mu2e{
...
+      G4int prec = G4cout.precision(15);
       G4cout << __func__ << " : "
              << GetProcessName()
              << " : current step "
              << aTrack.GetCurrentStepNumber()
              << ", max step limit "
              << static_cast<G4int>(mu2elimits_.maxStepsPerTrack())
              << ", proposed step "
-             << proposedStep
+             << std::setw(24) << std::scientific << proposedStep
```
kutschke, line 76: I am pretty sure that std::scientific is sticky.  So save and restore as you did for precision.  There are two places.

kutschke (changes requested):
Except for the  issue with a sticky iomanip, this looks good.

resnegfk:
BTW, to enable the diagnostic printouts, set
physics.producers.g4run.debug.steppingVerbosityLevel : 2
physics.producers.g4run.debug.trackingVerbosityLevel : 2
It is quite verbose though (one can go above 2 if desired, but the output volume increases a lot)

Mu2eG4/src/Mu2eSpecialCutsProcess.cc
resnegfk: Thanks, you are right. I added << std::defaultfloat

kutschke (approved):
Thanks for fixing the printout.

gaponenko (commented):
Hi,  the code looks fine, but I would like to understand what information was missing.    What is the difference between endGlobalTIme and preLastStepGlobalTime, for example?  Is this e.g. muon decay time vs stop time?
Andrei

MCDataProducts/inc/SimParticle.hh
```diff
@@ -125,6 +129,8 @@ namespace mu2e {
                      CLHEP::HepLorentzVector aendMomentum,
                      float                  aendGlobalTime,
                      float                  aendProperTime,
+                     float                  apreLastStepGlobalTime,
```
brownd1978, line 132: I am worried about inflating the size of general-purpose objects, used in great numbers, to support specialty use cases.  Can you please explain why this information is of such general interest that it needs to be stored for every single SimParticle?

kutschke: Let's discuss this following Krzysztof's presentation this afternoon.

resnegfk: Note that this is per particle, not per step though.
I suggest we discuss it at our meeting. Meanwhile, do you see an alternative approach?

brownd1978: Here's the payload breakdown for a typical mixing input from MDC2020: 63% comes from SimParticle.  This size causes a real limitation on mixing (cosmic mixing takes >8GB memory), though we use extremely little of the payload.
Details for branch: Events
Number of entries:  91192
        Size     Size/Entry   Fraction  Data Product Name
     (bytes)       (bytes)

   193504968         2122.0      0.633  mu2e::SimParticlemv_compressDetStepMCs__NeutralsResampler.
   102588848         1125.0      0.336  mu2e::CrvSteps_compressDetStepMCs__NeutralsResampler.
     3906805           42.8      0.013  mu2e::GenParticles_compressDetStepMCs__NeutralsResampler.
     1659538           18.2      0.005  mu2e::StatusG4_g4run__NeutralsResampler.
     1257933           13.8      0.004  art::EventIDs_neutralsResampler__NeutralsResampler.
      999378           11.0      0.003  mu2e::StrawGasSteps_compressDetStepMCs__NeutralsResampler.
      862598            9.5      0.003  mu2e::CaloShowerSteps_compressDetStepMCs__NeutralsResampler.
      728436            8.0      0.002  EventAuxiliary
--------------------------------------
   305508504         3350.2      1.000  Total

kutschke: Suppose we split this into a separate data product:

1. What other parts of SimParticle, if any, should move with it?
2. Can we do it in a way that does not require maintenance on the compression routines or the stage>1 G4 code?

kutschke: We had a similar discussion about Bob's request to store information prior to the interaction for protons in the stopping target.  Does anyone remember what we did for that case?  I don't see it in SimParticle.hh .

rlcee: > We had a similar discussion about Bob's request to store information prior to the interaction for protons in the stopping target. Does anyone remember what we did for that case? I don't see it in SimParticle.hh .

From a thread 9/17/18, it looks like in Bob's case the variable "lastMomentum" was always zero, and we changed it to be the momentum at the last step before the interaction

gaponenko: Note that the _preLastStepKE member is still taking space, even though it does not seem to be used:  the value is hardcoded to be -1 and there are no setters.   May be it is time to take it out, along with the accessor.

resnegfk:
@FNALbuild run build test

MCDataProducts/inc/SimParticle.hh
resnegfk: As we discussed at the meeting I stored the pre-last step info in the previously existing data members;
BTW, the endMomentum is also based on the pre-last step info as is endKE, 
so it all may be actually more consistent now

MCDataProducts/inc/SimParticle.hh
```diff
@@ -68,6 +68,8 @@ namespace mu2e {
       _endMomentum(),
       _endGlobalTime(0.),
       _endProperTime(0.),
+      // _preLastStepGlobalTime(0.),
```
brownd1978, line 71: Please remove these commented lines

MCDataProducts/inc/SimParticle.hh
```diff
@@ -109,6 +111,8 @@ namespace mu2e {
...
```
brownd1978, line 114: and here

MCDataProducts/inc/SimParticle.hh
```diff
@@ -125,6 +129,8 @@ namespace mu2e {
                      CLHEP::HepLorentzVector aendMomentum,
                      float                  aendGlobalTime,
                      float                  aendProperTime,
+                     // float                  apreLastStepGlobalTime,
```
brownd1978, line 132: and here

MCDataProducts/inc/SimParticle.hh
```diff
@@ -135,6 +141,8 @@ namespace mu2e {
       _endMomentum     = aendMomentum;
       _endGlobalTime   = aendGlobalTime;
       _endProperTime   = aendProperTime;
+      // _preLastStepGlobalTime = apreLastStepGlobalTime;
```
brownd1978, line 144: and here

MCDataProducts/inc/SimParticle.hh
```diff
@@ -284,6 +294,8 @@ namespace mu2e {
     XYZTVectorF                    _endMomentum;
     float                  _endGlobalTime;
     float                  _endProperTime;
+    // float                  _preLastStepGlobalTime;
```
brownd1978, line 297: and here

brownd1978 (changes requested):

MCDataProducts/inc/SimParticle.hh
resnegfk: Will do. I was going to remove them and tidy up the white space in a separate commit.

kutschke: Can you turn that around quickly tomorrow?

kutschke:
@FNALbuild run build test

MCDataProducts/inc/SimParticle.hh
resnegfk: Done

PR #823, opened 2022-06-27
Improvements to SimpleTimeCluster_module

rosebranson88:
- Updated SimpleTimeCluster to identify N>1 TimeClusters per event
- Replaced all timestamps with correctedTime
- Replaced a cut on the time width of a cluster (TimeWindow<100ns) with a cut on the time between consecutive hits in a cluster (TimeStep<20ns)

### DIFF
--- a/CosmicReco/src/SimpleTimeCluster_module.cc
+++ b/CosmicReco/src/SimpleTimeCluster_module.cc
@@ -81,7 +81,6 @@ private:
   bool _useoneplane;
   art::InputTag _chToken;
   art::InputTag _shfToken;
-  const StrawHitFlagCollection* _shfcol;
   const ComboHitCollection* _chcol;
   StrawHitFlag _hsel, _hbkg;
 
@@ -103,14 +102,6 @@ void SimpleTimeCluster::produce(art::Event& event) {
 
   auto const& chH = event.getValidHandle<ComboHitCollection>(_chToken);
   _chcol = chH.product();
-
-  if (_testflag) {
-    auto shfH = event.getValidHandle<StrawHitFlagCollection>(_shfToken);
-    _shfcol = shfH.product();
-    if (_shfcol->size() != _chcol->size())
-      throw cet::exception("RECO")
-          << "SimpleTimeCluster: inconsistent flag collection length " << std::endl;
-  }
 
   std::unique_ptr<TimeClusterCollection> tccol(new TimeClusterCollection);
   findClusters(*tccol);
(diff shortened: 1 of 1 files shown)


CosmicReco/src/SimpleTimeCluster_module.cc
```diff
@@ -1,231 +1,261 @@
...
+  ordChCol.reserve(_chcol->size());
 
-    for (size_t i=0; i<_chcol->size(); ++i) {
-      const ComboHit& ch  = _chcol->at(i);
-      if (_testflag && !goodHit((*_shfcol)[i]))
-        continue;
-      ordChCol.push_back(ComboHit(ch));
+  for (size_t i = 0; i < _chcol->size(); ++i) {
+    const ComboHit& ch = _chcol->at(i);
+    if (_testflag && !goodHit(ch.flag()))
```
bonventre, line 142: Looks like you should remove _shfcol and related fcl configuration

bonventre:
Have you tested this on VST data? I do worry that a time step is less robust than the window (does it miss any VST events that the current module doesn't?), and a couple noisy channels could break things (maybe just a check on the length at the end would be enough to fix that)

bonventre:
You can run on raw.mu2e.TRK_VST-cosmics.vst003.art runs 100025 through 100033 or some subset of that, and can use VSTStraightLineReco.fcl, just change the reco path to end at SimpleTimeCluster and change the endpath to just TimeClusterDiag

CosmicReco/src/SimpleTimeCluster_module.cc
rosebranson88: Thanks! fixed

rosebranson88:
> Have you tested this on VST data? I do worry that a time step is less robust than the window (does it miss any VST events that the current module doesn't?), and a couple noisy channels could break things (maybe just a check on the length at the end would be enough to fix that)

We are still getting the VST tests set up to check on efficiency, but I like the idea of a length check at the end, especially in a noisy context. I tried implementing it, and with the very clean cosmic-CORSIKA data we've been testing on, it results in some efficiency loss. I wonder if it would make sense to have it as an option that can be uncommented (or just the value adjusted) when the module is used in a noisier environment (e.g. as part of an OnSpill support track trigger as opposed to just OffSpill CosmicReco).

CosmicReco/src/SimpleTimeCluster_module.cc
```diff
@@ -160,16 +167,20 @@ void SimpleTimeCluster::findClusters(TimeClusterCollection& tccol) {
...
         break;
+      if (_usetimeWindow && endTime - startTime > _timeWindow)
+        break;
     }
     count -= ordChCol[endIndex].nStrawHits();
     endIndex--;
     endTime = ordChCol[endIndex].correctedTime();
     if (_hasmaxnsh && count > _maxnsh)
       continue;
+    if (!_usetimeWindow && endTime - startTime > 100)
```
bonventre, line 182: I think 100 here is supposed to be _timeWindow

rosebranson88: Yes, thanks! Fixed

bonventre (approved):

brownd1978:
@FNALbuild run build test

PR #826, opened 2022-07-01
fetching pre step point times for particles born at rest

resnegfk:
This PR should improve the code created in PR#821 to more consistently handle the case of particles created at rest
and adds clarifying comments

### DIFF
--- a/Mu2eG4/src/Mu2eG4UserHelpers.cc
+++ b/Mu2eG4/src/Mu2eG4UserHelpers.cc
@@ -73,7 +73,9 @@ namespace mu2e {
     // pre step point. Note that for the particles decaying in flight,
     // the end times are the ones after they decay, while for the ones
     // decaying at rest, the recorded times are the ones before they
-    // decay
+    // decay. Also note, that while the particle decays at rest the
+    // track length is not changing, so no special treatment is needed.
+    // Similarly, when the KE is zero, the momentum direction does not matter
 
     // global time at the point of interaction/decay
     double getEndGlobalTime(G4Track const* const trk) {
--- a/Mu2eG4/src/Mu2eG4UserHelpers.cc
+++ b/Mu2eG4/src/Mu2eG4UserHelpers.cc
@@ -64,15 +64,28 @@ namespace mu2e {
 
     }
 
+    // The following functions attempt to get the particle end
+    // parameters. For the one which have a chance to travel, the
+    // Mu2eRecorderProcess stores the information in the
+    // Mu2eG4UserTrackInformation just before they interact
+    // inelastically or decay. For the particles born at rest, such
+    // information is not provided and their times are taken from the
+    // pre step point. Note that for the particles decaying in flight,
+    // the end times are the ones after they decay, while for the ones
+    // decaying at rest, the recorded times are the ones before they
+    // decay
+
     // global time at the point of interaction/decay
     double getEndGlobalTime(G4Track const* const trk) {
       auto const* uti = dynamic_cast<Mu2eG4UserTrackInformation*>(trk->GetUserInformation());
-      return (uti->GetGlobalTime() >= 0.) ? uti->GetGlobalTime() : trk->GetGlobalTime();
+      return (uti->GetGlobalTime() >= 0.) ? uti->GetGlobalTime() :
+        trk->GetStep()->GetPreStepPoint()->GetGlobalTime();
     }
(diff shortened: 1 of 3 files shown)


Mu2eG4/src/Mu2eG4UserHelpers.cc
```diff
@@ -64,15 +64,28 @@ namespace mu2e {
...
+    // information is not provided and their times are taken from the
+    // pre step point. Note that for the particles decaying in flight,
+    // the end times are the ones after they decay, while for the ones
+    // decaying at rest, the recorded times are the ones before they
+    // decay
+
     // global time at the point of interaction/decay
     double getEndGlobalTime(G4Track const* const trk) {
       auto const* uti = dynamic_cast<Mu2eG4UserTrackInformation*>(trk->GetUserInformation());
-      return (uti->GetGlobalTime() >= 0.) ? uti->GetGlobalTime() : trk->GetGlobalTime();
```
kutschke, line 70: I would like to make sure that I understand how this interacts with resampling stopped particles.  In particular how it interacts with PR #810.  In the job that creates stopped particles, uti->GetGlobalTime is always  >=0. because the particle has taken some steps to get to the point where it stopped and the recorder process has been run at least once.  In the resampling job if we resample a stopped particle, and it decays on it's first step, then the recorder process will not have been run and GetGlobalTIme will return -1.  Is that correct?

kutschke:
Hmmm.  I have no idea why my question appears twice.  Sorry for the noise - just answer it once.

Mu2eG4/src/Mu2eG4UserHelpers.cc
resnegfk: Yes. For the particles created at rest, only the at rest process(es) are tried/invoked, 
therefore the recorder process which is an along process is not invoked.
The Mu2eG4UserTrackInformation initializes both times to -1. at creation.
Without this PR the time after the decay is stored in SimParticle is such cases.
I tested this with g4study, but it would be good if someone else than me also tested it.

kutschke: I am not aware that we have any working code to test this.  We will when PR #810 is updated to reflect the recent changes.  I think that is the right time to test it:

@brownd1978  Do you agree that testing this PR needs to be integrated into testing #810 or do you see another way to test?

resnegfk:
It looks like, currently, after we decided to update the times in place, and not add the new data members, 
the track length stored is not consistent with the times for the particles decaying at rest

kutschke:
> It looks like, currently, after we decided to update the times in place, and not add the new data members, the track length stored is not consistent with the times for the particles decaying at rest

Does this mean that a particle born at rest has a non-zero length?  Something else?

resnegfk:
> > It looks like, currently, after we decided to update the times in place, and not add the new data members, the track length stored is not consistent with the times for the particles decaying at rest
> 
> Does this mean that a particle born at rest has a non-zero length? Something else?

It is currently the track length, which is after the decay

resnegfk:
> It is currently the track length, which is after the decay

It looks like I am tired. The particle is at rest, so it is not moving, so it is ok now.

I still want to make some small updates to remove commented out code and add some debug printout, 
so I will push anohter commit.

resnegfk:
@FNALbuild run build test

Mu2eG4/src/Mu2eG4UserHelpers.cc
brownd1978: Doesn't this code affect all stops?   If so, validation should test this through the pileup job, which uses stopped muons.

resnegfk: > Doesn't this code affect all stops? If so, validation should test this through the pileup job, which uses stopped muons.

Yes, it does. It actually runs for all the tracks.

kutschke (approved):

Mu2eG4/src/Mu2eG4UserHelpers.cc
kutschke: > Doesn't this code affect all stops? If so, validation should test this through the pileup job, which uses stopped muons.

I believe that the nightly ceMix does pileup which will test the mu- stops.   So I merged it to let the nightly do the test.  Now that it is merged Shihua's use case will be covered when he updates his PR and tests.

PR #828, opened 2022-07-22
KinKal drift updates

brownd1978:
Functional drift fit including crude D2T calibration


brownd1978:
@FNALbuild run build test with Mu2e/Production#196

brownd1978:
@FNALbuild run build test with Mu2e/Production#196

Mu2eKinKal/test/KKSeed.fcl
```diff
@@ -1,4 +1,4 @@
-#include "Offline/fcl/minimalMessageService.fcl"
+nclude "Offline/fcl/minimalMessageService.fcl"
```
rlcee, line 1: Looks like a typo - missing "#i", not sure how this is getting through CI, or is it a github glitch?

Mu2eKinKal/src/ANNStrawHitUpdater.cc
```diff
@@ -0,0 +1,71 @@
...
+#include "Offline/TrackerConditions/inc/StrawResponse.hh"
+#include "Offline/TrackerGeom/inc/Straw.hh"
+#include "Offline/RecoDataProducts/inc/ComboHit.hh"
+#include <cmath>
+
+namespace mu2e {
+  using KinKal::ClosestApproachData;
+  using KinKal::VEC3;
+  ANNStrawHitUpdater::ANNStrawHitUpdater(ANNSHUConfig const& annshutuple) {
+    mva_  = new MVATools(std::get<0>(annshutuple));
```
rlcee, line 11: This looks like a memory leak.  My preference would be to unique_ptr

Mu2eKinKal/inc/KKStrawXing.hh
```diff
@@ -87,9 +108,37 @@ namespace mu2e {
...
     }
     // find the material xings from gas, straw wall, and wire
-    smat_.findXings(tpca_.tpData(),sxconfig_,EXING::matXings());
+    smat_.findXings(tpca_.tpData(),sxconfig_,mxings_);
+    // reset
+    fparams_ = Parameters();
+    if(mxings_.size() > 0){
+      // compute the parameter effect for forwards time
+      std::array<double,3> dmom = {0.0,0.0,0.0}, momvar = {0.0,0.0,0.0};
+      this->materialEffects(TimeDir::forwards, dmom, momvar);
```
rlcee, line 121: Is "this" necessary?  I see other places, not introduced in this PR, where it also might not be needed.

CommonMC/src/SelectRecoMC_module.cc
```diff
@@ -229,12 +240,26 @@ namespace mu2e {
       tshmc._mom = mcstep.momentum();
       tshmc._time = fmod(mcstep.time(),_mbtime);
       tshmc._strawId = sdmc.strawId();
+      tshmc._earlyend = sdmc.earlyEnd();
+      // compute the signal propagation time and drift time
+      double vprop = 2.0*srep->halfPropV(sdmc.strawId(),1000.0*hit.energyDep());
```
rlcee, line 245: the 1000.0, which appears many places, would ideally be designed out, or maybe be a constexpr with a good name, or commented.

rlcee (changes requested):
Due to the size and complexity of this PR, I couldn't give it a thoughtful review with less than weeks of effort, so I spent an hour looking through the code and found a few things, commented inline, which might be improvements.  I will request changes, but I don't feel strongly.

Mu2eKinKal/test/KKSeed.fcl
brownd1978: fixed.

Mu2eKinKal/inc/KKStrawXing.hh
brownd1978: 'this' is need to access base class members of templated classes, even if the functions aren't overridden.  IMHO It is a flaw in C++, though I've read plenty of online commentary that 'explains' it.  The alternative is explicitly call the function, ie BaseClass::materialEffects(.  There is no consensus in the community on best practices, I chose to use 'this'.

CommonMC/src/SelectRecoMC_module.cc
brownd1978: This is an unfortunate unit convention conflict: Mu2e uses MeV, but tracker calibration uses KeV.  Fixing this out-of-scope for this PR.

brownd1978:
@FNALbuild run build test with Mu2e/Production#196

Mu2eKinKal/src/ANNStrawHitUpdater.cc
brownd1978: correct, a leak, I fixed it.  Sady, unique_ptr is incompatible with std::any, which is used to store objects of this class in the config.  see https://stackoverflow.com/questions/66353692/stdany-of-stdunique-ptr-doesnt-work

brownd1978:
@FNALbuild run build test with https://github.com/Mu2e/Production/pull/196

brownd1978:
I believe i have addressed all rlc's comments, please let me know if there's anything else

kutschke:
> I believe i have addressed all rlc's comments, please let me know if there's anything else

I will look at this after lunch.

RecoDataProducts/inc/TrkStrawHitSeed.hh
```diff
@@ -9,67 +9,142 @@
...
+    Float_t         _werr;    // raw hit U position error estimate
+    Float_t         _tottdrift;   // drift time from TOT for this hit
+    float_t         _ptoca;    // reference particle time of closest approach (TOCA)
+    float_t         _stoca;    // reference sensor time of closest approach (TOCA)
+    Float_t         _rdoca, _rdocavar;   // reference (biased) DOCA from the track to the wire, signed by the angular momentum WRT the wire and the measurement end (and variance)
+    Float_t         _rdt, _rtocavar;   // reference (biased) time difference (and variance) at POCA
+    Float_t         _udoca, _udocavar; // unbiaed DOCA (and variance)
+    Float_t         _udt, _utocavar;   //unbiased dt and variance
+    Float_t         _rupos; // reference POCA position along the straw WRT the straw middle
+    Float_t         _uupos; // unbiased POCA position along the straw WRT the straw middle
```
kutschke, line 133: _uupos is computed but it has no accessor.  Is one needed?

RecoDataProducts/inc/TrkStrawHitSeed.hh
```diff
@@ -9,67 +9,142 @@
...
-    Float_t         _edep;        // reco energy deposition
-    Float_t         _wdist;       // wire distance
-    Float_t         _werr;        // wire distance error estimate
-    StrawEnd      _end;         // straw end used for hit time measurement
+    StrawId         _sid;   // which straw has the hit
+    StrawEnd        _end;         // straw end used for hit time measurement
     StrawHitFlag    _flag;    // flag describing the status of this hit (active, ....)
+    Int_t           _ambig;   // hit state, including LR ambiguity
+    Int_t           _algo;     // hit updater algorithm
+    Bool_t          _frozen; // hit state was frozen
```
kutschke, line 120: _frozen is not used anywhere.  It has no accessor and it's not a c'tor argument.  I think it's safe to delete it.

RecoDataProducts/inc/TrkStrawHitSeed.hh
```diff
@@ -9,67 +9,142 @@
...
+    float_t         _stoca;    // reference sensor time of closest approach (TOCA)
+    Float_t         _rdoca, _rdocavar;   // reference (biased) DOCA from the track to the wire, signed by the angular momentum WRT the wire and the measurement end (and variance)
+    Float_t         _rdt, _rtocavar;   // reference (biased) time difference (and variance) at POCA
+    Float_t         _udoca, _udocavar; // unbiaed DOCA (and variance)
+    Float_t         _udt, _utocavar;   //unbiased dt and variance
+    Float_t         _rupos; // reference POCA position along the straw WRT the straw middle
+    Float_t         _uupos; // unbiased POCA position along the straw WRT the straw middle
+    Float_t         _rdrift;  // drift radius for this hit
+    Float_t         _rerr;    // intrinsic radial error
+    Float_t         _dvel;  // instantaneous drift velocity
```
kutschke, line 136: The data members from _dvel to _rdesidpvar are like _frozen.  They are not used anywhere.  They have no accessor and  they are not c'tor or function arguments.   I think it's safe to delete them.

kutschke (changes requested):
I have one substantive comment - possibly missing code in TrkFitFlag.cc . The others are observations and suggestions.

rlcee (approved):
Sorry, forgot to close out..

kutschke:
@bonventre Dave tells me that this PR is now complete from his point of view.  When do you expect to be able to review it?

RecoDataProducts/inc/TrkCaloHitSeed.hh
```diff
@@ -13,37 +15,70 @@
 #include <Rtypes.h>
 namespace mu2e {
   struct TrkCaloHitSeed {
-    TrkCaloHitSeed() :  _trklen(0), _hitlen(0), _cdoca(0),_rerr(0)  {}
-    // construct from the information
+    TrkCaloHitSeed() :  _cdepth(0), _cdoca(0), _ctoca(0), _ctocavar(0), _dt(0), _tresid(0), _tresidmvar(0), _tresidpvar(0), _trklen(0), _hitlen(0),  _time(0), _terr(0), _rerr(0) {}
```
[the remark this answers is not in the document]
brownd1978: Are you suggesting a constructor that takes all the data members as arguments?  That is error prone when there are lots of struct members, as its hard to see if the order of arguments is wrong.  Filling the arguments later requires each data member be filled by name, which I find clearer.    It's true that leaves open the possibility of omitting setting a variable; the default constructor settings make that obvious (downstream).

RecoDataProducts/inc/TrkCaloHitSeed.hh
```diff
@@ -13,37 +15,70 @@
...
-      _cpos(cpos), _tmom(tmom), _flag(flag)
+        Float_t time, Float_t terr, XYZVectorF const& cpos, XYZVectorF const& tmom, StrawHitFlag const& flag) :
+      _flag(flag), _cdepth(0), _cdoca(cdoca), _ctoca(0), _ctocavar(0), _dt(0),
+      _tresid(0), _tresidmvar(0), _tresidpvar(0),
+      _cpos(cpos), _tmom(tmom), _trklen(trklen), _hitlen(hitlen), _time(time), _terr(terr),
+      _t0(t0), _rerr(rerr)
     {}
     // accessors
     art::Ptr<CaloCluster> const& caloCluster() const { return _cluster; }
+    auto clusterDepth() const { return _cdepth; }
```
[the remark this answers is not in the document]
brownd1978: I've switched to using auto return type on all accessors, as it's one fewer maintenance point.  As the data member type is unambiguous I personally don't find this obfuscating.

RecoDataProducts/inc/TrkStrawHitClusterSeed.hh
```diff
@@ -0,0 +1,14 @@
...
+//  Original author: Dave Brown (LBNL) 1 Aug 2022
+//
+#ifndef RecoDataProducts_TrkStrawHitClusterSeedSeed_HH
+#define RecoDataProducts_TrkStrawHitClusterSeedSeed_HH
+#include "Offline/RecoDataProducts/inc/TrkStrawHitCluster.hh"
+#include <functional>
+namespace mu2e {
+  struct TrkStrawHitClusterSeedSeed {
+  };
+}
```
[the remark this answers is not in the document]
brownd1978: Yes, another stub for future development.  I also see it has a typo: fixed.

RecoDataProducts/inc/TrkFitFlag.hh
```diff
@@ -18,7 +18,7 @@ namespace mu2e {
     enum bit_type {hitsOK=0,circleOK,phizOK,helixOK,seedOK,kalmanOK,circleInit,phizInit,
     circleConverged,phizConverged,helixConverged,seedConverged,kalmanConverged,
     MatCorr, BFCorr, FitOK,
-    KSF=16, KFF, TPRHelix, CPRHelix, Straight, KKLoopHelix,KKCentralHelix,KKLine};
+    KSF=16, KFF, TPRHelix, CPRHelix, Straight, KKLoopHelix,KKCentralHelix,KKLine, MCSeed=31};
```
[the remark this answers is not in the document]
brownd1978: Yes, thanks for catching this, I'll make it consistent.  MCSeed is incomplete development in this branch: the structure is in place but the implementation is incomplete.  It will be addressed in a future PR.

RecoDataProducts/inc/TrkCaloHitSeed.hh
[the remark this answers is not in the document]
kutschke: No.  If a data member is intialized in the declaration, like:

  int var = 2;

Then it can be omitted from the intializer list and they will intialized to the correct value in the correct order.

This definitely works for integral types and I think it works for floating types to; I will check for float.  I believe that the rhs obeys the same rules as the rhs of a constexpr.

RecoDataProducts/inc/TrkStrawHitSeed.hh
```diff
@@ -9,67 +9,142 @@
...
+
+    // legacy interface
+    auto index() const { return _index; }
+    auto const&  strawId() const { return _sid; }
+    auto const& flag() const { return _flag; }
+    auto const& algorithm() const { return _algo; }
+    auto strawHitState() const { return _ambig; }
+    auto hitTime() const { return _htime; }
+    auto energyDep() const { return _edep; }
+    auto const& driftEnd() const { return _end; }
```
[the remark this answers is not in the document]
brownd1978: StrawEnd is a type, with content int.  It could be returned by value, I don't think there's a big difference.

RecoDataProducts/inc/TrkStrawHitSeed.hh
brownd1978: The accessor functions are legacy.  Updated and future development uses the struct data members directly, which I find clearer and less maintenance.

RecoDataProducts/inc/TrkStrawHitSeed.hh
brownd1978: _frozen is used (in TrkAna) and is set (non-trivially) on line 42.

RecoDataProducts/inc/TrkCaloHitSeed.hh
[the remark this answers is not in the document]
kutschke: I just checked - it does work for floating types too.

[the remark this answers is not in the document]
kutschke: Doing it this way also ensures that, when there are multiple c'tors, this produces a single point of maintenance for data members that should be initialized to the same value in all c'tors; of one c'tor of several wants to initialize to a different value they can override it in the intializer list.

RecoDataProducts/inc/TrkStrawHitSeed.hh
brownd1978: Like _frozen (and most of the TrkStrawHitSeed variables) these are diagnostics.  They are set on lines 48 through 51.

RecoDataProducts/inc/TrkCaloHitSeed.hh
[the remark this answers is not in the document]
kutschke: Your call.  My comment was about obfuscation for end users.

RecoDataProducts/inc/TrkStrawHitSeed.hh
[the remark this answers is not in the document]
kutschke: There is a small difference.  If you return an int by value it returns 4 bytes.   If you return by const& it returns 8 bytes and then copies the value from that address.

RecoDataProducts/inc/TrkStrawHitSeed.hh
kutschke: Thanks.  I had forgotten that these are structs with public data, so accessors are not needed.   I see that's the same answer for all of the other things that I thought were not useful.

bonventre (approved):

RecoDataProducts/inc/TrkCaloHitSeed.hh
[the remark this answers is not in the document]
brownd1978: OK I see now.  It's a valid suggestion, I'll give it a try.

brownd1978:
@FNALBUILD run build test with Mu2e/Production#196

kutschke (approved):

PR #829, opened 2022-07-27
STM Offline Infrastructure

AndrewEdmonds11:
Hi Everyone,

We've been putting together some of the Offline infrastructure for the STM. This is in no way final but I think now is as good a time as any to open the PR. All of this is independent of the rest of Offline so should be safe to merge in but I would understand if this was left unmerged until we are further into this project. Anyway, an explanation of the code as it stands is below.

Thanks,
Andy

==============================

On the simulation side, we have the following:

 - a **new module**: ```MakeSTMSteps``` (in **new package** ```STMMC```) that takes ```StepPointMCs``` and creates the **new data product** ```STMStep``` (in package ```MCDataProducts```)
   - this just sums all the energy deposits in the same event and gets the time from the first step
 - a **new module**: ```MakeSTMDigis``` (in package ```STMMC```) that takes ```STMSteps``` and create the **new data product** ```STMDigi``` (in package ```RecoDataProducts```)
   - this creates a digi with just a single ADC value that we smear by a Gaussian
  - a **new module**: ```MakeSTMHits``` (in new package ```STMReco```) that takes ```STMDigis``` and creates the **new data product** ```STMHit``` (in package ```RecoDataProducts```)
    - at the moment, this module only works on ```STMDigis``` made from the simulation

On the data side, we have the following:

 - a **new module**: ```FromSTMTestBeamData``` (in package ```Sources```) that takes raw binary files from the test beam and creates ```STMDigis```
   - this uses structs defined ```STMTestBeamHeaders.hh``` (in package ```Sources```)
   - the fcl file for this is module is in ```Sources/test/```
   - I don't think this is the long-term home for this module. Ideally, I will get set up so that we read the binary files with otsdaq
 - a **new module**: ```STMZeroSuppression``` (in package ```STMReco```) that wraps the zero-suppression algorithm (stored in ```ZPAlg.hh/cc```) into an art module. This module runs on the unsuppressed digis
    - note that ```ZPAlg.hh/cc``` is a copy of code that was run at the test beam
    - Offline is definitely not the long-term home for either of these files but it's where we are currently running it
    - I imagine we will want some way to run zero-suppression offline but that requires some thinking
  - a **new module**: ```STMMovingWindowDeconvolution``` (in package ```STMReco```) that wraps the MWD algorithm (stored in ```MWDAlg.hh/.cc```) into an art module. This module runs on the zero-suppressed digis.
    - like with the zero-suppression ```MWDAlg.hh/.cc``` is a copy of code run during the test beam, and Offline is not the long-term home for this

A couple of other miscellaneous things:
 - we have some enums defined in ```STMTypes.hh``` (in package ```DataProducts```)
 - we have a couple of plotting modules 
   - ```PlotSTMDigis``` (in package ```STMReco```) plots the digitized waveforms
   - ```PlotSTMSpectrum``` (in package ```STMReco```) plots the energy spectrum of ```STMHits```


rlcee:
Andy, I have to follow up on one detail.  I see there were whitespace format errors (like spaces at the end of the line) in your PR.  Muse is setup to block any commit which has whitespace errors, so nominally there should never be a PR with a whitespace format problem, but recently we have seen several, so I'd like to ask how did you avoid the Muse commit hook, if you know?  Thanks

AndrewEdmonds11:
Hi Ray,

I'm not sure to be honest - we did run into those commit hooks a couple of times so I thought it was all OK. Some of these commits are from a long time ago (~6 months) so could it be from before the commit hook went in? 

I will clean up the whitespace issues shortly.

Thanks,
Andy

AndrewEdmonds11:
So I've fixed the whitespace errors and found a couple of things with my emacs set up:

 - I couldn't see the whitespace errors highlighted in the SConscript file until I added the following to my .emacs file:
 `(add-hook 'python-mode-hook
            '(lambda ()
                    (whitespace-mode)))`
 - I also couldn't see the whitespace errors in the fcl file because I hadn't followed the instructions in the comments of the .emacs file to get art-fhicl mode (which now gives my fcl files nice highlighting too). I also needed to add this to my .emacs
`(add-hook 'art-fhicl-mode-hook
            '(lambda ()
                    (whitespace-mode)))`

I can add this information to the wiki for future reference

rlcee:
>  Some of these commits are from a long time ago

That's probably the main thing.  The hook went in May 4.

> I can add this information to the wiki for future reference

I think probably all the main points are on [Editors page](https://mu2ewiki.fnal.gov/wiki/Editors), but please check if there is anything to add.  Thanks

brownd1978:
@FNALbuild run build test

RecoDataProducts/inc/STMDigi.hh
```diff
@@ -0,0 +1,78 @@
...
+    STMDigiFlag flag() const { return _flag; }
+    const std::vector<int16_t>& adcs() const { return _adcs; }
+
+  private:
+    uint32_t _trigNum;  // trigger number
+    STMTrigType _trigType; // combination of: trigger mode (external (beam) or internal (source)),
+                           // channel (HPGe or LaBr), and
+                           // data type (unsuppressed, zero-suppressed, MWD etc.
+    uint64_t _trigTime; // trigger time [ct]
+    uint32_t _trigTimeOffset; // time offset from trigger to first ADC value [ct]
```
brownd1978, line 70: Is this offset variable or is it fixed at begin run?  If it is fixed I suggest to make it part of a prodition.

RecoDataProducts/inc/STMDigi.hh
```diff
@@ -0,0 +1,78 @@
...
+    uint32_t _extra;  // type-dependent (i.e. for MWD digis will be mean and RMS of baseline,for PQ digis would be quality)
+    STMDigiFlag _flag;    // various error flags
+    std::vector<int16_t> _adcs; // vector of ADC values for the waveform
```
brownd1978, line 73: What defines the time bin for the ADC waveform?  That might also be a good candidate for a prodition.

RecoDataProducts/inc/STMDigi.hh
```diff
@@ -0,0 +1,78 @@
...
+      }
+    }
+
+
+    STMDigiFlag flag() const { return _flag; }
+    const std::vector<int16_t>& adcs() const { return _adcs; }
+
+  private:
+    uint32_t _trigNum;  // trigger number
+    STMTrigType _trigType; // combination of: trigger mode (external (beam) or internal (source)),
```
brownd1978, line 66: Are trigger type and number provided by TDAQ, or are they internal to the STM readout?  If they come from TDAQ are these values redundant with the packet header?

RecoDataProducts/inc/STMDigi.hh
```diff
@@ -0,0 +1,78 @@
...
+                           // channel (HPGe or LaBr), and
+                           // data type (unsuppressed, zero-suppressed, MWD etc.
+    uint64_t _trigTime; // trigger time [ct]
+    uint32_t _trigTimeOffset; // time offset from trigger to first ADC value [ct]
+    uint32_t _extra;  // type-dependent (i.e. for MWD digis will be mean and RMS of baseline,for PQ digis would be quality)
+    STMDigiFlag _flag;    // various error flags
+    std::vector<int16_t> _adcs; // vector of ADC values for the waveform
+  };
```
brownd1978, line 74: Is there nothing in STM that defines a channel?  I see that you have defined a  STMDigiCollection, but there doesn't seem to be anything that would distinguish 1 digi from another.

STMReco/src/MWDAlg.cc
```diff
@@ -0,0 +1,214 @@
...
+    if (cut_mode == 1)
+      threshold_cut = baseline_mean - baseline_rms*nsigma_cut;
+    else if (cut_mode == 2)
+      threshold_cut = fixed_cut_parameter;
+    else {
+      std::cout << "ERROR:: find_peaks() : unknown cut mode .... " << cut_mode << std::endl;
+      exit(-1);
+    }
+
+    //    if (cut_mode == 1)
```
brownd1978, line 57: If these printouts are useful I suggest to add a 'diaglevel' member and enable them when set.

STMReco/src/STMZeroSuppression_module.cc
```diff
@@ -0,0 +1,94 @@
...
+  };
+
+  STMZeroSuppression::STMZeroSuppression(const Parameters& config )  :
+    art::EDProducer{config}
+    ,_stmDigisTag(config().stmDigisTag())
+    ,_zpAlg(config().samplingFrequency(), config().tbefore(), config().tafter(), config().threshold())
+    ,_ctPerNs(1.0/(config().samplingFrequency()*1e-3))
+  {
+    consumes<STMDigiCollection>(_stmDigisTag);
+    produces<STMDigiCollection>();
```
brownd1978, line 65: Are the digi objects in the input and output collection identical, or does this module process the data as well as suppress zeros?  If there is content-level change you might consider a separate class for output (recodigi).

STMReco/src/ZPAlg.cc
```diff
@@ -0,0 +1,83 @@
...
+
+namespace mu2e {
+
+  int16_t* ZPAlg::ZeroSuppress(const int16_t* ADC, unsigned long int n, std::vector<size_t>& starts, std::vector<size_t>& ends){
+    //n number of elements to apply zero sup
+
+    //----------------------- PEAKS SUPPRESSION ALGORITHM-----------------------------------------
+
+    //Calculate the gradient for ADC values separated a distance window (in time would be window * tadc)
+    int window=100;
```
brownd1978, line 17: is the window size intended to be constant?  If so, consider making the ADC data an array.  If it might change, consider making it a vector.   If it must be passed by pointer, please pass the length with it.  These changes would avoid assuming the array length (which is unconstrained by the interface).

Sources/src/FromSTMTestBeamData_source.cc
```diff
@@ -0,0 +1,316 @@
+//
```
brownd1978, line 1: I like using a source as the interface to non-art testbeam data.

brownd1978 (commented):
Can you suggest other reviewers for this PR?

RecoDataProducts/inc/STMDigi.hh
AndrewEdmonds11: This one is variable. It's the time of this hit relative to the trigger time. There will be other offsets to time aligne the STM with other detectors and that would be a prodition

RecoDataProducts/inc/STMDigi.hh
AndrewEdmonds11: It will be defined by the sampling frequency in the ADC. I agree that it should be a prodition

RecoDataProducts/inc/STMDigi.hh
AndrewEdmonds11: I think the trigger number will be provided by TDAQ so I will remove it from here. The trigger type is a combination of things, some of which will be provided by TDAQ and some will be provided by the STM readout (like channel and data type). I'll split these out as necessary

RecoDataProducts/inc/STMDigi.hh
AndrewEdmonds11: So the channel number (HPGe or LaBr) is defined in the STMTrigType class at the moment. My plan was to keep digis from both detectors in the same collection for as long as possible

STMReco/src/STMZeroSuppression_module.cc
AndrewEdmonds11: As we discussed in the Comp-Soft meeting, my idea was to have one class for everything. We've agreed that's not necessarily the best design and that I need to rethink this part.

STMReco/src/ZPAlg.cc
AndrewEdmonds11: I'm not sure to be honest. This code is likely to move to online so we will keep this in mind

RecoDataProducts/inc/STMDigi.hh
kutschke: Can an event have many STMDigis?  If so, can they have different trigger types?

STMReco/src/ZPAlg.cc
kutschke: This code returns a pointer to something.  You should annotate the code to make it clear whether or not the caller receives ownership of memory ( ie is it their job to delete anything?).

AndrewEdmonds11:
Hi Everyone, I'm going to close this draft PR for the STM. It is just being used as a working branch for me and Scott. It's contents will be migrated to cleaner PRs as we go

PR #830, opened 2022-08-08
STMChannel enum class

AndrewEdmonds11:
Hi Everyone,

I will break down the big PR #829 into smaller PRs as I can / when it's needed. This PR adds an enum class for the STMChannel. This is needed to start work on the proditions interface (which Ray has kindly agreed to work on) since we will index the rows in the conditions database to the channel number.

I copied the code from the GenId class and made any necessary changes. One non-obvious change is that the "unknown" member is not at position 0. I did this because we want to start the channel numbering at 0. Let me know if there is a better way to do this.

Thanks,
Andy

### DIFF
--- a/DataProducts/inc/STMChannel.hh
+++ b/DataProducts/inc/STMChannel.hh
@@ -1,0 +1,129 @@
+#ifndef DataProducts_STMChannel_hh
+#define DataProducts_STMChannel_hh
+//
+// An enum-matched-to-names class for generator Id's.
+// based on GenId.hh
+//
+
+#include <cstdlib>
+#include <iostream>
+#include <string>
+#include <vector>
+
+namespace mu2e {
+
+  class STMChannel {
+
+  public:
+
+    // Need to keep the enum and the _name member in sync.
+    // Also, restrict to 8-bits because we will never need that many
+    // and this may be in every STMDigi
+    enum enum_type : int8_t { HPGe, LaBr, unknown, lastEnum };
+
+#ifndef SWIG
+    // Keep this in sync with the enum. Used in STMChannel.cc
+#define STMCHANNEL_NAMES \
+    "HPGe", "LaBr", "unknown"
+#endif
+
+  public:
+
+    // The most important c'tor and accessor methods are first.
+    STMChannel( enum_type id):
+      _id(id)
+    {}
+
+    enum_type id() const { return _id;}
+
+    const std::string name() const {
+      return std::string( _name[_id] );
+    }
+
+    // ROOT requires a default c'tor.
+    STMChannel():
+      _id(unknown){
+    }
+
+    virtual ~STMChannel(){}
+
+    bool operator==(const STMChannel g) const{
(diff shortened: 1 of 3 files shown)


AndrewEdmonds11:
P.S. I would suggest at least Rob and Ray as reviewers. Rob because he wrote the original GenId class, and Ray because he will be using this in the proditions code so I'd like to check it has everything he needs. Thanks

DataProducts/inc/STMChannel.hh
```diff
@@ -0,0 +1,129 @@
...
+    static int version() { return _version; }
+
+    // Static version of the name method.
+    static const std::string name( enum_type id ){
+      return std::string( _name[id] );
+    }
+
+    // Check validity of an Id. Unknown is defined to be valid.
+    static bool isValid( enum_type id){
+      if ( id <  unknown  ) return false;
```
kutschke, line 76: This needs maintenance because of the change in the numeric value of unknown.  The right test is <0 or >=unknown returns false.

kutschke (changes requested):
I really like that you added a test.   How is it invoked ?  Just root -l STMChannel.cc ?

AndrewEdmonds11:
> I really like that you added a test. How is it invoked ? Just root -l STMChannel.cc ?

No, it should be ```root -l STMReco/test/testSTMChannel.C```

DataProducts/inc/STMChannel.hh
AndrewEdmonds11: Thanks, Rob. Done :)

rlcee (approved):
You might want to fix the comment at the top of the hh and cc.  Consider adding printAll and findByName to the test.  Consider making the != operator the negation of the == operator.

DataProducts/inc/STMChannel.hh
kutschke: Thanks.

kutschke (approved):

kutschke:
A few last comments.  First, since this is an int8_t, think about alignment in classes/structs that contain this type as member data.  There are data types that must be 8 byte aligned or 4 byte aligned.  The general rule is to put all member data that must be 8 byte aligned first, then all data that must be 4 byte aligned and so on.  So single byte objects go last.  We have not always been careful about this to date but let's move in that direction.

The second question is: are there other small data members that could sensibly be packed along with this into a single word?  The total size of the STM data may be small enough that this is moot - or maybe not. What is known at this time?

AndrewEdmonds11:
Thanks, Rob. I'll keep the alignment in mind going forward.

On your second question: at the moment in the test beam data, we pack the channel number in a 16-bit word with the trigger type (internal or external) and the data type (unsuppressed/zero-suppressed etc.). It's unclear to me at the moment whether we will keep these packed or how much will be stored in the STMDigis themselves

kutschke:
Thanks for the information.  I will run the CI one more time, and, if all goes well, merge it.

@FNALbuild run build test

PR #836, opened 2022-08-26
replace 3.14 with M_PI

rlcee:
per #235 replace 3.14 literal with M_PI. It is possible that some of these uses for histogram boundaries were strategically just not quite pi, so they should be reviewed.  Tested identical with validation 1K ceSimReco and 200 reco.

### DIFF
--- a/Analyses/src/CaloClusterCompare_module.cc
+++ b/Analyses/src/CaloClusterCompare_module.cc
@@ -19,6 +19,8 @@
 #include "TLegend.h"
 #include "TTree.h"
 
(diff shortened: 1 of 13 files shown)


brownd1978 (approved):

kutschke (approved):

kutschke:
I thought that I had already sent this but I guess I forgot to click something - my apologies if you see it twice.

I suggest that we merge this today and not merge any other PRs that might affect the nightly validation.  This is the least work and the risk that it creates additional work is very low.

PR #838, opened 2022-08-29
cleanup fcl with Production references

rlcee:
per discussion on #426


kutschke (approved):
For now, just approving as a test.  See the conversation in PR #839

kutschke:
My test is complete and the approval is real.  Merging now.

PR #839, opened 2022-08-29
error exit on bad table name

rlcee:
no effect on validation

### DIFF
--- a/DbService/src/DbTool.cc
+++ b/DbService/src/DbTool.cc
@@ -88,18 +88,14 @@ int mu2e::DbTool::refresh() {
 int mu2e::DbTool::printContent() {
   int rc = 0;
 
-  std::string name;
-  std::string user;
-  std::vector<int> cids;
-
   map_ss args;
   args["name"] = "";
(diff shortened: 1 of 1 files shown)


kutschke (changes requested):
I am making a fake review requesting changes to test my understanding of this comment in the github docs:

https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/about-protected-branches#require-pull-request-reviews-before-merging

> Even after all required reviewers have approved a pull request, collaborators cannot merge the pull request if there are other open pull requests that have a head branch pointing to the same commit with pending or rejected reviews.  Someone with write permissions must approve or dismiss the blocking review on the other pull requests first.
>  

This PR and PR838 both point to the same head commit in Offline/main.

kutschke (approved):
"Request Changes" status did not block merging for PR #838.  So the text quoted in the previous comment means something different.

Approving for real.

kutschke:
@FNALbuild run build tests

PR #840, opened 2022-09-01
switch to new envset for art v3_12 + g4_v4_10_7_p03b

macndev:


kutschke-non-admin:
The time stamp on /cvmfs/mu2e.opensciencegrid.org/DataFiles/Muse/p032 is 11:28 AM and the time stamp on the fail message above is 11:28 AM.  So it seems likely that the CI failed because it could not see the envset.  I will resubmit the CI to check.

@FNALbuild run build test

kutschke:
Ooops.  Helenka pointed out that the last two command did not run because I was using my non-admin account that I created so that I could see what non-admin users see.   This time for sure.

@FNALbuild run build test

kutschke (approved):
Mackenzie has performed the validation and her report is below.  This is ready to be merged.

> I just finished documenting the art v3_12 validation. In all of my jobs, I see perfect matches across all plots. The number of stopped muons is exactly identical between the new and old areas.
> 
> https://mu2ewiki.fnal.gov/wiki/GeantChecklist#Art_v3_12_00
> 
> I double checked my log files to make sure the old area uses art v3_11 and the new area uses art v3_12 and indeed, it looks like everything ran correctly.

PR #846, opened 2022-09-02
added missing parameter configuration

gianipez:
Working with a student, we found out that the SeedFit (aka the tack Kalman-based fit with no material nor BField correction) uses a constant BField assumption and that we need to use this parameter to set it properly.

### DIFF
--- a/Mu2eG4/geom/bfgeom_DS70_no_tsu_ps_v01.txt
+++ b/Mu2eG4/geom/bfgeom_DS70_no_tsu_ps_v01.txt
@@ -23,6 +23,7 @@ vector<string> bfield.innerMaps = {
   "BFieldMaps/Mau13/DSExtension.header"
 };
 
+double bfield.scaleFactor = 0.7;
 
 //
 // This tells emacs to view this file in c++ mode.


gianipez:
In the seedFit case, the BField value is set in the following way: https://github.com/Mu2e/Offline/blob/main/TrkReco/src/KalFit.cc#L1004 

TBH, I would prefer it to be set in this other way:
https://github.com/Mu2e/Offline/blob/main/TrkFilters/src/HelixFilter_module.cc#L139 

Maybe @brownd1978 can comment on this.

Mu2eG4/geom/bfgeom_DS70_no_tsu_ps_v01.txt
```diff
@@ -23,6 +23,7 @@ vector<string> bfield.innerMaps = {
   "BFieldMaps/Mau13/DSExtension.header"
 };
 
+double bfield.scaleFactor = 0.7;
```
kutschke, line 26: Can you explain what you are doing here.  I think that this will make a field in the DS that is 0.49 times the nominal field?  The factor multiples the existing map.

kutschke: I looked at the code you pointed to in your comment.   I like your suggestion of looking up a nominal value in the map rather than using this scale factor. 

I am 99% sure that the PR as is should be rejected.  I will wait to hear from Dave and I added him as a reviewer.

kutschke:
Here is a suggestion that will let you keep going for the time being but I don't think that I want it committed to Offline.  I prefer that you work with Dave to get a good solution.   

double toyDS.bz  = 0.7; // Tesla

and leave the scale factor at 1.0.  This will correctly set the field for the seed fit and I don't think that it will break anything else.

For this study do you need any field real field maps - ie are you running G4 or doing Fits with the full field in the same job.

PR #854, opened 2022-09-22
Remove indirection cache.

kutschke:
The Tracker object internally had a redirection layer to map StrawId (a non-dense index) to the dense indices of the Straw objects in the collection of all straw objects.  The original plan was that the ordering in the dense array could be arbitrary so that we could, at some future date, we could optimize ordering to minimize cache misses.

With the introduction of the conditions system, the ordering in the dense array needs to be in the order of StrawId::uniqueStraw().  This removes the need for the indirection layer and indexing by StrawId::uniqueStraw() is now hardcoded.

Therefore the indirection layer has been removed.


kutschke:
After making the changes, I verified that all of the following build: Offline, DQM, REve, TrackerAlignment, TrkAna, Tutorial.   I also used valCompare to compare 1K events of ceSimReco before and after; all histograms matched exactly.

brownd1978 (approved):

kutschke:
@FNALbuild run build test

PR #856, opened 2022-09-23
STM Waveform Class

AndrewEdmonds11:
Hi Everyone,

As I started putting together the classes that we discussed at the recent Comp-Soft meeting, I realised that it would become a very large PR so I am going to try and break it down into smallre chunks for easier reviews. Let me know if you would rather have everything in one PR so that you can see a bigger picture.

This first PR adds the ```STMWaveform``` class which represents the digitized waveform. I have also added a source module that reads the STM test beam data (```raw.mu2e.STM_ELBE_2022.*.dat```) and creates a file of STMWaveforms. I have also added the test beam raw data format (```STMTestBeamHeaders```) and a class to store event-level information from the test beam (```STMTestBeamEventInfo```, currently unused). I have also added an analyzer module that plots the ```STMWaveforms```. I will put in a Production PR with the fcl to run this analyzer module.

After this PR is merged, the next PR will add the zero-suppression algorithm. This module will also create ```STMWaveforms```.

Let me know if you have any comments or questions.

Thanks,
Andy


kutschke:
> As I started putting together the classes that we discussed at the recent Comp-Soft meeting, I realised that it would become a very large PR so I am going to try and break it down into smallre chunks for easier reviews. Let me know if you would rather have everything in one PR so that you can see a bigger picture.

I like the size of this PR.

DataProducts/inc/STMTestBeamEventInfo.hh
```diff
@@ -0,0 +1,22 @@
...
+    STMTestBeamEventInfo() : _trigType(TrigType::unknown), _triggerTime(0) { }
+
+    void trigType(uint16_t type) { _trigType = TrigType(type);}
+    void triggerTime(uint64_t time) { _triggerTime = time;}
+
+    TrigType trigType() const { return _trigType;}
+    uint64_t triggerTime() const { return _triggerTime;}
+
+  private:
+    TrigType _trigType;
```
kutschke, line 17: A style preference you might consider.  You can write these as:

TrigType _trigType = TrigType::unknown;  // or maybe just unknown ?
uint64_t _triggerTime = 0;

Then they are automatically set in every c'tor and you can get rid of the initializer list.

DataProducts/inc/STMTestBeamEventInfo.hh
```diff
@@ -0,0 +1,22 @@
...
+
+namespace mu2e {
+  class STMTestBeamEventInfo {
+  public:
+    enum TrigType { unknown=-1, internal=0, external=1 };
+    STMTestBeamEventInfo() : _trigType(TrigType::unknown), _triggerTime(0) { }
+
+    void trigType(uint16_t type) { _trigType = TrigType(type);}
+    void triggerTime(uint64_t time) { _triggerTime = time;}
+
```
kutschke, line 12: When you create an object of this type, are the two data members known at c'tor time?  If so, please provide a c'tor that takes both of these as arguments.   In general classes should only have setter functions if it's not possible to construct the object completely when the c'tor is called.

DataProducts/inc/STMTestBeamEventInfo.hh
```diff
@@ -0,0 +1,22 @@
...
+    TrigType trigType() const { return _trigType;}
+    uint64_t triggerTime() const { return _triggerTime;}
+
```
kutschke, line 15: I suggest symmetrizing the names of the accessors - ie both trig or both trigger.

DataProducts/inc/STMTestBeamHeaders.hh
```diff
@@ -0,0 +1,109 @@
...
+#define DataProducts_STMTestBeamHeaders_hh
+//
+// Header definitions for STM test beam data
+//
+
+// C++ includes
+#include <iostream>
+#include <vector>
+#include <array>
+#include <Rtypes.h>
```
kutschke, line 11: What in this file needs Rtypes?

kutschke (commented):
I have made a few comments. I need to leave now but will come back later to finish.

DataProducts/inc/STMTestBeamEventInfo.hh
AndrewEdmonds11: Done in f78aa57

DataProducts/inc/STMTestBeamEventInfo.hh
AndrewEdmonds11: I move a few things in the source module to create the object when we know what its contents will be (6d6f873)

DataProducts/inc/STMTestBeamEventInfo.hh
AndrewEdmonds11: Good idea - done in 882fe8a

DataProducts/inc/STMTestBeamHeaders.hh
```diff
@@ -0,0 +1,109 @@
...
+        os << "Correct fixed header? " << std::boolalpha << header.checkFixedHeader() << std::endl;
+        os << "Data Size: " << header.getDataSize() << " bytes" << std::endl;
+        os << "No. of Slices: " << header.getNSlices() << std::endl;
+        os << "Trigger #" << header.getTriggerNumber() << std::endl;
+        os << "\tMode: " << header.getTriggerMode() << std::endl;
+        os << "\tTime: " << header.getTriggerTime() << " ct" << std::endl;
+        os << "\tTime Offset: " << header.getTriggerOffset() << " ct" << std::endl;
+        os << "\tNo. of Dropped Packets: " << header.getNDroppedPackets() << std::endl;
+        return os;
+      }
```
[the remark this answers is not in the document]
AndrewEdmonds11: Done here: 529916e

DataProducts/inc/STMTestBeamHeaders.hh
AndrewEdmonds11: Must have come along for the ride when I copied another file. I've deleted it in this commit a19559e

DataProducts/inc/STMTestBeamHeaders.hh
```diff
@@ -0,0 +1,90 @@
+#ifndef DataProducts_STMTestBeamHeaders_hh
```
brownd1978, line 1: Is this class intended as a prototype for eventual use with full Mu2e readout of STM data?  Or is the testbeam data so specific it won't get reused?

RecoDataProducts/inc/STMWaveform.hh
```diff
@@ -0,0 +1,51 @@
+#ifndef RecoDataProducts_STMWaveform_hh
```
brownd1978, line 1: Is there a potential name clash with the raw STM waveform 'digi' object?

RecoDataProducts/inc/STMWaveform.hh
```diff
@@ -0,0 +1,51 @@
...
+
+    uint32_t trigTimeOffset() const { return _trigTimeOffset; }
+    const std::vector<int16_t>& adcs() const { return _adcs; }
+
+  private:
+    uint32_t _trigTimeOffset; // time offset from EWT? to first ADC value [ct]
+    std::vector<int16_t> _adcs; // vector of ADC values for the waveform
+  };
+
+  class STMWaveformCollection {
```
brownd1978, line 36: since STMWaveformCollection is its own class (not just a typedef) I recommend splitting to its own header file.  Also, do you ever forsee a data product that's a collection (std::vector) of STMWaveformCollection?  That could result in some awkward names.

brownd1978 (commented):
Mostly comments about names, with an eye towards future applications.

DataProducts/inc/STMTestBeamHeaders.hh
AndrewEdmonds11: I didn't intend for this to become used in the full Mu2e readout. The same is true for the STMTestBeamEventInfo class. The final data packets will look pretty similar to this but I think that will all live in the online world

RecoDataProducts/inc/STMWaveform.hh
AndrewEdmonds11: This is intended to be the raw STM waveform 'digi' object. Should I add "Digi" to the class name?

RecoDataProducts/inc/STMWaveform.hh
AndrewEdmonds11: I'm happy to move STMWaveformCollection into its own header file. I don't ever forsee us making a collection of STMWaveformCollections. I'd like to keep things fixed so that we have one STMWaveformCollection for HPGe and another for LaBr. Each will be treated slightly differently so I don't see a reason we would want to loop through a collection of collections.

RecoDataProducts/inc/STMWaveformCollection.hh
```diff
@@ -0,0 +1,28 @@
...
+    STMWaveformCollection(STMChannel ch) : _channel(ch) {}
+
+    STMChannel channel() const { return _channel; }
+
+    const size_t size() const { return _waveforms.size(); }
+    void push_back(const mu2e::STMWaveform& waveform) { _waveforms.push_back(waveform); }
+    const std::vector<mu2e::STMWaveform>& waveforms() const { return _waveforms; }
+
+  private:
+    mu2e::STMChannel _channel;
```
rlcee, line 24: Is it possible to remove this channel member and operate by identifying the channel using the "instance" field of the product name?

AndrewEdmonds11: It's definitely possible, and I'm open to changing it. My worry with identifying the channel with the instance name is that a user could misconfigure later modules.

Just to check I understand your proposal with an example. At the moment, in a module that would calibrate the energies we would have something like this:
```
void MakeSTMHits::produce(art::Event& event) {
    auto waveformsHandle = event.getValidHandle<STMWaveformCollection>(_stmWaveformsTag);
    STMEnergyCalib const& stmEnergyCalib = _stmEnergyCalib_h.get(event.id()); // get calibration
    STMChannel ch = waveformsHandle ->channel();
    const auto& pars = stmEnergyCalib.calib(ch);
    for (const auto & waveform : waveformsHandle->waveforms()) { // I don't like the way this line look
        // do calibration
    }
}
```
(Admittedly we won't be calibrating the waveforms themselves but the object that will represent the pulse area).
And then your proposal would be something like this (changed lines highlighted with ```**```):
```
void MakeSTMHits::produce(art::Event& event) {
    auto waveformsHandle = event.getValidHandle<STMWaveformCollection>(_stmWaveformsTag);
    STMEnergyCalib const& stmEnergyCalib = _stmEnergyCalib_h.get(event.id()); // get calibration
**    STMChannel ch = STMChannel(_stmWaveformsTag.instance())
    const auto& pars = stmEnergyCalib.calib(ch);
**    for (const auto & waveform : *waveformsHandle) {
        // do calibration
    }
}
```
Is that right?

I was worried about misconfiguring modules when I thought we would need to pass in an extra fcl parameter and so a user could get their wires crossed. But I think this gets around that, right?

rlcee: Yes, you are demonstrating the idea I suggested.  Please note that I am not really promoting the suggestion, more just asking if it makes sense.  I'd have to understand your larger plans to see if I might develop a stronger opinion.

AndrewEdmonds11: Thanks, Ray. I like the suggestion. One thing I realized that might get a little wordy is in the collection names later on. For example, we will have a module like ```STMZeroSuppression``` and we will have two instances of it, one for each channel. So our fcl would look something like this:

```
zeroSuppressHPGe : { module_type: STMZeroSuppression
 # HPGe specific parameters
}
zeroSuppressLaBr : { module_type: STMZeroSuppression
 # LaBr specific parameters
}
```
If we put the channel name in the instance name, then we will have output collections ```zeroSuppressHPGe:HPGe``` and ```zeroSuppressLaBr:LaBr```, which is a little unsightly but not a big deal.

Let me know if this gives you a stronger opinion, otherwise I will implement your idea (I like that we can just have a vector of STMWaveforms) and we can revisit this in future PRs when there is more of the bigger picture in place.

gaponenko:
Hi,

> 
> If we put the channel name in the instance name, then we will have 
> output collections ```zeroSuppressHPGe:HPGe``` and 
> ```zeroSuppressLaBr:LaBr```, which is a little unsightly but not a big 
> deal.

The approach that we agreed to some time ago is to only use instance names 
where they are unavoidable.  I.e. where a single module produces more than 
one output collection of the same data type.  In other cases instance 
names just add unnecessary complication.

Andrei

RecoDataProducts/inc/STMWaveformCollection.hh
rlcee: >If we put the channel name in the instance name, then we will have output collections zeroSuppressHPGe:HPGe and zeroSuppressLaBr:LaBr, which is a little unsightly but not a big deal.

this is even better - the collections are identified by the module name field of the product name, so as your example here indicated, adding an instance name is redundant and can be skipped.  Then this also satisfies Andrei's rule.   And the comment that the internal class field is not needed continues to be relevant.

AndrewEdmonds11: There are two challenges with idenitfying the channel from the module label and taking it out of the internal class field.

The first is that, to get the channel for e.g. the calibration constants, we would need to parse the module label. This isn't a big challenge (since we can write a utility that just looks at the last 4 characters) but I don't think it's ideal.

The second challenge is that, for the test beam data, I have written the source module to create collections for both detectors. I could rewrite it to only create one collection and have two fcl files (one for each detector). Then we would run over the raw HPGe or the raw LaBr data with the appropriate fcl file. This is fine for the test beam data since we only had one detector connected at a time so don't need to worry about storing data for both.  

Having said that, I don't know what the situation will be when taking data. In the online/otsdaq/artdaq world, will data products be written with different module labels? I'll have to find out...

Anyway, a potential compromise for the time being would be to identify the channel by the module label (most cases) *or* the instance name (if one exists like for the source module).

rlcee: I didn't mean to drag out this comment so much.  I don't really have an opinion, it was just a suggestion/observation.  Any time you are sure you are on the right track, you should just go with it.  I think for me to comment on your latest comments, I'd need to understand the whole processing plan.  If you would like me to do that, let me know.  I don't think I would have any unique insight, just another pair of eyes.  I'm happy to do that if it would help.

As far as I know, the trigger process will act like a normal art job and your products will belong to your module which reads fragments.  I think the datalogger can alter events, but I don't see why it would in this case. But you should  probably confirm.

RecoDataProducts/inc/STMWaveform.hh
brownd1978: I think having 'Digi' in the name would be helpful, but its up to you.

brownd1978:
@FNALbuild run build test

RecoDataProducts/inc/STMWaveform.hh
AndrewEdmonds11: I agree. Done in commit 3be8644

AndrewEdmonds11:
Hi everyone - is there a hold-up with this PR? I believe I have responded to everything but let me know if I overlooked anything. Thanks

AndrewEdmonds11:
Thanks, Rob. No time pressure - would just be nice to have this merged in so I can get a clean start on the next parts

rlcee (approved):
A few comments for thought, I don't feel strongly

brownd1978 (approved):
The latest commits address my concerns.

DataProducts/inc/STMTestBeamHeaders.hh
```diff
@@ -0,0 +1,90 @@
...
+      uint16_t datasize[2]; // data size
+      uint16_t n_slices[2]; // number of slices
+      uint16_t trigger_number[2]; // trigger number
+      uint16_t trigger_mode[1]; // trigger mode
+      uint16_t trigger_time[4]; // trigger time
+      uint16_t trigger_offset[2]; // time of first ADC value relative to trigger time
+      uint16_t n_dropped_packets[1]; // number of dropped packets
+
+      uint32_t getFixedHeader() const {
+        return ((uint32_t) deadbeef[1] << 16) | ((uint32_t) deadbeef[0]);
```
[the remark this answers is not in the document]
AndrewEdmonds11: Thanks, Ray. I've changed to c++-style casting in this commit: d4e60ae8f.

On your other question: this is the format of the data in the binary files so I guess this is more like a fragment than an offline object. This object itself isn't being stored and I'm just using this struct to help read the binary file. Let me know if there is a better place for this than DataProducts (maybe Sources or even the source module itself?)

RecoDataProducts/src/classes_def.xml
```diff
@@ -375,4 +375,10 @@
  <class name="mu2e::RecoCount"/>
  <class name="art::Wrapper<mu2e::RecoCount>"/>
 
+<!--  ********* STM ********* -->
+ <class name="mu2e::STMWaveformDigi"/>
+ <class name="std::vector<mu2e::STMWaveformDigi>"/>
```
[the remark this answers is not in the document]
AndrewEdmonds11: Removed in commit 81630348e

DataProducts/inc/STMTestBeamHeaders.hh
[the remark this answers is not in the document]
rlcee: I didn't know it is not written out, that makes the comment less important.  Is it only used in the one module?  If so, it could be specific to that module.  If it is kept separate, but is only used a temp structure in the source module, I'd probably keep it in the Sources dir, but still no strong opinion, just a thought.

Sources/src/FromSTMTestBeamData_source.cc
```diff
@@ -0,0 +1,335 @@
...
+    // Open the input file
+    currentFileName_ = filename;
+    currentFile_ = new std::ifstream(currentFileName_.c_str(), std::ios::in | std::ios::binary);
+    if (!currentFile_->is_open()) {
+      throw cet::exception("FromSTMTestBeamData") << "A problem opening binary file " << currentFileName_ << std::endl;
+    }
+
+    // Extract the run number and subrun number from the file name
+    std::string delimeter = ".";
+    size_t currentPos = 0;
```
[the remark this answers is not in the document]
rlcee: If this is just for the test beam, then OK, that can get messy, but I don't think we can use as a technique in the long term.

Sources/src/FromSTMTestBeamData_source.cc
```diff
@@ -0,0 +1,335 @@
...
+
+    if(runNumber_ < 101000 || runNumber_ > 101999) {
+      throw cet::exception("FromSTMTestBeamData") << "Run number is outside of our reserved run range (101000 -- 101999)" << std::endl;
+    }
+
+    // There are two different versions of the binary file:
+    //  - v2 contains a unix timestamp after the trigger header
+    //  - v1 does not
+    // For the time being, just hardcode which runs belong to which (ideally would be in a DB?)
+    if (runNumber_ == 101001) {
```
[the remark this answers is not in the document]
rlcee: Also here, OK for the test beam run, but not permanently, IMO.  Ideally the data would contain the version number and the code can react to that.  This is a general problem in all the data taking, but I don't know of a general rule for a solution.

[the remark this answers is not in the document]
kutschke: If you backed in a corner and need to test on run numbers, I suggest writing a small class or function to manage the tests; give the functions good names so that people can just read the code; and document that class/function well. This captures all of the hard coded run numbers in one place.

DataProducts/inc/STMTestBeamHeaders.hh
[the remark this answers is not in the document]
AndrewEdmonds11: I think it's better located in the Sources module so I've moved it in this commit: b67531583

DataProducts/inc/STMChannel.hh
```diff
@@ -107,6 +109,19 @@ namespace mu2e {
       return lastEnum;
     }
 
+    // Want to take an input tag and return an STMChannel
+    static STMChannel getChannel(art::InputTag const& tag) {
```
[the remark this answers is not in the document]
kutschke: I agree that this belongs as a free function in Mu2eUtilities not part of this class.

RecoDataProducts/inc/STMWaveformDigi.hh
```diff
@@ -0,0 +1,35 @@
...
+
+    STMWaveformDigi(uint32_t trigTimeOffset, std::vector<int16_t> adcs) : _trigTimeOffset(trigTimeOffset), _adcs(adcs) {};
+
+
+    uint32_t trigTimeOffset() const { return _trigTimeOffset; }
+    const std::vector<int16_t>& adcs() const { return _adcs; }
+
+  private:
+    uint32_t _trigTimeOffset; // time offset from EWT? to first ADC value [ct]
+    std::vector<int16_t> _adcs; // vector of ADC values for the waveform
```
kutschke, line 30: Will different digis within one event have different lengths of waveform data?

At nominal intensity, roughly how many digis do you expect for each of LaBr and HPGe?

STMReco/src/SConscript
```diff
@@ -0,0 +1,74 @@
...
+                       'tbb',
+                       'cetlib',
+                       'cetlib_except',
+                       'CLHEP',
+                       'boost_filesystem',
+                       rootlibs,
+                       'pthread'
+                     ] )
+
+# helper.make_dict_and_map( [
```
kutschke, line 59: I prefer that you remove the commented code.  If and when you might want to re-enable it, the pattern might have changed.  So you should start from a known working example.

STMReco/src/SConscript
```diff
@@ -0,0 +1,74 @@
...
+                                  'art_Utilities',
+                                  'canvas',
+                                  'cetlib',
+                                  'cetlib_except',
+                                  'CLHEP',
+                                  rootlibs
+                                ] )
+
+helper.make_plugins( [ mainlib,
+                       'mu2e_Mu2eUtilities',
```
kutschke, line 31: Have you stripped down both link lists to the minimal needed Mu2e libraries?

Sources/src/FromSTMTestBeamData_source.cc
```diff
@@ -0,0 +1,335 @@
...
+    }
+
+    // Create the STMWaveformDigiCollection that we will write to the art event
+    // NB although we are creating one collection for each detector, in the test beam only one
+    //    detector was connected at a time so one of these collections will be empty
+    std::unique_ptr<mu2e::STMWaveformDigiCollection> outputHPGeWaveformDigis(new mu2e::STMWaveformDigiCollection());
+    std::unique_ptr<mu2e::STMWaveformDigiCollection> outputLaBrWaveformDigis(new mu2e::STMWaveformDigiCollection());
+
+    // Read the trigger header
+    STMTestBeam::TriggerHeader trigger_header[1];
```
kutschke, line 224: I am curious why you made this an array of length 1 and not just a plain on object of that type.  The code as it exists now would work fine if it were just an object.  Are you preparing for a future development?

kutschke (commented):
I have looked at about half of the files carefully and here are my comments to date.  I will get back to it later in the day.

Sources/inc/STMTestBeamHeaders.hh
```diff
@@ -0,0 +1,90 @@
...
+      uint32_t getTriggerOffset() const {
+        return ( uint32_t(trigger_offset[1]) << 16 | uint32_t(trigger_offset[0]) );
+      }
+
+      uint16_t getNDroppedPackets() const {
+        return uint16_t(n_dropped_packets[0]);
+      }
+
+      bool checkFixedHeader() const {
+        return (this->getFixedHeader() == (uint32_t) 0xDEADBEEF);
```
kutschke, line 60: I prefer that you use c++ style casts, not c style.  So uint32_t(0xDEADBEEF).  There are a few other places where you do this.  Please fix them all.

Sources/src/FromSTMTestBeamData_source.cc
```diff
@@ -0,0 +1,335 @@
...
+    fhicl::Atom<std::string> module_type{Name("module_type"), Comment("Art module type"), ""};
+  };
+  typedef fhicl::WrappedTable<Config> Parameters;
+
+
+  //================================================================
+  class STMTestBeamDataDetail : private boost::noncopyable {
+    std::string myModuleLabel_;
+    art::SourceHelper const& pm_;
+    unsigned runNumber_; // from file name
```
kutschke, line 59: Write this as   ```unsigned runNumber_ = 0;```  Then it is implicitly initialized in every c'tor.  This gets the boiler plate out of the c'tor, making the c'tor shorter and easier to read.  The same for a lot of other member data.   For data members of type string, it's OK to let them default construct (if that's what you want) and not mention them in the c'tor initializer list.

Sources/src/FromSTMTestBeamData_source.cc
```diff
@@ -0,0 +1,335 @@
...
+    std::ifstream* currentFile_ = nullptr;
+
+    unsigned currentSubRunNumber_; // from file
+    unsigned currentEventNumber_;
+    unsigned maxEvents_;
+    int verbosityLevel_;
+    uint16_t channel_; // the channel (HPGe or LaBr) will be different for each file
+    int binaryFileVersion_;
+
+    int printAtEvent;
```
kutschke, line 74: Is this a leftover or something for the future.  It's not used for anything now.

Sources/src/FromSTMTestBeamData_source.cc
```diff
@@ -0,0 +1,335 @@
...
+        art::EventPrincipal*& outE);
+
+    void closeCurrentFile();
+  };
+
+  //----------------------------------------------------------------
+  STMTestBeamDataDetail::STMTestBeamDataDetail(const Parameters& conf,
+      art::ProductRegistryHelper& rh,
+      const art::SourceHelper& pm)
+    : myModuleLabel_("FromSTMTestBeamData")
```
kutschke, line 105: This should come from conf().

Sources/src/FromSTMTestBeamData_source.cc
```diff
@@ -0,0 +1,335 @@
...
+    , currentEventNumber_(0)
+    , maxEvents_(conf().maxEvents())
+    , verbosityLevel_(conf().verbosityLevel())
+    , binaryFileVersion_(0)
+  {
+    rh.reconstitutes<mu2e::STMWaveformDigiCollection,art::InEvent>(myModuleLabel_, "HPGe");
+    rh.reconstitutes<mu2e::STMWaveformDigiCollection,art::InEvent>(myModuleLabel_, "LaBr");
+    rh.reconstitutes<mu2e::STMTestBeamEventInfo,art::InEvent>(myModuleLabel_);
+
+    currentSubRunNumber_ = 0;
```
kutschke, line 119: These lines, and a lot in the initialize list can go away if you initialize in the declaration.

Sources/src/FromSTMTestBeamData_source.cc
```diff
@@ -0,0 +1,335 @@
...
+  void STMTestBeamDataDetail::readFile(const std::string& filename, art::FileBlock*& fb) {
+
+    // Open the input file
+    currentFileName_ = filename;
+    currentFile_ = new std::ifstream(currentFileName_.c_str(), std::ios::in | std::ios::binary);
+    if (!currentFile_->is_open()) {
+      throw cet::exception("FromSTMTestBeamData") << "A problem opening binary file " << currentFileName_ << std::endl;
+    }
+
+    // Extract the run number and subrun number from the file name
```
kutschke, line 137: The code to extract information from the file name is a great candidate for putting it in a small helper function or helper class.  That way the mainline of readFile reads cleanly.  If the helper classs/function is of general use, but it in Mu2eUtilities.  If it is of use only in this file, put it in an anonymous namespace at the top of the file.

Sources/src/FromSTMTestBeamData_source.cc
```diff
@@ -0,0 +1,335 @@
...
+  class STMTestBeamDataDetail : private boost::noncopyable {
+    std::string myModuleLabel_;
+    art::SourceHelper const& pm_;
+    unsigned runNumber_; // from file name
+    art::SubRunID lastSubRunID_;
+    std::set<art::SubRunID> seenSRIDs_;
+
+    unsigned currentFileNumber_;
+    std::string currentFileName_;
+    std::ifstream* currentFile_ = nullptr;
```
kutschke, line 65: You are doing everything right with this bare pointer.  I strongly recommend changing the type to
``std::unique_ptr<std:ifstream>``` .   That way a future maintainer does not need to think about calls to close or calling delete.  You can just reset the unique_ptr and both are done.  Or if you copy into the unique_ptr, the object being replaced has it's d'tor called.  That automatically calls close.   Basically this future proofs the code.

I imagine you followed a model in which it was done with bare pointers.  We should fix them all as targets of opportunity.

STMReco/src/PlotSTMWaveformDigis_module.cc
```diff
@@ -0,0 +1,70 @@
...
+    auto waveformsHandle = event.getValidHandle<STMWaveformDigiCollection>(_stmWaveformDigisTag);
+
+    std::stringstream histname;
+    int count = 0;
+    for (const auto& waveform : *waveformsHandle) {
+      histname.str("");
+      histname << "evt" << event.event() << "_waveform" << count;
+      TH1F* _hWaveform = tfs->make<TH1F>(histname.str().c_str(), "", waveform.adcs().size(),0,waveform.adcs().size());
+      int i_bin = 1;
+      for (const auto& adc : waveform.adcs()) {
```
kutschke, line 61: Another FYI.  I have no idea what the compiler is going to do with this.   My rule of thumb is that if an object is smaller than a pointer, work with it by value, not by reference or pointer.  Since an adc value is smaller than 8 bytes, I would write this without the reference.  I bet that in this case the whole discussion is moot - the code is simple enough that the optimizer will probably do the obvious thing.

kutschke:
@AndrewEdmonds11 I am finished with my comments for tihs PR.    Some are just suggestions and FYIs.  

I have a few other comments that I will save for later so that we can get this one through.   There is a standard solution to holding a vector of variable length vectors in a single contiguous chunk of memory.  I will see if I can find example code to crib from.

kutschke:
@FNALbuild run build test

Sources/src/FromSTMTestBeamData_source.cc
```diff
@@ -135,9 +135,6 @@ namespace mu2e {
   //----------------------------------------------------------------
   void STMTestBeamDataDetail::closeCurrentFile() {
     currentFileName_ = "";
-    currentFile_->close();
```
kutschke: I recommend calling currentFile->reset(); here.  That actually closes the file so that your internal state keeps in sync with the intent of the callback model.

STMReco/fcl/plotSTMWaveformDigis.fcl
```diff
@@ -4,18 +4,20 @@
...
+process_name: PlotSTMWaveformDigis
 
 source : {
   module_type : RootInput
 }
 
-services : @local::Services.Reco
+services : {
+  message : @local::Services.Core.message
+  TFileService : { }
```
kutschke, line 15: If you don't need TFileService, don't include it.

kutschke:
Thanks @AndrewEdmonds11 .  I see that your recent PR address most (maybe all?) of my comments.

Sources/src/FromSTMTestBeamData_source.cc
```diff
@@ -100,14 +100,14 @@ namespace mu2e {
   STMTestBeamDataDetail::STMTestBeamDataDetail(const Parameters& conf,
       art::ProductRegistryHelper& rh,
       const art::SourceHelper& pm)
-    : myModuleLabel_("FromSTMTestBeamData")
+    : myModuleType_(conf().module_type())
```
kutschke, line 104: Ah.  I was mislead by the variable name.  Since it's really myModuleType, it's perfectly OK to hard code the string literal here.  Your call on whether to leave it as is or change it back.

RecoDataProducts/inc/STMWaveformDigi.hh
AndrewEdmonds11: Hi Rob. For the unsuppressed waveform digis, I think they will all be the same length. For zero-suppressed waveform digis (coming in a future PR) they could be different lengths.

Based on Jijun's most recent simulations (DocDB-42969) it looks like we expect a hit rate of ~140 kHz in the LaBr, and ~25 kHz for HPGe at nominal intensity. So that would ~0.2 hits per event in the LaBr and ~0.04 hits per event in the HPGe.

STMReco/src/SConscript
AndrewEdmonds11: I've stripped the link lists down to what I can in [8ec13f4](https://github.com/Mu2e/Offline/pull/856/commits/8ec13f455a3f6f81c572e3f581b8d2b59425018e)

Sources/src/FromSTMTestBeamData_source.cc
AndrewEdmonds11: I think I was just copying an example that used an array of length 1. I have changed to a plain object in [6ceadcb](https://github.com/Mu2e/Offline/pull/856/commits/6ceadcbf6c3dca6beccfc5cb6389cbb5513019fd)

Sources/inc/STMTestBeamHeaders.hh
AndrewEdmonds11: Done in [128b677](https://github.com/Mu2e/Offline/pull/856/commits/128b677e431aaf6953d160d90ef0a47f6c9383ae). I couldn't find a way to do a c++-style cast for the (char*)s that I use in FromSTMTestBeamData_source.cc though

Sources/src/FromSTMTestBeamData_source.cc
AndrewEdmonds11: Done in [86c7106](https://github.com/Mu2e/Offline/pull/856/commits/86c7106ce0d9319e26f96d34d4fdbda49c37ab29)

Sources/src/FromSTMTestBeamData_source.cc
AndrewEdmonds11: Done in [86c7106](https://github.com/Mu2e/Offline/pull/856/commits/86c7106ce0d9319e26f96d34d4fdbda49c37ab29)

DataProducts/inc/STMTestBeamHeaders.hh
kutschke: Thanks.  I will follow up on this outside of this ticket.

DataProducts/inc/STMChannel.hh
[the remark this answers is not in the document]
kutschke: @AndrewEdmonds11 Please make this change too.  We want to keep data product classes as simple as possible.

Sources/src/FromSTMTestBeamData_source.cc
AndrewEdmonds11: It was leftover from the example I copied. Removed in [a6ab5ad](https://github.com/Mu2e/Offline/pull/856/commits/a6ab5ad6fbacb45333d41f5be57795c8eaa35c84)

Sources/src/FromSTMTestBeamData_source.cc
AndrewEdmonds11: Done in [51c5470](https://github.com/Mu2e/Offline/pull/856/commits/51c5470641a24465142c7656e928bcb85c0f6b77) and renamed myModuleLabel to myModuleType since the module label is "source" and the module type is "FromSTMTestBeamData"

Sources/src/FromSTMTestBeamData_source.cc
AndrewEdmonds11: Done in [b453042](https://github.com/Mu2e/Offline/pull/856/commits/b453042068b1c4de70437c348d074f87485acb2d). This won't be generally useful since it is just the binary test beam filename that we need to decipher but I have put it in its own class in Sources

Sources/src/FromSTMTestBeamData_source.cc
```diff
@@ -0,0 +1,335 @@
...
+      binaryFileVersion_ = 1;
+    }
+    else {
+      binaryFileVersion_ = 2;
+    }
+
+    std::string subrunNo = sequencer.substr(7, 7+8); // sequencer always has 6 characters for run followed by an underscore and then 8 characters for the sub run
+    currentSubRunNumber_ = std::stoi(subrunNo);
+    currentEventNumber_ = 1;
+    fb = new art::FileBlock(art::FileFormatVersion(1, "STMTestBeamDataInput"), currentFileName_);
```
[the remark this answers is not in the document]
AndrewEdmonds11: Added the comment in [2286a09](https://github.com/Mu2e/Offline/pull/856/commits/2286a0936d2f2e477a572b27fcc5c0b471651e7e)

Sources/src/FromSTMTestBeamData_source.cc
AndrewEdmonds11: Done here [2d0e936](https://github.com/Mu2e/Offline/pull/856/commits/2d0e936dcc79cfd39f811ed2a179cbede517c07f)

Sources/src/FromSTMTestBeamData_source.cc
```diff
@@ -0,0 +1,335 @@
...
+          else {
+            throw cet::exception("FromSTMTestBeamData") << "Trying to create a waveform with an invalid STMChannel (" << channel_ << ")" << std::endl;
+          }
+        }
+        else { return false; }
+      }
+    }
+    else { return false; }
+
+    art::put_product_in_principal(std::move(outputHPGeWaveformDigis), *outE, myModuleLabel_, "HPGe");
```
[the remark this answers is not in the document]
AndrewEdmonds11: Hi Rob, I think downstream we want to assume that both HPGe and LaBr collections are in the event (even if they are empty). This should allow us to have one fcl file for processing so that if one detector is physically off that fcl file will still run

Sources/test/FromSTMTestBeamData.fcl
```diff
@@ -0,0 +1,36 @@
...
+     module_type : RootOutput
+     fileName    : @nil
+     outputCommands : [ "keep *_*_*_*"]
+  }
+}
+
+physics.end_paths : [ EndPath ] # needed for generate_fcl
+outputs.Output.fileName: "dig.owner.description.version.sequencer.art"
+# no histogram output
+services.TFileService.fileName : "/dev/null"
```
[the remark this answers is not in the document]
AndrewEdmonds11: This line is gone when I reduce the number of services I create in [c227f96](https://github.com/Mu2e/Offline/pull/856/commits/c227f96627e412e064b4f900c16a1c7e18d18479)

Sources/test/FromSTMTestBeamData.fcl
```diff
@@ -0,0 +1,36 @@
...
+
+#include "Offline/fcl/standardServices.fcl"
+
+process_name: STMTestBeamData
+source : {
+  module_type : FromSTMTestBeamData
+  verbosityLevel : 0 # or 1
+}
+
+services : @local::Services.Core
```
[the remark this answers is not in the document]
AndrewEdmonds11: Done in [c227f96](https://github.com/Mu2e/Offline/pull/856/commits/c227f96627e412e064b4f900c16a1c7e18d18479)

STMReco/src/PlotSTMWaveformDigis_module.cc
```diff
@@ -0,0 +1,70 @@
...
+#include "Offline/RecoDataProducts/inc/STMWaveformDigi.hh"
+
+namespace mu2e {
+
+  class PlotSTMWaveformDigis : public art::EDAnalyzer {
+    public:
+      using Name=fhicl::Name;
+      using Comment=fhicl::Comment;
+      struct Config {
+        fhicl::Atom<art::InputTag> stmWaveformDigisTag{ Name("stmWaveformDigisTag"), Comment("InputTag for STMWaveformDigiCollection")};
```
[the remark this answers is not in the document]
AndrewEdmonds11: Thanks, Rob. I'd like to be specific here

STMReco/src/PlotSTMWaveformDigis_module.cc
```diff
@@ -0,0 +1,70 @@
...
+
+    art::ServiceHandle<art::TFileService> tfs;
+    auto waveformsHandle = event.getValidHandle<STMWaveformDigiCollection>(_stmWaveformDigisTag);
+
+    std::stringstream histname;
+    int count = 0;
+    for (const auto& waveform : *waveformsHandle) {
+      histname.str("");
+      histname << "evt" << event.event() << "_waveform" << count;
+      TH1F* _hWaveform = tfs->make<TH1F>(histname.str().c_str(), "", waveform.adcs().size(),0,waveform.adcs().size());
```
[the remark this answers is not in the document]
AndrewEdmonds11: I agree - done here [8cec6f9](https://github.com/Mu2e/Offline/pull/856/commits/8cec6f9b41614dba845fab107fa1189950638ebb)

STMReco/src/PlotSTMWaveformDigis_module.cc
AndrewEdmonds11: I've decided to keep this as it is. Hopefully the compiler is optimizing it. If we see performance issues, I can change it

DataProducts/inc/STMChannel.hh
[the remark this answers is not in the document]
AndrewEdmonds11: Yep, done in [111b4cd](https://github.com/Mu2e/Offline/pull/856/commits/111b4cdc188fc42704636bca1e900716b328dc58)

Sources/inc/STMTestBeamHeaders.hh
AndrewEdmonds11: I tried static_cast and dynamic_cast. Let me see if reinterpret_cast works

AndrewEdmonds11: reinterpret_cast worked. Done in 8e33f68eb

Sources/src/FromSTMTestBeamData_source.cc
AndrewEdmonds11: Thanks, Rob. Done in 6d6022a85

STMReco/fcl/plotSTMWaveformDigis.fcl
AndrewEdmonds11: For plotSTMWaveformDigis.fcl, I do use but I define the file name later. In commit e102cd663 I moved the file name into this block to make it obvious

Sources/src/FromSTMTestBeamData_source.cc
AndrewEdmonds11: Thanks, Rob. I will leave it as it is for now

Sources/src/FromSTMTestBeamData_source.cc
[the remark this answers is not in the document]
AndrewEdmonds11: I agree that this is not a permanent solution. In the future we will hopefully be reading art files as our inputs and not our own binary files (which had a unix time stamp added during the test beam)

Sources/src/FromSTMTestBeamData_source.cc
```diff
@@ -0,0 +1,335 @@
...
+    std::string sequencer = tokens.at(4);
+    std::string runNo = sequencer.substr(0, 6); // sequencer always has 6 characters for run
+    runNumber_ = std::stoi(runNo);
+
+    if(!art::RunID(runNumber_).isValid()) {
+      throw cet::exception("BADCONFIG", " FromSTMTestBeamData: ")
+        << " fhicl::ParameterSet specifies an invalid runNumber = "<<runNumber_<<"\n";
+    }
+
+    if(runNumber_ < 101000 || runNumber_ > 101999) {
```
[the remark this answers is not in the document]
AndrewEdmonds11: I just added this check so that we keep within our allowed run numbers for the test beam data. In the future, we won't need this source module and all the run numbering will be done with art and the run database

AndrewEdmonds11:
I think I've responded to all the comments but let me know if there are any I missed. Thanks!

kutschke:
> I think I've responded to all the comments but let me know if there are any I missed. Thanks!

I am looking at them now.

@FNALbuild run build test

kutschke:
@FNALbuild run build test

kutschke:
@AndrewEdmonds11 The compiler error is that in the loop over adcs, the type of i_adc needs to be size_t.  Once you fix it, I will rerun the CI.  If it passes we are done.

kutschke (approved):
There is a trivial compiler error to fix.  Otherwise this is ready to be merged.

AndrewEdmonds11:
Thanks, Rob. I've fixed the compilation error.

kutschke:
@FNALbuild run build test

PR #858, opened 2022-09-27
remove meco interpolation style

rlcee:
- work on Issue #766 
- validated with ceSimReco, potSim and cosmicSimReco exact matches
- no change expected in nightly validation


kutschke (approved):
Thanks for doing this.

PR #860, opened 2022-09-29
remove MECO field map

rlcee:
Remove obsolete MECO style B field map.  Tested with ceSimReco, potSim and cosmicsSimReco.  Expect no change in validation. Work on #766


BFieldGeom/inc/BFMapType.hh
```diff
@@ -27,11 +27,11 @@ namespace mu2e {
 
     class BFMapType {
        public:
-        enum enum_type { unknown, GMC, G4BL, PARAM, lastEnum };
+        enum enum_type { unknown, unused, G4BL, PARAM, lastEnum };
```
kutschke, line 30: This class is not persisted in any data product.  So it is safe to remove unused entries instead of renaming them to unused.

rlcee: Done

kutschke (approved):

kutschke:
@FNALbuild run build test

PR #861, opened 2022-09-29
Upgrade PDGCode and add helper functions.

kutschke:
Change the PDGCode enum to be a class using the EnumToStringSparse class template.

Add a new set of utility functions for reading enum types as strings and returning enum type objects; one is a class template that will work for any of our enum types that are constructable from string.

Add a c'tor from string to GenId.hh

Tweaks to the dictionary makers to accommodate this.

I have verified that there are exact matches for 1k events in each of ceSimReco.fcl and reco.fcl.


rlcee (approved):
All looks good.

One minor comment, up to you..
Would it be a little cleaner appearance to initialize the map without std::make_pair, like
` ... nam {
{PDGCodeDetail::d,            "d"},
{PDGCODEDetail::anti_d,   "anti_d"}
};`

kutschke:
> One minor comment, up to you.. Would it be a little cleaner appearance to initialize the map without std::make_pair, like ` ... nam { {PDGCodeDetail::d, "d"}, {PDGCODEDetail::anti_d, "anti_d"} };`

Good idea.  I'm doing it now.  Will push after tests.

kutschke:
> One minor comment, up to you.. Would it be a little cleaner appearance to initialize the map without std::make_pair, like ` ... nam { {PDGCodeDetail::d, "d"}, {PDGCODEDetail::anti_d, "anti_d"} };`

Done.

kutschke:
@FNALbuild run build test

MCDataProducts/inc/GenId.hh
```diff
@@ -72,6 +72,12 @@ namespace mu2e {
       _id(id)
     {}
 
+    // c'tor from a string.
+    // If the name is not recognzied it returns the value "unknown"
```
gaponenko, line 76: Is there a strong reason to allow garbage value propagate downstream instead of throwing an exception as soon as a problem is detected?

Mu2eUtilities/inc/fromStrings.hh
```diff
@@ -0,0 +1,59 @@
...
+  // The functions belows will allow you to specify PDGCodes using the name or the alias.
+
+  // Optimization to avoid repeated creatation of Handles to ParticleDataList
+  mu2e::PDGCode::type PDGCodefromString( std::string const& s, mu2e::ParticleDataList const& pdt ){
+    mu2e::ParticleData const& p = pdt.particle(s);
+    return (static_cast<mu2e::PDGCode::type>(p.id()));
+  }
+
+  mu2e::PDGCode::type PDGCodefromString( std::string const& s ){
+    mu2e::GlobalConstantsHandle<mu2e::ParticleDataList> pdt;
```
gaponenko, line 42: I thought pulling servce handles "out of air" was deprecated in art since it became multithreaded.   If I remember this right, should we avoid doing this in our new code?  Please comment.

Mu2eUtilities/inc/fromStrings.hh
```diff
@@ -0,0 +1,59 @@
...
+    return PDGCodefromString( s, *pdt);
+  }
+
+  // Specialziation of the templated function fromStrings for PDGCode::type.
+  // Cannot do this PDGCode without maintenance on that class.
+  std::vector<mu2e::PDGCode::type> PDGCodesfromStrings( std::vector<std::string> const& v){
+    mu2e::GlobalConstantsHandle<mu2e::ParticleDataList> pdt;
```
gaponenko, line 49: Another service handle creation outside of a module.

gaponenko (commented):
Hi and sorry for taking so long to review.  I have a few questions, please see below.

brownd1978 (commented):
Has this change been validated to be backwards-compatible when reading PDGCode in existing data products?

kutschke:
@gaponenko  and @rlcee a reminder that this PR is in your input queue.

gaponenko:
I did not get answers to my questions on this thread,  posted a week ago.

Andrei

kutschke:
> I did not get answers to my questions on this thread, posted a week ago. Andrei
> […](#)

Thanks for reminding me.  They are still in my vacation backlog.

MCDataProducts/inc/GenId.hh
kutschke: Good catch.  Just pushed the fix.  The fix is inside GenId::findByName which is used in several other places in the code base.  Those places previously propagated a value of unknown when it encountered an unrecognized input.  Now those places will throw an exception too.  I don't know of a good way to test that these still work but I looked at them all and they were not doing any checking for validity on their own.  So this is very likely an improvement everywhere.

kutschke:
> Has this change been validated to be backwards-compatible when reading PDGCode in existing data products?

I used new code to run val.fcl on an input file made by old code.  Then I used valCompare and got an exact match.

kutschke:
@FNALbuild run build test

kutschke:
Following a discussion off of GitHub we decided, that for PDGCode, we would support only the names as given in PDGCode.hh; we will not support the alternate names supported by ParticleDataList.  So this PR is now just the free function template plus a free function to handle the special case of returning ```vector<PDGCode::type>```, not ```vector<PDGCode>```; this is useful for backwards compatibiity. We can decide later if we want to update the old code and remove this function.

kutschke:
@FNALbuild run build test

brownd1978 (approved):

kutschke:
@gaponenko I think that I have answered all of your questions.  There is no more pulling handles out of thin air.

gaponenko (approved):
Rob, thank you for making the changes!
Andrei

PR #863, opened 2022-10-04
Add TrackerBFieldInfo.

kutschke:
The the comments in  Mu2eUtilities/inc/TrackerBFieldInfo.hh for usage.  The class is designed so that we can do transformations to other coordinate systems in the class in order to clean up client code.

### DIFF
--- a/Mu2eUtilities/inc/TrackerBFieldInfo.hh
+++ b/Mu2eUtilities/inc/TrackerBFieldInfo.hh
@@ -13,8 +13,8 @@
 //
 // To use this to look at limits of a map:
 // TrackerBFieldInfo info
-//  cout << info.xmin() << " " << info.xmax() << endl;  // and so on
-//  cout << info.getKey() << endl;                      // name of the map that was selected
+//  cout << info.map().xmin() << " " << info.map().xmax() << endl;  // and so on
+//  cout << info.map().getKey() << endl;                            // name of the map that was selected
 //
 // Original author Rob Kutschke
 //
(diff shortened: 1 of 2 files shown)


brownd1978 (commented):
Is this just a draft?  I don't see much content.

kutschke:
> Is this just a draft? I don't see much content.

This is it.  It solves the problem of finding the right map.   Look at the comments on lines 14 to 17 of the header file.  If you want me to forward the xmin() xmax() etc functions I can do it quickly this evening. 

Other things that I could add quickly if you are inrerested are:

1. zmin and zmax in tracker coordinates
2. radius of the cylinder that's inscribed inside the rectangular prism on which the map is defined

I was not sure what you wanted and did not want to lard it up with accessors that would never be used.

Mu2eUtilities/inc/TrackerBFieldInfo.hh
```diff
@@ -0,0 +1,44 @@
...
+// I made this a class so that we can add other accessors that
+// transform the information received from the map.
+//
+// If a map cannot be located it throws an exception.
+// If there is more than one map among the BField innerMaps, it
+// also throws and exception.
+//
+// To use this to look at limits of a map:
+// TrackerBFieldInfo info
+//  cout << info.xmin() << " " << info.xmax() << endl;  // and so on
```
brownd1978, line 16: I am confused: the comment speaks of an 'xmin' function in TrackerBFieldInfo, but I don't see that defined or implemented in this PR.  Did you perhaps forget to push recent changes?

kutschke:
ooops 

info.map().xmin() etc

Do you want to zoom about this to resolve it quickly?

Mu2eUtilities/inc/TrackerBFieldInfo.hh
kutschke: oops - I replied in the general comment box

kutschke: I updated the comments.

brownd1978 (approved):

PR #867, opened 2022-10-11
Improved neural net updaters

brownd1978:
Improved training and configuration of straw hit updater neural nets used in the KinKal fit.


brownd1978:
@FNALbuild run build test

rlcee (approved):
As usual, I can't effectively comment on the larger structures or algorithms, so I spent an hour looking for anything unexpected and didn't find anything worth commenting on.

PR #875, opened 2022-10-26
Fixes and improvements to drift residuals

brownd1978:
This PR also adds TOT data to ComboHit TrkStrawHitSeed to facilitate calibration using drift fits.


Mu2eKinKal/inc/KKFit.hh
```diff
@@ -452,10 +456,14 @@ namespace mu2e {
...
       if(kktrk.fitStatus().usable()) {
-        // compute unbiased residuals
-        udres = strawhit->residual(Mu2eKinKal::dresid);
-        if(strawhit->refResidual(Mu2eKinKal::tresid).active()) utres = strawhit->residual(Mu2eKinKal::tresid);
+        try {
+          udres = strawhit->residual(Mu2eKinKal::dresid);
+          utres = strawhit->residual(Mu2eKinKal::tresid);
+        } catch (std::exception const& error) {
+          std::cout << "Unbiased residual calculation failure, nDOF = " << fstatus.chisq_.nDOF();
+        }
```
bonventre, line 466: Is there any flag to know that this failed?

brownd1978: If the residual calculation throws, the Residual object will be 'inactive', and the variances will be negative.  I should note this is an extremely rare condition, when the fit has just enough coverage to converge, so that removing even 1 hit (which the unbiased residual calculation effectively does) leaves the fit undercovered for some parameter, resulting in an unphysical covariance matrix.

kutschke:
@FNALbuild run build test

bonventre (approved):

PR #877, opened 2022-10-27
added functions to McUtils

gianipez:


TrkReco/src/TrkRecoMcUtils_tool.cc
```diff
@@ -29,229 +29,214 @@
...
-int TrkRecoMcUtils::getPdgID(const SimParticle* Sim) { return Sim->pdgId(); }
+  //-----------------------------------------------------------------------------
+  const XYZVectorF* TrkRecoMcUtils::getMom(const art::Event* Event, int HitIndex) {
 
-//-----------------------------------------------------------------------------
-float TrkRecoMcUtils::getStartMom(const SimParticle* Sim) {
-  CLHEP::HepLorentzVector const& p = Sim->startMomentum();
-  return sqrt(p.x() * p.x() + p.y() * p.y() + p.z() * p.z());
-}
+    if (Event->event() != _lastEvent) initEvent(Event);
```
kutschke, line 224: This test is not robust.  You should compare Event->id();  .  In a sparse skim in which there are sequential events from different run/subrun but with the same event number, you will not detect that you have a new event and will fail to call initEvent.  I see that this is ubiquitous in this file.  Please fix this - it's straightforward.

kutschke (changes requested):
Please let me know if the following is correct.  This PR has 3 things.

1. Add one new member function to base and derived class - this is the thing that Matthew needs, right?
2. Remove 3 member functions in the derived classs that are not in the concrete class
3. Lots of code reformatted with no functional change

Is that right?  In the future please submit reformatting changes on their own, not mixed with real changes.

I also found an error in the code that I would like you to fix before I merge.

TrkReco/src/TrkRecoMcUtils_tool.cc
gianipez: done

gianipez:
> Please let me know if the following is correct. This PR has 3 things.
> 
> 1. Add one new member function to base and derived class - this is the thing that Matthew needs, right?
> 2. Remove 3 member functions in the derived classs that are not in the concrete class
> 3. Lots of code reformatted with no functional change
> 
> Is that right? In the future please submit reformatting changes on their own, not mixed with real changes.
> 
> I also found an error in the code that I would like you to fix before I merge.

Hello Rob,

I didn't remove any function. I'm not sure why the web interface messed up the changes introduced with the indentation (which I didn't change). I just added a new function.

TrkReco/src/TrkRecoMcUtils_tool.cc
```diff
@@ -29,229 +29,214 @@
...
+      float getStartMom(const SimParticle* Sim) override;
+
+  };
+
+  //-----------------------------------------------------------------------------
+  TrkRecoMcUtils::TrkRecoMcUtils(const fhicl::ParameterSet& PSet) :
+    _comboHitCollTag   { PSet.get<std::string>("comboHitCollTag"   ) },
+    _strawDigiMCCollTag{ PSet.get<std::string>("strawDigiMCCollTag") }
+  {
+    _lastEvent   = art::EventID(0,0,0);
```
kutschke, line 80: This will compile but will cause a run time error.  Just default construct it.  Internally the art::EventID has a reserved values for each field meaning "invalid".   When you default construct it, each field gets it's invalid value.  No legal event ID can have any field invalid.  So they have already covered this case for you.

kutschke (changes requested):
Thanks for making the change everywhere.

There is one important detail you need to fix.

kutschke:
> I didn't remove any function. I'm not sure why the web interface messed up the changes introduced with the indentation (which I didn't change). I just added a new function.

I clone both repos and did a test merge.  I confirm what's shown in the diff listing on github.  You have removed 3 member functions from the derived class:

```
 const XYZVectorF getMomentum(const art::Event* Event, int Index);
 const CLHEP::Hep3Vector getPosition(const art::Event* Event, int Index);
 const CLHEP::Hep3Vector getStartPosition(const SimParticle* Sim);
```

Both the declaration and the definition are removed.

There was a typo in my earlier email - there are removed from the derived class.  They were never present in the base class. So there was never any way for them to be visible to end users, who can only see the base interface, not the derived interface.

kutschke:
@FNALbuild run build test

kutschke (approved):
Thanks for fixing the initialization.

I plan to merge at 11:45 AM today ( about 45 min from now).  If anyone objects, let me know.

PR #878, opened 2022-10-27
example of using strings for PDGid

rlcee:


kutschke (changes requested):
Please also update: 
    Offline/Analyses/test/vd_readStage2.fcl
    Production/JobConfig/pbar/vertex_s1.fcl
     - in this file, there are two savePDG entries, only one is for ReadVirtualDetector 

I looked at the following other repos and do not see "savePDG" at all: TrkAna, TrackerCalibration, TrackerAlignment, DQM, Tutorial, REve

In a separate PR can you look for an opportunity to exercise fromStrings?

kutschke:
@FNALbuild run build test

kutschke (approved):
Good catch on the tvdDropPDG in vd_reaadStage2.fcl .

PR #880, opened 2022-10-28
start STMWaveforms

rlcee:
Print and val routines for new STM product

### DIFF
--- a/Print/fcl/printLong.fcl
+++ b/Print/fcl/printLong.fcl
@@ -156,6 +156,11 @@ physics :{
         #inputTags : [ "KFFDeM" ] # select instances
       }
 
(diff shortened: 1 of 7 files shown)


AndrewEdmonds11 (approved):
Ray pointed me to an example Validation file and Print output. It looks great to me

kutschke:
No idea why CI did not run.  Doing it now.

@FNALbuild run build test

kutschke:
... I also see that the first buildbot message that announces the proposed changes did not run until after my recent comment

kutschke:
Helenka says that the problem with the CI not running appears to be glitch on the GitHub side.  Let's watch to see if it repeats on other PRs.

PR #881, opened 2022-10-31
align PDGCode names with geant

rlcee:
- remove p_plus in favor of proton (both were in PDGCode, but proton which was effectively disabled)
- remove null in favor of unknown
- bring PDGCode number and names into alignment with ParticleDataList second column, which is based on geant. Most notably, this changes the sign of some anti particles.  For example, anti of 3112 Sigma_minus is anti_Sigma_minus (was plus)
- example of something which is listed in PDG, in PDGCode, but not in ParticleDataList: 200553 Upsilon(3S).
- added a few that were in ParticleDataList but missing from PDGCode (ex. eta_b) (more remain missing)
- there are others in PDGCode and missing from ParticleDataList, like a00 10111
- corrected a few apparent typos, ex. Xi_primeanti__c0 ->Xi_prime_anti_c0
- there remain a dozen or so plus/minus name changes to make
however this is getting unwieldy so that will come in a second PR.


kutschke:
@gaponenko I added you as a reviewer to look at Sources/src/ExtMonFNALMARSUtils.cc.  Feel free to look at other parts of the PR.  I will send an email shortly with a discussion about the changes.

kutschke:
@FNALbuild run build test

rlcee:
I removed some parts of this PR, as discussed in the 11/2 comp/soft meeting.  I now am thinking doing this cleanup in maybe 4 small parts.
@FNALbuild run build test

kutschke:
When Ray refactored this PR, he removed the code that I wanted Andrei to review.  So I will remove him as a reviewer.

kutschke (approved):
One note for the record:  BTrk/TrkBase/TrkParticle.hh also defines enum values for p_plus and anti_p_minus.  I think this is OK since BTrk is being phased out and I don't see any places that compare a TrkParticle::type with PDGCode::enum_type or PDGCode::type.  I don't see a similar use of names inside KinKal.

I am approving this PR and merging.

PR #887, opened 2022-11-05
Calofastfix

brownd1978:
Add a time offset to the CaloClusters produced by CaloClusterFast, to make them consistent with the time of clusters reconstructed by CaloClusterMaker.  Disable the ad-hoc downstream corrections for cluster time offset (61ns!!!) in downstream trigger config.  With these changes the trigger configuration of the KinKal fit converges and produces good fits.  This addresses issue 884.  I'm sure the smart people working on the calorimeter code can do a better job of this eventually by getting the value from the conditions, this is just a patch to allow further progress with KinKal integration.

### DIFF
--- a/CalPatRec/fcl/prolog_trigger.fcl
+++ b/CalPatRec/fcl/prolog_trigger.fcl
@@ -111,7 +111,8 @@ CprTrigger : { @table::CprTrigger
             BackgroundSelectionBits                     : ["Background"]
             minClusterEnergy                            : 50.    # MeV
             minClusterSize                              : 2      # number of crystals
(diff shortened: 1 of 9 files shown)


kutschke:
@FNALbuild run build test

bechenard (approved):
I have a solution that is m ore robust, but that will do for the moment. Leave a note somewhere so that I know to get back to this.

kutschke:
> I have a solution that is m ore robust, but that will do for the moment. Leave a note somewhere so that I know to get back to this.

I created issue #888 to remind Bertrand to provide the better solution.  Merging now.

PR #894, opened 2022-11-19
add geant name

rlcee:
This is the second in 4 commits to implement the plan in doc 43671. The 4 steps are

1. fix duplicates
2. add geant name
3. incorporate difference between PDG and geant names
4. additional cleanup to get closer to PDG list


kutschke (approved):
Looks good.

PR #897, opened 2022-11-25
add geant names

rlcee:
- add geant particle names that are different from PDG
- fix one typo in a name


kutschke (approved):
I checked that all changes match what we discussed.  I did not check to see if any were missed.

PR #905, opened 2022-11-30
added calibration proditions to CRV reconstruction

ehrlich-uva:
<ul>
<li>added calibration proditions to crv reconstruction</li>
<li>bug fix in crv photon generator</li>
<li>added prolog_extracted_v02.fcl</li>
</ul>
This pull request should not change the results of the validation scripts. It re-uses the previous calibration constant.


rlcee (approved):
Looks good, I'll approve, with thoughts for your consideration.  

This is a small technical point.  Since the conditions can change on each event (in highly-skimmed datasets, literally every event) we recommend not caching the calibration table anywhere, but always passing it as CrvCalib&. It looks like you re-cache it on every event, so this is technically OK, but still passing it along emphasizes its volatile nature.  I didn't track the display call chain, but if it is really inconvenient to always pass it, then you can get the table with _calib_h.getPtr() and that gives you a shared pointer to the table, to pass and cache which is a little better than holding a bare pointer.

The long term plan is to remove ConditionsService and it looks like CRV is down one parameter now, which maybe can come from DAQConditions?  If so, CRV could be out of the ConditionsService.  (Obviously this is not required to be part of this PR.)

ehrlich-uva:
@rlcee 
CRVCalib gets updated at every event in the EventDisplay and other modules where I use it, i.e. it is not cached anywhere. I will make a change in the EventDisplay to use _calib_h.getPtr().
The only remaining item for the CRV in ConditionsService is the constant digitization period (independent of time and channel number). It is used in the MC and reco part. Can DAQConditions be used for MC and reco?

rlcee:
> @rlcee CRVCalib gets updated at every event in the EventDisplay and other modules where I use it, i.e. it is not cached anywhere. I will make a change in the EventDisplay to use _calib_h.getPtr(). The only remaining item for the CRV in ConditionsService is the constant digitization period (independent of time and channel number). It is used in the MC and reco part. 

Doesn't it get cached here, in DataInterface?
void setCRVCalib(const mu2e::CRVCalib &calib) {_calib=&calib;}

> Can DAQConditions be used for MC and reco?
Dave and Richie set up the values in DAQConditions, so they'd have to work with you in this.

ehrlich-uva:
> Doesn't it get cached here, in DataInterface? void setCRVCalib(const mu2e::CRVCalib &calib) {_calib=&calib;}

This function gets called at every new event (with a new CRVCalib). I will still switch to using the smart pointer for _calib_h as you suggested.

PR #914, opened 2022-12-22
remove site_tools

rlcee:
no longer needed since scons v4 includes this function and Muse calls it from there


kutschke (approved):
Looks good.

PR #915, opened 2023-01-03
Enabling the use of Geant4 11.1 etc.

resnegfk:
This pull request enables the use of Geant4 11.1
It introduces two optional parameters which affect EM energy fluctuations
(their functionality, by definition of them being optional, is disabled by default)


kutschke (approved):
This looks good to me.

PR #918, opened 2023-01-11
inverted the search direction of the calo-seeded pattern-reco alg

gianipez:
Now calPatRec starts from the tracking face closer to the calorimeter for seeding the helix search


kutschke (approved):
I am approving this and will merge.

A suggestion for the future.  Unless you think that the decision about the direction is final, it would be a good idea to refactor this to make the direction a parameter.

For future reference, it would be wise to parameterize this

PR #919, opened 2023-01-11
Remove +prof/+debug from xerces_c in the table file, as it no longer

eflumerf:
uses those qualifiers.

Commit e483353 also tagged with d10_20_00 for use in the online, using artdaq_core_mu2e v1_08_04.


eflumerf:
[d004.txt](https://github.com/Mu2e/Offline/files/10396439/d004.txt)

For reference, here is the "d004" environment I used for building the offline d10_20_00 UPS products for TDAQ use

kutschke (approved):
This looks correct to me.  I will talk with you separately about how to reengineer this to use Muse and thereby avoid multiple points of maintenance.  

This change has no coverage in the nightly tests.

PR #921, opened 2023-01-17
Move STM Sampling Frequency to Proditions

AndrewEdmonds11:
After discussing with Ray, we decided that the sampling frequency parameter that is used in the STMZeroSuppression module should move to the ProditionsService for the time being. We also decided that this would just be defined in fcl and not yet put in a database table. At the moment, the STM test beam data has a different sampling frequency for different detectors; in the real experiment, these will be the same.

### DIFF
--- a/STMConditions/fcl/prolog.fcl
+++ b/STMConditions/fcl/prolog.fcl
@@ -4,6 +4,7 @@ STMEnergyCalib : {
    useDb : false
    verbose : 0
    pedestals : [ ["HPGe",0.0], ["LaBr",0.0] ]
+   samplingFrequencies : [ ["HPGe",1.0], ["LaBr",1.0] ]
 }
 
 END_PROLOG
--- a/STMConditions/inc/STMEnergyCalib.hh
+++ b/STMConditions/inc/STMEnergyCalib.hh
@@ -20,9 +20,10 @@ class STMEnergyCalib : virtual public ProditionsEntity {
 
   typedef std::map<STMChannel, STMEnergyCorr> CalibMap;
   typedef std::map<STMChannel, float> PedestalMap;
+  typedef std::map<STMChannel, float> SamplingFrequencyMap;
 
-  STMEnergyCalib(CalibMap const& cmap, PedestalMap const& pmap) :
-      ProditionsEntity(cxname), _cmap(cmap), _pmap(pmap) {}
+  STMEnergyCalib(CalibMap const& cmap, PedestalMap const& pmap, SamplingFrequencyMap const& sfmap) :
+    ProditionsEntity(cxname), _cmap(cmap), _pmap(pmap), _sfmap(sfmap) {}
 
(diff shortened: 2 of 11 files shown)


kutschke:
The CI did not auto-run.  Launching it by hand.

@FNALbuild run build test

kutschke:
@rlcee a gentle reminder that this review is in your input queue.

kutschke:
@FNALbuild run build test

rlcee (approved):
One small comment.  In the zero suppression module, you derive a few quantities from the proditions, then store them as module members.  This pattern sort-of implies that they have a lifetime longer the event, which they shall not (a strong requirement), so ideally they would be local variables.  Also note that they can be provided, and/or precomputed/stored, as you need them, by the prodition class.  Then, if you need to pass them to other methods, you just pass the proditions class smart pointer.  Conceptually, the prodition is the source of everything time-dependent, including derived quantities.

kutschke:
@AndrewEdmonds11 Please address Ray's comments in a future PR.  The Prodition system knows the interval of validity of the quantities that it manages and it recomputes them as needed.  That way the module code does not need to be aware of intverals of validity.

AndrewEdmonds11:
Thanks for the suggestion. One follow up question: in these modules I also always convert from the sampling frequency to the nsPerClockTick. I remember there was some discussion about using the prodition to get derived quantities but I'm not sure where that landed. Can I put in a function ```STMEnergyCalib::ctPerNs(STMChannel)``` which would internally convert the sampling frequency to the nsPerClockTick, or should I add a function to STMUtils to do that?

rlcee:
> Thanks for the suggestion. One follow up question: in these modules I also always convert from the sampling frequency to the nsPerClockTick. I remember there was some discussion about using the prodition to get derived quantities but I'm not sure where that landed. Can I put in a function `STMEnergyCalib::ctPerNs(STMChannel)` which would internally convert the sampling frequency to the nsPerClockTick, or should I add a function to STMUtils to do that?

I think the best approach would be for the proditions object to hold or compute any quantities derived from time-dependent database values, like your ctPerNs method.  But I do think you could take it too far, for example, all tracks are technically derived in part from the database values.  So there is some logical limit where it would make sense to take database quantities and, in the module event method, give them to a utility to produce a derived quantity.  Some questions might be whether the derived quantity depends only on other database quantities or is the database ancillary to the quantity, or will the utility be used in other places/ways?  In any case, the most important principle is that all quantities derived from a database quantity should not be cached, since then it is too easy to accidentally miss the time-dependence.

kutschke:
I agree with Ray's previous comment - one follow up.  If you create a derived Proditions class that, among other things, has a member function that turns ticks into ns then no other code needs to know the details of how that's done; also, it's guaranteed to be up to date wrt all of the objects on which it depends.  Maybe the member function will need a second argument to specify the channel.  There are other patterns but the goal is to put the ugly part in exactly one place.

PR #923, opened 2023-01-19
changed variable name

gianipez:


kutschke (commented):
Hi Giani,

  I did a test merge in my local clone and then looked for remaining instances of minNHits as folows:

`ack  minNHits | grep -v _minNHits | grep -v minNHitsTimeCluster`

The output is:

  ```
TrkFilters/src/SeedFilter_module.cc:40:        fhicl::Atom<double>             minNHits          {     
  Name("minNStrawHits"),           Comment("minNStrawHits     ") };
  TrkFilters/src/SeedFilter_module.cc:85:    _minnhits  (config().minNHits()),
  CalPatRec/inc/CalTimePeakFinder_types.hh:35:      int                             minNHits;
  CalPatRec/src/CalTimePeakFinderDiag_tool.cc:279:      if (nhits >= _data->minNHits) n_good_seeds += 1;
 CalPatRec/src/CalTimePeakFinderDiag_tool.cc:303:      if (nhits > _data->minNHits) {
 CalPatRec/src/CalTimePeakFinder_module.cc:273:          if (nsh >= _data.minNHits) {

```

Do any of these need changing too?   I checked for the string minNHits in Production, TrackerCalibration, TrackerAlignment and TrkAna; all look OK.

gianipez:
Hi @kutschke , 
Thanks for double-checking. Yes, they can stay as they are. I just wanted to make more clear the user interface in the fhicl file used to configure the trigger filters.

Cheers,

Giani

brownd1978 (approved):
This is fine with me.   Do you foresee cuts on other kinds of hits in the future (trkcalohits)?

kutschke:
Checking to see if the last commit fixed the failure in the cosmic part of CI

@FNALbuild run build test

gianipez:
In this specific case, the filter for the `TimeCluster`, Matthew noticed that there are a few cases where we have a very small amount of ComboHits (<5) that are made of a very large number of StrawHits (4-5). One possibility to explore would be to add a cut on the # of ComboHits separately (at the TimeCluster level). At the trigger level, we don't apply the requirement to have a CaloCluster associated with the track because we don't want to be limited by the Online-reco calorimeter energy threshold. The possibility to do it is already in the filter, but it is not exploited in the trkPatRec sequences.

PR #924, opened 2023-01-19
enabling the use of Geant4 HP thermal neutron scattering

resnegfk:
This pull requests enables the use of Geant4 HP thermal neutron scattering (disabled by default)


kutschke (approved):
Thanks for setting this up for us.

PR #925, opened 2023-01-20
STM Prodition Update

AndrewEdmonds11:
As a small update to #921, I have taken the suggestion there and moved the sampling frequency to nsPerCt conversion into the STM prodition. I have also added utility functions to do clock tick to time conversions into STMUtils, which are used in the appropriate places


rlcee (approved):
Again, fine to proceed now, but since I'm asked to comment, I'll just mention that the new STMUtils methods would also be fine as methods on the prodition class

kutschke:
I second Ray's comment that the two functions that convert between clock times and time fit naturally as a Proditions class - that user code does not need to access the underlying Proditions data to pass it around.  This can be dealt with in a future PR.  Merging now.

PR #926, opened 2023-01-21
removed the default configurations of the filters used in the trigger sequences from the TrkFilers/fcl/prolog_trigger file

gianipez:
These filters must be configured by the script used to generate the Menu. Otherwise when we run the Online reconstruction in OTS mistakes could happen when configuring the trigger paths.

### DIFF
--- a/GeneralUtilities/inc/PhiPrescalingParams.hh
+++ b/GeneralUtilities/inc/PhiPrescalingParams.hh
@@ -1,7 +1,7 @@
 #ifndef GeneralUtilities_PhiPrescalingParams_hh
 #define GeneralUtilities_PhiPrescalingParams_hh
 
-#include "fhiclcpp/types/Atom.h"
+#include "fhiclcpp/types/OptionalAtom.h"
 
 namespace mu2e {
   struct PhiPrescalingParams {
@@ -10,9 +10,9 @@ namespace mu2e {
     float    _phase;
 
     struct Config {
-      fhicl::Atom<float> amplitude{fhicl::Name("amplitude"), fhicl::Comment("amplitude"), 0.};
-      fhicl::Atom<float> frequency{fhicl::Name("frequency"), fhicl::Comment("frequency"), 0.};
-      fhicl::Atom<float> phase    {fhicl::Name("phase"),     fhicl::Comment("phase"), 0.};
+      fhicl::OptionalAtom<float> amplitude{fhicl::Name("amplitude"), fhicl::Comment("amplitude")};
+      fhicl::OptionalAtom<float> frequency{fhicl::Name("frequency"), fhicl::Comment("frequency")};
+      fhicl::OptionalAtom<float> phase    {fhicl::Name("phase"),     fhicl::Comment("phase")};
     };
 
     PhiPrescalingParams():
@@ -25,10 +25,12 @@ namespace mu2e {
       _frequency(Frequency),
       _phase    (Phase) {}
 
-    PhiPrescalingParams(const Config& conf):
-      _amplitude(conf.amplitude()),
-      _frequency(conf.frequency()),
-      _phase    (conf.phase()) {}
+    PhiPrescalingParams(const Config& conf){
+      float x;
+      if (conf.amplitude(x)) _amplitude = x;
+      if (conf.frequency(x)) _frequency = x;
+      if (conf.phase(x))     _phase     = x;
+    }
   };
 
 }
--- a/GeneralUtilities/inc/PhiPrescalingParams.hh
+++ b/GeneralUtilities/inc/PhiPrescalingParams.hh
@@ -10,9 +10,9 @@ namespace mu2e {
     float    _phase;
 
     struct Config {
-      fhicl::Atom<float> amplitude{fhicl::Name("amplitude"), fhicl::Comment("amplitude")};
-      fhicl::Atom<float> frequency{fhicl::Name("frequency"), fhicl::Comment("frequency")};
-      fhicl::Atom<float> phase    {fhicl::Name("phase"),     fhicl::Comment("phase")};
+      fhicl::Atom<float> amplitude{fhicl::Name("amplitude"), fhicl::Comment("amplitude"), 0.};
+      fhicl::Atom<float> frequency{fhicl::Name("frequency"), fhicl::Comment("frequency"), 0.};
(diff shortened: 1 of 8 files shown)


gianipez:
None of the changes in this PR will affect the nightly. 

I took advantage of this PR to fix a bug in the fragment-to-caloHits module; it was a trivial typo... nothing that needs a careful review. The effect was the duplication of the caloHits.

gianipez:
@FNALbuild run build test

kutschke:
> These filters must be configured by the script used to generate the Menu. Otherwise when we run the Online reconstruction in OTS mistakes could happen when configuring the trigger paths.

I like the idea of having a single authoritative source for parameters.  I would like to understand more about "the script used to generate the Menu".

The cartoon I have in my head is that when we run in the online environment we export information from the OTS Mongo DB and a script will complete the parameter sets - is that correct?    

> None of the changes in this PR will affect the nightly.

How will it affect the POMS campaigns that run the trigger code?  Will they continue to work as is or do we need to add an extra step to the POMS campaigns to complete the fcl files?

kutschke:
I added Dave and Sophie as reviewers to understand if any maintenance will be needed on the POMS campaigns that run the trigger.

GeneralUtilities/inc/PhiPrescalingParams.hh
```diff
@@ -10,9 +10,9 @@ namespace mu2e {
     float    _phase;
 
     struct Config {
-      fhicl::Atom<float> amplitude{fhicl::Name("amplitude"), fhicl::Comment("amplitude")};
-      fhicl::Atom<float> frequency{fhicl::Name("frequency"), fhicl::Comment("frequency")};
-      fhicl::Atom<float> phase    {fhicl::Name("phase"),     fhicl::Comment("phase")};
+      fhicl::Atom<float> amplitude{fhicl::Name("amplitude"), fhicl::Comment("amplitude"), 0.};
```
kutschke, line 13: Is the meaning of 0 that every event will be passed?  ( Just checking that understand the convention you have chosen).

We have a policy that fcl parameters that "affect physics" should not be given default values in code.   I would like to discuss with you and @brownd1978 if we should think of prescale factors as parameters that affect physics.

gianipez:
> 

In the online environment, the info is read from the set of tables configured at run time (these can be bright new or loaded from a previous configuration). We have a C++ module that creates the set of FHICL files used for the trigger configurations. In the Offline case instead, we are using a python script for reading the configurations from `Trigger/data` and creating the Menu files accordingly. 

I (and @dnbrow01 ) checked that the package `Production` is not affected as it was already including only the Menu files from `gen/Trigger`.

GeneralUtilities/inc/PhiPrescalingParams.hh
gianipez: These parameters are used for a specific case; prescaling the IPA electrons using their helices' `phi0` parameter. These parameters are not used if the `HelixSeedFilter_module` doesn't have the option `prescaleUsingD0Phi` set to `true`.  Prescaling based on this formula is supposed to provide a uniformly distributed helix population, which otherwise will be skewed following the muon beam orientation. Thus, I don't affect physics. 
Also, being forced to set these parameters in the trigger paths where they are not used would be confusing in my opinion.

kutschke: In that case I recommend the following.  Change their type from Atom<T> to OptionalAtom<T> and remove the default value from the code; in the comment text say that they are required when prescaleUsingD0Phi is true.  That way they do not clutter the fcl definitions in which they are not needed and they still obey our policy.

It's a good instinct not to clutter the fcl with irrelevant information.

gianipez: thanks for the suggestion. I just pushed the change.

sophiemiddleton (approved):
Looks fine and @gianipez mentions he's checked Production, since there's no effect I'm fine with the changes

GeneralUtilities/inc/PhiPrescalingParams.hh
kutschke: Thanks for getting this done.  Please watch for other targets of opportunity to use OptionalAtom.

kutschke:
> I (and @brownd1978 ) checked that the package `Production` is not affected as it was already including only the Menu files from `gen/Trigger`.

@brownd1978 do you want to say anything more?  If not I am ready to merge now.

kutschke:
@FNALbuild run build test

Rerunning CI after merge of #928

TrkFilters/src/HelixFilter_module.cc
```diff
@@ -59,7 +59,7 @@ namespace mu2e
         fhicl::Atom<double>             minNLoops            {     Name("minNLoops"),               Comment("minNLoops      ") };
         fhicl::Sequence<std::string>    helixFitFlag         {     Name("helixFitFlag"),            Comment("helixFitFlag   "), std::vector<std::string>{"HelixOK"} };
         fhicl::Atom<int>                debugLevel           {     Name("debugLevel"),              Comment("debugLevel")     , 0 };
-        fhicl::Atom<bool>               prescaleUsingD0Phi   {     Name("prescaleUsingD0Phi"),      Comment("prescaleUsingD0Phi") };
+        fhicl::Atom<bool>               prescaleUsingD0Phi   {     Name("prescaleUsingD0Phi"),      Comment("prescaleUsingD0Phi"), false };
```
brownd1978, line 62: Is this no a deprecated option?  The general rule is that all config that could reasonably have different settings i a production job be set in prolog, not code default.

gianipez: The point of this PR is to have no defaults for the filters. This is not a deprecated parameter. It is used only in the `trkHelix_ipa_phiScaled` trigger path, that's why I thought it was better to have the default here instead of being forced to add 1 line in each HelixFilter module configuration used in each track trigger.

gianipez: I'll make it an `OptionalAtom` if that's ok

gianipez: done

gianipez:
@FNALbuild run build test

brownd1978 (approved):

PR #928, opened 2023-01-24
Added support to save unfiltered helices in TimeAndPhiClusterFinder

bechenard:
For cases leaving a large number of hits in the tracker (e.g. track bouncing back and forth).


brownd1978 (approved):
Looks good

PR #931, opened 2023-02-01
DeltaFinder, TZFinder and PhiClusterFinder modules added to CalPatRec

NamithaChitrazee:
The following modules and their diagnostic tools, some fcl files to test them out have been added to CalPatRec : 

1.  DeltaFinder : For compton hit removal @pavel1murat 
2. TZClusterFinder : Form time clusters using the time and z information of combo hits @matthewstortini 
3.  PhiClusterFinder : Form time clusters separated in phi. Motivation: Identify and reconstruct tracks which are simultaneous in time but well separated in phi. @NamithaChitrazee 

A minor change made to the MergeHelices module as well. When comparing helices, `if(nover >= _minnover && nover/float(minh) > _minoverfrac)` condition changed to `if(nover/float(minh) > _minoverfrac)`. The motivation for this change can be found in [mu2e-docdb 43996](https://mu2e-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=43996) slide 12.


TrkHitReco/src/CombineStrawHits_module.cc
```diff
@@ -228,6 +231,15 @@ namespace mu2e {
     combohit._tres       = _terr/sqrt(combohit._nsh); // error proportional to # of straws (roughly)
     //      float wvar           = sqrtf((wacc2/weights-wacc/weights*wacc/weights));//define quality as variance/average ratio
     combohit._qual       = 1.0; // quality isn't used and should be removed FIXME
+//-----------------------------------------------------------------------------
+// if combohit._wres is not consistent with the spread of the individual hits,
+// increase it
+//-----------------------------------------------------------------------------
+    if (_checkWres) {
+      float wdist2 = wacc2/weights;
+      float sigw   = sqrt((wdist2-combohit._wdist*combohit._wdist)/(combohit.nCombo()-1+1e-12));
```
brownd1978, line 240: Since this test only makes sense with nCombo()>1, please just add that as a criteria in the 'if'.  Then you won't need the fudge factor, and you won't get bizarre, huge results.

TrkReco/src/TrkRecoMcUtils_tool.cc
```diff
@@ -64,7 +64,7 @@ namespace mu2e {
       //                    const art::InputTag& Tag) override;
 
       virtual const SimParticle* getSimParticle(const art::Event* Event, int HitIndex) override;
-      virtual const XYZVectorF* getMom(const art::Event* Event, int HitIndex) override;
+      virtual const XYZVectorF*  getMom        (const art::Event* Event, int HitIndex) override;
```
brownd1978, line 67: please remove spurious whitespace

TrkReco/src/TrkRecoMcUtils_tool.cc
```diff
@@ -32,23 +32,23 @@
 namespace mu2e {
 
   class TrkRecoMcUtils : public mu2e::McUtilsToolBase {
```
brownd1978, line 34: Mu2e software policy is that modules (and tools) must be upgraded to use validated configuration instead of parameter sets when modified.  Please update accordingly.

TrkReco/src/MergeHelices_module.cc
```diff
@@ -149,7 +149,8 @@ namespace mu2e {
     // Calculate the chi-sq of the helices
     findchisq(evt,h2,chih2xy,chih2zphi);
     // overlapping helices: decide which is best
-    if(nover >= _minnover && nover/float(minh) > _minoverfrac) {
```
brownd1978, line 152: It seems this change is equivalent to setting _minnover to 0.  Why instead did you remove that test?

brownd1978 (changes requested):
This PR has conflicts with the head that must be resolved before the rest of the CI can proceed.  Please fix that, then relaunch the CI.

CalPatRec/src/TZClusterFinderDiag_tool.cc
```diff
@@ -0,0 +1,296 @@
+
+#include "fhiclcpp/ParameterSet.h"
+
+#include "CLHEP/Matrix/Vector.h"
+#include "CLHEP/Matrix/SymMatrix.h"
+#include "CLHEP/Vector/ThreeVector.h"
+
+#include "BTrk/BbrGeom/HepPoint.h"
```
gianipez, line 8: Quite a few of these includes are not necessary. Please, clean up

CalPatRec/src/TZClusterFinderDiag_tool.cc
```diff
@@ -0,0 +1,296 @@
...
+#include "art/Framework/Principal/Event.h"
+
+#include "TH1.h"
+#include "TH2.h"
+#include "TProfile.h"
+#include "TEfficiency.h"
+
+#include "TMath.h"
+
+using CLHEP::HepVector;
```
gianipez, line 38: I don't think these are used anywhere. Please, remove

CalPatRec/inc/PhiClusterFinder_types.hh
```diff
@@ -0,0 +1,51 @@
...
+
+namespace art {
+  class Event;
+};
+
+namespace fhicl {
+  class ParameterSet;
+};
+
+class TH1F;
```
gianipez, line 16: These `class` instances are not needed. Please remove

CalPatRec/inc/PhiClusterFinder_types.hh
```diff
@@ -0,0 +1,51 @@
...
+class TH2F;
+
+#include "Offline/RecoDataProducts/inc/StrawHitFlag.hh"
```
gianipez, line 19: A lot of these includes are not needed. Please, clean up this list

CalPatRec/inc/PhiClusterFinder_types.hh
```diff
@@ -0,0 +1,51 @@
...
+#include "Offline/RecoDataProducts/inc/StrawHitPosition.hh"
+#include "Offline/RecoDataProducts/inc/TimeCluster.hh"
+#include "Offline/RecoDataProducts/inc/CaloCluster.hh"
+#include "Offline/Mu2eUtilities/inc/LsqSums4.hh"
+
+
+
+namespace mu2e {
+
+  class KalFitResultNew;
```
gianipez, line 30: this is not needed here.

CalPatRec/inc/PhiClusterFinder_types.hh
```diff
@@ -0,0 +1,51 @@
...
+      fhicl::Atom<int> diagLevel{fhicl::Name("diagLevel"), fhicl::Comment("diagnostic level")};
+      fhicl::Atom<std::string> tool_type{fhicl::Name("tool_type"), fhicl::Comment("tool type: Phi Cluster Finder Diag")};
+      fhicl::Atom<int> mcTruth{fhicl::Name("mcTruth"), fhicl::Comment("MC truth")};
+    };
+
+    struct Data_t {
+
+      const art::Event*               _event;
+      const TimeCluster*              _timeCluster;
+      TimeClusterCollection*          _tccol;
```
gianipez, line 45: `_tccol` should be a `const TimeClusterCollection*` to avoid the `cast` in the other module

CalPatRec/inc/TZClusterFinder_types.hh
```diff
@@ -0,0 +1,147 @@
+#ifndef __TrkPatRec_TZClusterFinder_types_hh__
```
gianipez, line 1: fix these names please.

CalPatRec/inc/TZClusterFinder_types.hh
```diff
@@ -0,0 +1,147 @@
...
+
+namespace art {
+  class Event;
+};
+
+namespace fhicl {
+  class ParameterSet;
+};
+
+class TH1D;
```
gianipez, line 14: this is not needed

CalPatRec/inc/TZClusterFinder_types.hh
```diff
@@ -0,0 +1,147 @@
...
+
+#include "fhiclcpp/types/Atom.h"
+#include "fhiclcpp/types/Table.h"
+#include "fhiclcpp/types/Sequence.h"
+
+#include "Offline/RecoDataProducts/inc/StrawHitFlag.hh"
```
gianipez, line 20: please, review the list of these includes

CalPatRec/src/PhiClusterFinderDiag_tool.cc
```diff
@@ -0,0 +1,121 @@
...
+#include "art/Framework/Principal/Event.h"
+
+#include "TH1.h"
+#include "TH2.h"
+#include "TProfile.h"
+#include "TEfficiency.h"
+
+#include "TMath.h"
+
+using CLHEP::HepVector;
```
gianipez, line 36: I don't think these are needed.

CalPatRec/src/PhiClusterFinderDiag_tool.cc
```diff
@@ -0,0 +1,121 @@
+
+///////////////////////////////////////////////////////////////////////////////
+//
+///////////////////////////////////////////////////////////////////////////////
+#include "fhiclcpp/ParameterSet.h"
+
+#include "CLHEP/Matrix/Vector.h"
```
gianipez, line 7: please, review the list of these includes.

CalPatRec/src/PhiClusterFinder_module.cc
```diff
@@ -0,0 +1,566 @@
...
+#include "art/Framework/Core/EDProducer.h"
+#include "art_root_io/TFileService.h"
+#include "art/Utilities/make_tool.h"
+
+#include "fhiclcpp/types/Atom.h"
+#include "fhiclcpp/types/Sequence.h"
+#include "fhiclcpp/ParameterSet.h"
+#include "CLHEP/Units/PhysicalConstants.h"
+
+#include "Offline/GeometryService/inc/GeometryService.hh"
```
gianipez, line 15: please, review the list of these includes.

CalPatRec/src/PhiClusterFinder_module.cc
```diff
@@ -0,0 +1,566 @@
...
+
+#include "Offline/TrkReco/inc/TrkUtilities.hh"
+#include "Offline/TrkReco/inc/TrkTimeCalculator.hh"
+#include "Offline/GeneralUtilities/inc/Angles.hh"
+#include "Offline/ConfigTools/inc/ConfigFileLookupPolicy.hh"
+#include "Offline/Mu2eUtilities/inc/ModuleHistToolBase.hh"
+#include "Offline/Mu2eUtilities/inc/polyAtan2.hh"
+#include "Offline/CalPatRec/inc/PhiClusterFinder_types.hh"
+
+#include "TH1.h"
```
gianipez, line 31: you are using only the `TH1F` class in the module. All the variables where you use `Double_t` can be converted to `float`, so you don't need any other ROOT include.

CalPatRec/src/PhiClusterFinder_module.cc
```diff
@@ -0,0 +1,566 @@
...
+
+#include <iostream>
+#include <fstream>
+#include <memory>
+#include <algorithm>
+#include <utility>
+#include <functional>
+#include <vector>
+
+using namespace std;
```
gianipez, line 52: please, remove this. Use the `std::` prefix when you use the `std` functions.

CalPatRec/src/PhiClusterFinder_module.cc
```diff
@@ -0,0 +1,566 @@
...
+using namespace boost::accumulators;
```
gianipez, line 53: the same applies here. The risk is that a user may use a variable that has the same name as one of the functions defined in these libraries and in some rare cases that can create problems because the compiler doesn't spot it.

CalPatRec/src/PhiClusterFinder_module.cc
```diff
@@ -0,0 +1,566 @@
...
+    float                               _mindphi,_pitch;
+    float                               _tmin, _tmax, _tbin;
+    TrkTimeCalculator                   _t0calc;
+    float                               _ymin,_maxdt,_minsigma;
+    int                                 _npeak;
+    art::ProductToken<ComboHitCollection> const _chToken;
+    art::ProductToken<TimeClusterCollection> const _tcToken;
+    bool                                _useCC;
+    const ComboHitCollection*           _chcol;
+    TH1F*                               _hist1;
```
gianipez, line 112: For this PR it is OK for me to leave around these TH1F, BUT in a future PR you should switch to `std::array` or `std::vector`.

CalPatRec/src/PhiClusterFinder_module.cc
```diff
@@ -0,0 +1,566 @@
...
+    _ymin         (config().yMin()),
+    _maxdt        (config().maxDt()),
+    _minsigma     (config().minSigma()),
+    _npeak        (config().peakWidth()),
+    _chToken      {consumes<ComboHitCollection>(config().ComboHitCollection()) },
+    _tcToken      {consumes<TimeClusterCollection>(config().TimeClusterCollection()) },
+    _useCC        (config().useCC())
+    {
+      art::ServiceHandle<art::TFileService> tfs;
+      _hist1 = tfs->make<TH1F>( "hist1" , "phi spectrum",_nphibins,_phimin,_phimax);
```
gianipez, line 159: is there a specific reason to save this TH1?

CalPatRec/src/PhiClusterFinder_module.cc
```diff
@@ -0,0 +1,566 @@
...
+    _nhitotal.resize(_nphiclusters);
+
+    for(int m=0;m<_nphiclusters;m++){
+      _phitotal[m] = 0.;
+      _nhitotal[m] = 0.;
+    }
+    auto const& tcH = event.getValidHandle(_tcToken);
+    const TimeClusterCollection& tccol(*tcH);
+    // For the diagnostics
+    _data._tccol =const_cast<TimeClusterCollection*>(tcH.product());
```
gianipez, line 194: As I wrote in the other comment, `._tccol` should be a `const TimeClusterCollection*` to avoid this `const_cast`

CalPatRec/src/PhiClusterFinder_module.cc
```diff
@@ -0,0 +1,566 @@
...
+      }
+    }
+    if(_debug > 0 and tccol.size() > 0){
+      int newsize = tccol1->size();
+      int oldsize = tccol.size();
+      int eventno = _iev;
+      printf("Output tccol size = %i Input tccol size =  %i event = %i\n",newsize,oldsize,eventno);
+    }
+
+    TimeClusterCollection& tccolnew(*tccol1);
```
gianipez, line 215: if you change the interface of the function `addCaloCluster` you don't need this variable anymore.

CalPatRec/src/PhiClusterFinder_module.cc
```diff
@@ -0,0 +1,566 @@
...
+        float phi1(0);
+        _hist1->Reset("ICSEM");
+
+        if (_debug > 1) printf("Cluster : %2i\n",counter);
+        // Fill the phi histogram with all the unused hits
+        for(int i=0; i<nh; i++){
+          int ind = ordchcol[i];
+          const mu2e::ComboHit* ch = &_chcol->at(ind);
+          if(usedhit[i]==true) continue;
+          Double_t phi = ch->phi();
```
gianipez, line 261: why this can't be a float?

CalPatRec/src/PhiClusterFinder_module.cc
```diff
@@ -0,0 +1,566 @@
...
+          if(phi < 0) phi += 2*M_PI;
+          _hist1->Fill(phi,ch->nStrawHits());
+          // if(_debug>3) std::cout<<"Phi "<<phi<<" ch phi "<< phi << "n straw hits = "<<ch->nStrawHits()<<std::endl;
+        }
+
+        // Find the min and max phi around the highest phi bin in the phi spectrum
+        float cluphimin(0),cluphimax(0);
+        clusterminmax(cluphimin, cluphimax);
+        if (_debug > 2) printf("Min phi = %f Max phi = %f \n",cluphimin,cluphimax);
+        float bin = _hist1->GetBinWidth(0);
```
gianipez, line 272: this is a constant as the binning is fixed. It can be moved outside the loop.

TrkReco/src/MergeHelices_module.cc
NamithaChitrazee: Yes, I realize now that it is like setting _minnover to 0. The change was made when I noticed events where two tracks are reconstructed for a single particle trajectory. In all such events, the overlapping fraction of hits was > 0.5 but the no. of overlapping hits was just 1 or 2 hits short of _minnover (set to 10). 
[Uploading Screenshot 2023-02-02 at 12.06.26.png…]()

CalPatRec/inc/PhiClusterFinder_types.hh
NamithaChitrazee: Yes, removed it. [CalPatRec/inc/PhiClusterFinder_types.hh](https://github.com/Mu2e/Offline/pull/931/files/60d80792e35b9b409fb2c89b187e928a9f543ed8#diff-db4f7728590a32e1a74031a52ad38b4e056b3a34f18f08990cca4865e46d5347)

CalPatRec/inc/PhiClusterFinder_types.hh
NamithaChitrazee: Removed the unnecessary includes. [CalPatRec/inc/PhiClusterFinder_types.hh](https://github.com/Mu2e/Offline/pull/931/files/60d80792e35b9b409fb2c89b187e928a9f543ed8#diff-db4f7728590a32e1a74031a52ad38b4e056b3a34f18f08990cca4865e46d5347)

CalPatRec/src/PhiClusterFinderDiag_tool.cc
NamithaChitrazee: Yes, removed them. [CalPatRec/src/PhiClusterFinderDiag_tool.cc](https://github.com/NamithaChitrazee/Offline/commit/6da4c6b6b23a5d9cdae1c3d40d8475aea48a731b#diff-455ed377eb30267abd879308114e4fd8d4d3456b9b0b5bee5e3110900ef18b0f)

CalPatRec/src/PhiClusterFinderDiag_tool.cc
NamithaChitrazee: I've cleaned the list of includes.  [CalPatRec/src/PhiClusterFinderDiag_tool.cc](https://github.com/NamithaChitrazee/Offline/commit/6da4c6b6b23a5d9cdae1c3d40d8475aea48a731b#diff-455ed377eb30267abd879308114e4fd8d4d3456b9b0b5bee5e3110900ef18b0f)

CalPatRec/src/PhiClusterFinder_module.cc
```diff
@@ -0,0 +1,559 @@
...
+
+//------------------------------------------------------------------------------
+// Function to assign hits to the time cluster
+//-----------------------------------------------------------------------------
+  void PhiClusterFinder::assignHits(TimeCluster& otc, const std::vector<StrawHitIndex>& ordchcol) {
+    // assign hits to the closest time peak
+    for(unsigned istr=0; istr<ordchcol.size();++istr) {
+      float time = _t0calc.comboHitTime((*_chcol)[istr],_pitch);
+      float dt = fabs(time - otc._t0._t0);
+      if (dt < _maxdt+otc._t0._t0err){
```
gianipez, line 470: At this stage, `otc._t0._t0err` is a constant. I think `dt < _maxdt` should be enough.

CalPatRec/src/PhiClusterFinder_module.cc
```diff
@@ -0,0 +1,559 @@
...
+// Function to add calo cluster to the time clusters
+//-----------------------------------------------------------------------------
+  void PhiClusterFinder::addCaloClusters(TimeClusterCollection& tccolnew, const TimeClusterCollection& tccol) {
+    for(size_t ipeak=0; ipeak<tccol.size(); ipeak++) {
+      auto& tc = tccol[ipeak];
+      if (tc.hasCaloCluster()) {
+       // if(_debug>1) std::cout<<"TC calo x = "<<tc._caloCluster->cog3Vector().x()<<" y = "<<tc._caloCluster->cog3Vector().y()<<" time = "<<tc._caloCluster->time()<<" nhits = "<<tc.nhits()<<std::endl;
+        for(size_t inewpeak = 0; inewpeak<tccolnew.size();inewpeak++){
+         auto& newtc = tccolnew[inewpeak];
+         float dphi = fabs(polyAtan2( newtc._pos.y(), newtc._pos.x()) - polyAtan2( tc._caloCluster->cog3Vector().y(), tc._caloCluster->cog3Vector().x()));
```
gianipez, line 546: the evaluation of `polyAtan2( tc._caloCluster->cog3Vector().y(), tc._caloCluster->cog3Vector().x())` can be performed outside of the inner `for` loop to avoid re-doing it multiple times.

CalPatRec/src/PhiClusterFinder_module.cc
```diff
@@ -0,0 +1,559 @@
...
+      if (_timespec.GetBinContent(ibin) >= _ymin) bcv.push_back(make_pair(_timespec.GetBinContent(ibin),ibin));
+    std::sort(bcv.begin(),bcv.end(),[](const BinContent& x, const BinContent& y){return x.first > y.first;});
+    for (const auto& bc : bcv) {
+      float nsh(0.0);
+      float t0(0.0);
+      for (int ibin = std::max(1,bc.second-_npeak);ibin < std::min(nbins,bc.second+_npeak+1); ++ibin) {
+        nsh += _timespec.GetBinContent(ibin);
+        t0 += _timespec.GetBinCenter(ibin)*_timespec.GetBinContent(ibin);
+      }
+      t0 /= nsh;
```
gianipez, line 453: I'm not entirely sure if maybe it is not necessary, but just to be sure, I would move this line inside the `if` block below to avoid the nan case.

CalPatRec/src/PhiClusterFinder_module.cc
NamithaChitrazee: Yes, this has been changed. [const TimeClusterCollection*](https://github.com/NamithaChitrazee/Offline/commit/6da4c6b6b23a5d9cdae1c3d40d8475aea48a731b#diff-fdcfc4df7cada41e0f990c317d0561385316cce5ccc0ab47414f76a89d1f29b3:~:text=_data._tccol%20%3D%20tcH.product()%3B)

CalPatRec/src/PhiClusterFinder_module.cc
NamithaChitrazee: This is outside the loop.

gianipez: I meant outside of the ‘while’ loop

CalPatRec/src/PhiClusterFinder_module.cc
NamithaChitrazee: I've removed the other ROOT includes.

CalPatRec/src/PhiClusterFinder_module.cc
```diff
@@ -0,0 +1,566 @@
...
+      if (nsh >= _minnsh){
```
gianipez, line 454: the way this nested loop is set up, it uses always the last entry of the vector `bcv` if `nsh >_minnsh`, so it doesn't take the highest one because `bcv` is previously sorted such that the highest bin is the first entry. Am I missing anything?

CalPatRec/src/PhiClusterFinder_module.cc
NamithaChitrazee: [std::](https://github.com/NamithaChitrazee/Offline/commit/5bd924e22ca2cb1bdcf00fb256e410a5ac953d79#:~:text=using%20namespace%20std%3B) removed.

CalPatRec/src/PhiClusterFinder_module.cc
NamithaChitrazee: Hi @gianipez I've cleaned up the module quite a bit. Removed the functions like assignHits, findPeak, fillTimeSpectrum, all of which are from the time when we were using the ComboHitCollection as input.

brownd1978:
@FNALbuild run build test

TrkHitReco/src/CombineStrawHits_module.cc
brownd1978: I'm still confused: it seems this cut can be applied even if nCombo=1, in which case the value of sig2 will be huge (~10^12).  Shouldn't the logic be to exclude this case: ie (if (_checkWres && combohit.nCombo() > 1){...?

TrkPatRec/src/RobustHelixFinder_module.cc
```diff
@@ -262,6 +264,7 @@ namespace mu2e {
     _useTripletAreaWt(config().UseTripletArea()),
     _chToken     {consumes<ComboHitCollection>(config().ComboHitCollection()) },
     _tcToken     {consumes<TimeClusterCollection>(config().TimeClusterCollection()) },
+    _chfCollTag  {config().chfCollTag()},
```
brownd1978, line 267: External flag collections are not used consistently between CalPatRec and TrkPatRec.  Fixing that is beyond the scope of this PR, so its OK to add this external collection for now, but it will need to be cleaned up in a future PR.

TrkReco/src/MergeHelices_module.cc
brownd1978: Right.  That suggests that cut needs to be re-optimized for your sample, but the cut might still be useful, especially once you include pileup background.  Please leave the test in, and you can adjust the value in fcl .

TrkReco/src/TrkRecoMcUtils_tool.cc
```diff
@@ -74,19 +74,28 @@ namespace mu2e {
...
+//-----------------------------------------------------------------------------
+  TrkRecoMcUtils::TrkRecoMcUtils(const McUtilsToolBase::Config& config) :
+    _comboHitCollTag   { config.comboHitCollTag()    },
+    _strawDigiMCCollTag{ config.strawDigiMCCollTag() }
   {
     _lastEvent   = art::EventID();
     _mcdigis     = nullptr;
     _chColl      = nullptr;
     _mbtime      = -1;
-    //    _timeOffsets = new SimParticleTimeOffset(*TimeOffsets);
```
brownd1978, line 84: thanks for deleting these obsolete lines

TrkReco/src/TrkRecoMcUtils_tool.cc
```diff
@@ -74,19 +74,28 @@ namespace mu2e {
 
   //-----------------------------------------------------------------------------
   TrkRecoMcUtils::TrkRecoMcUtils(const fhicl::ParameterSet& PSet) :
```
brownd1978, line 76: Is this constructor still needed, and if so, why?  Can't this tool always be configured from the Config struct?

brownd1978 (changes requested):

CalPatRec/src/PhiClusterFinder_module.cc
```diff
@@ -0,0 +1,566 @@
...
+    _nhitotal.clear();
+    _phitotal.resize(_nphiclusters);
+    _nhitotal.resize(_nphiclusters);
+
+    for(int m=0;m<_nphiclusters;m++){
+      _phitotal[m] = 0.;
+      _nhitotal[m] = 0.;
+    }
+    auto const& tcH = event.getValidHandle(_tcToken);
+    const TimeClusterCollection& tccol(*tcH);
```
gianipez, line 161: this variable is not necessary if you use `_data._tccol` or alternatively, delete `_data._tccol` as it is used nowhere in the module.

CalPatRec/src/PhiClusterFinder_module.cc
```diff
@@ -0,0 +1,566 @@
...
+       initCluster(tc);
+       if (_debug > 1) {
+         int nsh = tc._nsh;
+         printf("Time cluster : %i No. straw hits : %i \n",counter,nsh);
+       }
+       // Sigma of the phi spectrum
+       float sigma(0);
+       if (tc._nsh > _minnsh) {
+         if (tc._nsh > _nphiclusters) {
+           sigma = checkdelta(tc, j, ordchcol);
```
gianipez, line 326: could you please change the name of the variable `sigma`? `::checkdelta` doesn't evaluate a standard deviation, so I think it's confusing. Alternatively, you can change the math exploited in that function, but then its threshold might need to be re-calibrated.

TrkReco/src/MergeHelices_module.cc
NamithaChitrazee: Yes, I understand and can easily adjust the parameter in fcl. I have reverted it back.

CalPatRec/src/PhiClusterFinder_module.cc
NamithaChitrazee: `_data._tcol` is used for the diagnostics so that one can check the no. of input time clusters and compare with the no. of output phi clusters. I've changed it to `if(_diag > 0)` then initialize `_data._tccol`.

TrkReco/src/TrkRecoMcUtils_tool.cc
pavel1murat: thanks for raising this issue ! .. I tried to switch ot a config-based constructor some time ago and failed.  Spent more time now - turned out that a config-based constructor exists but takes more parameters, which is absolutely not obvious from the error diagnostics...

Anyway, the tool changed to use a config-based construction.

TrkReco/src/TrkRecoMcUtils_tool.cc
pavel1murat: done.

pavel1murat (approved):
overall, the changes are ready to be merged. Work on performance improvement will continue

brownd1978:
@FNALbuild run build test

TrkHitReco/src/CombineStrawHits_module.cc
pavel1murat: if nCombo=1, the numerator is zero.

TrkReco/src/TrkRecoMcUtils_tool.cc
pavel1murat: those are not spurious - the text alignment makes the code better readable

TrkHitReco/src/CombineStrawHits_module.cc
```diff
@@ -228,6 +231,15 @@ namespace mu2e {
...
+      if (combohit._wres < sigw/2) combohit._wres = sigw;
```
brownd1978, line 241: What is the '2' here?  It looks like a tuned parameter.  If so, it should be a fcl parameter.  Making a statistical consistency check makes sense, but this seems an ad-hoc algorithm.  How about using the PDG formula for how errors are expanded when measurements are inconsistent?  That's well-documented and has a smooth behavior.  Here's an interesting paper on another technique: https://arxiv.org/pdf/2004.01219.pdf

TrkPatRec/src/RobustHelixFinder_module.cc
brownd1978: 0/0 is undefined.  Best is to use logic to avoid this situation.

matthewstortini (approved):

gianipez (approved):
I think the modules are now in a good state. Further improvements will be part of future PRs.

brownd1978 (approved):

PR #933, opened 2023-02-03
TrkDiag updates

brownd1978:
Modernize some of the TrkDiag analysis modules


brownd1978:
@FNALbuild run build test

TrkDiag/fcl/prolog.fcl
```diff
@@ -8,30 +8,23 @@ BD : {
   StrawDigiMCCollection : "compressDigiMCs"
   ComboHitCollection : "makePH"
   StrawHitFlagCollection : "FlagBkgHits:ComboHits"
-  BackgroundHitClusterCollection:"FlagBkgHits"
-  UseFlagCollection : true
-  TimeOffsets : {
-    inputs : [ @sequence::DigiCompression.TimeMaps ]
-  }
+  BkgQualCollection:"FlagBkgHits"
```
bechenard, line 11: Just out of curiosity, what is replacing the TimeOffset? Is that done without any config now?

TrkDiag/inc/BkgHitInfo.hh
```diff
@@ -1,18 +1,15 @@
...
   // extend StrawHitInfo struct with additional info on background hits
   struct BkgHitInfo : public StrawHitInfo {
-    Float_t _gdist; // generalized distance
-    Float_t _rrho; // transverse distance to cluster center
-    Float_t _rerr; // estimated error on transverse distance
-    Int_t _index; // index into StrawHit collection
-    CLHEP::Hep3Vector _rpos; // relative position of hit WRT cluster center
-    Bool_t _active; // hit is active in cluster
-    Bool_t _cbkg; // hit is classified backg
+    float _gdist = 0; // generalized distance
```
bechenard, line 7: it would look so much better if the '=' were all aligned :-) I would like to see your face right now....

TrkDiag/src/BkgDiag_module.cc
```diff
@@ -29,23 +29,39 @@
...
 #include "Offline/TrkDiag/inc/BkgHitInfo.hh"
-/// Utilities
-#include "Offline/Mu2eUtilities/inc/SimParticleTimeOffset.hh"
-// diagnostics
-#include "Offline/TrkDiag/inc/StrawHitInfo.hh"
+// art
+#include "canvas/Utilities/InputTag.h"
+#include "fhiclcpp/types/Atom.h"
+#include "fhiclcpp/types/Sequence.h"
 using namespace std;
```
bechenard, line 36: you would need to add a few std::string here and there and you could get rid of "using namespace std", which I believe is always a good thing. Or just use using std::string

TrkDiag/src/BkgDiag_module.cc
```diff
@@ -75,121 +90,115 @@ namespace mu2e
...
       // MC truth variables
-      Int_t _ppid, _ppdg, _pgen, _pproc, _ncontrib, _icontrib[512];
-      Float_t _pmom;
-      Int_t _nconv, _ndelta, _ncompt, _ngconv, _nebkg, _nprot, _nprimary;
+      int _ppid, _ppdg, _pgen, _pproc, _ncontrib, _icontrib[512];
+      float _pmom;
+      int _nconv, _ndelta, _ncompt, _ngconv, _nebkg, _nprot, _nprimary;
       std::vector<BkgHitInfo> _bkghinfo;
 
-      Int_t   _nindex,_hitidx[8192];
```
bechenard, line 96: I don't believe clusters with a single hits are saved, so we would need this tree to store the information of *all* hits. I saved some information, but the alternative would be to make a tree with a single BkhHitInfo branch, and record the info of all hits. Then in your original tree, you just have a pointer to these hits. Or we keep it as is, or we make dummy clusters for unassociated hits. Pick you poison

TrkDiag/src/BkgDiag_module.cc
```diff
@@ -201,36 +210,31 @@ namespace mu2e
...
 
-
     _nhits=0;
+    _hitPos.clear();
+    _hitPos.reserve(_chcol->size());
     for(size_t ich=0;ich<_chcol->size();++ich){
-      _hitRad[_nhits]    = sqrtf(_chcol->at(ich).pos().perp2());
-      _hitPhi[_nhits]    = _chcol->at(ich).pos().phi();
-      _hitHPhi[_nhits]   = _chcol->at(ich).helixPhi();
+      _hitPos.push_back(_chcol->at(ich).pos());
```
bechenard, line 217: To be honest, Radius, Phi and Z is more useful than X,Y and Z, but we could go this way if you want.

TrkDiag/src/BkgDiag_module.cc
```diff
@@ -360,21 +362,19 @@ namespace mu2e
   bool BkgDiag::findData(const art::Event& evt){
     _chcol = 0; _shfcol = 0; _bkgccol = 0; _bkgqcol = 0; _mcdigis = 0;
```
bechenard, line 363: To be honest, I'm not fond of having a million pointers for the data products, it makes the class bloated. But you get a free pass for this time

TrkDiag/src/BkgDiag_module.cc
```diff
@@ -474,22 +474,20 @@ namespace mu2e
...
       }
     }
   }
 
   void BkgDiag::fillStrawHitInfoMC(StrawDigiMC const& mcdigi, art::Ptr<SimParticle>const& pptr, StrawHitInfo& shinfo) const {
     // use TDC channel 0 to define the MC match
-    auto const& spmcp = mcdigi.earlyStrawGasStep();
-    art::Ptr<SimParticle> const& spp = spmcp->simParticle();
-    shinfo._mct0 = _toff.timeWithOffsetsApplied(*spmcp);
+    auto const& sgsp = mcdigi.earlyStrawGasStep();
```
bechenard, line 488: There are a ton of places that could uses an "auto" instead of the variable declaration, this would make the code much more elegant - and I know aristocratic gentlemen like yourself like elegance....

TrkDiag/src/ComboHitDiag_module.cc
```diff
@@ -41,55 +40,71 @@ namespace mu2e
 {
   class ComboHitDiag : public art::EDAnalyzer {
     public:
-      explicit ComboHitDiag(fhicl::ParameterSet const&);
+
```
bechenard, line 43: Here too, a bit of formatting make reading the code much easier.... Yes yes, I know.

bechenard (approved):
I won't pretend I understand the modifs to the MC information, but this looks fine. I made a few cosmetic suggestions, and there is a question regarding the tree structure we would like to save information for all hits (I would need this).

TrkDiag/fcl/prolog.fcl
brownd1978: TimeOffset is no longer neede.  Proton bunch time profile is applied in the beam campaign, muon lifetime in the Primary and pileup.  References to TimeOffset should be removed opportunistically.

TrkDiag/inc/BkgHitInfo.hh
brownd1978: Not a requirement.

TrkDiag/src/BkgDiag_module.cc
brownd1978: I removed the general 'using namespace' call, I agree that's ugly.

TrkDiag/src/BkgDiag_module.cc
brownd1978: There are 2 trees, one with cluster information (and information about the hits in that cluster), and one generic tree for all hits.  I'm not sure which you mean by 'original tree'.  Trees can't save pointers, so I also don't know what you mean by that.  Are you worried about the length of the array?  Root overwrites that in the streamer.  If it's something else please clarify.  This is an old section of the code and, though it's ugly, as far as I can tell it does the right thing.

TrkDiag/src/BkgDiag_module.cc
brownd1978: This is old code, it works, I prefer to leave it.

brownd1978:
addressing some general comments: iterators are required instead of implicit loops as identity comparisons are part of the logic, and iterators are needed to navigate to ComboHit daughters.

brownd1978:
@FNALbuild run build test

TrkDiag/src/BkgDiag_module.cc
bechenard: Sorry for the confusion. The BkgDiag tree has info about clusters and the hits that make them (note that this is only a subset of all hits). The BkgDiag2 tree has summary info about all the hits in the event. One could only store the hit information in the BkgDiag2 tree, and save the indices of the hits forming each cluster in BkgDiag. This would save space, but introduce coupling between the two trees, and this will probably cause problems with other scripts now that I think about it. So we will leave it this way.

PR #934, opened 2023-02-08
Revert TOT change

brownd1978:
Back this out until the effect of absolute timing is understood


brownd1978:
I inadvertently committed this change in PR 933:.  I’m backing it out as it caused a +0.9 ns net shift in the hit times.  Does that make sense to you Richie?

bonventre:
Yeah, the old way matches the calibration generation which is the only way the absolute timing will be right

bonventre (approved):

kutschke:
Thanks Richie - modulo my lunch time and the CI completing cleanly, I will merge this as soon as the CI completes.

PR #935, opened 2023-02-09
Remote handing geom

sridhar130:
`remoteHandlingInteriorWall5.txt`-> positioned just below `remoteHandlingInteriorWall2.txt`
`ExtShieldUpstreamBox_38` -> was removed, `remoteHandlingInteriorWall7.txt` was used instead as a shield. 

`remoteHandlingInteriorWallNotch1 & 2` (which were present earlier) in combination with wall7 creates a small passage for communication cables.

`mu2eHall_v05.txt` accomodates both `mu2eHall_v03.txt` & `mu2eHall_v04.txt`. The newly written walls (5-7) are included here.
v05 has been included in `geom_2021_PhaseI_v03.txt` which is in turn called by `geom_common_current.txt`.


kutschke:
Thanks for the PR.  There is some sort of illegal geometry that was caught by the G4 overlap checker but not by the root one.  To learn how to run these tests yourself, see https://mu2ewiki.fnal.gov/wiki/Validation#Overlaps .  The g4 overlap checker now runs much faster than it used to - it's much less than 2 hours now.

To download the log file do the following when logged in to any mu2e machine:

` curl -O https://buildmaster.fnal.gov/buildmaster/job/GitHubPRTests/job/mu2e-offline-build-test//1555/artifact/g4surfaceCheck.log`

If you are at the lab or on VPN you can just click on the log file link above.

There are two complaints:

```
Checking overlaps for volume remoteHandlingInteriorWall5 (G4SubtractionSolid) ... 
Checking overlaps for volume remoteHandlingInteriorWall7 (G4ExtrudedSolid) ... 
```

There are additional details in the log file.  Once you have fixed the problem, push the fix to your branch and it will become part of the PR

kutschke:
Sridhar presented this PR at a meeting of the shielding assessment group.  Several changes were requested.  In the discussion following this meeting we decided to close this PR and Sridhar will make a new one.

PR #942, opened 2023-02-16
KinKal options

brownd1978:
Add some options for the KinKal fit: these currently don't demonstrate improved resolution so are turned off by default, but they might prove useful in future.


kutschke:
@FNALbuild run build test

There has been an intervening merge, not yet reflected in the status of this PR.

PR #946, opened 2023-02-22
STM Hits

AndrewEdmonds11:
Hi Everyone,

This should be the last big commit for the STM Offline infrastructure. It adds the ```STMHit``` data product which respresents a calibrated energy and time of a hit in the STM. It also adds a ```MakeSTMHits``` module to turn ```STMMWDDigis``` into ```STMHits``` using energy calibration constants through the proditions service, and a ```PlotSTMEnergySpectrum``` module as an example of how to analyze these hits.

Thanks,
Andy

### DIFF
--- a/RecoDataProducts/inc/STMHit.hh
+++ b/RecoDataProducts/inc/STMHit.hh
@@ -1,0 +1,29 @@
+#ifndef RecoDataProducts_STMHit_hh
+#define RecoDataProducts_STMHit_hh
+//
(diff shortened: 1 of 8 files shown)


brownd1978 (approved):

brownd1978:
Can you suggest anyone else who might need to review the STM code?

brownd1978:
@FNALbuild run build test

PR #950, opened 2023-02-24
enforce deterministic ordering

rlcee:
The CrvRecoPulse collection was printing in a random order which was caused by it being a random order, which was SelectRecoMC seeing a random order of pulse lists inside coincidences.  This was caused by using a std::set containing vector iterators, which seems to have caused a sort on memory addresses (did not track that down fully).  This PR has a solution which is minimally disrupting, but the internet seems to prefer a free-standing function, which would require more edits.

### DIFF
--- a/CRVResponse/src/CrvCoincidenceFinder_module.cc
+++ b/CRVResponse/src/CrvCoincidenceFinder_module.cc
@@ -63,6 +63,15 @@ namespace mu2e
       fhicl::Atom<int> bigClusterThreshold{Name("bigClusterThreshold"), Comment("no coincidence check for clusters with a number of hits above this threshold")};
     };
 
+    // a vector that checks for duplicates
+    template <typename T> class uniqueVector : public std::vector<T> {
+    public:
+      void insert(const T& x) {
+        using Base = std::vector<T>;
+        if( std::find(Base::begin(),Base::end(),x) == Base::end() ) Base::push_back(x);
+      }
+    };
+
     typedef art::EDProducer::Table<Config> Parameters;
 
     explicit CrvCoincidenceFinder(const Parameters& config);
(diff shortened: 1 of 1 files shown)


ehrlich-uva:
Interesting. Thank you for finding this.

CRVResponse/src/CrvCoincidenceFinder_module.cc
```diff
@@ -63,6 +63,15 @@ namespace mu2e
       fhicl::Atom<int> bigClusterThreshold{Name("bigClusterThreshold"), Comment("no coincidence check for clusters with a number of hits above this threshold")};
     };
 
+    // a vector that checks for duplicates
+    template <typename T> class uniqueVector : public std::vector<T> {
```
kutschke, line 67: It's always a little bit scary inheriting from an std:: container type.  They do not have virtual d'tors so there are use cases that will slice the object.

Another way to do this might be to use std::set but do define it with a user supplied sort function.  @ehrlich-uva : is there a sort ordering that will accomplish what we need?

kutschke (commented):
It's always a little

CRVResponse/src/CrvCoincidenceFinder_module.cc
ehrlich-uva: Yes, I can try to come up with a ordering function and add it to the set definition.

rlcee:
The other approach is to define a fre function
void push_back_nodups(std::vector<T>& v, T& t)
this could even go in GeneralUtilities.  Once you reject the vector subclass, and agree to edit all points of use, then I think I'd lean towards this solution rather than creating a wrapper class for a iterator (if I understood).   Another approach is to write out the if(find) at each point of use.

rlcee:
Another approach would be to create a utility class called unsorted_set (maybe unset or uset) which has the interface and properties of a set but does not sort.  You only lose a little efficiency, which you are losing anyway with this constraint.  I see there is a std::unordered_set, but that sorts on a hash of the key, which would have the same problem.

ehrlich-uva:
I will make a few changes to this part of the code which will eliminate the need for the set.

ehrlich-uva:
I found that the best solution is adding a comparison function to the set. It's in PR 951. I tested it and found that it fixes the problem (i.e. same order of CrvRecoPulses after running the same script multiple times).

kutschke:
> I found that the best solution is adding a comparison function to the set. It's in PR 951. I tested it and found that it fixes the problem (i.e. same order of CrvRecoPulses after running the same script multiple times).

Does this mean that if we accept and merge #951 we should close this PR?

ehrlich-uva:
@kutschke Yes, it can be closed, if PR 951 is accepted/merged.

kutschke:
This is superseded by #951 and will be closed.

Issue #936, opened 2023-02-13
CrvRecoPulses random order


rlcee:
Running `Production/Validation/nightly/reco_00.fcl` repeatedly and printing the products shows that CrvRecoPulses are stored in a random order.  The product should be reproducible.   Also I see that  the chi2 field is often a nan, but I believe we should avoid storing nan's in  products.

ehrlich-uva:
I have no explanation for the random order. There is nothing random in the CRV reco sequence.
The chi2 gets set to NAN, if the fit fails. In this case, an error flag gets set for the reco pulse and the pulse parameters are calculated in a different way (so that the pulse can still be used).

rlcee:
random order addressed in #950  In discussions there was some agreement to change chi2=NaN to chi2=-1

kutschke:
> random order addressed in #950 In discussions there was some agreement to change chi2=NaN to chi2=-1

If we do change chi2 to -1, do we need to add protection to the logchi2 plot in ValCrvRecoPulse.cc ?

kutschke:
NaN to -1 works for me.  Won't there still be an undefined sort order if there is more than one -1 in the group?  I think that stable sort will resolve that.

ehrlich-uva:
I can change it to -1. There are CrvRecoPulseFlags. So I could add an additional flag for these types of chi2s.

ehrlich-uva:
Yes, there could be multiple pulses with a chi2 of -1 within a single event. However, each reco pulse has its own chi2 and its own CrvRecoPulseFlags. Downstream modules need to decide what to do with pulses that have a chi2 of -1. So far, no module uses this variable.

kutschke:
Do you understand the source of the random order - I thought it was from sorting on chi2 and sorting on a NaN produces undefined behaviour - maybe I am confused.

ehrlich-uva:
The random order problem was solved in PR 951. An explanation why we saw this random order was found by Ray - see top of PR 950.

rlcee:
@ehrlich-uva I'll assume you will complete this shortly.  If not, I'll ping it in a few days.  I'm happy to help.

kutschke:
> The random order problem was solved in PR 951. An explanation why we saw this random order was found by Ray - see top of PR 950.

Thanks for the reminder.

Ray: what's the plan to move forward?  Will you make the change or ask Ralf to?  

Ralf: is there any code that tests for NaNs?

rlcee:
[...]

PR #951, opened 2023-02-26
added a sort function to a set used in CrvCoincidenceFinder

ehrlich-uva:
This change prevents the CrvRecoPulses belonging to a CRV coincidence cluster to be stored in a random order. It fixes Issue #936 (CrvRecoPulses random order). I verified that it solves the problem by running Validation/cosmicSimReco.fcl multiple times before and after this change.

### DIFF
--- a/CRVResponse/src/CrvCoincidenceFinder_module.cc
+++ b/CRVResponse/src/CrvCoincidenceFinder_module.cc
@@ -570,7 +570,8 @@ namespace mu2e
     //we want to collect all hits belonging to coincidence groups,
     //but avoid collecting hits multiple times, if they belong to different coincidence groups.
     //can be done by placing the hit interator into a set.
-    std::set<std::vector<CrvHit>::const_iterator> coincidenceHitSet;
+    auto setComp = [](const std::vector<CrvHit>::const_iterator &a, const std::vector<CrvHit>::const_iterator &b) {return a->_crvRecoPulse < b->_crvRecoPulse;};
+    std::set<std::vector<CrvHit>::const_iterator,decltype(setComp)> coincidenceHitSet(setComp);
 
     int minCoincidenceLayers = std::min_element(hits.begin(),hits.end(),
                                [](const CrvHit &a, const CrvHit &b){return a._coincidenceLayers < b._coincidenceLayers;})->_coincidenceLayers;


kutschke (commented):
Can more than one CrvHit point to the same CrvRecoPulse object?

I see that this code contains both srt::vector<CrvHit> and std:list<CrvHit>.  Was the issue that the std::list caused the random ordering in memory?  Or something else?

kutschke:
@ehrlich-uva Do you have any advice about the merge ordering of this PR and #943?  Will #943 need additional work if this is merged first?

ehrlich-uva:
There is only one CrvHit per CrvRecoPulse, that's why it should work as sorting parameter. 

The randomness came from the std::set, because iterators from different vectors (= different Crv layers) were inserted into the set. The goal was to avoid duplicate hits from the same layer, which was accomplished by comparing the iterators from each layer separately (automatically by the std::set mechanism). However, this created a fixed order of CrvHits only for each layer separately. Since the iterators were used for comparison (i.e. memory addresses were compared), the order of iterators coming from different layers was not fixed.

I see that PR 947 was just merged. So I would like to wait one nightly validation to verify that it didn't cause any changes.
Next we could merge this PR 951 and again wait one nightly validation (it should not cause any changes either). After that, we can merge PR 943, which will cause some changes in the validation results for CRV data products, because the coincidence settings have changed.

kutschke:
> I see that PR 947 was just merged. So I would like to wait one nightly validation to verify that it didn't cause any changes.
> Next we could merge this PR 951 and again wait one nightly validation (it should not cause any changes either). After that, we can merge PR 943, which will cause some changes in the validation results for CRV data products, because the coincidence settings have changed.

I like that plan. I would have thought that 951 might make a change because it can be a different sort order than before?

rlcee (approved):
nice solution

ehrlich-uva:
@kutschke Yes, the order of the reco pulses will change, but I think the nightly validation doesn't care about it.

kutschke:
@FNALbuild run build test

ehrlich-uva:
I would like to wait with this PR and the other PR until I convinced myself that nothing went wrong with the geometry change.

kutschke (approved):

PR #953, opened 2023-03-01
added new CRV epilog following the changes in the CrvCoincidenceFinder

ehrlich-uva:
The changes that were made to the CRV prolog also need to be done in the CRV epilog_extracted.
So I just added a new epilog. This shouldn't break anything.

### DIFF
--- a/CRVResponse/fcl/epilog_extracted.fcl
+++ b/CRVResponse/fcl/epilog_extracted.fcl
@@ -1,0 +1,1 @@
+#include "Offline/CRVResponse/fcl/epilog_extracted_v03.fcl"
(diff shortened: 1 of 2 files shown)


brownd1978 (commented):
Hi Ralf, the old version of this file is #included in https://github.com/Mu2e/Production/blob/main/JobConfig/reco/Extracted.fcl
Can you create an indirection script in CrvResponse that doesn't have a version #?  Then Production won't need to be updated every time you change this version.  Or, consider just letting git take care of versioning for you.

ehrlich-uva:
I could add an epilog_exctract.fcl that includes epilog_extracted_v03.fcl (or any newer version), and Production can include epilog_exctract.fcl. That's how it is already done for the CRV prolog.

brownd1978 (approved):

brownd1978:
Thanks for updating the epilog.  Can you also update the Production epilog to source this new, generic file, and make a PR of that?

PR #961, opened 2023-03-13
Added new variables to BkgDiag

myucel-fnal:


TrkDiag/src/BkgDiag_module.cc
```diff
@@ -300,9 +309,17 @@ namespace mu2e
...
       bool pce = _pgen==2; // primary from a CE
       _nindex=0;
+      double sumEdep(0.);
+      double sqrSumDeltaTime(0.);
+      double sqrSumDeltaX(0.);
+      double sqrSumDeltaY(0.);
       for(auto const& ich : cluster.hits()){
         ComboHit const& ch = _chcol->at(ich);
         BkgClusterHit const& bhit = _bkghitcol->at(ich);
+        sumEdep +=  ch.energyDep();
```
brownd1978, line 319: edep is measured per straw.  I think this sum should be weighted by the # of straw hits (ch.nStrawHits()

TrkDiag/src/BkgDiag_module.cc
```diff
@@ -300,9 +309,17 @@ namespace mu2e
...
+        sqrSumDeltaX += std::pow(ch.pos().X() - _cpos.X(),2);
```
brownd1978, line 320: The RMS calculation is mathematically correct, but I suspect we will have better discrimination computing the rms of the transverse distance to the straw, not the hit.  That will remove the uncertainty from the position along the straw, and it automatically combines the 2 dimensions.  You can compute that using 2 line POCA, one line being in Z through the cluster position, the other being along the straw (udir) through the hit position.  Or its simple enough to compute analytically.

brownd1978:
@FNALbuild run build test

TrkDiag/src/BkgDiag_module.cc
```diff
@@ -353,6 +370,10 @@ namespace mu2e
         if(bkghinfo._mcpdg == PDGCode::proton)_nprot += ch.nStrawHits();
         _bkghinfo.push_back(bkghinfo);
       }
+      _avecedep = sumEdep/_nchits;
+      _rmscposx = sdt::sqrt(sqrSumDeltaX/_nchits);
```
kutschke, line 374: This line is the build error sdt --> std

kutschke (changes requested):

kutschke:
@myucel-fnal I am checking in.  I think that I am waiting for you to make a commit to fix the build issue and push it to your fork.  Are we on the same page?

kutschke:
@FNALbuild run build test

kutschke (approved):

kutschke:
The CI tests passed.

Dave made a suggestion about using a 2 line POCA - it was just a suggestion not a requirement.  Do you plan to incorporate that in this PR - if not then this PR is ready to merge.  Please let me know your plans.

myucel-fnal:
Let me work on those changes, I need to check results to make sure. Hopefully I am finished with it today.

PR #968, opened 2023-04-19
Update background training

brownd1978:
Train Background hit ANN inference code using root 6.28.  Include training for the trigger.


kutschke (approved):
I have reviewed the bookkeeping changes and it it looks good.  I think that the most reasonable way to review the physics content is to merge it and let the nightly do it's work. 

@brownd1978 Do you agree?

PR #972, opened 2023-04-22
Trigger MC truth fcl

brownd1978:
Add fcl to configure KalSeedMC construction in the trigger reco sequences.  This also fixes a configuration bug which allowed background hits into the pattern recognition.

### DIFF
--- a/Mu2eKinKal/test/TTTA.fcl
+++ b/Mu2eKinKal/test/TTTA.fcl
@@ -10,7 +10,7 @@
 #include "Offline/Mu2eKinKal/fcl/prolog_trigger.fcl"
 #include "Offline/Trigger/fcl/prolog_trigger.fcl"
 #include "TrkAna/fcl/prolog.fcl"
-#include "Offline/CommonMC/fcl/prolog.fcl"
+#include "Offline/CommonMC/fcl/prolog_trigger.fcl"
 
 process_name: TTKKSeed
 source : {module_type : RootInput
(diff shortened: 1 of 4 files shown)


kutschke:
@FNALbuild run build test

Mu2eKinKal/test/TTTA.fcl
gianipez, line 1: I noticed a tiny bug in this configuration file; when you re-define the trigger sequences, we should add a module instance of the `MergeTriggerInfo` module. This is done automatically at `build` time by the `genTriggerFcl.py` script for generating the Menu files. you would need to add a line like the following to instantiate the module:
`physics.producers."+<trigger_path_name>+"TriggerInfoMerger"+" : { module_type : MergeTriggerInfo }"`

and then of course, append the `<trigger_path_name>+"TriggerInfoMerger"` module instance at the end of the sequence.

TrkHitReco/fcl/prolog.fcl
```diff
@@ -13,8 +13,8 @@ makeSH : {
   MaximumTime             : 1690.0 # ns
   MinimumEnergy           : 0.0001 # MeV
   MaximumEnergy           : 0.0035 # MeV
-  MinimumRadius           : 0.0 # mm
-  MaximumRadius           : 800.0 # mm
+  MinimumRadius           : 395.0 # mm
+  MaximumRadius           : 650.0 # mm
```
gianipez, line 17: I would change the setting of the same parameters in the `makePH` module instance to avoid confusion. 
Also, I think that maybe a de-activation of X straws at smaller(larger) radial distance w.r.t. the solenoid axis would allow to obtain a more uniform selection (I'm not asking to make this change in this PR, but I wanted to mention it to keep it in mind for the future)

Mu2eKinKal/test/TTTA.fcl
brownd1978: This is a diagnostic script used for making TrkAna output.   It needs to be replaced with an official trigger path that includes MC truth matching as we discussed, maybe that's the place to put MergeTriggerInfo.

kutschke:
@brownd1978 are you planning to followup on Giani's comments in this PR or in a follow-on PR?

kutschke:
@FNALbuild run build test

kutschke:
@gianipez please clarify if you are waiting for work from Dave before you approve this.  If not please approve when ready.  Thanks.

gianipez (approved):

PR #973, opened 2023-04-23
Allow for Multiple CrvCoincidenceCollections

AndrewEdmonds11:
@oksuzian is studying the CRV reconstruction with different PE thresholds and was running into various issues including missing MC information. These issues were appearing at later stages but were due to the fact that we are only taking a single ```CrvCoincidenceCollection``` in ```SelectRecoMCs``` and ```compressRecoMCs```.

In this PR I've updated these modules to be able to take multiple ```CrvCoincidenceCollections```. The names of the output collections change from:

```
mu2e::CrvCoincidenceClusterMCs_compressRecoMCs__Reconstruct.
mu2e::CrvCoincidenceClusters_SelectRecoMC__Reconstruct.
```
to 
```
mu2e::CrvCoincidenceClusterMCs_compressRecoMCs_CrvCoincidenceClusterMatchMC_Reconstruct.
mu2e::CrvCoincidenceClusters_SelectRecoMC_CrvCoincidenceClusterFinder_Reconstruct.
```
which could have some downstream effects. There will be corresponding PRs in Production and TrkAna for these but there may be other places that need changing.

I've tested these updates by running reco and TrkAna on 10 events of  dig.mu2e.CeEndpointMix2BBSignal.MDC2020r_perfect_v1_0.art both with and without additional ```CrvCoincidenceCollections``` and everything seems to work.

Let me know if you have any questions / comments

### DIFF
--- a/CommonMC/src/SelectRecoMC_module.cc
+++ b/CommonMC/src/SelectRecoMC_module.cc
@@ -67,7 +67,7 @@ namespace mu2e {
         fhicl::Atom<bool> useSHFC                 { Name("UseStrawHitFlagCollection"),                  Comment("Use the StrawHitFlag collection"), false};
         fhicl::Atom<art::InputTag> PP             { Name("PrimaryParticle"),                  Comment("PrimaryParticle")};
         fhicl::Atom<art::InputTag> CCC            { Name("CaloClusterCollection"),          Comment("CaloClusterCollection")};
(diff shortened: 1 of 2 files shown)


kutschke:
@FNALbuild run build test with Mu2e/Production#244

kutschke:
@AndrewEdmonds11  Did you expect these failures even when the matching PR from Production is included?

AndrewEdmonds11:
Hi Rob, I guess I need to propagate the new collection names to other fhicl files. I'll get to it soon

kutschke (approved):

PR #976, opened 2023-05-01
DAQ dev

gianipez:
added threshold to have a better control over the CaloHits produced from the fragments

### DIFF
--- a/DAQ/src/CaloHitsFromFragments_module.cc
+++ b/DAQ/src/CaloHitsFromFragments_module.cc
@@ -68,8 +68,17 @@ public:
         fhicl::Name("nSigmaNoise"),
         fhicl::Comment("Maxnumber of sigma Noise to combine digi") };
     fhicl::Atom<float> hitEDepMax{
-        fhicl::Name("hitEDepMax"),
+        fhicl::Name("hitEDepMin"),
         fhicl::Comment("Maximum hit energy in MeV (to reject saturated pulses)")};
+    fhicl::Atom<float> hitEDepMin{
+        fhicl::Name("hitEDepMin"),
+        fhicl::Comment("Minimum hit energy in MeV")};
+    fhicl::Atom<float> caphriEDepMax{
+        fhicl::Name("caphriEDepMax"),
+        fhicl::Comment("Maximum CAPHRI hit energy in MeV")};
+    fhicl::Atom<float> caphriEDepMin{
+        fhicl::Name("caphriEDepMin"),
+        fhicl::Comment("Minimum CAPHRI hit energy in MeV")};
   };
 
   // --- C'tor/d'tor:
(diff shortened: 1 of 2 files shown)


gianipez:
tagging @brownd1978 @kutschke

kutschke:
Giani.  I checked the source repo and it is 83 commits behind the head of Offline.   I have checked that the merge will work correctly.  This seems like an odd choice so I am checking that this is what you intended.

DAQ/src/CaloHitsFromFragments_module.cc
```diff
@@ -68,8 +68,17 @@ class art::CaloHitsFromFragments : public EDProducer {
         fhicl::Name("nSigmaNoise"),
         fhicl::Comment("Maxnumber of sigma Noise to combine digi") };
     fhicl::Atom<float> hitEDepMax{
-        fhicl::Name("hitEDepMax"),
+        fhicl::Name("hitEDepMin"),
```
kutschke, line 71: There is now a duplicate definition for the fhicl::Name("hitEDepMin"). I presume this is a typo?

DAQ/src/CaloHitsFromFragments_module.cc
```diff
@@ -357,8 +368,10 @@ void art::CaloHitsFromFragments::analyze_calorimeter_(
       //      }
       float time = hits[hitIdx].first.Time + peakIndex * digiSampling_ + timeCalib_[sipmID];
 
+      bool  isCaphri = std::find(caphriCrystalID_.begin(), caphriCrystalID_.end(), crystalID) != caphriCrystalID_.end();
       // FIX ME! WE NEED TO CHECK IF TEH PULSE IS SATURATED HERE
-      if (eDep < hitEDepMax_) {
+      if ( ( (eDep >= hitEDepMin_) || (isCaphri && (eDep >= caphriEDepMin_))) &&
+           ( (eDep <  hitEDepMax_) || (isCaphri && (eDep <  caphriEDepMax_))) ) {
```
kutschke, line 374: The first line looks like it expects caphriEDepMin < histEDepMin ( ie a lower threshold for caphri hits).  Will this logic do the right thing if caphriEDepMin > histEDepMin.  If that's an illegal configuration, then protect against it in the c'tor.  A similar comment for the second line.

DAQ/src/CaloHitsFromFragments_module.cc
gianipez: fixed. Thanks for noticing it.

DAQ/src/CaloHitsFromFragments_module.cc
gianipez: I understand your point. We are still in a developing stage for the proper CAPHRI simulation/configuration, so I think we can revisit it once that will be more clear. During the summer time we should have collaborators optimizing this part of the code and this change is intended to help the process.

kutschke: You can address this in a later PR.  Please be sure that, as this work evolves, the c'tor has tests for logical consistency of the configuration. Then you can assume consistency in the rest of the code.

And thanks for fixing the typo.

kutschke:
@FNALbuild run build test

kutschke (approved):

brownd1978 (approved):

PR #979, opened 2023-05-05
Simplify KK fit instance names, for downstream compatibility

brownd1978:
This PR must go with the equivalent PR in Production


kutschke:
@FNALbuild run build test with Mu2e/Production#250

PR #982, opened 2023-05-23
remove duplicate binary write

rlcee:
At some point, probably when I was re-organizing this code, we were left with writing binary files in two places in the code.  I noticed it the last time I converted Bfields.  This PR just removes one of the two writes.  Checked that the conversion works properly.  Checked that ceSimReco is not affected.


kutschke (approved):
It looks right.

PR #983, opened 2023-05-24
adding files used in diagnosing muon stopping problem

resnegfk:
Just archiving the files. Also for a future use should we need them to verify that the problem
reported in https://bugzilla-geant4.kek.jp/show_bug.cgi?id=2531 is fixed


kutschke (approved):
I had a quick look and these look OK.  These are all new files (no changed files) and there is no test coverage in valication.

kutschke:
Krzysztof reminded me that these files were used to prepare the plots for his presentation to the G4 group that describes the bug:
  https://indico.cern.ch/event/1256711/#11-problem-of-muon-transport

PR #993, opened 2023-06-19
update for next-gen data-handling

rlcee:
Revamp these files for next-gen data-handling.  No one currently uses these so there should be no effect anywhere.


brownd1978 (approved):

brownd1978:
@FNALbuild run build test

PR #999, opened 2023-07-06
Fix flag bug

brownd1978:
Fix bug in conditional nesting that resulted in the 'dead' bit begin set for almost all straw hits.


kutschke (approved):
I can't comment on the changes in the test files.  I see that the only real change is in StrawHitRecoUtils.cc and I understand logic change.

Issue #970, opened 2023-04-22
Double semicolon causes -Werror=pedantic


eflumerf:
https://github.com/Mu2e/Offline/blob/ba920e9ff8da82de72dfdf779a50238a8226f806/RecoDataProducts/inc/ComboHit.hh#L106

eflumerf:
@brownd1978

rlcee:
all pedantic warning are now fixed and the pedantic check is on my default

PR #1000, opened 2023-07-07
Pendantic fixes

rlcee:
Fix most issues associated with -WError=pedantic (#970)
Fixed:
* double semicolon
* semicolon after namespace, function, or macro
* format %p requires void* argument

Unfixed:
* extra semi-colon in BTrk
* variable length arrays (70 places). Probably could be vectors, but hesitate to just blindly swap
* intialize to -1: mask(0xffff).  Probably should be unsigned, but complicated use.


kutschke (approved):
I used git diff with a long string of awk/grep t o convince myself that this PR does only what Ray says it does.

I also checked about merge conflicts with the other outstanding PRs.  If we merge in the order 1000, 994, 989, there is one conflict that occiurs doing the last merge and we can likely resolve it with the online editor.

I will start a slack conversation about merge order

PR #1002, opened 2023-07-11
avoid overflow

rlcee:
Initializing a short to 0xFFFF causes a pedantic warning


kutschke:
@pavel1murat git blame says that this line is your's  and that's why I asked you to review.  We would like to increase the level of compiler checking by adding --pedantic and this line is flagged by that.  The alternative is modify this code, and the related data product code, to changed the variable type to unsigned short.  I think it's your call.

kutschke (approved):
Pasha approved in a private email to me.

PR #1004, opened 2023-07-13
use CaloSiPMId instead of CaloIDMapper

rlcee:
This is the third and final (from me) update concerning the new calorimeter id classes.  Here I remove CaloIDMapper in favor of CaloSiPMId, and make sure the calo geometry uses CaloConst.hh values where possible.


kutschke:
Who are appropriate reviewers for this: Giani, Sophie, (Stefano?)?  Others?

rlcee:
> Who are appropriate reviewers for this: Giani, Sophie, (Stefano?)? Others?

I believe CaloIDMapper is Bertrand's code, so I think he would be best, then maybe Giani, then I'm not sure.

brownd1978 (approved):
This looks like a valuable simplification.

bechenard (approved):
Looks good (still on vacation so only gave a cursory look but don't want to hold this up)

kutschke:
@FNALbuild run build test

PR #1006, opened 2023-07-14
fix pedantic arrays in Analyses

rlcee:


kutschke (approved):
THanks for looking after these.

kutschke:
@FNALbuild run build test

PR #1007, opened 2023-07-14
fix CRV pedantic errors

rlcee:
Tested with small samples of ceSimReco cosmicSimReco and reco


ehrlich-uva (approved):

CRVResponse/src/CrvCoincidenceFinder_module.cc
```diff
@@ -757,7 +757,7 @@ namespace mu2e
                       [](L &a, L &b){return a->_maxSlope < b->_maxSlope;}))->_maxSlope;
     double maxSlopeDifference = (*std::max_element(layerIterators,layerIterators+n,
                                 [](L &a, L &b){return a->_maxSlopeDifference < b->_maxSlopeDifference;}))->_maxSlopeDifference;
-    double slopes[n-1];
+    std::vector<double> slopes(n-1);
```
ehrlich-uva, line 760: I'm curious to see, if this impacts the speed, since this part is called very often - especially for events with a large number of hits. Perhaps an std::array may be better than a vector.

CRVResponse/src/CrvCoincidenceCheck_module.cc
```diff
@@ -459,10 +460,10 @@ if(crvCoincidenceCollection->size()==0)
...
     if(n==0) return;
 
-    double maxTimeDifferences[n];
-    double times[n];
-    double timesPulseStart[n], timesPulseEnd[n];
-    double x[n], y[n];
+    std::vector<double> maxTimeDifferences(n);
+    std::vector<double> times(n);
+    std::vector<double> timesPulseStart(n), timesPulseEnd(n);
+    std::vector<double> x(n), y(n);
```
ehrlich-uva, line 466: This module is not used anymore. I kept it just in case some older scripts still use it.

CRVResponse/src/CrvCoincidenceFinder_module.cc
brownd1978: Array requires knowing the dimensions at compile time

rlcee: Without a proper profiler, I don't think we can accurately answer this question.  The art total CPU time is from the OS, but I'm not sure how accurate it is and whether we could sense the difference in the overall exe time.  The module times art prints are elapsed times and therefore inadequate for small CPU changes.  I don't know what the compiler optimizer is capable of.  If there is a reasonable value for the maximum n, then we can make this a local std::array.  Or you could imagine using a vector created outside of loops, though that is a little non-standard.  I'm willing to do whatever the consensus is.  Ralf, you mentioned an array, do you have a max possible n?

ehrlich-uva: The maximum value for n is 4 (=4 CRV layers). This means that the maximum array size is 3.

kutschke: Thanks for the discussion.  Dave is right - I thought I had remembered that the the dimension could be specified at run time so long as it was const and had scope and lifetime that was greater than or equal to the array.  I was wrong about that.   

My sense is that it makes more sense to leave this as std::vector - then it's obvious what the dimension means and it's correct even if a future upgrade puts >4  SiPMS on some counters.  

Sorry for the noise.

CRVResponse/src/CrvCoincidenceCheck_module.cc
kutschke: Thanks Ralf.

CRVResponse/src/CrvCoincidenceFinder_module.cc
```diff
@@ -757,7 +758,7 @@ namespace mu2e
                       [](L &a, L &b){return a->_maxSlope < b->_maxSlope;}))->_maxSlope;
     double maxSlopeDifference = (*std::max_element(layerIterators,layerIterators+n,
                                 [](L &a, L &b){return a->_maxSlopeDifference < b->_maxSlopeDifference;}))->_maxSlopeDifference;
-    double slopes[n-1];
+    std::array<double,3> slopes;
```
kutschke, line 761: Let's wrap this up.  I am uncomfortable leaving it as it since the 3 is a magic number and, in principle, could need maintenance in the future.  So let's put this one back to vector.  Sorry for the noise.  Ray, can you look after this.

ehrlich-uva: We already have `constexpr static const int nLayers = 4;` in line 163.
So we could use nLayers to create `constexpr int maxSlopes = (nLayers-1);`

rlcee: DataProducts/inc/CRVId.hh is the logical place for this, but it does not have nlayers. Let me know if you want it there, or Ralf's suggestion.

kutschke: I like Rays idea of move the constexpr to CRVid.hh, removing it from this file and then using std::array with maxSlopes.

@ehrlich-uva is that OK with you?  

In a separate PR we can scrub for other definitions of nLayers  .

ehrlich-uva: Yes, that's Ok with me.

kutschke:
@FNALbuild run build test

kutschke (approved):

PR #1009, opened 2023-07-15
fix trk pedantic

rlcee:
Tested with small samples of ceSimReco cosmicSimReco and reco


brownd1978 (approved):

matthewstortini (approved):

gianipez (approved):
Everything looks good to me. The only thing I noticed is that RobustHelixFit uses `using namespace std` that I personally don't like, but this is out of the scope of this PR.

PR #1013, opened 2023-07-18
Update artdaq-core-mu2e library names to match updated names in

eflumerf:
artdaq_core_mu2e v1_09_02

### DIFF
--- a/Analyses/src/SConscript
+++ b/Analyses/src/SConscript
@@ -72,7 +72,7 @@ helper.make_plugins( [ mainlib,
   'pthread',
   #'hep_concurrency',
   'DTCInterface',
-  'Overlays'
+  'artdaq-core-mu2e_Overlays'
 ] )
 # this tells emacs to view this file in python mode.
 # Local Variables:
--- a/.muse
+++ b/.muse
@@ -1,5 +1,5 @@
 # prefer to build with this environment
-ENVSET p043
+ENVSET p044
 # add these to python path
 PYTHONPATH Trigger/python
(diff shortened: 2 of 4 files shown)


kutschke (approved):

rlcee:
We will need to immediately turn around and update the envset for pedantic BTrk version.  @kutschke do you want to keep that a separate step or try to merge that in here?

kutschke:
> We will need to immediately turn around and update the envset for pedantic BTrk version. @kutschke do you want to keep that a separate step or try to merge that in here?

Good catch.  I will ask Mackenzie to include it in her G4 update envset which should come later today.

Analyses/src/SConscript
pavel1murat, line 1: I wonder it that was sufficient for linking - I had to link an analysis module earlier today and had to add one more library:
 
                  'artdaq-mu2e-core_Overlays',
                  'artdaq-mu2e-core_Data',
                  'artdaq-core_Data',
                  'DTCInterface'

However , if the linker finds everything it needs w/o libartdaq-mu2e-core_Data.so, so be it

kutschke: Did you have a link-time or run-time error?

pavel1murat (approved):
looks good, I made a minor comment, but it may not require any action...

Analyses/src/SConscript
pavel1murat: yes

kutschke: Thanks.  Since the CI links I will go ahead and merge.

PR #1014, opened 2023-07-18
removed NANs and isnans from CRVResponse modules

ehrlich-uva:


ehrlich-uva:
I would like to have this PR merged separately so that we would notice any potential differences in the nightly validations. (It shouldn't result in any differences.)

kutschke:
> I would like to have this PR merged separately so that we would notice any potential differences in the nightly validations. (It shouldn't result in any differences.)

Will do.

CRVResponse/src/MakeCrvRecoPulses.cc
```diff
@@ -277,7 +278,8 @@ void MakeCrvRecoPulses::SetWaveform(const std::vector<unsigned int> &waveform,
     double pulseTime    = fitParam1;
     float  pulseHeight  = fitParam0/TMath::E();
     float  pulseBeta    = fitParam2;
-    float  pulseFitChi2 = (fr->Ndf()>0?fr->Chi2()/fr->Ndf():NAN);
+    float  pulseFitChi2 = (fr->Ndf()>0?fr->Chi2()/fr->Ndf():-1);
+    bool   zeroNdf      = (fr->Ndf()>0?0:1);
```
rlcee, line 282: I'd use false:true here for readability

rlcee (approved):
Approved, with one note for your consideration

CRVResponse/src/MakeCrvRecoPulses.cc
ehrlich-uva: I will change it.

kutschke:
@FNALbuild run build test

PR #1016, opened 2023-07-19
start removing SimParticleTimeMap

rlcee:
per #852 , just one initial step while clarifying next steps


brownd1978 (commented):
The exposed part of SimParticleTimeMap is SimParticleTimeOffset.    How does this change interact with that?

brownd1978 (approved):

kutschke:
@FNALbuild run build test

kutschke:
@FNALbuild run build test

PR #1020, opened 2023-07-24
Calo DB next round of changes

sophiemiddleton:
The next round of changes for the calorimeter DB information.


kutschke:
@sophiemiddleton  is this PR still infrastructure or will it modify the behaviour of nightly validation?

sophiemiddleton:
its still infrastructure

sophiemiddleton:
@FNALbuild run build test

DbTables/inc/CalCosmicT0Align.hh
```diff
@@ -0,0 +1,70 @@
...
+    private:
+      CaloSiPMId   _roid;
+      float _Tcorr;
+      float _Terr;
+      float _chisq;
+    };
+
+    constexpr static const char* cxname = "CalCosmicT0Align";
+
+    CalCosmicT0Align():DbTable(cxname,"cal.cosmicT0align","roid,Tcorr,Terr,chisq"){}
```
rlcee, line 34: In the quotes here is the database table name and column names, which are case insensitive, please make these "T" into "t".  We will want to make this change in existing tables, but that can wait for another PR.

DbTables/inc/CalCosmicT0Align.hh
```diff
@@ -0,0 +1,70 @@
...
+
+    const Row& row(std::uint16_t roid) const { return _rows.at(roid); }
```
rlcee, line 36: In the other tables, we had a CaloSIPMId here

CaloConditions/src/CalCalibMaker.cc
```diff
@@ -1,29 +1,71 @@
...
   typedef std::shared_ptr<CalCalib> ptr_t;
 
+/*******************************************************************/
   ptr_t CalCalibMaker::fromFcl() {
-  CalCalibPar nominal(_config.ADC2MeV(), _config.timeoffset(), _config.algID());
+  
+  if (_config.verbose()) {
+    std::cout << "CRalCalibMaker::fromFcl making nominal CalCalib\n";
+  }
+  CalCalibPar nominal(_config.ADC2MeV(), _config.timeoffset(), _config.ECombAlgID());
```
rlcee, line 16: Is the order of arguments incorrect here?

CaloConditions/src/CalCalibMaker.cc
```diff
@@ -1,29 +1,71 @@
...
 
   size_t nChan = CaloConst::_nChannel;
 
+  if (_config.verbose()) {
+    std::cout << "CalCalibMaker::fromFcl filling " << nChan << " channels\n";
+    std::cout << "CalCalibMaker::fromFcl nominal " << fixed << setprecision(3)
+         << setw(10) << nominal.ADC2MeV() << setprecision(3) << setw(10)
+         << nominal.timeOffset() << setprecision(3) << setw(10)
```
rlcee, line 24: order here should probably follow argument order

rlcee (changes requested):
One general comment.  The idea of embracing CaloSiPMId is that these: "std::uint16_t roid" become "CaloSiPMId roid".   It is up to you and the other calo coders how to do this, but that is the idea.  A fe w other comments inline.

kutschke:
@FNALbuild run build test

DbTables/inc/CalCosmicT0Align.hh
sophiemiddleton: I dont see that in other tables, but maybe it needs to be implemented for all?

rlcee: I was comparing to CalEnergyCalib, which uses it, but it looks like that is the only table which does.  But I think this is part of the general comment, how much do you want to embrace CaloSiPMID, and how much right now?  I won't insist in this PR, but eventually there should be a consistent pattern.   I also see some tables use int and some uint16_t

sophiemiddleton:
@FNALbuild run build test

DbTables/inc/CalCosmicEnergyCalib.hh
```diff
@@ -37,7 +37,8 @@ namespace mu2e {
 
     CalCosmicEnergyCalib():DbTable(cxname,"cal.cosmicenergycalib","roid,EPeak,ErrEPeak,Width,ErrWidth,chisq"){}
 
-    const Row& row(std::uint16_t roid) const { return _rows.at(roid); }
+    const Row& row(CaloSiPMId  _roid) const {
```
rlcee, line 40: Sorry one more - this underscore is misleading

sophiemiddleton: fixed

sophiemiddleton:
@FNALbuild run build test

kutschke:
@sophiemiddleton The CI found whitespace issues. Please address these.  Thanks.

rlcee (approved):

kutschke:
@sophiemiddleton Ray has approved this now.  I will wait for the whitespace fixes.

sophiemiddleton:
@FNALbuild run build test

kutschke (approved):
Thanks for the final fixes.  I will merge now.

PR #1025, opened 2023-08-02
db columns names

rlcee:
postgres is case insensitive, so use non-caps for column names.  No change to any functionality or validation.


sophiemiddleton (approved):
Thanks, looks fine

PR #1027, opened 2023-08-03
remove SimParticleTimeOffset (6)

rlcee:
Maybe Ralf should review


brownd1978 (approved):
The fact that the config was using an empty set of maps means removing this should have no effect.

PR #1030, opened 2023-08-03
improve string cleanup

rlcee:
Test code:
```
  std::string test("\t1 2\t3\f\n4 \n");
  string test2 = test;
  cout << "X" << test << "X "<< test.size()<<endl;
  test.erase(test.begin(), std::find_if(test.begin(), test.end(), std::not1(std::ptr_fun<int, int>(std::isspace))));
  cout << "X" << test << "X "<< test.size()<<endl;
  test2.erase(std::remove_if(test2.begin(), test2.end(), ::isspace), test2.end());
  cout << "X" << test2 << "X "<< test2.size()<<endl;

```

output:
```
X       1 2     3

4 
X 11
X1 2    3

4 
X 10
X1234X 4
```
showing the new code is probably doing what was originally intended

### DIFF
--- a/GeneralUtilities/inc/BitMap.hh
+++ b/GeneralUtilities/inc/BitMap.hh
@@ -289,7 +289,7 @@ namespace mu2e {
       if(invalid){
       // try to interpret as a (text) hex string; make sure to edit out spaces first!
         std::string cname(name);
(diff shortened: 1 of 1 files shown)


kutschke:
@brownd1978 I am approving and merging this PR now.  It concerns the piece of BitMap.hh that interprets a string as either a the name of a bit, or failing that, as a hex string.   When you get back you might want to take a look at it in case both Ray and I misunderstood the intent of the original code which did not do as advertised in the comments.

PR #1032, opened 2023-08-04
Kkgeom

brownd1978:
This PR introduces a new geometry package used for extending and sampling track fits.  It relies on features of a new version of KinKal.


brownd1978:
I have left this PR as a draft as some debugging and optimization are still needed.  The basic structure of the geometry classes can still be reviewed.

kutschke:
> I have left this PR as a draft as some debugging and optimization are still needed. The basic structure of the geometry classes can still be reviewed.

I will get people to look at it while you are away.

If I merge the outstanding PR in KinKal, tag and install KinKal on /cvmfs, should I expect it to resolve the build issue?

brownd1978:
> If I merge the outstanding PR in KinKal, tag and install KinKal on /cvmfs, should I expect it to resolve the build issue?
Yes.  But I prefer you not do that, as I'm still finding corner-cases that need to be resolved inside KinKal.

kutschke:
> > If I merge the outstanding PR in KinKal, tag and install KinKal on /cvmfs, should I expect it to resolve the build issue?
> 
> Yes. But I prefer you not do that, as I'm still finding corner-cases that need to be resolved inside KinKal.

Ok. I will leave it as is.

brownd1978:
This PR depends on KinKal v2.4.2

brownd1978:
@FNALbuild run build test

rlcee:
clang-tidy has some potentially useful warnings (of course, as a lower priority..)

brownd1978:
@FNALbuild run build test

kutschke:
@FNALbuild run build test

KinKalGeom/inc/SurfaceId.hh
```diff
@@ -0,0 +1,51 @@
...
+        TCRV=200 // CRV test planes
+      };
+
+    static std::string const& typeName();
+    static std::map<enum_type,std::string> const& names();
+  };
+  using SurfaceIdEnum = EnumToStringSparse<SurfaceIdDetail>;
+  class SurfaceId {
+    public:
+      // copy the constructors
```
kutschke, line 30: Have you considered forwarding the enum_type from SurfaceIdDettail or SurfaceIdEnum ( via using or typedef?)? That way when you need to use the enum names, you can access them as SurfaceId::TT_Mid etc.  

For example at line 33, the type of the first argument would just be SurfaceId::enum_type.

kutschke (approved):
This is a huge step forward and I am looking forward to the collaboraiton using the new capabilities that this adds.

As you requested, I had a close look at SurfaceId.  I have not followed details of when wildcarding is useful but I like the overall design - the top level thing has the simple name.  I left one inline comment for you to consider in a future PR - forwarding the enum_type to the top level.

I skimmed the other classes and left a comment for a future PR inline - in header files that might reasonably be viewed by colleagues of intermediate skill, I prefer not to use auto for return types of accessors.

I have two more general comments to consider for a future PR.  First, is shared_ptr the right kind of pointer to use.  It's important to use when there are issues of lifetime management but I don't see that you have that here (maybe I missed it).   Or do you see multithreading future in a shared_ptr could be important?

Finally, there are more  header-only dependency loops between RecoDataProducts and the Mu2e KinKal code.  This is not fatal but it is icky and I would like to discuss options to clean it up. I think it's the only place that we still have these loops.

I will approve and merge now.

Issue #868, opened 2022-10-12
CRVResponse package contains a mix of reco and MC modules


brownd1978:
This package should be split to separate simulation from reconstruction, and (perhaps) calibration from reconstruction.

rlcee:
It looks like we only run two crv modules in reco: CrvRecoPulsesFinder and CrvCoincidenceClusterFinder.  I could move these two to a new directory called CRVReco.  Neither of them has any MC dependence at first glance.  @ehrlich-uva please correct me..

ehrlich-uva:
The reco modules we are running are the CrvRecoPulseFinder and the CrvCoincidenceFinder. There is also a CrvCoincidenceClusterMatchMC that follows the CrvCoincidenceFinder. It matches the reco clusters with the MC.

The CrvCoincidenceCheck and CrvCoincidenceClusterFinder are outdated modules that shouldn't be used anymore. Perhaps we could remove them.

rlcee:
Hmm, I was fooled by the fact that the reco instance of CrvCoincidenceFinder is called CrvCoincidenceClusterFinder.  Was this on purpose to preserve product names or is it a mistake?

I would go ahead a remove the obsolete modules, they would be easy to get back. 

May I proceed?  I will run all the usual checks.  Thanks

ehrlich-uva:
I don't remember why I kept the name CrvCoincidenceClusterFinder in the fcl, but it's possible that I did it to preserving the product name.
Yes, you may proceed to remove the obsolete modules.
Thank you.

rlcee:
I will close this tomorrow unless there is any objection

rlcee:
reco modules split off to a new CRVReco directory, simulation stays in CRVResponse - issue complete

PR #1034, opened 2023-08-08
split CRVReco

rlcee:
Per #868 create CRVReco and split reco modules from MC in CRVResponse.  Removed two modules at Ralf' s request.  Split CrvHelper into two classes for MC and reco-only.  No change to physics.  Validated with ceSimReco, comicSimReco and reco.


kutschke (approved):
I am not an expert in this code but I had a quick look and I do not see any red flags.

kutschke:
@ehrlich-uva  I am merging without waiting for your review.  I will appreciate it if you look at the PR when you have a chance.  Then let us know if whether or not you see anything that should be addressed.

ehrlich-uva (commented):
Looks Ok. Most of the changes are just moving files to different directories and changing some of the #includes and class names.
@rlcee Thank you for doing this.

PR #1035, opened 2023-08-09
remove three clang-tidy checks

rlcee:
These are ubiquitous and will likely never be fixed, so turn them off for now to reduce the noise and the remainder can actually be examined.

readability-braces-around-statements
`if(condition) expression;  // no braces`
bugprone-narrowing-conversions
`int = size_t  or float = double // coordinate or make conversion explicit`
bugprone-easily-swappable-parameters
`function(float x, float y) // suggestion is to pass a struct`


kutschke (approved):
This looks OK.

PR #1038, opened 2023-08-16
remove some reserve-word include guards

rlcee:
clang-tidy flags variables named with reserve-word style.  The changes in this commit are all of little mnemonic consequence and were non-standard anyway.  Also fixed some typos, and removed guards from some cc files.

Many other flagged variables are a little awkward to change, like `_ADCValue` so I'm skipping them, and really should be eventually addressed by the original author, or will never be fixed, so we could turn off this tidy check.


rlcee:
The goal of working on clang-tidy is to reduce the number of errors to a manageable number so the warnings can be looked at.  For example, in this commit, it is telling us that `HitData_t` is forward declared as both `mu2e::HitData_t` and `mu2e::DeltaFinderTypes::HitData_t` .

kutschke:
> The goal of working on clang-tidy is to reduce the number of errors to a manageable number so the warnings can be looked at. For example, in this commit, it is telling us that `HitData_t` is forward declared as both `mu2e::HitData_t` and `mu2e::DeltaFinderTypes::HitData_t` .

Are both definitiions of HitData_t visible within a single compilation unit?

PR #1044, opened 2023-08-18
Kkgfix

brownd1978:
Fixes for problems seen in the nightly validation of PR 1032.


kutschke (approved):
I am not an expert on everything here. I don't see any redflags.  I am merging now and will do this as the only merge of the day so we can compare the Sat AM nightly to Thursday AM nightly.

PR #1047, opened 2023-08-23
Move CRV MC Assns to Offline

AndrewEdmonds11:
There is a module in TrkAna that makes a big Assns between CrvCoincidenceClusters and CrvCoincidenceClusterMCs in order to allow easy access to data from multiple CrvCoincidenceCluster/CrvCoincidenceClusterMC collections when simulating with different thresholds. That module belongs in Offline and should be run at the end of the reco sequence.

This PR moves the module in CRVResponse and adds the fcl configuration to prolog_v11.fcl. Let me know if anything needs moving around.


brownd1978 (approved):
Looks good.  There will need to be a corresponding change to Production/JobConfig/reco/prolog.fcl to pass validation

brownd1978:
@FNALbuild run build test

AndrewEdmonds11:
@FNALbuild run build test with Mu2e/Production#264

PR #1048, opened 2023-08-25
switch from cryptopp to std::hash

rlcee:
In preparation for e28, we investigated whether cryptopp UPS product was necessary and it appears it can be replaced with std::hash.  By finding the source code and by simple tests, we confirmed that std::hash will produce the same random seed for the same event.  After install std::hash, we ran potSimMT twice and confirmed identical results.


kutschke (approved):
I will want to merge this on a day when there are no other merges because random numbers will change and I want to cleanly establish a new baseline.

kutschke:
There were on other merges to either Production or Offline today.  So I am merging it now.

PR #1050, opened 2023-08-29
removed & in reference to dangling cases

macndev:
This is the first iteration of fixes to make Offline compatible with the new e28 compiler.


kutschke (approved):
It looks good.

PR #1060, opened 2023-08-31
avoid very strict compiler check on amibuous terminated string

rlcee:
The e28 compiler warnings included a strict check on strncpy.  If the source and destination strings are the same size, there is no way that it can be guaranteed that you do not truncate the source string.  If you make the destination string longer, and add an insurance \0, then the compiler accepts it.  I didn't want to make the destination string longer since that might be copied somewhere else in the code, the easiest thing to do is to copy the string and enforce a termination, which is the obvious intended behavior by the author.

In general though, I do not like this code, and this is good example why.  It is c-ish, threaded and locked, which I don't think we want in the middle of code base.  It is also complete overkill - 1100 lines where 30 lines will do.  If Rob and Dave give me the go-ahead I'll just replace it with a short, readable utility routine.

### DIFF
--- a/GeneralUtilities/inc/csv.hh
+++ b/GeneralUtilities/inc/csv.hh
@@ -75,8 +75,12 @@ namespace io {
             with_file_name() { std::memset(file_name, 0, max_file_name_length + 1); }
 
             void set_file_name(const char* file_name) {
-                std::strncpy(this->file_name, file_name, max_file_name_length+1);
-                this->file_name[max_file_name_length] = '\0';
+                size_t i = 0;
+                while( file_name[i] != '\0' && i < max_file_name_length) {
+                  this->file_name[i] = file_name[i];
+                  i++;
+                };
+                this->file_name[i] = '\0';
             }
 
             char file_name[max_file_name_length + 1];


brownd1978 (approved):
I have no problem replacing this utility with something simpler.  Note std::string and related classes contain many parsing features

kutschke:
I don't see a replacement as a priority but I don't have any objection.  There are lots of corner cases to deal with, such as commas inside quoted strings should be ignored for splitting into fields.  If you decide to do it, please test to see that it correctly parses one of Brian Polliack's Bfield parameter files.  it looks like he left us two copies of the same file:

./BFieldTest/test/Mau10_800mm_long.csv
./BFieldGeom/test/Mau10_800mm_long.csv

gaponenko:
Can we find this in an available library instead of re-implementing the parsing of a standard format?

Andrei

kutschke:
> Can we find this in an available library instead of re-implementing the parsing of a standard format? Andrei

The one we have now is an available library that pops up near the top of Google searches.  Everything else that shows up is also open source.  I presume that a really good one is not available because it's too easy to write in house - so no one will pay for it.   The one we have does all of the string handling in C, not C++.  So I agree with Ray that's clunky.  As I said, improving it is not high on my priority list.

rlcee:
Recall that the database tables are CSV, and they are parsed by custom code [here](https://github.com/Mu2e/Offline/blob/26c9f5bfdcd949b2bbcb255a354541663ea9ed91/DbTables/src/DbUtil.cc#L194)
When I came to this point, I believe I looked for library code and did not find anything I was satisfied with.  With these few lines, I have complete control over the file interpretation, and can also respond to things like how query engine delivers text. It handles comments, commas in quotes, and allows quotes in fields, if they are \\" or "" see [the rules](https://mu2ewiki.fnal.gov/wiki/ConditionsData#strings)   The bfield case is actually simpler because it never has imbedded commas or quotes, so can always just be split on commas (have to check if it allows comments) so reading those files should be a few lines of code.

kutschke:
Can you pull that code out to a utility library and use it in both places?  Or is it operating on member data?

rlcee:
> Can you pull that code out to a utility library and use it in both places? Or is it operating on member data?

I can pull it out.  It is in the Db library because I originally had a vision that the db code could be extracted from Offline as a unit and used other places.  That vision has been worn down over the years, and was always unlikely.   There is also already GeneralUtilities dependence (though this would be the first via infrastructure code).
I see there are only two technicalities different from the commonly referenced [spec](https://datatracker.ietf.org/doc/html/rfc4180)
which I think I can patch up.  Really I don't understand why csv.hh is 1100 lines. (Seems to be doing parallel optimized high-volume reads, maybe.)  The replacement will be less than 100.

kutschke:
Yes.  It does do parallel read.  When Brian first deployed it, he wiped out a few grid nodes because he set the number of threads to the number of available cores on the machine, not the number available in slot.   So he disabled MT.

When your find a natural time, go ahead with the refactoring.

PR #1071, opened 2023-09-07
remove CrvParams from ConditionsService

rlcee:
- the digitization period (12.55ns) is now taken from a header
- validated with cosmicSimReco


brownd1978 (approved):

CRVConditions/inc/CRVDigitizationPeriod.hh
```diff
@@ -0,0 +1,10 @@
+#ifndef CRVConditions_CRVDigitizationPeriod_hh
+#define CRVConditions_CRVDigitizationPeriod_hh
+
+namespace mu2e {
+
+  constexpr double CRVDigitizationPeriod = 12.55; // ns
```
ehrlich-uva, line 6: I would like to check this number. We have been using 12.55ns for a long time, but I want to check that it is not actually 12.50ns.

ehrlich-uva: It turns out that 12.55 is an older value, and should be updated to 12.5. 
@rlcee Could you please change it.

ehrlich-uva (approved):

brownd1978:
@FNALbuild run build test

PR #1074, opened 2023-09-08
Add NDOF in KalSeed for BTrk for backwards-compatibility

brownd1978:


rlcee:
I've been thinking we should look at the clang-tidy warnings, since the noise is currently acceptable and they do point out things that make you go "hmmm"
if with identical then and else branches
C++ object pointer is null

kutschke:
> I've been thinking we should look at the clang-tidy warnings, since the noise is currently acceptable and they do point out things that make you go "hmmm" if with identical then and else branches C++ object pointer is null

I agree but let's make that a separate PR  (or series of PRs).

PR #1080, opened 2023-09-12
deprecated arithmetic fixes for e28

macndev:
Another PR for e28 compiler compatibility, enum types have been deprecated in some arithmetic statements.

### DIFF
--- a/TrkReco/src/RobustHelixFit.cc
+++ b/TrkReco/src/RobustHelixFit.cc
@@ -599,7 +599,7 @@ bool RobustHelixFit::initFZ_2(RobustHelixFinderData& HelixData) {
   // Part 2: perform a more accurate estimate - straight line fit
   //-----------------------------------------------------------------------------
   if (nstations_with_hits < 2) return false;//hdfdz = _mpDfDz;
-  else                         hdfdz = xmp*rhel.helicity()._value;
+  else                         hdfdz = xmp*double(rhel.helicity()._value);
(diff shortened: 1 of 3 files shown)


TrkReco/src/RobustHelixFit.cc
```diff
@@ -599,7 +599,7 @@ bool RobustHelixFit::initFZ_2(RobustHelixFinderData& HelixData) {
   // Part 2: perform a more accurate estimate - straight line fit
   //-----------------------------------------------------------------------------
   if (nstations_with_hits < 2) return false;//hdfdz = _mpDfDz;
-  else                         hdfdz = xmp*rhel.helicity()._value;
+  else                         hdfdz = xmp*double(rhel.helicity()._value);
```
kutschke, line 602: Please make this float, not double.  I just checked how hdfdz and xmp are defined.

kutschke (changes requested):
One small change.  I think I gave you the wrong advice before.

TrkReco/src/RobustHelixFit.cc
macndev: done!

kutschke (approved):

TrkReco/src/RobustHelixFit.cc
kutschke: Thanks.

kutschke:
@FNALbuild run build test

PR #1082, opened 2023-09-13
remove deprecated std::binary_function from code using BTrk

macndev:
Last PR for e28 compatibility before changing to new envset. Clean validation with 2000 ceSimReco events.


kutschke:
@FNALbuild run build test

kutschke (approved):

PR #1081, opened 2023-09-12
uninitialized variable changes and a bug fix for e28 compatibility

macndev:
Another PR for e28 compatibility.

PR #1084, opened 2023-09-13
envset change for e28 compiler upgrade

macndev:


kutschke:
The CI failed becuase we are still waiting for PR #1081.

@FNALbuild run build test with #1081

kutschke (approved):
The "test with" CI run done earlier today covers the intervening merge.  So there is no need for a final CI run.

PR #1090, opened 2023-09-18
Dh 230916b

rlcee:
- force runs entry calculation to long long, needed for data run numbers
- separate fcl's for sim and data for printMetadata (only used in next-gen data-handling)

### DIFF
--- a/Print/src/RunSubrunEvent_module.cc
+++ b/Print/src/RunSubrunEvent_module.cc
@@ -96,7 +96,7 @@ void mu2e::RunSubrunEvent::beginSubRun(art::SubRun const& subRun) {
   if (!_printSam) return;
 
   // add this subrun to the list, if it is not already there
-  unsigned long long id = subRun.run()*1000000 + subRun.subRun();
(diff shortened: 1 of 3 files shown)


Print/src/RunSubrunEvent_module.cc
```diff
@@ -96,7 +96,7 @@ void mu2e::RunSubrunEvent::beginSubRun(art::SubRun const& subRun) {
   if (!_printSam) return;
 
   // add this subrun to the list, if it is not already there
-  unsigned long long id = subRun.run()*1000000 + subRun.subRun();
+  unsigned long long id = subRun.run()*1000000ull + subRun.subRun();
```
kutschke, line 99: Something to consider for a future PR.  Art has a art::SubRunID class.  Objects of that type are sortable and printable.  Did you consider using those instead of constructing your own identifier?   I think that your choice of a multiplier of 1E6 is safe but using art::SubRunID is explicitly safe.  It will be epsilon slower but I don' think that's a consider for this code.

kutschke (approved):

PR #1104, opened 2023-10-11
Calo prolog fix

bechenard:
We might want to restructure the whole calo code in the future, but this should solve the issue for now


bechenard:
Somehow I also got the code changes for the previous pull request copied in. I'll close this one and make a new

PR #1108, opened 2023-10-13
Add option to keep all CrvSteps

AndrewEdmonds11:
This PR adds an option to keep all CrvSteps from specific collections during digi compression. This will be used during the digitization of the cosmic datasets where we want to keep CrvSteps from all cosmic ray particles even if they do not get added to a CrvDigiMC.

There will be a corresponding PR in the Production repo to add the new fhicl parameter with its default (which will not use this option).

So far I have tested by digitizing 10 mixed events with just one background frame. Counting the number of CrvSteps, things look like they are working:
 - without this option, first event has 448 steps
 - with this option and only keeping cosmic ray steps, first event has 450 steps
 - with this option and also keeping the single background frame's steps, first event has 5904 steps

(Obviously we wouldn't do this last one but it shows the code is doing what I expect.#)

I've also tested that the reconstruction still runs (it does), and TrkAna can run on the output of reconstruction (it does).

I will do some final validation tests now but if anyone has any comments so far, let me know


kutschke:
@AndrewEdmonds11 Please describe what tests you did to ensure that this is working correctly?  Are there reviewers for this PR who know enough to give a meaningful review?

AndrewEdmonds11:
Hi Rob,

This still isn't finalized yet. I'm working with Yuri and Dave to make sure
this is the way we want to go to solve the problem at hand

Cheers,
Andy

kutschke:
> Hi Rob, This still isn't finalized yet. I'm working with Yuri and Dave to make sure this is the way we want to go to solve the problem at hand Cheers, Andy

@AndrewEdmonds11    has the status changed?

brownd1978:
I suspect this is no longer needed after PR 1111 and 275.

AndrewEdmonds11:
@FNALbuild run build test with Mu2e/Production#277

Compression/src/CompressDigiMCs_module.cc
```diff
@@ -429,6 +434,31 @@ void mu2e::CompressDigiMCs::produce(art::Event & event)
       throw cet::exception("CompressDigiMCs") << "The number of CrvDigiMCs before and after compression does not match ("
                                               << crvDigiMCs.size() << " != " << _newCrvDigiMCs->size() << ")" << std::endl;
     }
+
```
brownd1978, line 437: My understanding of the use case for saving these  CrvSteps was for when they are generated by the primary particle, to study Crv pathologies.  However I don't see any logic here to select steps from the primary (or its daughters).  Isn't that necessary?

brownd1978 (commented):
As with Prodution PR 277 I'd like to understand the plan for using these new options to achieve the specific physics goal.

Compression/src/CompressDigiMCs_module.cc
AndrewEdmonds11: Hi Dave, so there's no need for any logic here because we are just identifying whole CrvStepCollections that we don't want compressed.

Up to this point in the code, we are writing out CrvSteps from all sources (primary, pileup, etc.) that have got into a CrvDigiMC. Here then, we go through any CrvStepCollections that we still want to keep, check if that CrvStep has already been seen, and if it hasn't been seen, we add it to the output collection. 

For the CRV studies, they want all the CrvSteps from the primary stage and so we set _crvStepsToKeep to  ["CosmicS1"] in epilog_cosmic.fcl in Mu2e/Production#277). For all other datasets, we can leave the _crvStepsToKeep array empty.

I hope that makes sense.

brownd1978 (approved):
I understand the intent of this change now.  It looks fine technically, at least for digitization.  I don't think it fully solves the problem as these additional CRV steps will subsequently be removed by SelectRecoMC after reconstruction.  There are several possible fixes for that, depending on the required physics use case.  I suggest we discuss that outside this PR.

brownd1978:
@FNALbuild run build test with Mu2e/Production#277

PR #1110, opened 2023-10-22
Add clarifying comments to the base  G4 prolog.

kutschke:
Also updated the comment field in the validated fhicl source.

### DIFF
--- a/Mu2eG4/inc/Mu2eG4Config.hh
+++ b/Mu2eG4/inc/Mu2eG4Config.hh
@@ -118,7 +118,8 @@ namespace mu2e {
       fhicl::Atom<double> protonProductionCut {Name("protonProductionCut")};
 
       fhicl::Atom<std::string> physicsListName {Name("physicsListName")};
-      fhicl::Atom<bool> turnOffRadioactiveDecay {Name("turnOffRadioactiveDecay"), false};
+      fhicl::Atom<bool> turnOffRadioactiveDecay {Name("turnOffRadioactiveDecay"),
+          Comment("turnOn/Off radioactive decay. If both are false, we get the G4 default, which may be physics list dependent. If both are true, it is a configuration error."), false};
       fhicl::Atom<bool> turnOnRadioactiveDecay {Name("turnOnRadioactiveDecay"), false};
       fhicl::Atom<bool> turnOnThermalNeutronPhysics {Name("turnOnThermalNeutronPhysics"), false};
       fhicl::Atom<bool> decayMuonsWithSpin {Name("decayMuonsWithSpin"), false};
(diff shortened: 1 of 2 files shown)


kutschke:
@resnegfk Is my comment that the state may be physics list dependent, correct?
@pavel1murat Does this resolve your question on slack?

resnegfk (approved):

pavel1murat:
Hi Rob, I'm still at a loss about how 
https://github.com/Mu2e/Offline/pull/1110/commits/1973e1ee0508e4088c6985693bd7eb057ee9a10d#diff-e741fa4f81ba3dac9ec59c44435c20ce2fd40d09bf263edc809708fee1c9cc89R24-R25 

could simultaneously make sense. both say 'turnRadioactiveDecay' , and one is ON, and another one is OFF  . 
Both are booleans, so for me, the two are mutually exclusive options. 

To save time,  I approve but I'm still in the dark

pavel1murat (approved):

kutschke:
@FNAL build run build test

Mu2eG4/inc/Mu2eG4Config.hh
```diff
@@ -118,7 +118,8 @@ namespace mu2e {
       fhicl::Atom<double> protonProductionCut {Name("protonProductionCut")};
 
       fhicl::Atom<std::string> physicsListName {Name("physicsListName")};
-      fhicl::Atom<bool> turnOffRadioactiveDecay {Name("turnOffRadioactiveDecay"), false};
+      fhicl::Atom<bool> turnOffRadioactiveDecay {Name("turnOffRadioactiveDecay"),
+          Comment("turnOn/Off radioactive decay. If both are false, we get the G4 default, which may be physics list dependent. If both are true, it is a configuration error."), false};
       fhicl::Atom<bool> turnOnRadioactiveDecay {Name("turnOnRadioactiveDecay"), false};
```
brownd1978, line 123: I suggest adding a comment here too.  Maybe the same comment?  Or just refer to the above comment?

kutschke: The --print-description output looks like:

             ## turnOn/Off radioactive decay. If both are false, we get the G4 default, which may be physics list dependent. If both are true, it is a configuration error.

              turnOffRadioactiveDecay: false  # default

              turnOnRadioactiveDecay: false  # default

Do you know how to add a properly indented line break?

kutschke: I guessed the line break correctly.  More in a minute.

kutschke: Done and pushed.

brownd1978 (approved):

brownd1978:
@FNALbuild run build test

PR #1113, opened 2023-10-27
Chi2 distance calculation for TNTClusterer, FlagBkgHits with StereoHits

myucel-fnal:


TrkHitReco/data/TrainBkgDiagSTH.dat
```diff
@@ -0,0 +1,16 @@
+tensor_dense3kernel0 18
```
brownd1978, line 1: Can we name this TrainBkgDiagStation?  That will make it clearer

TrkHitReco/data/TrainBkgDiagSTHplane.dat
```diff
@@ -0,0 +1,16 @@
+tensor_dense3kernel0 18
```
brownd1978, line 1: And name this TrainBkgDiagPlane?  Then the existing TrkBiagDiag can be called TrkBkgDiagPanel and everything is explicit.

TrkHitReco/src/TNTClusterer.cc
```diff
@@ -102,22 +108,38 @@ namespace mu2e
...
 
+    int method = distMethod_;
     int ditime(int(maxHitdt_/tbin_));
     unsigned nchanged(0);
     for (size_t ihit=0;ihit<BkgHits.size();++ihit) {
 
       auto& hit  = BkgHits[ihit];
       const auto& chit = chcol[hit.chidx_];
 
+      float dhit(0.), dseed(0.), mindist(0.);
```
brownd1978, line 119: Please move this if outside the loop.  It only needs to be executed once/cluster, not once/hit.

TrkHitReco/src/TNTClusterer.cc
```diff
@@ -102,22 +108,38 @@ namespace mu2e
       clusters[ic].clearHits();
     }
 
+    int method = distMethod_;
```
brownd1978, line 111: Do we need to use different ANN trainings for the different distance methods?

brownd1978 (commented):
Once the chisq based distance method is fully tested we might want to move to that permanently, as the existing distance method is only correct for panel hits.  It also complicates the config having to keep 2 kinds of training files.

brownd1978:
@FNALbuild run build test

TrkHitReco/src/TNTClusterer.cc
```diff
@@ -110,8 +125,9 @@ namespace mu2e
       const auto& chit = chcol[hit.chidx_];
 
       // -- if hit is ok, reassign it right away
-      if (hit.distance_ < dhit_) {
+      if (hit.distance_ < dhit_ || hit.dchi2_ < dhit_) {
```
brownd1978, line 128: This statement makes the choice of distance scale appear ambiguous.  Can you instead use distance_ in both cases?  then you just cut on 1 variable.

brownd1978 (commented):
This is getting closer.  I'd like to further consolidate treatment of distance method, it might require adding a bitmask member.  We should discuss in slack.

TrkHitReco/src/TNTClusterer.cc
myucel-fnal: Yes, this should work

brownd1978:
@FNALbuild run build test

TrkHitReco/data/TrainBkgDiagSTH.dat
brownd1978, line 1: please remove this file

TrkHitReco/data/TrainBkgDiagSTHplane.dat
brownd1978, line 1: And this one

TrkHitReco/src/TNTClusterer.cc
```diff
@@ -6,29 +6,42 @@
...
+
+    tbin_     =1.0;
+    dd2_      = dd_*dd_;
+    maxwt_    = 1.0f/minerr;
+    md2_      = maxdist*maxdist;
+    trms2inv_ = 1.0f/trms/trms;
+
+    switch(distMethod_) {
+      case TNTClusterer::useDistance:
+        dhit_ = 5.;//mm
```
brownd1978, line 35: I am confused why these hard-coded numbers are needed.  Shouldn't they come from the config object at initialization?

myucel-fnal: Misunderstanding on my part, I'll change it back

TrkHitReco/src/TNTClusterer.cc
```diff
@@ -6,29 +6,46 @@
...
+
+    tbin_     =1.0;
+    dd2_      = dd_*dd_;
+    maxwt_    = 1.0f/minerr;
+    md2_      = maxdist*maxdist;
+    trms2inv_ = 1.0f/trms/trms;
+
+    switch(distMethod_) {
+      case TNTClusterer::useDistance:
+        dhit_ = dhitDist_;
```
brownd1978, line 39: i think its better to reuse the same config functions for both dist and chi2 mode parameters, otherwise we must set values that are never used.  It will be clear from the fcl which one  you are configuring, so what unit they are in.  'Distance' is effectively an abstract quantity.

TrkDiag/src/BkgDiag_module.cc
```diff
@@ -138,7 +140,9 @@ namespace mu2e
     _bkghToken{ consumes<BkgClusterHitCollection>(config().BkgClusterHitCollection() ) },
     _mcdigisToken{ consumes<StrawDigiMCCollection>(config().StrawDigiMCCollection() ) },
     _mcprimaryToken{ consumes<PrimaryParticle>(config().MCPrimary() ) }
-  {}
+  {
+    _station = config().ComboHitCollection().label() == "SFlagBkgHits";
```
brownd1978, line 144: Note this logic will fail in the reconfigured sequence I'm working on, as the module instance will allways be FlagBkgHits, regardless of stereo or not.  I suggest instead you add a config parameter describing the expected level of the FlagBkgHits ComboHit collection (uniquepanel, etc ) and then test below on it matching.  Just copy how the config is setup for the outputlevel of FlagBkgHits.

TrkDiag/src/BkgDiag_module.cc
```diff
@@ -221,15 +228,19 @@ namespace mu2e
   void BkgDiag::analyze(const art::Event& event ) {
     if(!findData(event))
       throw cet::exception("RECO")<<"mu2e::BkgDiag: data missing or incomplete"<< std::endl;
-    // check consistency,bkgqcol is eliminated from the code, we should add a different check TO DO
-    //if(_bkgccol->size() != _bkgqcol->size())
-    //  throw cet::exception("RECO")<<"mu2e::BkgDiag: data inconsistent"<< std::endl;
+    if(_station){
```
brownd1978, line 231: You can test directly if you make the expected level a config parameter, see above.

brownd1978 (commented):
Getting closer.  Please move the PR out of draft once you address these issues.

RecoDataProducts/inc/BkgCluster.hh
```diff
@@ -15,27 +17,34 @@ namespace mu2e
    {
        //Default hit count chosen for compuational efficiency
        BkgCluster() {_hits.reserve(16);}
-       BkgCluster(XYZVectorF const& pos, float time) : _pos(pos), _time(time) {_hits.reserve(16);}
-
+       BkgCluster(TwoDPoint point, float time) : _point(point), _time(time) {_hits.reserve(16);_cpoints.addPoint(_point,0);}
+       enum  distMethod {distance,chi2};
```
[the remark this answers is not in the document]
myucel-fnal: I agree completely. I was thinking between Cartesian or Spatial.

TrkHitReco/src/TNTClusterer.cc
myucel-fnal: Ok, will combine the two. I felt like it will be confusing when looking only at the config file.

TrkDiag/src/BkgDiag_module.cc
myucel-fnal: I see how that doesn't work. I was thinking for this check to happen automatically without requiring input from user. I guess this is easier, I'll change it.

brownd1978 (approved):
Looks good, thanks for making the requested changes!

brownd1978:
@FNALbuild run build test

PR #1114, opened 2023-10-28
CentralHelix fit for offspill with temporary seeding

bonventre:
Requires KinKal#183 for fits to converge properly


brownd1978 (approved):
This looks like a pure addition, save for the SimpleTimeCluster change.  Since this requires a KinKal update as well it will take longer to process, should the SimpleTimeCluster change be put into 1120 instead?  Or is it tied to these changes?

brownd1978:
Since this passes CI without the KinKal update we can process it as-is.

PR #1117, opened 2023-10-28
standardize services link pattern

rlcee:
leave only the art macro definition in the "_service.cc" plugin and keep the rest of the code in the mainlib. No functional change.


kutschke:
@FNALbuild run build test

ConditionsService/src/SConscript
```diff
@@ -18,9 +18,16 @@ mainlib = helper.make_mainlib ( [ 'mu2e_RecoDataProducts',
...
+                                  'fhiclcpp_types',
+                                  'tbb',
+                                  'canvas',
                                   'cetlib',
                                   'cetlib_except',
+                                  'CLHEP',
+                                  'boost_filesystem'
                                   ] )
 
 helper.make_plugins( [ mainlib,
```
kutschke, line 33: I bet that we can greatly shorten the link list for the plugin.so.  It will get all of the Mu2e libs via loading the mainlib .so.  Same comment for the other modified SConscripts.

DbService/src/SConscript
```diff
@@ -29,7 +35,7 @@ helper.make_plugins( [ mainlib,
                        'art_Framework_Services_Registry',
                        'art_root_io_tfile_support',
                        'art_root_io_TFileService',
-                       'art_Framework_Services_Optional_RandomNumberGenerator_service',
+                       'art_Framework_Services_Optional_RandomNumberGenerator',
```
kutschke, line 38: Do we really use random numbers in the DbService ?  If so, what's the reason?

ProditionsService/src/SConscript
```diff
@@ -50,7 +69,7 @@ helper.make_plugins( [ mainlib,
                        'art_Framework_Services_Registry',
                        'art_root_io_tfile_support',
                        'art_root_io_TFileService',
-                       'art_Framework_Services_Optional_RandomNumberGenerator_service',
```
kutschke, line 53: I would be surprised if we use random numbers in this service.

kutschke (approved):
I made a few comments for cleaning up the SConscript files.  At least some of the cruft  originally comes from me.  I would like to defer that to a follow-on PR so that I can merge this tonight.

PR #1118, opened 2023-10-28
fix tracking type-limits warnings

rlcee:
these should be reviewed, TrkExt change might have an effect?


kutschke (approved):
This looks OK to me.  To be safe I will merge it on it's own on Sunday.

PR #1120, opened 2023-10-30
Need dead straw mask in makePH when not filtering in makeSH

bonventre:


Mu2eKinKal/test/TTTA.fcl
```diff
@@ -67,7 +67,7 @@ physics.producers.TTmakeSH.FilterHits : false
 physics.producers.TTmakePH.TestFlag : true
 #physics.producers.TTmakePH.FilterHits : true
 physics.producers.TTmakePH.StrawHitSelectionBits : ["EnergySelection","TimeSelection","RadiusSelection"]
-physics.producers.TTmakePH.StrawHitMask          : ["Noisy"]
+physics.producers.TTmakePH.StrawHitMask          : ["Dead","Noisy"]
```
brownd1978, line 70: When simulating we have:
https://github.com/Mu2e/Offline/blob/4e3de16eacd23d3e9e8de089b8f107d1a5aaa470/TrackerMC/src/MakeStrawGasSteps_module.cc#L347
Which I think guarantees we have no steps or digis from dead straws.  Do we need this for real data, where the 'dead' channels still produce noise?  Or is there a gap in the logic?

brownd1978 (commented):
One question about the new flags, otherwise this looks good.

Mu2eKinKal/test/TTTA.fcl
bonventre: We could plausibly have a channel marked dead that we still included in the readout mask and could generate noise, but also this just guarantees in sim that if the dead straw condition is added at the digi or reco stage it will still be applied

brownd1978 (approved):

brownd1978:
@FNALbuild run build test

brownd1978:
@FNALbuild run build test

PR #1125, opened 2023-11-08
Add RecoNullProducer

kutschke:
A variant on NullProducer that is MC free.  (NullProducer has an option to consume random seeds).  And a general cleanup of SConscript.


brownd1978 (approved):
Nice trick with the helper!

PR #1127, opened 2023-11-11
clean up rule of 3 warnings

rlcee:
- fixes warnings from -Wdeprecated-copy -Wdeprecated-copy-dtor
- does not require these switches since root and boost have many warnings
- Why Mu2eInterface/inc/Detector.hh and ConditionsEntity.hh look like they do.  The virtual dtor was declared in header and implemented in the cc (I assume to force object into the lib). I changed it to default, copy ctor and assignment were needed, added those.  The copy ctor cause the default ctor to be deleted (I think).  I added the default ctor declaration in the header, implemented in the cc.  This caused dozens of link errors probably because there were many uses of the ctor without linking to the lib.  So moved the implementation back to the header. At this point the ctor is acting just like the previous default ctor.   If you want the implementation in the cc and all the link lib added several places, I can do that. Just not sure how important it is.
- tested with cesSimReco
- all secondary repos are also fixed


rlcee:
CalPatRec/inc/CalHelixFinderData.hh has several pointers as class members, one which is deleted in the dtor, but none were copied in the copy ctor and assignment operator.  The cleanup here does not change this behavior, just declares it correct.  Someone may want to check this.

Mu2eInterfaces/inc/ConditionsEntity.hh
```diff
@@ -15,7 +15,10 @@ namespace mu2e
   class ConditionsEntity
   {
   public:
+    ConditionsEntity() {}
     virtual ~ConditionsEntity();
+    ConditionsEntity(const ConditionsEntity&);
```
kutschke, line 20: Shouldn't the default c'tor and copy c'tor also be virtual?  I think they need to be override-able in the derived classes.

The same for Detector.hh and ProditionsEntity (and others?).

kutschke:
> CalPatRec/inc/CalHelixFinderData.hh has several pointers as class members, one which is deleted in the dtor, but none were copied in the copy ctor and assignment operator. The cleanup here does not change this behavior, just declares it correct. Someone may want to check this.

Please add this as an issue.

kutschke:
In the description you talk about putting the =default in the .cc file.    I thought it would be OK to have it in the header.  Did you try that?

rlcee:
> In the description you talk about putting the =default in the .cc file. I thought it would be OK to have it in the header. Did you try that?

In Mu2eInterfaces/src/Detector.cc  The original author put the empty dtor in a cc file all by itself.  I assumed this was for a specific reason.  I guess this is micro-optimization, but it might be for some reason I don't know about.  If you are fine with removing Detector.cc and putting it all (empty defaults) in the hh, I'm fine with that.

Mu2eInterfaces/inc/ConditionsEntity.hh
rlcee: Reminding myself about these dusty corners - ctor's can't be virtual because there is no vtable at construction.  I hear this as you are always constructing base or derived but never in a situation where you are constructing maybe dervied_1 or derived_2.  The assignment operator shouldn't generally be virtual since any derived class should provide a assignment operator with a signature of the derived type.  The user has to understand what types they are assigning. The derived default assignment will copy the base and derived class, and a custom derived assignment should deal with the base class members explicitly.  I see the virtual assignment operators of the base type do have a use case, but this is getting quite obscure.  It might be safest to delete the assignment operator of a base class unless some code is actually assigning between base class objects, which we shouldn't be doing in these cases.  Anyway, I will wait for your reply on the Detector.cc question below and fix both together.

kutschke:
@FNALbuild run build test

rlcee:
With the removal of Mu2eInterfaces library, there will need to be coordinated PR's in secondary repos, which I will put in now.

rlcee:
> With the removal of Mu2eInterfaces library, there will need to be coordinated PR's in secondary repos, which I will put in now.

REve, Stntuple, TrkAna, TrackerAlignment PR's should be merged right after this PR

kutschke:
@FNALbuild run build test

kutschke (approved):
Thanks for looking after this.

PR #1132, opened 2023-11-14
Small fix to use TNTClusterer with Spatial method

myucel-fnal:


RecoDataProducts/inc/BkgCluster.hh
```diff
@@ -17,13 +17,13 @@ namespace mu2e
    {
        //Default hit count chosen for compuational efficiency
        BkgCluster() {_hits.reserve(16);}
-       BkgCluster(TwoDPoint point, float time) : _point(point), _time(time) {_hits.reserve(16);_cpoints.addPoint(_point,0);}
+       BkgCluster(TwoDPoint point, float time) : _point(point), _time(time) {_hits.reserve(16);_cpoints.addPoint(_point,0);_pos = point.pos3();_pos = point.pos3();}
```
brownd1978, line 20: _pos is set twice.  It can also be set on initialization.

brownd1978 (changes requested):
small fix needed

RecoDataProducts/inc/BkgCluster.hh
myucel-fnal: Ooops, didn't see that in time.

brownd1978:
@FNALbuild run build test

PR #1135, opened 2023-11-15
Streamline handling of mf::LogInfo messages from GenEventCountReader.

kutschke:
This is a candidate to replace the fcl/messageService.fcl part of  PR #1134


rlcee:
This is a better way to use categories (you can see I did this in DbService, for example).  But why use messagelogger at all?  Either 1) the printout should always be there (which you are doing with this change) then just print it.  2) if it needs to be turned on and off, why not a fcl verbose switch?  **Please see my BIG objection on the other PR.**

kutschke:
@FNALbuild run build test

brownd1978 (approved):

PR #1143, opened 2023-11-20
Remove extra assert the refers to Mu2e driver internals

eflumerf:


brownd1978 (commented):
Is this related to the debug build issue?

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
```diff
@@ -1233,8 +1233,6 @@ void ArtBinaryPacketsFromDigis::fillCrvDataStream(DTCLib::DTC_Event& currentEven
   size_t sz =
       crvData.header.TransferByteCount; // byte count was increased to get full chunks of 16 bytes
 
-  assert(sz < sizeof(mu2e_databuff_t));
```
kutschke, line 1236: Is this a real fix, since the type is no longer defined?  Or is it papering over the issue and we will need to revisit?

eflumerf: The type is no longer defined in Offline-visible code, but it should not matter because the DTC_Event type handles ensuring that the buffers don't go over the 32KB limit

kutschke (approved):

DAQ/src/ArtBinaryPacketsFromDigis_module.cc
eflumerf: https://github.com/Mu2e/artdaq_core_mu2e/blob/8ef7971818275ecc194fa312b837a4a8e6fb0f68/artdaq-core-mu2e/Overlays/DTC_Packets.cpp#L966

eflumerf: https://github.com/Mu2e/artdaq_core_mu2e/blob/8ef7971818275ecc194fa312b837a4a8e6fb0f68/artdaq-core-mu2e/Overlays/DTC_Packets.h#L1179

brownd1978 (approved):

PR #1146, opened 2023-11-23
move constant values to GlobalConstants

rlcee:
Next step in preparation to remove Conditions. Validated with ceSimReco (no changes in validation)


brownd1978 (approved):
Do we expect these values to be bit-identical with previous?

rlcee:
> Do we expect these values to be bit-identical with previous?

yes, tested

kutschke (approved):

PR #1147, opened 2023-11-27
do not use gslcblas

rlcee:
gslcblas has some of the same symbols as openblas.  Since we need opendblas for SOFIE, use openblas everywhere blas is needed

### DIFF
--- a/Mu2eUtilities/src/SConscript
+++ b/Mu2eUtilities/src/SConscript
@@ -47,8 +47,7 @@ mainlib = helper.make_mainlib ( [ 'mu2e_ConditionsService',
                                   'cetlib_except',
                                   'hep_concurrency',
                                   rootlibs,
-                                  babarlibs,
-                                  'gslcblas'
+                                  babarlibs
                                   ] )
 
 # This tells emacs to view this file in python mode.


Mu2eUtilities/src/SConscript
```diff
@@ -47,8 +47,7 @@ mainlib = helper.make_mainlib ( [ 'mu2e_ConditionsService',
                                   'cetlib_except',
                                   'hep_concurrency',
                                   rootlibs,
-                                  babarlibs,
-                                  'gslcblas'
+                                  babarlibs
```
brownd1978, line 50: whys is openblas not needed?

kutschke:
> If it's not a real dependency then it shouldn't be there. It looks like the gsl routines we're using don't involve blas, so its probably correct.
>

I think you  are right.  I am 99% sure that art is configured to load all symbols from libraries when they are loaded.  So it loads all of libgsl.so , this generates references to gslcblas.   I checked all of the undefined symbols in libgsl.so and they are all defined in libopenblas.so  .

kutschke:
@FNALbuild run build test

PR #1148, opened 2023-11-28
move to p054, art v3_14_01

rlcee:
- this will move art version to v3_14_01
- in a quick validation, some differences in simulation were observed, but might be due to a change in clhep, which affects geant
- reco showed no differences
- all secondary repos built OK except REve; the author has been informed and is working on it.
- when REve is understood and fixed, its PR should be merged at the same time as this

Here are three summaries of recent envsets.   The  story is that there were 3 separate changes in flight at the same time but, due to durations of reviews, the merge order was different than the envset creation order.

Sequence of envset use:
P050
P053
P051
P054 = p052

This is the story behind the number sequence:
- p051 was planned as the successor to p050 but was stuck in review. 
--When the review was complete, added the p053 changes to this
- p052 was planned as the successor to p051 and waited for p051 to complete its review
--When the review was complete, added the p053 and p051 changes to this
- p053 was the actually successor to p050 because it had to jump head of p051
- p054 is a clone of p052. So that the working head is again the highest available envset.

This is the change in envset content:
- p050 old art, old ADCM, old KinKal
- p051 old art, new ADCM, new kinkal
- p052 new art, new ADCM, new kinkal
- p053 old art, old ADCM, new kinkal
- p054 copy of p052 (to put it latest in sequence)

### DIFF
--- a/.muse
+++ b/.muse
@@ -1,5 +1,5 @@
 # prefer to build with this environment
-ENVSET p051
+ENVSET p054
 # add these to python path
 PYTHONPATH Trigger/python
 # add Offline/bin to path


rlcee:
- this PR will also bring in root v6_28_06b, which fixes pasha's complaint about this root script seg faulting
```
{
  gInterpreter->AddIncludePath(".");
  int nargs(-100);
}

```

kutschke:
Rerun CI following intervening merges.

@FNALbuild run build test

kutschke (approved):
With the muse upgrade to v4_09_00, the main and all secondary repos build.

PR #1149, opened 2023-11-28
move OnSpillBins to GlobalConstants

rlcee:
This is the last item to move to GlobalConstants because it is true constant.  This will help with other conditions cleanups.  I also change _period in CommonMC/src/EventWindowMarkerProducer_module.cc from float back to double.  This inadvertent change may have triggered and "unstable' in the nightly validation.  Checked with small sample of ceSimReco.


kutschke:
@FNALbuild run build test

kutschke (approved):

PR #1151, opened 2023-11-29
remove ExtMonFNALConditions

rlcee:
- continuing removing obsolete text conditions 
- should be bit compatible
- tested with ceSimReco
- if there is a ExtMon sim test I can run, let me know


gaponenko (approved):
The changes look reasonable, thank you.

gaponenko (approved):

kutschke:
If the fix of the art suite upgrade converges today I will merge that PR but not this.  If it does not converge today, I will merge this PR this evening.

kutschke:
I don't expect the memory increase with p054 to be resolved today.  So merging this now.

PR #1155, opened 2023-12-04
migration of the prolog_trigger files to a separate repository

gianipez:
All the prolog_trigger files were migrated to https://github.com/gianipez/mu2e_trig_config/tree/main , which needs to be moved to the Mu2e project.


Trigger/src/PrescaleEvent_module.cc
```diff
@@ -25,48 +28,68 @@ namespace mu2e
...
-    bool         useFilteredEvts_;
-    int          _debug;
-    //    TriggerFlag  _trigFlag;
-    unsigned     _nevt, _npass;
+    bool filter(art::Event & e) override;
+    virtual bool endRun(art::Run& run ) override;
+
+  private:
+
+    std::vector<EventModeConfig> _eventMode;
```
brownd1978, line 63: Why is this a vector?  Can a single filter instance be used in multiple event modes?

brownd1978 (commented):
As we discussed today, this fails the python build since it removes many of the includes the Offline trigger python depends on.

Also, since the new trigger prolog is independent of any existing prolog, this is the moment to make the cpr and tpr configurations consistent.

Trigger/src/PrescaleEvent_module.cc
gianipez: That's the current idea. The current design is to have the possibility to run the same trigger path in different event-modes with different prescale factors.

brownd1978 (commented):
All the prolog from CRVResponse, which is a simulation package, please return that.
Similarly return all the conditions, like ConditionsService/data/czarnecki_Ti.tbl, these are not related to DAQ.  Let me know if we need a 2nd discussion to define the boundary conditions for what should/should not be migrated to common_config

rlcee:
Dave's comment prompted to take a look and I see this is much more extensive than I thought it was.  
- how was this list made? It seems like you took ~every non-c++ file.  
- I don't understand why these files move (just a sampling of the question)
```
CRVConditions/data/nominal.txt
CaloConditions/data/nominal.txt
DbService/data/connections.txt
DbService/data/create.sql
GlobalConstantsService/data/ParticleList.txt
```
- do you have a plan for the organization of where these files go?
- I think I am not up to date on the plan, so not sure how to comment

gianipez:
> All the prolog from CRVResponse was removed. That is a simulation package. Similarly conditions like ConditionsService/data/czarnecki_Ti.tbl are not related to DAQ, I don't understand why they were removed. Let me know if we need a 2nd discussion to define the boundary conditions for what should/should not be migrated to common_config

Apologies for the mistake. in ConditionsService/data/ there is a file I was typically including in the fcl (conditions_01.txt) and thus I moved the entire folder. For CRVRespnse, the issue was that I got mislead by the missing "MC" component in the folder name.

brownd1978:
We need to talk about this PR.  I have a much clearer picture of what needs to be done, in particular:
- all fcl related to MonteCarlo must remain in Offline
- all fcl and data structures related to the Offline reconstruction exclusively must remain in Offline
- Offline build procedures and all the Musings need to be updated so that we can still run triggers in the Offline environment
- Affected modules  not currently using validated fcl need to be updated
Getting this right requires more work than simply moving directories but it will result in a more maintainable and logical structure.
If done right, I suspect common_config will no longer be necessary, we will only need trigger_config.
I suggest a meeting and go over each package affected by this PR one-by-one to understand what needs to be done.

brownd1978:
Is there anything holding up processing this PR?

gianipez:
> Is there anything holding up processing this PR?

Hi Dave,

I think I sent a slack message to set up a meeting to clarify the points you mentioned. Could you please take a look when you have time?

rlcee:
If musings, proditions or validation are affected, please keep me in the loop

PR #1161, opened 2023-12-09
expand clang-tidy

rlcee:
run on header files and add some checks, most notably, uninitialized class members.


kutschke (approved):
Thanks for adding this.

brownd1978 (approved):

PR #1164, opened 2023-12-13
Production target geometry fixes

mhedges:
This PR addresses 2 concerns regarding the geometry of the production target: the orientation of the production target and the locations of the support tabs and rods.

- The production target was previously put in "backwards", meaning it was rotated 180 degress about y-axis from the nominal orientation. This is the most important and significant issue addressed in this PR.

- The tabs, support rods, and wires that hold the production target were not in the correct positions. These have been moved much closer to the as-built locations, but I suspect are not perfect (especially the support rods). I do not expect these changes to impact much of anything.

In a future PR, I intend to consolidate the production target to a single geometry file, but I feel there is no need to wait to do that before inserting these fixes. Ideally, one should not have to update any FCL to use the fixes in this PR, whereas consolidating geom files would require FCL changes.

### DIFF
--- a/Mu2eG4/geom/ProductionTarget_Hayman_v2_0.txt
+++ b/Mu2eG4/geom/ProductionTarget_Hayman_v2_0.txt
@@ -33,7 +33,7 @@ vector<double> targetPS_heightOfRectangularGapPerSection = {12.135,12.135,12.640
 vector<double> targetPS_thicknessOfGapPerSection                 = {2.0,2.0,1.0}; //mm
 
 int targetPS_nHaymanFins                        = 4;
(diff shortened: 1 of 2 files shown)


krlynch:
Michael and I discussed these changes pretty extensively before he submitted the PR.  We've looked closely at all the visualization relative to both the actual target and the drawings in DocDB.  This is not perfect as he mentions, but it's much closer to the real thing.

krlynch (approved):
Michael and I discussed these changes pretty extensively before he submitted the PR. We've looked closely at all the visualization relative to both the actual target and the drawings in DocDB. This is not perfect as he mentions, but it's much closer to the real thing, and the discrepancies are small enough that they should not affect any physics studies.

PR #1168, opened 2024-01-09
Reseeder prototype 

kutschke:
Prototype of a class to reseed a random engine on each event,using the event id and a user supplied salt.

And an example of using it, using the SeedService seed for the module as the salt.

It still contains debugging printout and should be considered a draft for comment.  I would like to make the default that it only does this when the number of scheudles is > 1.  I would like to do it without having to put a nscheudles argument into each module fcl config.


kutschke:
Closing this PR.  it contains cruft that came along by accident.  I will put a new one in shortly.

PR #1169, opened 2024-01-09
Cleaned up prototype of a re-seeder helper class.

kutschke:
Prototype of a class to reseed a random engine on each event,using the event id and a user supplied salt.

And an example of using it, using the SeedService seed for the module as the salt.

It still contains debugging printout and should be considered a draft for comment. I would like to make the default that it only does this when the number of scheudles is > 1. I would like to do it without having to put a nscheudles argument into each module fcl config.

### DIFF
--- a/Mu2eUtilities/inc/ReSeedByEventID.hh
+++ b/Mu2eUtilities/inc/ReSeedByEventID.hh
@@ -3,30 +3,33 @@
 //
 // Reseed a random engine using the event id and a user supplied salt.
 //
+// Does not work with all CLHEP::RandomEngine concrete types.
+//   - see description in .cc file
+//
 
 #include "Offline/SeedService/inc/SeedService.hh"
 
 #include "canvas/Persistency/Provenance/EventID.h"
 #include "art/Framework/Services/Optional/RandomNumberGenerator.h"
-#include "CLHEP/Random/MixMaxRng.h"
+#include "CLHEP/Random/RandomEngine.h"
 
 namespace mu2e {
 
   class ReSeedByEventID{
 
   public:
-    ReSeedByEventID( CLHEP::HepRandomEngine& engine, SeedService::seed_t salt);
+    ReSeedByEventID( CLHEP::HepRandomEngine& engine, SeedService::seed_t salt, int verbosity=0 );
 
     void reseed ( art::EventID const& id ) const;
 
   private:
 
-    CLHEP::MixMaxRng *  _engine; // the engine to be reseeded
-    SeedService::seed_t _salt;   // user supplied salt.
+    CLHEP::HepRandomEngine& engine_;     // The engine to be reseeded
+    SeedService::seed_t     salt_;       // User supplied salt.
+    int                     verbosity_;  // Control amount of printout; 0=none.
 
   };
 
 } // namespace mu2e
-
 
 #endif
--- a/Mu2eUtilities/inc/ReSeedByEventID.hh
+++ b/Mu2eUtilities/inc/ReSeedByEventID.hh
@@ -1,0 +1,32 @@
+#ifndef Mu2eUtilities_ReSeedByEventID_hh
+#define Mu2eUtilities_ReSeedByEventID_hh
+//
+// Reseed a random engine using the event id and a user supplied salt.
+//
+
+#include "Offline/SeedService/inc/SeedService.hh"
+
+#include "canvas/Persistency/Provenance/EventID.h"
+#include "art/Framework/Services/Optional/RandomNumberGenerator.h"
+#include "CLHEP/Random/MixMaxRng.h"
+
+namespace mu2e {
+
+  class ReSeedByEventID{
+
+  public:
+    ReSeedByEventID( CLHEP::HepRandomEngine& engine, SeedService::seed_t salt);
+
+    void reseed ( art::EventID const& id ) const;
+
+  private:
+
+    CLHEP::MixMaxRng *  _engine; // the engine to be reseeded
+    SeedService::seed_t _salt;   // user supplied salt.
+
+  };
+
+} // namespace mu2e
+
+
+#endif
--- a/Mu2eUtilities/src/ReSeedByEventID.cc
+++ b/Mu2eUtilities/src/ReSeedByEventID.cc
@@ -1,0 +1,35 @@
+#include "Offline/Mu2eUtilities/inc/ReSeedByEventID.hh"
+//#include "art/Framework/EventProcessor/Scheduler.h"
+
+#include "art/Framework/Services/Registry/ServiceHandle.h"
+
+#include "fhiclcpp/exception.h"
+
+#include <iostream>
+#include <string>
+
+namespace mu2e {
+  ReSeedByEventID::ReSeedByEventID( CLHEP::HepRandomEngine& engine, SeedService::seed_t salt):_salt(salt){
+    std::string engineType{art::ServiceHandle<art::RandomNumberGenerator>{}->defaultEngineKind()};
+    std::cout << "Kind:    " << engineType << std::endl;
+    if ( engineType != "MixMaxRng" ){
+      throw cet::exception("BADCONFIG")
+        << "ReSeedByEventID requires that RandomNumberGenerator::defaultEngineKind be MixMaxRng.  It is: " << engineType << "\n";
+    }
+    // Ideally only do this when the number of schedules is >1.
+    // The next line is not working.
+    //std::cout << "Threads: " << art::ServiceHandle<art::Scheduler>{}->num_threads() << std::endl;
+    _engine = static_cast<CLHEP::MixMaxRng*>(&engine);
+  }
+
+  void  ReSeedByEventID::reseed ( art::EventID const& id ) const{
+    std::cout << "Reseed: " << id  << " " << _salt << std::endl;
+    std::array<long,4> seeds;
+    seeds[0] = id.run();
+    seeds[1] = id.subRun();
+    seeds[2] = id.event();
+    seeds[3] = _salt;
+    _engine->setSeeds( seeds.data(), seeds.size() );
+  }
+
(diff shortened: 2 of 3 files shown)


Mu2eUtilities/src/ReSeedByEventID.cc
```diff
@@ -0,0 +1,35 @@
...
+        << "ReSeedByEventID requires that RandomNumberGenerator::defaultEngineKind be MixMaxRng.  It is: " << engineType << "\n";
+    }
+    // Ideally only do this when the number of schedules is >1.
+    // The next line is not working.
+    //std::cout << "Threads: " << art::ServiceHandle<art::Scheduler>{}->num_threads() << std::endl;
+    _engine = static_cast<CLHEP::MixMaxRng*>(&engine);
+  }
+
+  void  ReSeedByEventID::reseed ( art::EventID const& id ) const{
+    std::cout << "Reseed: " << id  << " " << _salt << std::endl;
```
brownd1978, line 26: add if(verbose_)

Mu2eUtilities/src/ReSeedByEventID.cc
```diff
@@ -0,0 +1,35 @@
...
+#include "fhiclcpp/exception.h"
+
+#include <iostream>
+#include <string>
+
+namespace mu2e {
+  ReSeedByEventID::ReSeedByEventID( CLHEP::HepRandomEngine& engine, SeedService::seed_t salt):_salt(salt){
+    std::string engineType{art::ServiceHandle<art::RandomNumberGenerator>{}->defaultEngineKind()};
+    std::cout << "Kind:    " << engineType << std::endl;
+    if ( engineType != "MixMaxRng" ){
```
brownd1978, line 15: Why is this restriction necessary?  it seems setSeeds is defined in the base class.

brownd1978 (commented):
How will this interact with the art framework RN checkpointing?

Mu2eUtilities/src/ReSeedByEventID.cc
kutschke: Good catch.  Some time ago it was not in the base class and I did not check again.

Mu2eUtilities/src/ReSeedByEventID.cc
kutschke: Yes.  That's why it was marked draft.

kutschke:
> How will this interact with the art framework RN checkpointing?

The checkpointing captures the state of the engine at the start of each event and writes it to the output file as a data product; on readback, if asked, it will restore the state at the start of each event.  The code in this PR unconidtionallly overwrites the state.  So RN checkpoint is not necessary at all if all engines use this technique.  If some engines, but not others, use this techinque, then the checkpointing will continue to work for the other engines.

The only hesitation that I have about using this technique everywhere is that it might dominate CPU time for very light weight modules.  Ray did some measurements I can dig up.

Mu2eUtilities/src/ReSeedByEventID.cc
kutschke: Actually, I need to check if all engines actually use all 4 seeds.  If some silently look at one only entry in the array then we need some sort of test, perhaps excluding this engines.  Perhaps alternate behaviour.

kutschke: I confirm that HepJamesRandom::setSeeds( const long*, int)  silently ignores all but the first seed.

https://gitlab.cern.ch/CLHEP/CLHEP/-/blame/develop/Random/src/JamesRandom.cc#L262

So the question is whitelist or blacklist?

brownd1978:
Does mixmax take exactly 4?   Are there any engines that can take 4?
It’s too bad the abstract interface doesn’t provide a ‘seedsize’ function
to allow flagging this without having to know how individual algorithms are
coded.

For our use case I guess we will use mixmax for a while, so whitelist is
OK, but this is a point of manual maintenance if we change in future.

Office Phone (510) 486-7261

brownd1978:
I'm not sure putting the run number into the seed is necessary: that's fixed at the campaign level (ie MDC2020 is essentially 1 run number).  If mixmax reallly uses all seeds I guess it doesn't hurt.

kutschke:
> Does mixmax take exactly 4? 
It takes up to 4, with the number to be used specified as the second argument.

> Are there any engines that can take 4? 
I have not yet checked any except HepJamesRandom and MinMax.  There are 13 more to check.

> It’s too bad the abstract interface doesn’t provide a ‘seedsize’ function to allow flagging this without having to know how individual algorithms are coded. 
We can ask for it.  I will find out whom to ask.

> For our use case I guess we will use mixmax for a while, so whitelist is OK, but this is a point of manual maintenance if we change in future. 
I agree.  I will keep it as a whitelist so that it fails hard if we move to another engine withiout doing the maintenance.  I will improve the error message and added comments to the code.

> I'm not sure putting the run number into the seed is necessary: that's fixed at the campaign level (ie MDC2020 is essentially 1 run number). If mixmax reallly uses all seeds I guess it doesn't hurt.
Yes.  It does not hurt.

gaponenko:
Hi,

Minmax is pretty unique in accepting 4 independent seeds in an efficient way.
I think CERNLIB RANLUX allowed for more than one seed, in an inefficient way (could take
dozens of minutes to initalize), unless I misremember.  The current version has only
a single seed, https://luscher.web.cern.ch/luscher/ranlux/guide.pdf

Mersenne Twister is also a single seed.

Andrei

Mu2eUtilities/src/ReSeedByEventID.cc
gaponenko: I think whitelisting MINMAX here is the right thing to do.
But why is the check done  on an object pulled from the service handle instead of using the constructor arg (engine)?

Mu2eUtilities/inc/ReSeedByEventID.hh
```diff
@@ -0,0 +1,32 @@
...
+
+  public:
+    ReSeedByEventID( CLHEP::HepRandomEngine& engine, SeedService::seed_t salt);
+
+    void reseed ( art::EventID const& id ) const;
+
+  private:
+
+    CLHEP::MixMaxRng *  _engine; // the engine to be reseeded
+    SeedService::seed_t _salt;   // user supplied salt.
```
gaponenko, line 25: Identifiers that start with an underscore  and are in the external scope are reserved  in gcc
https://www.gnu.org/software/libc/manual/html_node/Reserved-Names.html
Of course class member names are not global, but to a reader of a function that uses _salt it may not be obvious whether this is a member or a reserved global.     (And  underscore followed by a capital letter are reserved by C++ standard.)  I would advocate to just avoid names that start with an underscore in our new code;  use salt_ instead.

gaponenko: The formatting got messed up:  I suggested to put the underscore at the end of salt instead of at the beginning

kutschke:
Strictly speaking reseeding is only needed when we run with nschedules > 1. 

I have decided to reseed for all values of nschedules so that nschedules=1 has the same behaviour as other values.

Does anyone feel that nschedules=1 should retainn the old behaviour  ( continue the previous engine state )?

brownd1978:
On Tue, Jan 9, 2024 at 6:19 PM Rob Kutschke ***@***.***>
wrote:

> Strictly speaking reseeding is only needed when we run with nschedules > 1.
>
Reproducing a problem in a single-threaded digitization job by jumping to a
particular event requires reseeding the detector response simulation
modules.  Am I missing something?

Dave

-- 
Office Phone (510) 486-7261

kutschke:
Thanks for the comments and suggestions made during the review.  I believe that I have implemented them all.  I will remove the draft status now.

kutschke:
>> On Tue, Jan 9, 2024 at 6:19 PM Rob Kutschke ***@***.***> wrote: Strictly speaking reseeding is only needed when we run with nschedules > 1.

> Reproducing a problem in a single-threaded digitization job by jumping to a particular event requires reseeding the detector response simulation modules. Am I missing something? Dave

Until now, the only way to do that was to run all events once and save the engine state in a data file; then subsequent runs could jump to the event of interest.

If we adopt the reseeder for all engines then it will be easier to jump to a given event; no need to run a job to save engine state.

As always, for jumping to an event to work, all code must be compliant with the red text in: https://mu2ewiki.fnal.gov/wiki/RandomNumbers#CLHEP_distributions

gaponenko (approved):
Looks good!

brownd1978 (approved):

kutschke:
@FNALbuild run build test

rlcee:
Two quick questions (I don't feel strongly - I'll merge today either way)
- I was expecting this reseed to be dependent on  fcl parameter choosing it.  Is it your intention to make this method default?
- could you hash the eventID into a single seed and allow all engines? Or does that reduce the random quality?

kutschke:
> Two quick questions (I don't feel strongly - I'll merge today either way)
> 
> * I was expecting this reseed to be dependent on  fcl parameter choosing it.  Is it your intention to make this method default?

I thought about that.  We  only need to do this in modules that might sometimes need to be run with nschedules > 1; for now that means only modules that are run in a job that also runs G4.  So, for now, it just the digi maker modules and maybe some of the event generator modules.  For those modules we want them to behave the same regardless of the value of nschedules.  So there is no reason to ever turn it off.  So I did not make a fcl parameter to control it.

> * could you hash the eventID into a single seed and allow all engines? Or does that reduce the random quality?

I am concerned about the size of the seed space for engines with 32 bit seeds.   For 32 bits, you have a 50% chance of identical seeds in about  32K tries. 

  As I understand MixMax, all values of the seed pace produce independent patterns of  random variates.  So using all 4 seeds means that we never need to worry about the small seed space.

   It looks likely to me that MixMax will be our choice for some time and we don't need to consider other options until it is no longer our choice.

PR #1171, opened 2024-01-10
Update TrkQual for python-based trainings

AndrewEdmonds11:
Hi Everyone,

In this PR, I've made code changes so that we can use python-based trainings for TrkQual.

A few updates from DocDB-47519:
 - I haven't removed the ```TrkQual``` data product since it is being used by a couple of modules. I'm waiting to hear back from their maintainers about what to do. I can put in a separate PR for that work
 - I couldn't find an already existing ```MVAResult```-type class so I've kept it here
 - I haven't removed the ProditionsService stuff that exists for TrkQual. I will think about removing it again when I know what to do about the TrkQual data product

Cheers,
Andy


kutschke (commented):
I think that you also need to edit:
  RecoDataProducts/src/classes.h
  RecoDataProducts/src/classes_def.xml 

Without that you won't be able to write data products that use this class.  Or am I missing something

Any thoughts on a policy to document where the trkqual.dat and .hxx came from?  For example a tag in your repo, plus names of input files used for training?    Where would it go - in the commit comments?  Somewhere else?

RecoDataProducts/inc/MVAResult.hh
```diff
@@ -0,0 +1,17 @@
+//
+// A datsa product that stores the result of any MVA/ML/AI algorithm
```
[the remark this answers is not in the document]
AndrewEdmonds11: Fixed in 778a36b76

AndrewEdmonds11:
Thanks, Rob. Answers below:

> I think that you also need to edit: RecoDataProducts/src/classes.h RecoDataProducts/src/classes_def.xml
> 
> Without that you won't be able to write data products that use this class. Or am I missing something
> 

Good catch. We weren't writing out TrkQual to an art file (it runs in the same job as TrkAna) so I didn't catch this in any of my tests. I've updated the classes.h and classes_def.xml (commit 87983abca) in case we want to do this in the future

> Any thoughts on a policy to document where the trkqual.dat and .hxx came from? For example a tag in your repo, plus names of input files used for training? Where would it go - in the commit comments? Somewhere else?

I've added a reference to the TrkQual repo commit to the top of the .hxx file (commit 6db7e33e0). Unfortunately, we can't add a comment to the .dat file, which may be more important (since we can retrain the same model for a different .dat but use the same .hxx). The input dataset is given in the python notebook so that is documented.

Longer term, we probably want a policy for all our ML algorithms to have the python notebook in DocDB with the input variable distributions, training dataset etc. all documented. Perhaps also referencing it on the Mu2eWiki?

kutschke:
@FNALbuild run build test

kutschke:
> > Any thoughts on a policy to document where the trkqual.dat and .hxx came from? For example a tag in your repo, plus names of input files used for training? Where would it go - in the commit comments? Somewhere else?
> 
> I've added a reference to the TrkQual repo commit to the top of the .hxx file (commit [6db7e33](https://github.com/Mu2e/Offline/commit/6db7e33e03cee906c0bb2e732aba61e370d472e7)). Unfortunately, we can't add a comment to the .dat file, which may be more important (since we can retrain the same model for a different .dat but use the same .hxx). The input dataset is given in the python notebook so that is documented.
> 
> Longer term, we probably want a policy for all our ML algorithms to have the python notebook in DocDB with the input variable distributions, training dataset etc. all documented. Perhaps also referencing it on the Mu2eWiki?

I like the idea of docdb  and prefer it to the wiki - it gives you a natural way to document version evolution.  Then put the docdb number, including version number, as a comment in the .hxx file. Docdb and the wiki both give you a place to put as detailed a writeup as is needed.  Can you do that for this PR?

Can you asl the authors about comments in the .dat files?  If it's not supprted can you request it as a new feature?

Unless comments in the .dat come through for us, there is one thing that might not be robust in this system.  How to document which .dat files can be used with which .hxx file.   We can document it in docdb but people have to remember to do it ( I mean if you retrain with new input files but an otherwise identical model, you will produce a new .dat file that can be read with the same .hxx file, right?)

I guess the solution is that review must check for proper documentation.  Do you see a more robust solution?

rlcee:
> How to document which .dat files can be used with which .hxx file

Name the hxx
TrkQual_typeN.hxx
Name the dat:
TrkQual_typeN_vM.dat

Any typeN can read any typeN dat.  This also allows you to commit multiple version of hxx. typeN could increment N, or be some other mnemonic.

RecoDataProducts/src/classes_def.xml
```diff
@@ -245,6 +245,10 @@
  <class name="mu2e::TrkQualCollection"/>
  <class name="art::Wrapper<mu2e::TrkQualCollection>"/>
 
+ <class name="mu2e::MVAResult"/>
+ <class name="mu2e::MVAResultCollection"/>
+ <class name="art::Wrapper<mu2e::MVAResultCollection>"/>
+
```
kutschke, line 251: If you want to make a data product that is a single MVAResult, you also need:

```
<class name="art::Wrapper<mu2e::MVAResult>"/>
```

kutschke:
@FNALbuild run build test

kutschke:
@brownd1978 A gentle reminder that I asked you to review this PR.  Please let me know when you expect to be able to get to it.  Thanks.

RecoDataProducts/src/classes_def.xml
```diff
@@ -245,6 +245,10 @@
  <class name="mu2e::TrkQualCollection"/>
  <class name="art::Wrapper<mu2e::TrkQualCollection>"/>
 
+ <class name="mu2e::MVAResult"/>
+ <class name="mu2e::MVAResultCollection"/>
+ <class name="art::Wrapper<mu2e::MVAResultCollection>"/>
```
kutschke, line 250: Do you expect to put a single MVAResult into the event as a data product, or only collections?  If you want the former, you also need the Wrapper<mu2e::MVAResult>.

AndrewEdmonds11: Done in 2abe4fb

kutschke: Thanks!  One last CI run and I will merge.

kutschke:
@FNALbuild run build test

kutschke (approved):

PR #1172, opened 2024-01-17
Add CMake Build System

eflumerf:
Added by Rob Kutschke:

This PR also makes minor changes to Mu2eUtilities/inc/TriggerResultsNavigator.hh and .cc.


kutschke (approved):
A few things.

Merging this PR is not a commitment on the part of Offline to adopt cmake now or on any particular time scale.  My opinion is that there remains a lot of work to do before it would make sense for Offline to adopt cmake; at this time we have higher priorities.

Is there any documentation focused at new users that describes when it is necessary to update CMakeLists.txt files and how to do so?  Until that documentation is vetted by us we are not making any commitment to update CMakeLists.txt files when needed; you are on your own for that.

This PR should have been two PRs.  One to do the job that was advertised in the title and a second to make the modifications to TriggerResultsNavigator.   In the future please submit something like this as two PRs.   In the near future I plan to start rejecting PRs on these grounds.  For this PR I included this information in the PR descirption ( which had been empty).

kutschke:
@eflumerf I did not notice the whitespace errors before merge this PR.  See

https://buildmaster.fnal.gov/buildmaster/job/GitHubPRTests/job/mu2e-offline-build-test//1938/artifact/whitespace_errs.log

( You need to be onsite or on VPN to see this link.  Or you can curl -O it when logged into a lab machine )..

Please promptly put in a followup PR to fix these issues.

PR #1181, opened 2024-01-30
create sim version of tracker alignment prodition and db tables

rlcee:
In discussions it was decided to create a permanent alignment system for simulation purposes, in parallel to the reco system.  This PR implements that plan.   The new prodition adds minimal code since it can be templated or derived from the reco code.
It is off by default.

I tested that there is no change in default ceSimReco, and that one can load different alignments in the different proditions.  I'd advise an expert to further test it at the full detector physics level.

The alignment tables X are duplicated as tables XSim.  These can be loaded from text now for testing, and then we can decide when to create the tables in the db and add to conditions sets.

At the module level, one can access both proditions by instantiating :
  ProditionsHandle<Tracker> _trkr_h;
  ProditionsHandle<Tracker> _trks_h{"Sim"};
and in fcl
services.ProditionsService.alignedTracker.useDb: true
services.ProditionsService.alignedTrackerSim.useDb: true

This sets a pattern for variants of proditions, which will probably come up again.

### DIFF
--- a/DbTables/inc/TrkAlignElement.hh
+++ b/DbTables/inc/TrkAlignElement.hh
@@ -80,6 +80,8 @@ class TrkAlignPanel : public TrkAlignElement {
   constexpr static const char* cxname = "TrkAlignPanel";
   TrkAlignPanel() :
       TrkAlignElement(cxname, "trk.alignpanel", StrawId::_nupanels) {}
+  TrkAlignPanel(const char* name, const char* dbname) :
+    TrkAlignElement(name, dbname,  StrawId::_nupanels) {}
 };
 
 class TrkAlignPlane : public TrkAlignElement {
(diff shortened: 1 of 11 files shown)


kutschke:
@bonventre I just added you as a reviewer to this PR.  Please work with Ray to understand if it meets your needs for separate sim and reco tables.

kutschke:
@ehrlich-uva I just added you as a reviewer to this PR.  Please work with Ray to understand if this meets your needs for conditions for sim and reco.  If it does, I propose the following path forward.  When this review is finished, I will merge it.  Then you can update PRs Offline 1163 and Production 287 to use this system instead of the text files  Does that work you.

@rlcee Does that plan work for you?

rlcee:
> @ehrlich-uva I just added you as a reviewer to this PR. Please work with Ray to understand if this meets your needs for conditions for sim and reco. If it does, I propose the following path forward. When this review is finished, I will merge it. Then you can update PRs Offline 1163 and Production 287 to use this system instead of the text files Does that work you.
> 
> @rlcee Does that plan work for you?

I think, Ralf can confirm, that he does not need the sim and reco version of a single table in his current PR.  But that might be a future development.

ehrlich-uva:
@rlcee The CRV time offsets need to have different sim and reco tables.

bonventre:
Thanks Ray, this is exactly what I was imagining. I managed to get straw digi making and TrkAna mc truth working from this PR pretty easily. I'll note that it does include tables that we currently won't need (TrkAlignTracker/Panel/Plane will all be zero), but could possibly be used in the future if we end up revisiting the g4 geometry. For now we'll also want a config flag to sum the straw and wire misalignments in AlignedTrackerMaker for sim which is a small easy change.

bonventre (approved):

kutschke (approved):

PR #1185, opened 2024-02-09
Quick fix for the crashes in g4test_03MT.fcl and transportOnly.fcl

kutschke:
... that were introduced by Offline PR #1183.

See commit comments for details.


kutschke:
@FNALbuild run build test

rlcee (approved):
This look like what I expected.  One minor comment, do you think the services might be Sim instead of SimAndReco?

brownd1978 (approved):
I can't judge whether this does what's intended but the code structure looks good.

kutschke:
> This look like what I expected. One minor comment, do you think the services might be Sim instead of SimAndReco?

I turned off reco as an expediency to move 1185 forward.  I need to speak with Krzzsztof before I make it permanent.  I it is permanent I will change it to just Sim.  All before the end of the week.

I would like Ray to merge it so I am not merging my own PR.

PR #1188, opened 2024-02-12
Have TrackQuality follow MU2E_SEARCH_PATH for dat files

AndrewEdmonds11:
This solves a bug where the user has to be in the directory above Offline for TMVA::SOFIE to find the .dat file. This uses ConfigFileLookupPolicy so that it follows MU2E_SEARCH_PATH. I've done a test and it works.

### DIFF
--- a/TrkDiag/inc/TrkQual_ANN1.hxx
+++ b/TrkDiag/inc/TrkQual_ANN1.hxx
@@ -65,7 +65,7 @@ Session(std::string filename ="") {
    std::ifstream f;
    f.open(filename);
    if (!f.is_open()){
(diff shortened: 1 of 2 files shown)


brownd1978 (approved):

kutschke:
@AndrewEdmonds11 - I want to check one thing.  This always either worked right or failed hard.  So it should not change nightly validation.  Do I understand that correctly?

AndrewEdmonds11:
> @AndrewEdmonds11 - I want to check one thing. This always either worked right or failed hard. So it should not change nightly validation. Do I understand that correctly?

Hi Rob - that's right. I'm not sure if the TrackQuality module runs in nightly validation though. It's usually only run in the same job as TrkAna

PR #1194, opened 2024-02-17
Remove unneeded lib

rlcee:
Found this lib was not in our spack builds, then found we could link without it.  Let me know if there are link options or cases I'm not aware of.


kutschke:
@FNALbuild run build test

kutschke (approved):

PR #1195, opened 2024-02-18
Gamma conversion generator

sdifalco:
New complete generator for gamma conversion based on previous Michael Makenzie's module in Su2020 repository. As agreed with Rob the generator will go in the Mu2eG4 directory instead of the usual EventGenerator because it uses some G4 libraries. The SConscript in Mu2eG4/src also needs a fix: mu2e_DataProducts needs to be added in helper.make_plugins

### DIFF
--- a/Mu2eG4/src/GammaConvGenerator_module.cc
+++ b/Mu2eG4/src/GammaConvGenerator_module.cc
@@ -1,0 +1,214 @@
+// Generates e+/e- from converted photons.
+// This generator uses ...
+//
+// S. Di Falco, 2023
+
+#include <iostream>
+#include <string>
+#include <vector>
+#include <cmath>
+#include <memory>
+
+#include "CLHEP/Vector/LorentzVector.h"
+#include "CLHEP/Random/RandFlat.h"
+#include "CLHEP/Units/PhysicalConstants.h"
+
+#include "fhiclcpp/types/Atom.h"
+#include "fhiclcpp/types/Sequence.h"
+
+#include "art/Framework/Core/EDProducer.h"
+#include "art/Framework/Principal/Event.h"
+#include "art/Framework/Services/Registry/ServiceHandle.h"
+#include "Offline/SeedService/inc/SeedService.hh"
+#include "Offline/DataProducts/inc/PDGCode.hh"
+#include "Offline/MCDataProducts/inc/StageParticle.hh"
+#include "Offline/MCDataProducts/inc/PhysicalVolumeInfoMultiCollection.hh"
+#include "Offline/Mu2eUtilities/inc/simParticleList.hh"
+#include "Offline/MCDataProducts/inc/GenParticle.hh"
+#include "Offline/Mu2eUtilities/inc/GammaPairConversionSpectrum.hh"
+#include "Offline/Mu2eG4/inc/findMaterialOrThrow.hh"
+
+// Geant4 includes
+#include "Geant4/G4Material.hh"
+
+namespace mu2e {
+
+  //================================================================
+  class GammaConvGenerator : public art::EDProducer {
+  public:
+    struct Config {
+      using Name=fhicl::Name;
+      using Comment=fhicl::Comment;
+      fhicl::Atom<std::string> physVolModuleLabel{Name("physVolModuleLabel"),Comment("module creating physical volumes")};
+      fhicl::Atom<std::string> physVolProcessLabel{Name("physVolProcessLabel"),Comment("module creating physical volumes"),"Primary"};
+      fhicl::Atom<art::InputTag> inputSimParticles{Name("inputSimParticles"),Comment("A SimParticleCollection with input converted photons.")};
+      fhicl::Atom<std::string> stopMaterial{Name("stopMaterial"),Comment("G4 name of the material where the photon has converted (temporary)."),"ST_Wires"};
+      fhicl::Atom<unsigned> verbosity{Name("verbosity")};
+      fhicl::Atom<bool>   useCorrelatedAngleOverKE{Name("useCorrelatedAngleOverKE"), Comment("Flag to use correlated e+e- cos/KE"), false};
+    };
+
+    using Parameters= art::EDProducer::Table<Config>;
+    explicit GammaConvGenerator(const Parameters& conf);
+
+    virtual void produce(art::Event& event) override;
+
+    //----------------------------------------------------------------
+  private:
+
+    double particleMass_;
+
+    // Module label of the module that made the volume.
+    std::string physVolModuleLabel_;
+    // Module label of the last process that wrote the volume list.
+    std::string physVolProcessLabel_;
+
+    art::ProductToken<SimParticleCollection> const simsToken_;
+    std::string stopMaterial_;
+
+    unsigned verbosity_;
+
+    art::RandomNumberGenerator::base_engine_t& eng_;
+    CLHEP::RandFlat randFlat_;
+    bool   useCorrelatedAngleOverKE_; //correlate the cos/ke for the e+ and e-
+
+    GammaPairConversionSpectrum* spectrum_; //pair production spectrum
+    enum {kMaxConversionMaterialElements = 10};
+    ProcessCode process;
+  };
+
+  //================================================================
+  GammaConvGenerator::GammaConvGenerator(const Parameters& conf)
+    : EDProducer{conf}
+      //    , particleMass_(GlobalConstantsHandle<ParticleDataList>()->particle(static_cast<PDGCode::type>(PDGCode::e_minus)).mass())
+    , physVolModuleLabel_{conf().physVolModuleLabel()}
+    , physVolProcessLabel_{conf().physVolProcessLabel()}
+    , simsToken_{consumes<SimParticleCollection>(conf().inputSimParticles())}
+    , stopMaterial_{conf().stopMaterial()}
+    , verbosity_{conf().verbosity()}
+    , eng_{createEngine(art::ServiceHandle<SeedService>()->getSeed())}
+    , randFlat_{eng_}
+    , useCorrelatedAngleOverKE_(conf().useCorrelatedAngleOverKE())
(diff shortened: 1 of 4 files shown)


kutschke:
Adding a comment that I sent to Stefano directly in response to a question from him.  it may be of general interest.

> Two things about accessing the log file.  First, the log files takes a little while (minutes?) to reach their archival location.  Often I retry a few minutes later and it works.  It worksed for me a minute ago.
>  
>    Second, and you may already know this: read the section on accessing log files in https://mu2ewiki.fnal.gov/wiki/GitHubWorkflow#GitHub_Pull_Request_Procedures_and_FNALbuild
>  
>   The log file says that the problem occurs when compiling the module:
>  
> Offline/Mu2eG4/src/GammaConvGenerator_module.cc:28:10: fatal error: Offline/Mu2eUtilities/inc/GammaPairConversionSpectrum.hh: No such file or directory
>    28 | #include "Offline/Mu2eUtilities/inc/GammaPairConversionSpectrum.hh"
>  
> It looks like you are missing a header file – does it also have a matching .cc?

sdifalco:
@FNALbuild run build test

Mu2eG4/src/GammaConvGenerator_module.cc
```diff
@@ -0,0 +1,214 @@
...
+      throw   cet::exception("NO PHYSICALVOLUMEINFO COLLECTION")
+        <<"GammaConvGenerator::produce(): no suitable PhysicalVolume collection\n";
+    }
+    if (PhysicalVolumeInfos->empty()){
+      throw   cet::exception("EMPTY PHYSICALVOLUMEINFO COLLECTION")
+        <<"GammaConvGenerator::produce(): PhysicalVolume collection is empty\n";
+    }
+    //const auto& vols = PhysicalVolumeInfos->at(gammapart->simStage());
+    /*
+    PhysicalVolumeInfoSingleStage const* vols = &PhysicalVolumeInfos->at(gammapart->simStage());
```
kutschke, line 148: Please either remove the commented out code or put it behind a verbosity level.  The one exception to this request is if the commented out code is something in development.

Mu2eG4/src/GammaConvGenerator_module.cc
```diff
@@ -0,0 +1,214 @@
...
+    //const auto& endInfo = vols.at(gammapart->endVolumeIndex());
+    PhysicalVolumeInfo const& endInfo = vols->at(cet::map_vector_key(gammapart->endVolumeIndex()));
+
+    if (verbosity_>1){
+      std::cout << "Gamma stop found in: " << endInfo.materialName() << std::endl;
+    std::cout << "Requested material is:" << " " << stopMaterial_ << std::endl;
+    }
+    G4Material*stopMat = findMaterialOrThrow((G4String const& )endInfo.materialName());
+    */
+    G4Material*stopMat = findMaterialOrThrow((G4String const& )stopMaterial_);
```
kutschke, line 166: I am 99% sure that geometry and materials are defined at beginRun time.  So you can move the lookup of the element fractions to beginRun and save only the information you need.

Mu2eG4/src/GammaConvGenerator_module.cc
```diff
@@ -0,0 +1,214 @@
...
+    const auto simh = event.getValidHandle<SimParticleCollection>(simsToken_);
+    const auto gammas = simParticleList(simh,PDGCode::gamma,ProcessCode::conv);
+
+    if(gammas.empty()) {
+      throw   cet::exception("NO GAMMAS")
+        <<"GammaConvGenerator::produce(): no suitable converted gamma in the input SimParticleCollection\n";
+    }
+
+    // Take the converted gamma as SimParticle
+    const auto gammapart = gammas.at(eng_.operator unsigned int() % gammas.size());
```
kutschke, line 110: This is not guaranteed to uniformly sample the set of gammas.   The effect is negligible.  Did you do it this way since it's advertised as being faster than randFlat_,fireInt( gammas.size());

sdifalco: I have inherited this from Michael's code. If you think it's better I can use randFlat.

Mu2eUtilities/src/GammaPairConversionSpectrum.cc
```diff
@@ -0,0 +1,257 @@
...
+    const double u = (border > rndnum1) ? uu*a1 : uu*a2;
+    return u;
+  }
+
+  void GammaPairConversionSpectrum::initializeElementData() {
+    for(int z = 1; z < 121; ++z) {
+      double z3 = pow(z, 1./3.);
+      //from G4Element::ComputeCoulombFactor()
+      const double k1 = 0.0083, k2 = 0.20206, k3 = 0.0020, k4 = 0.0369;
+      const double alpha = 1./137.036; //fine structure constant
```
brownd1978, line 241: alpha is defined in
"CLHEP/Units/PhysicalConstants.h"
please use the value from there instead.

brownd1978 (changes requested):

Mu2eUtilities/src/GammaPairConversionSpectrum.cc
sdifalco: Replacing with CLHEP::fine_structure_constant

Mu2eG4/src/GammaConvGenerator_module.cc
sdifalco: changing to what is suggested

Mu2eG4/src/GammaConvGenerator_module.cc
sdifalco: geometry readout moved to beginrun

Mu2eG4/src/GammaConvGenerator_module.cc
sdifalco: removed

sdifalco:
@FNALbuild run build test

kutschke:
@sdifalco please let me know when your are finished with this set of commits and I will rerun the CI

sdifalco:
I have just resent the CI...

kutschke:
Your code looks correct to me.  I think this message is a result of design choices made by G4 in the mid 1990s - STL was still in development and the community was just learing to use C++.  So G4 has many design choices that would be made differenlty today.  

We might be stuck.   I will ask the local experts if they know a way around it.  But let's not hold up the PR just for this.

Mu2eG4/src/GammaConvGenerator_module.cc
sdifalco: unnecessary PhysVol collection unpacking deleted

sdifalco:
@FNALbuild run build test

sdifalco:
I have completed all the changes. I wasn't able to fix the last few warnings

brownd1978 (approved):

kutschke (changes requested):
Thanks Stefano.  This is a big improvement.  Thanks for cleaning up the clang tidy warnings.  I have made a note of some of the ones in the included files to address myself.

I do have two requested final changes.  One is inline in the code - to remove a memory leak.

The second is to remove the lines of commented out code.  I don't see any value in keeping them.

sdifalco:
I have deleted the commented code I have found.
The map is now private (no need for a set function, a get function is enough for what I see)
I don't see the requested change inline in the code to avoid memory leakage: was this one?
Thanks

sdifalco:
Now I have warnings from
 std::array <const double,3> sf1a={42.038,-8.29,0.958};
like "2.038 is a magic number; consider replacing it with a named constant"
should I have used 
const  std::array <double,3> sf1a={42.038,-8.29,0.958};
?

Thanks

sdifalco:
@FNALbuild run build test

kutschke:
Hi Stefano,

  The memory leak is in the new used to create the object of type GammaPairConversionSpectrum that is pointed to by the variable spectrum_ .   It is never deleted.  While it’s not a big memory leak it’s easy to fix.

   Rob


From: sdifalco ***@***.***>
Date: Thursday, February 22, 2024 at 4:45 PM
To: Mu2e/Offline ***@***.***>
Cc: Rob Kutschke ***@***.***>, Team mention ***@***.***>
Subject: Re: [Mu2e/Offline] Gamma conversion generator (PR #1195)


[EXTERNAL] – This message is from an external sender

I have deleted the commented code I have found.
The map is now private (no need for a set function, a get function is enough for what I see)
I don't see the requested change inline in the code to avoid memory leakage: was this one?
Thanks

kutschke:
> Now I have warnings from std::array <const double,3> sf1a={42.038,-8.29,0.958}; like "2.038 is a magic number; consider replacing it with a named constant" should I have used const std::array <double,3> sf1a={42.038,-8.29,0.958}; ?
> 
> Thanks

Let me check on thing - it will take 5 minutes.

kutschke:
.... I think it is suggesting:

constexpr std::array<double,3> sfa1{42.038,-8.29,0.958};

And a small detail.  Where you have:
const double FZ_const=8.;

If you write that as
constexpr double FZ_const=8.;

it gives the compiler more information.  It explicitly says that FZ_const is a compile time constant and the optimizer may treat it as such.   The way that your wrote it, it's very likely the optimizer will figure that out but it has to do some work to do so.  It's not a big deal but it is an example of just do it right all of the time so that when it matters you will just do it right.

kutschke:
You only need this much:

`spectrum_                (&randFlat_, useCorrelatedAngleOverKE_)
`

sdifalco:
I had to add static in front of constexpr to compile
I have fixed the memory leak as suggested
Thanks

sdifalco:
@FNALbuild run build test

kutschke:
> I had to add static in front of constexpr to compile I have fixed the memory leak as suggested Thanks

Thanks - I forgot about the static.  Good catch.  It's needed for constexpr data members but it is not needed for constexpr stack variables in function scope.

Presuming the CI goes well, I will merge tonight.

PR #1197, opened 2024-02-23
Add cut on NDOF for stereoline fit

brownd1978:
also add a flag bit to specify if a combohit was made using the stereo line fit, and record it in ComboHitDiag.
The ANN data file for stereo hits with SLINE was malformed (blocks in the wrong order) and cause an exception if used.  It also seemed to have identical numerical content as the regular stereo weights.   I removed it for now.


kutschke:
@brownd just checking.  This looks like it will affect nightly validation because of the added cut in MakeStereoHits.  I propose to delay merging until tomorrow so that we can see if Giani's PR produces any unexpected changes in ceMix.   Is that Ok for you?

myucel-fnal (approved):
Looks good to me. Had some observations I forwarded to Dave.

brownd1978:
The defaults don’t change

Office Phone (510) 486-7261

kutschke (approved):

PR #1198, opened 2024-02-23
Fix a typo in PTM geometry

gaponenko:
The HEAD of Offline throws on an attempt to print out the final SimpleConfig.    This PR fixes the issue.


kutschke (approved):
Thanks for catching and fixing this.  It did not generate an error during normal running because SimpleConfig systems any type to be returned as a string ( so that you can see the exact representation used for a number, for example).

PR #1199, opened 2024-02-24
Remove muonID for the shielding assessment

gaponenko:
The current implementation is literally a large piece of iron, so this commit simplifies things by removing the code instead of complicating things by adding an enable/disable switch.

The geometry version used for the Mu2e shielding assessment must have muonID taken out.


kutschke (approved):
I will merge this on a day by itself in order to get a new validation baseline.

I might merge some other commits that day if I am sure they will not change validation.

kutschke:
I have no idea why the CI was not run on creation of the PR.

@FNALbuild run build test

kutschke:
Hmmm.  The "You have proposed changes to files in these packages:" message posted a few minutes after I launched the CI by hand.  That was 16 hours after the PR was created - so it was just very, very slow.

PR #1203, opened 2024-02-27
updates in the trigger code

gianipez:
now the prescale set to -1 means false in all the events of that specific eventMode

### DIFF
--- a/Trigger/src/PrescaleEvent_module.cc
+++ b/Trigger/src/PrescaleEvent_module.cc
@@ -91,7 +91,7 @@ namespace mu2e
     bool retval(false);
     bool condition = e.event() % ps == 0;
 
-    if(condition) {
+    if(condition && (ps > 0)) {
       ++_npass;
       retval = true;
     }
--- a/Trigger/src/PrescaleEvent_module.cc
+++ b/Trigger/src/PrescaleEvent_module.cc
@@ -89,11 +89,12 @@ namespace mu2e
     int ps = _eventMode[0].prescale();
 
     bool retval(false);
-    bool condition = e.event() % ps == 0;
 
-    if(condition && (ps > 0)) {
-      ++_npass;
-      retval = true;
+    if(ps > 0) {
+      retval = e.event() % ps == 0;
+      if (retval){
+        ++_npass;
+      }
     }
     return retval;
   }
--- a/DAQ/test/generateBinaryFromDigi.fcl
+++ b/DAQ/test/generateBinaryFromDigi.fcl
@@ -7,7 +7,7 @@
 
 #include "Offline/fcl/minimalMessageService.fcl"
 #include "Offline/fcl/standardServices.fcl"
-#include "Offline/DAQ/fcl/prolog_trigger.fcl"
+#include "mu2e_trig_config/core/trigProducers.fcl"
 
 process_name : BinaryFromDigi
 
@@ -26,7 +26,7 @@ physics : {
 
    producers : {
       binaryOutput: {
-         @table::DAQ.producers.binaryOutput
+         @table::TrigDAQProducers.binaryOutput
          generateTimestampTable : 1
       }
    }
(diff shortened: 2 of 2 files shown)


gianipez:
@FNALbuild run build test

kutschke:
@brownd1978 Do  you want to review this PR or should I look after it on my own.  It looks Ok to me.

Trigger/src/PrescaleEvent_module.cc
```diff
@@ -91,7 +91,7 @@ namespace mu2e
     bool retval(false);
     bool condition = e.event() % ps == 0;
```
kutschke, line 92: The remainder operator will produce a NaN if ps is zero. The variable condition will never be true because NaN==0 is always false.  Please add some code to the c'tor that checks for legal values of the prescal and takes appropriate action.   

Also, I had to look up the order of operator precedence to be sure this was doing the right thing.  I would add () as: ( e.event() % ps) == 0;  your call.

Trigger/src/PrescaleEvent_module.cc
```diff
@@ -91,7 +91,7 @@ namespace mu2e
...
```
kutschke, line 93: Separate comment on the preceding line.  Have you thought about  using ( _nevt % ps ) instead of (e.event() % ps).  When we are running in the trigger farm there will be short term structure in the event numbers seen by any given trigger process. I don't have a good quantitative picture of what short term is.   I think in the long term event builders that pull events from off-machine board readers will introduce enough randomness.  The issue is the short term - could we get all events that pass landing on a small number of trigger processes.

This situation will never come up with MC as we structure it today.

kutschke (changes requested):
I am curious about PrescaleEvent_module.

You have

kutschke:
Another question.  Why does this module only look at element 0 of the _eventMode vector?  Is this one of the things waiting for the EventHeader to be populated?

brownd1978 (approved):

Trigger/src/PrescaleEvent_module.cc
gianipez: two lines below there is a check on `ps` (ps>0), which I think is enough. We will using `ps=-1` as a convention in the Trigger Menu configuration to indicate a trigger path that is non active (in a given event-mode).

Trigger/src/PrescaleEvent_module.cc
gianipez: I didn't think about this TBH. Probably we can figure this out when we will do a larger scale test later on.

gianipez:
> Another question. Why does this module only look at element 0 of the _eventMode vector? Is this one of the things waiting for the EventHeader to be populated?

Correct. Once we will have the `mu2e::EventHeader` in the event payload, we will uncomment those lines and do it properly.

gianipez:
Shall we proceed with this PR or is there anything left?

kutschke:
> Shall we proceed with this PR or is there anything left?

There is one thing left.  I want you to check for a misconfiguration in the constructor.  You replied to my original comment and I replied to that.  I do not want our flow to control to rely on generation and testing of NaNs, which is what the current code does when it is misconfigured.

I think that the right answer for now, and maybe long term, os to throw an exception in the c'tor if ps == 0 and let art do a graceful shutdown.

This touches the issue of use of exception handlng in the trigger.  Eric has said that he does not want the code to stop in mid run.  But this is at startup, which has different boundary conditions.   I have contacted Andrew Norman about when and where to have the conversation about a policy for handling of exceptional conditions in the trigger.   I suggested we record this as an issue in the appropriate issue tracker and asked him if we should use the Offline issue tracker or one  of the TDAQ ones.  If I don't hear back soon I will put it in the Offliine issue tracker.

kutschke:
> The configuration checks for the Trigger module instances happen in the python script that generates the configuration files. If a User doesn't use the standard workflow it's at their own risk and I don't think it should be taken into account.

I don't like leaving landmines in the code.  Please make the change.

kutschke:
Thanks for making the change.  When you update the documentation, please make it clear that any number <=0 will fail all events.

@FNALbuild run build test

kutschke (approved):

PR #1213, opened 2024-03-06
Make straw status testing in digi making optional, default off

brownd1978:
This makes the tracker digitization consistent with the convention agreed on with CRV

### DIFF
--- a/TrackerMC/src/StrawDigisFromStrawGasSteps_module.cc
+++ b/TrackerMC/src/StrawDigisFromStrawGasSteps_module.cc
@@ -114,6 +114,7 @@ namespace mu2e {
           fhicl::Atom<int> diagpath{ Name("DiagPath"), Comment("Digitization Path for waveform diagnostics") ,0 };
           fhicl::Atom<string> spinstance { Name("StrawGasStepInstance"), Comment("StrawGasStep Instance name"),""};
           fhicl::Atom<string> spmodule { Name("StrawGasStepModule"), Comment("StrawGasStep Module name"),""};
+          fhicl::Atom<bool> usestatus { Name("UseStatus"), Comment("Use TrackerStatus when making digis"), false};
 
         };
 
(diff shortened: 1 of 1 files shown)


TrackerMC/src/StrawDigisFromStrawGasSteps_module.cc
```diff
@@ -535,8 +535,13 @@ namespace mu2e {
 
     void StrawDigisFromStrawGasSteps::fillClusterMap(StrawPhysics const& strawphys,
         StrawElectronics const& strawele,
-        TrackerStatus const& trackerStatus,
         art::Event const& event, StrawClusterMap & hmap){
+// get status if needed
+      std::shared_ptr<const TrackerStatus> trackerStatus;
+      if(_usestatus) {
+        ProditionsHandle<TrackerStatus> _trackerStatus_h;
```
kutschke, line 542: @rlcee I have a question about the recommended use patterns of proditions .   Maybe I am out of date but I thought that the plan was to get the ProditionsHandle once at the start of the job and hold it as a data member.  It does not occupy much memory and but getting a new handle every event is a little bit of work.   If _usedb is false, you will never call getPtr(event.id()).  While it's not really important in this module I think we want to establish a recommended pattern and show it to others as a module.

rlcee: Yes, the ProditionsHandle should be created once as an art module class member.  On construction, it will get a pointer from the service, but nothing substantial happens.  calling get() or getPtr() triggers fromFcl or fromDb, depending on that setting, so if you are not going to use the result, you probably want to not call that.  Recreating the handle in the event will cause it to get that service pointer again, then also search for the cached content that you are losing the pointer to when you deleted the handle last event.  So if you are using the content, then deleting the handle frequently is a little cost.

kutschke: thanks for looking after this.

kutschke:
@FNALbuild run build test

PR #1215, opened 2024-03-08
Bug fixes for wireoffsets

bonventre:
Requires wire position to follow truncation to straw volume, fixes tau calculation

### DIFF
--- a/CommonMC/src/SelectRecoMC_module.cc
+++ b/CommonMC/src/SelectRecoMC_module.cc
@@ -309,25 +309,47 @@ namespace mu2e {
     TwoLinePCA_XYZ strawpca( XYZVectorF(straw.strawPosition()), XYZVectorF(straw.strawDirection()),
           tshmc._cpos, tdir );
 
-    auto mcsep = wirepca.point2() - wirepca.point1();
-    tshmc._wireLen = (wirepca.point2()-XYZVectorF(straw.wirePosition())).Dot(wdir);
-    auto mcperp = tdir.Cross(wdir).Unit();
-    tshmc._wireDOCA = -1*mcperp.Dot(mcsep);
-    auto pdir = wdir.Cross(bdir);
-    if (pdir.Dot(XYZVectorF(straw.wirePosition())) < 0.0) pdir *= -1.0; //sign radially out
-    tshmc._wirePhi = atan2(mcsep.Dot(pdir),mcsep.Dot(bdir));
-    tshmc._wireDot = tdir.Dot(wdir);
-    auto wperp = wdir.Cross(mcperp);
-    tshmc._wireTau = (tshmc._cpos-XYZVectorF(straw.wirePosition())).Dot(wperp);
+    // vector from middle of wire to POCA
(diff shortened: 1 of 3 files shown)


TrackerMC/src/StrawDigisFromStrawGasSteps_module.cc
```diff
@@ -1258,13 +1258,20 @@ namespace mu2e {
       float dw = delta.Dot(sdir);
       XYZVectorF cperp = delta - dw*sdir; // just perp part
       float phi = atan2(cperp.Dot(pdir),cperp.Dot(zdir)); // angle around wire WRT z axis in range -pi,pi
-      float rho = min(sqrt(cperp.mag2()),(float)_rstraw); // truncate!
+      float rho = sqrt(cperp.mag2());
```
brownd1978, line 1261: I am confused how cpos could ever be outside the straw, since it comes from the G4 energy deposits.

CommonMC/src/SelectRecoMC_module.cc
```diff
@@ -310,15 +310,15 @@ namespace mu2e {
           tshmc._cpos, tdir );
 
     auto mcsep = wirepca.point2() - wirepca.point1();
-    tshmc._wireLen = mcsep.Dot(wdir);
+    tshmc._wireLen = (wirepca.point2()-XYZVectorF(straw.wirePosition())).Dot(wdir);
```
brownd1978, line 313: I don't understand how wirepca.point1() could be off the wire position.  Is one of those really WRT  the straw axis?  If so, can the names be set so that it's more clear?

brownd1978 (commented):
Thanks for using unbiased doca to define the angles.  I don't fully understand the changes, I suspect the subtlety of the difference between straw and wire axis is getting obscured by the variable names.

TrackerMC/src/StrawDigisFromStrawGasSteps_module.cc
bonventre: The position is randomly adjusted here https://github.com/Mu2e/Offline/blob/030e4c8178023f2259a37d23c4a42d84d6bc763c/TrackerMC/src/StrawDigisFromStrawGasSteps_module.cc#L934

CommonMC/src/SelectRecoMC_module.cc
bonventre: wirepca.point1() is at the same longitudinal position so mcsep.Dot(wdir) ~ 0, need to compare to the wire middle instead

brownd1978:
I trust that you’ve tested this and are getting the right values.  My point
is that the code is hard to understand.   Can you define some intermediate
variables with clear names ( POCA_to_straw or similar) and add some
comments so that its easier to understand the intent.
Office Phone (510) 486-7261

bonventre:
Sorry, I was trying to keep names and calculations the same as current TrkAna to make it easier to compare on my end. I've added some comments which hopefully make it clearer

brownd1978 (approved):
thanks for the updates

brownd1978:
@FNALbuild run build test

PR #1217, opened 2024-03-08
Throw an exception rather than do nothing on a configuration mistake.

gaponenko:
The existing code in Mu2eStudyWorld throws an exception for unrecognized values of services.GeometryService.tool_type, but does nothing if the string is set to "Mu2e".   As long as "Mu2e" is not a valid name for a g4study setup, this is still a configuration mistake and the code should throw instead of trying to run with a bad configuration.
This PR removes the "do nothing" special case.  Please comment.


pavel1murat (approved):
makes sense.

resnegfk (approved):

PR #1219, opened 2024-03-10
New deltaFunctionGenerator module.

gaponenko:
Adding a new event generator, no effect on validation.

### DIFF
--- a/EventGenerator/src/deltaFunctionGun_module.cc
+++ b/EventGenerator/src/deltaFunctionGun_module.cc
@@ -1,0 +1,164 @@
+// This generator throws particles of the same type and the same
+// kinematics in each event, i.e. a delta-function distribution in the
+// phase space.  This is exactly what is needed for some simulations,
+// in particular those using G4study.  The explicit module name will
+// hopefully help to prevent future complexification of this code.
+// This generator is so basic that it does not even need random numbers.
+// It is fully configurable via fcl, without the need for an
+// additional txt file.
+//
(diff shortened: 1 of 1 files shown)


kutschke:
What features does this add that are not already prsentHow is this different that the existing particle gun generator?

gaponenko:
Hi Rob,

The new module supports one use case as simply and efficiently as possible, instead of trying to support multiple different use cases.
I want to use it with a g4study setup.   An existing module that could work is EventGenerator, but it requires a separate SimpleConfig file; an overhead I would like to avoid.
Some generators are configurable from fcl, like PrimaryProtonGun, but they are tied to the full Mu2e geometry and can not be used with g4study.

The new generator is fully multi-threaded, unlike any of the existing generators.  Perhaps a small impact in practice, but why compromise if we don't have to?
The laptop I am running this on has 20 cores, so enabling MT in general makes a lot of sense.

An important factor for me is also using "code that is so simple there are obviously no bugs, instead of code so complicated there are no obvious bugs" whenever I can.

Andrei

kutschke (commented):
In the fcl comments, please add units and, where appropriate, "in the Mu2e coordinate system".

gaponenko:
Done, please take a look.
Andrei

gaponenko:
Good catch, fixed!

kutschke (approved):

kutschke:
@FNALbuild run build test

PR #1220, opened 2024-03-10
Add an "empty" geometry for g4study

gaponenko:
This PR adds code to create an "empty" g4study geometry, with nothing placed into the mother box-with-a-skin volume.
It also includes related clean-ups that remove multiple lines of code without any functionality change.


Mu2eG4/inc/InitEnvToolBase.hh
```diff
@@ -1,28 +1,21 @@
...
 
+  class VolumeInfo;
+  class SimpleConfig;
+
   class InitEnvToolBase {
-  protected:
-    std::string _name;
   public:
 
     InitEnvToolBase() noexcept = default ;
```
kutschke, line 12: We should make it impossible to construct an object of base type.  I believe we can do this by making the default c'tor protected.

I see that you removed the _name functionality in it's entirety.

kutschke (commented):
Can you describe the use case for making an empty study geometry?

kutschke:
@pavel1murat and @resnegfk I added you as reviewers to this PR to answer one question because you are both stakeholders in this technology.  This PR  removes the _name featture because it is not currently used.  If you have no objections, please approve the PR.  if you do have objections, please explain on this thread why it's needed.

gaponenko:
> Can you describe the use case for making an empty study geometry?

Looking at any particle interaction in any single material.     One can just set world.materialName to the desired value, and not worry about defining extra parameters for nested volumes and their materials.

kutschke:
@FNALbuild run build test

resnegfk (approved):
Although using VolumeInfo makes it more Mu2e specific, I do not object to the changes.

resnegfk:
The above comment should have read (skip the Although intro, sorry):  I do not object to the changes.

PR #1221, opened 2024-03-11
Update CMakeLists.txt as needed. 

eflumerf:
Add script to show when files have been added or removed


eflumerf:
Commit [1560c76](https://github.com/Mu2e/Offline/pull/1221/commits/1560c76110e7f28367fb02993706460f1c1c4bd3) was tagged with "d11_01_00" and UPS products were deployed on the mu2edaq machines based on this tag for the DAQ Global Run 1.

kutschke:
@FNALbuild run build test

.gitignore
```diff
@@ -44,3 +44,4 @@ tags
 .clangd/*
 compile_commands.json
 spack-*
+Offline
```
kutschke, line 47: I am curious why you need this?  Is it because of the symlink you need to add?

kutschke (approved):
I see that you removed a lot of configure_file COPYONLY cmake commands.  Are these files now found in the source tree ( or a copy of it?).

kutschke:
@eflumerf I added a few questions that I would like answered - for my own background information.  I am merging now.

.gitignore
eflumerf: Yes

eflumerf:
> I see that you removed a lot of configure_file COPYONLY cmake commands. Are these files now found in the source tree ( or a copy of it?).

configure_file only puts things into the build tree. The install(DIRECTORY) command I replaced them with puts them into the appropriate place in the install tree (whether UPS product or Spack installation), and it also copies the files recursively so that we don't need to maintain a configure_file command for each data file.

PR #1222, opened 2024-03-12
implement TFileService access

rlcee:
Andrei suggested this is a better format, which makes sense.  It was not used anywhere yet.


kutschke (approved):
Looks good to me.

kutschke:
@FNALbuild run build test

There was a merge since the previous test completed.

PR #1229, opened 2024-04-06
Revert type change

brownd1978:
For unknown reasons this change produced corrupt data products.


brownd1978:
This problem also shows that validation isn't adequately covering MC truth matching.  We have exactly 1 plot sensitive to MC matching, it did not catch this problem. We need to add more coverage.

rlcee (approved):
I presume you've tested that this fixes the root problem, which is the main thing.   I think style prefers relation(_rel) instead of (relation)_rel.

PR #1231, opened 2024-04-12
compressPDGCodeHisto() tweak

gaponenko:
Passing a *handle* by address instead of by value probably does not affect code performance in a measurable way, but loses functionality: member initialization like the following example does not work with the extraneous '&'

```
MyAnalyzer::MyAnalyzer(const Parameters& conf, const art::ProcessingFrame&)
    : SharedAnalyzer{conf}
    , myhisto_{compressPDGCodeHisto(art::ServiceHandle<art::TFileService>())}
    , ....
{ ... }
```


rlcee (approved):
Looks useful, thanks

PR #1233, opened 2024-04-13
Add module to merge KalSeed collections

brownd1978:
The primary use case for this module as to support all-in-one processing in TrkAna, but it could have other applications.  Note that the KalSeedMC collection and the Assns are already all-in-one, as those are created by createRecoMC.


AndrewEdmonds11 (approved):
Looks good to me, Dave. Are you going to merge the PR now, or will you wait to make sure I can use the KalSeedPtrs in TrkAna? For what it's worth, I don't see a need to wait before merging

brownd1978:
@FNALbuild run build test

PR #1234, opened 2024-04-15
clean up of deprecated fcl files + fixed one bug in CalHelixFinderAlg

gianipez:
- removed old fcl files using the deprecated Trigger implementation. The examples for running the trigger exe live in the mu2e_trig_config area.
- fixed one parameter in CalHelixFinderAlg that was limiting to 500 the number of hits in a helix candidate


CommonMC/fcl/prolog_trigger.fcl
brownd1978, line 1: This file is referenced by https://github.com/Mu2e/mu2e_trig_config/blob/main/test/triggerTestMC.fcl
Do you plan to move it to mu2e_trig_config?  If so, this PR needs to be linked with a PR there

brownd1978 (commented):
There's an open question about the CommonMC/prolog_trigger.fcl file

CommonMC/fcl/prolog_trigger.fcl
gianipez: I restored the file. I prefer to not include any MC config in mu2e_trig_config for now

CalPatRec/inc/CalHelixFinderData.hh
```diff
@@ -50,9 +50,9 @@ namespace mu2e {
     //   // kNPanelsPerFace =  3
     // };
 
-    enum { kMaxResidIndex = 500 };
+    enum { kMaxResidIndex = 5000 };
 
-    constexpr static uint16_t        kNMaxChHits = 500;
+    constexpr static uint16_t        kNMaxChHits = 5000;
```
brownd1978, line 55: This change is acceptable as a short-term bug fix, but it opens questions about implications to performance.

As a possible long-term fix, can the storage associated with this value:
Offline/CalPatRec/inc/CalHelixFinderData.hh:      float    resid[kMaxResidIndex];
Offline/CalPatRec/inc/CalHelixFinderData.hh:      float    dist [kMaxResidIndex];
Offline/CalPatRec/inc/CalHelixFinderData.hh:      float    dz   [kMaxResidIndex];
Be converted to std::vector instead?

brownd1978 (approved):

CalPatRec/inc/CalHelixFinderData.hh
gianipez: I remember a long time ago that using an array instead of vector was visibly making a difference in the timing performance when I was testing the trigger reconstruction.

PR #1236, opened 2024-04-16
make ptr check more general

rlcee:
Histogramming a quantity from product B in validation for product A violates the validation philosophy of simplicity and robustness, and indeed it caused a crash.  However, rather than remove this histo I changed it to a more robust check.

### DIFF
--- a/Validation/src/ValStepPointMC.cc
+++ b/Validation/src/ValStepPointMC.cc
@@ -34,21 +34,15 @@ int mu2e::ValStepPointMC::fill(const mu2e::StepPointMCCollection& coll,
   _hVer->Fill(1.0);
 
   _hN->Fill(coll.size());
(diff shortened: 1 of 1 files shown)


brownd1978 (approved):
I hope this is a temporary stopgap.  Long term I would like workflows that don't persist StepPointMCs outside of sim stages.  It means creating new *Step classes for specific purposes (like crvStage1).

rlcee:
> I hope this is a temporary stopgap. Long term I would like workflows that don't persist StepPointMCs outside of sim stages. It means creating new *Step classes for specific purposes (like crvStage1).

This change is intended to be permanent, since it doesn't change the behavior, only stops it from crashing.  It also doesn't affect any workflows.  Buy maybe I don't understand what you're saying.

PR #1239, opened 2024-04-19
Chi2Clusterer SLine training

myucel-fnal:


TrkHitReco/src/FlagBkgHits_module.cc
```diff
@@ -72,9 +77,11 @@ namespace mu2e
       float                                       cperr2_;
       int const                                   debug_;
       std::string                                 kerasW_;
+      bool                                        useSLine_;
       float                                       kerasQ_;
       int                                         iev_;
-      std::shared_ptr<TMVA_SOFIE_TrainBkgDiag::Session> sofiePtr;
+      std::shared_ptr<TMVA_SOFIE_TrainBkgDiag::Session> sofiePtr1;
```
brownd1978, line 83: Where does this model definition come from?  I see the #include for the Chi2SLine model, but not for the other model.   Should this module #include all the models?

TrkHitReco/src/FlagBkgHits_module.cc
```diff
@@ -52,6 +56,7 @@ namespace mu2e
         fhicl::OptionalTable<TNTClusterer::Config>    TNTClustering{        Name("TNTClustering"),        Comment("TNT Clusterer config") };
         fhicl::OptionalTable<Chi2Clusterer::Config>   Chi2Clustering{       Name("Chi2Clustering"),       Comment("Chi2 Clusterer config") };
         fhicl::Atom<std::string>                      kerasWeights{         Name("KerasWeights"),         Comment("Weights for keras model") };
+        fhicl::Atom<bool>                             useSLine{             Name("UseSLine"),             Comment("Use SLine info") };
         fhicl::Atom<float>                            kerasQuality{         Name("KerasQuality"),         Comment("Keras quality cut") };
       };
```
brownd1978, line 61: Would it make sense to add a config for the ComboHit level of the Input collection?  Since that could determine what training to use?  Also you could check that the input collection actually has the expected level, and throw an exception if not.

brownd1978 (commented):
This looks basically fine.  I don't understand how FBH is getting the non-chi2 training.  I suggest all the trainings be explicitly named:
TMVA_SOFIE_TrainBkgDiag -> TMVA_SOFIE_TrainBkgTN2Station etc.  There's no need for 'Diag' in the training name.
I suggest to test the consistency of the input ComboHits level with the training.

kutschke:
Clear the cosmicOffSpill error

@FNALbuild run build test

TrkHitReco/src/FlagBkgHits_module.cc
myucel-fnal: Line #21 and #22 have the definitions for the models. Chi2 model has additional parameters.

TrkHitReco/src/FlagBkgHits_module.cc
myucel-fnal: Do you want to make it so that user don't choose the training specifically but rather specifies the clusterer and input level of the CH collection? I think it makes sense since with the combination of clusterers, trainings and inputs, there is an increasing probability for user error.

brownd1978 (commented):
additional comments

brownd1978 (approved):

brownd1978:
@FNALbuild run build test

brownd1978:
@FNALbuild run build test

brownd1978:
@FNALbuild run build test

PR #1255, opened 2024-05-04
val track charge fix

rlcee:
I found that some validation jobs were ended with a "tried to save duplicate histogram" error, which I fixed with adding the "2".  Then I reran my test and it seg-faulted on this line
`double ksCharge = ks.intersections().front().pstate_.charge();`
I see that the intersections access is protected against empty vectors in the other access, and if an empty vector is possible here, then this line could cause undefined behavior and I got lucky.  So I moved it into the protected block.  But I also lazily switched to the charge of the sid surface intersection instead of the front of the intersections vector.  I suspect this is OK for this purpose (an more consistent?), but if the front of the vector is important, then I can re-write to get that safely.

### DIFF
--- a/Validation/src/ValKalSeed.cc
+++ b/Validation/src/ValKalSeed.cc
@@ -38,7 +38,7 @@ namespace mu2e {
     _hpce = tfs.make<TH1D>("pce", "p CE", 100, 95.0, 110.);
     _hpcep = tfs.make<TH1D>("pcep", "p CE+", 100, 82.0, 97.);
     _hsignedp = tfs.make<TH1D>("signedp", "signedp", 200, -110., 110.);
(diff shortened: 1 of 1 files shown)


kutschke (approved):
This looks right to me.

I added Dave and Giani as reviewers as an FYI for them.  I plan to merge this evening so that the Sunday AM validation works.

gianipez (approved):
Eveything looks ok to me

brownd1978 (changes requested):
I have provided an alternate, more robust and definitive fix in PR 1257.  Please use that instead.

kutschke:
This is superseded by PR #1257.  Closing it.

kutschke:
> I have provided an alternate, more robust and definitive fix in PR 1258. Please use that instead.

Yes - I saw it.

PR #1257, opened 2024-05-04
Art v3_15_00 preparation

kutschke:
Make changes to that are required for art suite v3_15_00 that upgrades root to v6_30_06.   The submitted code works with both the new art suite and the with the current art suite v3_14_03.  I checked that 2000 events of ceSimReco produced identical results.

  In all cases the problem was that fhicl code was exposed to dictionary makers because of people having includes that were either unneeded or were in the .hh file rather than the .cc file, where they belonged.   I don't understand the details of what in fhicl is causing the problem but there is no reason for fhicl code to be exposed to dictionary makers.

This is something that likely would have been fixed by IWYU so that is back on my list as a priority.

There will be an accompanying PR in REve.  

There were no build-time issues with TrackerAlignment, TrkAna and Tutorial.

### DIFF
--- a/Mu2eBTrk/inc/ParticleInfo.hh
+++ b/Mu2eBTrk/inc/ParticleInfo.hh
@@ -8,8 +8,7 @@
 // This code looks after the translation and caches results.
 //
 
-#include "Offline/GlobalConstantsService/inc/GlobalConstantsHandle.hh"
-#include "Offline/GlobalConstantsService/inc/ParticleDataList.hh"
+#include "Offline/GlobalConstantsService/inc/ParticleData.hh"
 
 #include "BTrk/BaBar/ParticleInfoInterface.hh"
 
@@ -17,6 +16,8 @@
 #include <map>
 
 namespace mu2e {
+
+  class ParticleDataList;
 
   class ParticleInfo : public ParticleInfoInterface {
 
@@ -41,9 +42,8 @@ namespace mu2e {
     //the following has to be mutable because BTrk holds this
     //with a const pointer
 
-    // Handle to the full particle data table.
     // Guaranteed valid throughout the job.
-    mutable GlobalConstantsHandle<ParticleDataList> pdt_;
+    ParticleDataList const& pdt_;
 
     // Local cache of the information for particles that we care about;
     // indexed by TrkParticle::type, not by PDG::id.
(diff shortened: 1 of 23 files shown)


kutschke:
For the record, here is Kyle's explanation for why the new fhiclcpp caused problems.

`
with art 3.15, we are using C++20 concepts to constrain the types of arguments allowed for function templates. ROOT's genreflex and rootcling functionalities do not yet understand C++20 concepts; hence the failures. I'm glad you were able to get around the issue–we've tried to avoid using C++20 concepts in places where dictionaries will be generated...but that does require some reworking of code.`

The bottom line remains that we should never have been exposing fhicl to genreflex.

rlcee (approved):
All looks like good cleanups.

kutschke:
First test of CI after Jenkins migration:

@FNALbuild run build test

Mu2eBTrk/inc/ParticleInfo.hh
```diff
@@ -41,9 +42,8 @@ namespace mu2e {
     //the following has to be mutable because BTrk holds this
     //with a const pointer
 
-    // Handle to the full particle data table.
     // Guaranteed valid throughout the job.
-    mutable GlobalConstantsHandle<ParticleDataList> pdt_;
+    ParticleDataList const& pdt_;
```
brownd1978, line 46: presumably this is the same object as before.  Who holds that handle and makes sure the object stays in scope throughout the job?  Is that guaranteed by the service?

brownd1978 (approved):
I'd like to verify the memory of the ParticleDataList is pinned even without keeping the handle, just by having accessed the handle.  Presuming that's guaranteed I approve.

Mu2eBTrk/inc/ParticleInfo.hh
kutschke: > presumably this is the same object as before. Who holds that handle and makes sure the object stays in scope throughout the job? Is that guaranteed by the service?

Yes.  The ParticleDataList is created by GlobalConstantsService upon the first request.  It is created on the heap and stays in the same place for the duration of the job.  The object is owned by the GlobalConstantsService  and
it's destructor is called during the destructor of the GlobalConstantsService.

When you create a GlobalConstantsHandle<T> it gets an art::ServiceHandle<GlobalConstantsService> and  asks the service for a const pointer to the  object.
 
The above relies on the rules for what we allow into the GlobalConstantsService. 

Does that answer your question?

PR #1258, opened 2024-05-05
Use charge from KalSeed pdg code

brownd1978:
This is an alternate and more robust (definitive) fix to the problem addressed in PR 1255


Validation/src/ValKalSeed.cc
```diff
@@ -77,8 +79,8 @@ namespace mu2e {
     return 0;
   }
 
-  int ValKalSeed::fill(const KalSeedCollection& coll,
-      art::Event const& event) {
+  int ValKalSeed::fill(const KalSeedCollection& coll, art::Event const& event) {
+    auto const& ptable = GlobalConstantsHandle<ParticleDataList>();
```
kutschke, line 83: A question for a possible followup PR.

I suggest making ptable a datamember of type `ParticleDataList const&` and initializing it in the c'tor, as in:

https://github.com/kutschke/Offline/blob/art_v3_15_00_prep/Mu2eBTrk/inc/ParticleInfo.hh
https://github.com/kutschke/Offline/blob/art_v3_15_00_prep/Mu2eBTrk/src/ParticleInfo.cc

This works because information from the GlobalConstantsService does not change over the job.  Getting the handle every event is right for Proditions objects, which may change over the course of a job.

The trivial speed improvement is not the issue - it's providing a model of doing things correctly.

kutschke (approved):
I understand why this is better than Ray's.  I will merge this and close the other later this evening.  I will leave a window for Ray to comment in.

Validation/src/ValKalSeed.cc
kutschke: If you want to address this tonight, let me know and I will watch for it before I merge.

rlcee (approved):
I second Rob's comment on the service, but I am also wondering - does the KalSeed know the charge independent of the service?  Or will users of KalSeed always need to do this conversion with the service?  Even if it is not saved anywhere but implicitly in the PdgId, would you be willing to derive the sign from the PdgId internally?  Can the fit be configured to fit for a hypothetical particle not in the service?

kutschke:
I would like to hear the answers to Ray's questions but I want to merge this now for tonight's validation.

PR #1259, opened 2024-05-06
Move check_cmake.sh to bin

kutschke:
... which is where it belongs.


eflumerf (approved):

kutschke:
The failed jobs died because the database was unreachable.  There was a no known problem with the db server that has been resolved. 

@FNALbuild run build test

kutschke:
@rlc - I added you as a reviewer to check that bin is the right place for this in your vision of organization.

rlcee (approved):
I think in the long run we had talked about moving Offline/bin to codetools or elsewhere, but that is a way off, so this is perfectly fine.  Thanks

PR #1260, opened 2024-05-07
added DB to the eventdisplay fcl files

ehrlich-uva:
The event display requires the values of the CRV pedestals to display reconstructed CRV waveforms. These values are in the database.


kutschke (approved):
My first thought was that this is two points of maintenance so we should use   a common epilog file.  But on further consider I agree that having two separate configurations is right - today they happen to be the same but they won't in general be the same.  And that will certainly be true for data.  

For the time being we do need to remember to update both when that is the right thing to do.

rlcee (approved):
The collaboration's patterns of how to set db purpose and version should be discussed at a higher level, but I think this is fine here.

brownd1978 (commented):
If this pattern is copied it will result in many points of maintenance.
I suggest to create a file  Offline/ConditionsService/fcl/NominalDatabase.fcl
That will have this content.  Then Ralfs script, or any other that needs a default to keep from crashing, can reference it.  It won't be the correct answer for all use cases but will keep standard codes from crashing.

kutschke:
@ehrlich-uva Dave put in a PR that creates  Offline/DbService/fcl/NominalDatabase.fcl .  Please update your PR to remove this file and #include the new file instead.  Once Dave approves, I will merge.

kutschke:
THanks Ralf.  I won't run the PR since this change has no coverage in the PR.

ehrlich-uva:
I had to include the NominalDatabase.fcl at the end of the fcl files. Otherwise the settings would have been overwritten by DbEmpty (via @table::Services.Core), which is defined in DbService/fcl/prolog.fcl.

kutschke:
> I had to include the NominalDatabase.fcl at the end of the fcl files. Otherwise the settings would have been overwritten by DbEmpty (via @table::Services.Core), which is defined in DbService/fcl/prolog.fcl.

Yes.  That's right.

PR #1262, opened 2024-05-10
changed sort fnction implementation and added protections in the code

gianipez:
The fixes were tested on `mu2ebuild01` running n>60 times the following command:
`mu2e -n 100 -c Production/Validation/nightly/ceMix_14.fcl`

tagging: @rlcee @matthewstortini @kutschke

### DIFF
--- a/CalPatRec/src/AgnosticHelixFinder_module.cc
+++ b/CalPatRec/src/AgnosticHelixFinder_module.cc
@@ -127,6 +127,7 @@ namespace mu2e {
       bool    uselessTripletSeed;
       bool    notOnSegment;
       bool    debugParticle; // only filled in debug mode -- true if mc particle, false if background
+      float   z;
     };
 
     struct tripletPoint {
@@ -428,6 +429,8 @@ namespace mu2e {
       consumes<CaloClusterCollection>  (_ccLabel);
       produces<HelixSeedCollection>    ();
 
+      _tcHits.reserve(1000);
+
       if (_debug == 1) {
         _mcUtils = art::make_tool<McUtilsToolBase>(config().mcUtils, "mcUtils");
         if (_runDisplay == 1) {
@@ -447,7 +450,6 @@ namespace mu2e {
 
       if (_diagLevel == 1) _hmanager = art::make_tool<ModuleHistToolBase>(config().diagPlugin, "diagPlugin");
       else _hmanager = std::make_unique<ModuleHistToolBase>();
-
       if (_useStoppingTarget == true) { _stopTargPos.SetCoordinates(0.0, 0.0, std::numeric_limits<float>::max()); }
 
     }
@@ -645,7 +647,7 @@ namespace mu2e {
 
     if (hitIndice == HitType::CALOCLUSTER) {
       _tcHits[tcHitsIndex].circleError2 = _caloClusterSigma * _caloClusterSigma;
-    } else {
+    } else if (hitIndice >= 0){
       float transVar = _chColl->at(hitIndice).transVar();
       float x = getPos(tcHitsIndex).x();
       float y = getPos(tcHitsIndex).y();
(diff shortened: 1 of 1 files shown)


CalPatRec/src/AgnosticHelixFinder_module.cc
```diff
@@ -729,8 +730,6 @@ namespace mu2e {
   //-----------------------------------------------------------------------------
   void AgnosticHelixFinder::tcHitsFill(size_t tc) {
 
-    size_t sortStartIndex = 0;
-
     cHit hit;
     hit.circleError2 = 1.0;
```
brownd1978, line 734: Please move these initializations to the cHit declaration.  All the data members should always be initialized to a default value

CalPatRec/src/AgnosticHelixFinder_module.cc
```diff
@@ -785,8 +789,8 @@ namespace mu2e {
 
     // order from largest z to smallest z (skip over stopping target and calo cluster since they
     // aren't in _chColl)
-    std::sort(_tcHits.begin() + sortStartIndex, _tcHits.end(), [&](const cHit& a, const cHit& b) {
-        return _chColl->at(a.hitIndice).pos().z() > _chColl->at(b.hitIndice).pos().z();
+    std::sort(_tcHits.begin(), _tcHits.end(), [&](const cHit& a, const cHit& b) {
```
brownd1978, line 792: Since z doesn't change since construction, it would be more efficient to use std::set and sort on insertion.

brownd1978 (commented):
These comments aren't directly related to the fix.  The struct member initialization is required.

CalPatRec/src/AgnosticHelixFinder_module.cc
gianipez: Thanks for the comments. @matthewstortini will make another round of optimization in the future, but now it's out of the scope of this PR.

matthewstortini (approved):
i will look into if the reserve for _tcHits is necessary later and make another PR if it's not. for now, the changes Giani made fix the bug.

kutschke:
I am not aware of any imminent plans for new PRs that will require ceMix validation to be working.  So we can leave this open until Matthew is back and has time to complete the work.   In addition to the requests on the review so far, I would like:

1. Understand why the problem occurred.  Without that understanding the PR might just be moving the problem around by reordering memory, not fixing it.
2. Test using all of the 20 jobs in Production/Validation/nightly/ceMix_nn.fcl, for nn=00 ... 19.

kutschke:
and a 3rd item.  Requests that will come up during the ongoing review.

kutschke:
> i will look into if the reserve for _tcHits is necessary later and make another PR if it's not. for now, the changes Giani made fix the bug.

If adding this makes the problem go away, it is almost certainly just masking the true problem.

CalPatRec/src/AgnosticHelixFinder_module.cc
gianipez: done

kutschke:
We plan to leave the PR open.  If there is a need to validate something ceMix related in the next few days we will deal with it when it comes up.

CalPatRec/src/AgnosticHelixFinder_module.cc
```diff
@@ -428,6 +429,8 @@ namespace mu2e {
       consumes<CaloClusterCollection>  (_ccLabel);
       produces<HelixSeedCollection>    ();
 
+      _tcHits.reserve(1000);
```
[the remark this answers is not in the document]
gianipez: no. This is not what solved the problem, but I left it there because it's a good practice.

gianipez:
Matthew will take over this fix in a separate PR. Closing this one

CalPatRec/src/AgnosticHelixFinder_module.cc
```diff
@@ -729,47 +750,38 @@ namespace mu2e {
...
-      offset += _calorimeter->caloInfo().getDouble("BPHoleZLength");
-      offset += _calorimeter->caloInfo().getDouble("FEEZLength");
-      offset /= 2.0;
-      _caloPos.SetCoordinates(tpos.x(), tpos.y(), tpos.z() - offset);
-      sortStartIndex++;
+      const CaloCluster*cluster = cl.get();
+      if (cluster != nullptr){
+        CLHEP::Hep3Vector gpos = _calorimeter->geomUtil().diskToMu2e(cl->diskID(), cl->cog3Vector());
+        CLHEP::Hep3Vector tpos = _calorimeter->geomUtil().mu2eToTracker(gpos);
+        double offset = _calorimeter->caloInfo().getDouble("diskCaseZLength");
```
kutschke, line 770: The calculation of offset is invariant.  It could be done at beginRun time and cached.  It also involves string comparisons so it will be slow.

PR #1264, opened 2024-05-15
start ad-hoc table feature

rlcee:
Ad-hoc tables have arbitrary structure and the user can just insert and read rows as they chose.  They can't be accessed though a conditions set and will not be available in production.  The tables are still defined in DbTables code so they can be manipulated by existing tools.  The immediate use case is to record which CID in an archive table corresponds to which CID in a reco table.  There is an example in the dev database:

```
build01 db > dbTool print-adhoc --database mu2e_conditions_dev --full --pretty --name TstAdhoc1
TABLE TstAdhoc1
# exint   exfloat           exstring   create_user                        create_time   
      0,      1.3,     first comment,          rlc,  2024-05-14 19:15:37.715581-05:00
     21,      2.3,   another comment,          rlc,  2024-05-14 19:15:37.715581-05:00
```
This should have no effect on any validation

### DIFF
--- a/DbService/data/create.sql
+++ b/DbService/data/create.sql
@@ -175,6 +175,15 @@ CREATE TABLE tst.calib2
 GRANT SELECT ON tst.calib2 TO PUBLIC;
 GRANT INSERT ON tst.calib2 TO val_role;
 
+-- example adhoc table
+
+CREATE TABLE tst.adhoc1
+  (exint INTEGER , exfloat NUMERIC, exstring TEXT, 
+  create_time TIMESTAMP WITH TIME ZONE NOT NULL, 
+  create_user TEXT NOT NULL );
+GRANT SELECT ON tst.adhoc1 TO PUBLIC;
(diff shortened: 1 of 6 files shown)


kutschke (approved):
Is this anyone who can meaningfully review this.

I have had a look and nothing pops up but I am not an expert.

rlcee:
> Is this anyone who can meaningfully review this.

Richie might have some thoughts to add.  The code is mostly cut-and-paste.   Ii would be good to discuss in the calibration meeting, but no need to wait for that -  as-is, there is the example in dev and nothing in prd, so we can still easily change.

kutschke:
> Richie might have some thoughts to add. The code is mostly cut-and-paste. Ii would be good to discuss in the calibration meeting, but no need to wait for that - as-is, there is the example in dev and nothing in prd, so we can still easily change.

@bonventre I will merge Ray's PR now.  I am sending this as an FYI that it's something you might want to follow.  Of course all comments are welcome.

PR #1267, opened 2024-05-18
Allow missing collections when making merged KalSeed Ptr collections

brownd1978:
Allow missing KalSeed collections to support merging multiple trigger paths


kutschke (approved):
Looks good.

PR #1269, opened 2024-05-22
fixing bug in assigning uselessTripletSeed

matthewstortini:
Just fixing the bug we saw pop up in the nightly validation.

Dave, the other changes are very trivial ones that you asked for in the last PR from Giani that we closed. Hope it's fine I included them.

### DIFF
--- a/CalPatRec/src/AgnosticHelixFinder_module.cc
+++ b/CalPatRec/src/AgnosticHelixFinder_module.cc
@@ -115,19 +115,19 @@ namespace mu2e {
     };
 
     struct cHit {
-      int     hitIndice; // index of point in _chColl
-      float   circleError2;
-      float   helixPhi;
-      float   helixPhiError2;
-      int     helixPhiCorrection;
-      bool    inHelix;
-      bool    used; // whether or not hit is used in fits
-      bool    isolated;
-      bool    averagedOut;
-      bool    notOnLine;
-      bool    uselessTripletSeed;
-      bool    notOnSegment;
-      bool    debugParticle; // only filled in debug mode -- true if mc particle, false if background
+      int     hitIndice = 0; // index of point in _chColl
+      float   circleError2 = 1.0;
+      float   helixPhi = 0.0;
+      float   helixPhiError2 = 0.0;
+      int     helixPhiCorrection = 0;
+      bool    inHelix = false;
+      bool    used = false; // whether or not hit is used in fits
+      bool    isolated = false;
+      bool    averagedOut = false;
+      bool    notOnLine = true;
+      bool    uselessTripletSeed = false;
+      bool    notOnSegment = true;
+      bool    debugParticle = false; // only filled in debug mode -- true if mc particle, false if background
     };
 
     struct tripletPoint {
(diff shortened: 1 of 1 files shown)


CalPatRec/src/AgnosticHelixFinder_module.cc
```diff
@@ -1083,7 +1077,7 @@ namespace mu2e {
           }
         }
       }
-      _tcHits[i].uselessTripletSeed = uselessSeed;
+      if (i<_tcHits.size()) {_tcHits[i].uselessTripletSeed = uselessSeed;}
```
kutschke, line 1080: Just checking.  is it normal and correct behaviour for i to be >= _tcHits.size()?  Or is this test masking a deeper error?

CalPatRec/src/AgnosticHelixFinder_module.cc
```diff
@@ -730,44 +730,37 @@ namespace mu2e {
...
-    hit.used = false;
-    hit.isolated = false;
-    hit.averagedOut = false;
-    hit.notOnLine = true;
-    hit.uselessTripletSeed = false;
-    hit.notOnSegment = true;
-
     // push back stopping target if it is to be used
     if (_useStoppingTarget == true) {
+      cHit hit;
```
kutschke, line 735: For a future PR.  A better way to write these two lines to define a c'tor for cHit that takes just this argument.  

What's written is referred to as two phase construction.  The preferred idiom is single phase construction unless that's logically impossible.

kutschke (commented):
Thanks for getting to this quickly.  I have one question inline and one suggestion for a future PR.

CalPatRec/src/AgnosticHelixFinder_module.cc
matthewstortini: Ah, I see. I will change that in my next PR.

CalPatRec/src/AgnosticHelixFinder_module.cc
matthewstortini: Yes. Just a handful on lines up (line 1067), I set i = _tcHits.size(). This is done because at that line in the for loop we have found a helix and i want to escape the i, j, and k loops right away. The helix is saved in the k loop, so i just break that one in the standard way. I am then forcing the j and i loops to end by setting i and j to be out of range. This is why line 1080 was having issues, prior to adding the if statement protection. Maybe it would be cleaner to create a bool, like "helixFound" or something, that is false by default, and then have break statement in the i/j loops when that bool is true. I can change that in my next PR, if you prefer.

kutschke: For a future PR.  There are two recommended ways to break out of a multiply nested control structure.  One is to use a goto statement.  The other is to put the nested structure in a function - then you can return from the function instead of using break.    Dave made an earlier comment about preferring more smaller functions vs fewer bigger functions.  This is one of the (many) reasons why.

brownd1978, line 1080: is the condition i>=_tcHits.size() a sign of a logical error?  Or can it occur through the normal, correct processing of the preceeding logic?   If the former, this fix is inadequate, merely covering over an error not fixing it.  If the later is true please add a comment describing the situation by which i might legitimately be >= size()

kutschke: I asked the same question and he answered above.

brownd1978: I don't see either your question or the answer.  Perhaps that was in email?

brownd1978: Oh I see, that conversation was collapsed as 'resolved'.

kutschke: I just unresolved it

brownd1978 (approved):

CalPatRec/src/AgnosticHelixFinder_module.cc
kutschke: Issue #1270 has been created to follow up on  my comments and Dave's similar comments later in the review.

kutschke (approved):
Approving with the request to follow up in later PRs with the requests made during the review and in issue #1270 .

Please submit the follow up work in self contained, reviewable PRs rather than in fewer PRs each covering many topics.

PR #1276, opened 2024-05-29
Upgrade to envset p057.

kutschke:
This envset differs from p056 only in a hack to deal with different versions of root being shipped in the sl7/ups and al9/spack distributions of art v3_14_03.

### DIFF
--- a/.muse
+++ b/.muse
@@ -1,5 +1,5 @@
 # prefer to build with this environment
-ENVSET p056
+ENVSET p057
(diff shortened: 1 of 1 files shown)


kutschke:
I tested that all of Offline,  Production,  REve,  TrkAna,  Tutorial, mu2e_trig_config  build for both sl7/ups and al9/spack.
I have not done any run-time tests.

brownd1978 (approved):

PR #1280, opened 2024-06-03
move CRVResponse text files to data

rlcee:
The cmake Offline build did not install these files correctly.  This is a fix, and also a better separation of directory content.  Checked by running several validation exe's in both builds.


kutschke:
@oksuzian and @ehrlich-uva I added you as reviewers just to give you a heads up on some bookkeeping changes that Ray is making.  I don't plan to wait for you to review before merging but I would appreciate it if you approve once you have had a look.

kutschke (approved):
Looks good.

ehrlich-uva (approved):

PR #1290, opened 2024-06-24
implement a StageParticle generator which is aware of heterogenous materials

edcallaghan:
This PR implements an event generator which is aware that materials may be composed of multiple elements, and defers the actual generation of daughters to different routines each associated with a different element. This involves the definition of both base classes defining a somewhat generic interface and concrete implementations of samplers for both i) which element is relevant for a given generator call, and ii) the spacetime coordinates of the generated daughter. To enable the use of these routines for simulating Michel electrons originating in the inner proton absorber, updates are made to the global constants listing, and approximate (read: very approximate) DIO spectra are added to the repository.

Changes to existing build rules:
-`EventGenerator/src/SConscript`: compilation now includes from/links against `GEANT4`. This is to enforce a "one definition rule" connecting materials as relevant in simulation and in event the new event generators. This is described in more detail below.

Updates to existing functionality:
-`PhysicsParams`: This interface can now be used to query the atomic radius of various elements.

New modules:
-`CompositeMaterialGenerator`: This is the generator described above. The module contains both an `art`-tool which samples elements from an existing `G4Material`, and a map from atomic labels to `art`-tools which sample starting positions/times and PDG/momenta of the generated particle, respectively. At each call, an element is sampled according the configured routine, and the chosen element's tools are then called to actually define the generated particle, which is expressed as a `StageParticle`. That the elements for which sampling tools are defined and the elements which comprise the chosen `G4Material` match is verified, but this must be deferred until "runtime" to accommodate out-of-order construction, wherein the `G4Materials` are not actually defined when the module is initialized. The use of `art`-tools allows for a very flexible configuration interface --- I was really glad to discover this!

New functionality:
-`ElementSamplerTool`: This base class defines an interface for sampling an element from a `G4Material`. This is another place with a deferred-initialization pattern, in which the actual access of a `G4Material` is deferred until the tool is actually called, to ensure that the `G4Material` has been defined.
-`AtomicVolumeSamplerTool`: A concrete implementation of the above, in which elements are sampled according to their relative volumes. This is relevant to the choice of which element a muon will be atomically captured by; the volumes are computed from the estimated atomic radii, which are supplied by the `GlobalConstantsService`/`PhysicsParams` interface.
-`PositionSamplerTool`: This base class defines an interface for sampling spacetime coordinates, given a preexisting collection of `SimParticle`s. A more generic interface would require passing the full `art::Event` into this class's sampling method, which can cause subtle data corruption involving unresolved pointers into data product collections which may present in downstream modules. Access to the data products should stay at the module-level, so the tool interface can only cover up to collections of particles. This limits its use to the generation of `StageParticles`, and thus the sampling call returns the chosen `SimParticle` to be referenced by the generated `StageParticle`.
-`MuStopDecayPositionSamplerTool`: A concrete implementation of the above. A single muon is chosen from the collection of `SimParticle`s, and its position is used as the starting position of the generated `StageParticle`. The time coordinate is sampled exponentially according to the bound muon lifetime in the relevant atom.

New auxiliary data:
-`globalConstants_01.txt`: The atomic numbers and radii of carbon and hydrogen, as well the lifetimes and fractions for muons bound therein, have been added. FYI, some of the earlier numbers here are referenced from 2013, and may want to be revisited. There are preexisting entries labeled `IPA` which are just copies of the values for carbon --- AFAIK these aren't actively used for anything, but I've left them in place nonetheless. A `czarneckiCoefficient` entry for carbon already existed. A value for hydrogen is ~required to cleanly make use of the existing interface, but a value for hydrogen has not been calculated (to my knowledge). There is currently no intent to make use of the associated functionality, so I've entered a value of zero, but this is admittedly not great.
-`ConditionsService/data/heeck_finer_binning_2016_szafron-scaled-to-{6C,1H}.tbl`: Two new pretabulated spectra have been added, which are duplicates of `heeck_finer_binning_2016_szafron.tbl` with the abscissa (first column) scaled such that the endpoint value is that of the relevant isotope. This is a crude approximation, but allows us to simulate "a tail" for standard model decays of carbon- or hydrogen-bound muons as a first pass; they can be replaced with improved tabulations as they become needed and available.

One open question I have is this: in a few places, I found it convenient to call a `GEANT4`-supplied rng, e.g. `G4Uniform()`, as opposed to sorting through the `art` interface for that. I'm not expecting a yes, but is that alright?


brownd1978 (commented):
This PR adds important new capability for capture on heterogenous materials, directly relevant for several applications.  A presentation on the physics inputs underlying this is needed, for instance the assumptions behind using the Heeck spectrum outside of aluminum, and why atomic radius is the right scaling for relative capture rates. I have a few minor technical questions as well, but basically this looks really good.

EventGenerator/inc/PositionSamplerTool.hh
```diff
@@ -0,0 +1,43 @@
...
+  // in general, need to propagate SimParticles back to midstage generators
+  // for provenance tracking
+  typedef art::Ptr<SimParticle> SimParticlePtr ;
+  typedef std::pair<SimParticlePtr,CLHEP::HepLorentzVector> ParticlePositionPair;
+  typedef std::vector< SimParticlePtr > SimParticlePtrVector;
+
+  class PositionSamplerTool{
+    public:
+      PositionSamplerTool(){ /**/ };
+     ~PositionSamplerTool(){ /**/ };
```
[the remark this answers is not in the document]
edcallaghan: Done.

EventGenerator/inc/ElementSamplerTool.hh
```diff
@@ -0,0 +1,70 @@
...
+#include <Geant4/Randomize.hh>
+
+namespace mu2e{
+  class ElementSamplerTool{
+    public:
+      ElementSamplerTool(std::string name){
+        _name = name;
+        _initialized = false;
+      }
+     ~ElementSamplerTool(){ /**/ };
```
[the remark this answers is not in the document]
edcallaghan: Done.

brownd1978 (approved):

EventGenerator/inc/ElementSamplerTool.hh
```diff
@@ -0,0 +1,70 @@
...
+      }
+    protected:
+      std::string _name;
+      G4Material* _material;
+      std::vector<std::string> _elements;
+      bool _initialized;
+
+      virtual void finish_initialize() = 0;
+      void initialize(){
+        _material = G4Material::GetMaterial(_name, false);
```
kutschke, line 45: A function with this much code belongs in a .cc file, not in a header.  We have a lot of code that does this incorrectly.  Since this is new code, let's get it right.

EventGenerator/inc/PositionSamplerTool.hh
```diff
@@ -0,0 +1,43 @@
...
+
+#ifndef EventGenerator_PositionSamplerTool_hh
+#define EventGenerator_PositionSamplerTool_hh
+
+// stl
+#include <utility>
+#include <vector>
+
+// art
+#include <art/Framework/Principal/Event.h>
```
kutschke, line 12: Same comment about < vs " as in the other PR.

EventGenerator/src/SConscript
```diff
@@ -13,6 +13,8 @@ helper=mu2e_helper(env);
 
 cryinc = os.environ['CRYHOME'] + '/src'
 crylib = os.environ['CRY_LIB']
+g4inc = os.environ['G4INCLUDE']
```
kutschke, line 16: I would like to keep all references to G4 within Mu2eG4.  The motivation is to be able to make G4-free builds which are required by the trigger.  We eventually want to get to MC-free builds but we are not there yet. 

  One generator module is already in G4 for this reason:

  Mu2eG4/src/GammaConvGenerator_module.cc

Please look into what that would require for this PR.  

Maybe it's time to make a new top level directory for event generators that depend on G4.?

Trigger/src/SConscript
```diff
@@ -67,6 +67,8 @@ helper.make_plugins( [ mainlib,
                        'xerces-c',
                        'boost_filesystem',
                        'hep_concurrency',
+                       'KinKal_Trajectory',
```
kutschke, line 70: I don't see other trigger related changes in this PR. Does this belong in this PR?

EventGenerator/inc/ElementSamplerTool.hh
edcallaghan: Done.

EventGenerator/inc/PositionSamplerTool.hh
edcallaghan: Done.

EventGenerator/src/SConscript
edcallaghan: The fundamental "problem" is that materials are defined via hard-codes in `Mu2e/src/ConstructMaterials.cc`, so that interfacing with G4 is the only way to query what they are comprised of. If there were a Mu2e-owned `Service`/interface which maintained a table of configurable materials and their compositions, which G4 queried when constructing `G4Material`s, then that would provide the functionality I'm looking for (and presumably obviate the delayed-initialization tricks I've had to use here).

But I imagine that there is a reason that things are done the way they are (and if not, implementing the kind of ting I'm imagining is a separate project). In that case, I don't particularly care whether the generator lives under `EventGenerator/`, `Mu2eG4/`, or somewhere new. Unless there is some subtly in crossing project boundaries that I'm aware of, just moving it into `Mu2eG4` would probably satisfy the separation you'd like.

kutschke (commented):
See inline comments.

EventGenerator/inc/ElementSamplerTool.hh
kutschke: Thanks

EventGenerator/src/SConscript
kutschke: Please move it into Mu2eG4 and revert the changes to EventGenerator/src/SConscript.

Implementing a complete parallel materials system seemed like far too much work for something that would never be used except to feed G4.  And we did not have a use case to drive making a simplified materials system.  It's only been in the past year or so that we have had a use case.

edcallaghan:
FYI, a recent commit resolves a lingering deficiency in which, e.g. static CLHEP random sampling methods were called which used a global random number engine. All the random sampling introduced in this PR now relies on an `art`-backed engine following the usual pattern.

EventGenerator/src/SConscript
edcallaghan: Ok --- `CompositeMaterialGenerator`, `ElementSamplerTool`, and `AtomicVolumeSamplerTool`, which depend directly or indirectly on G4, now live under `Mu2eG4`, and the uses of G4 libraries or `Mu2eG4` in `EventGenerator/src/SConscript` have been removed. `PositionSamplerTool` and `MuStopPositionSamplerTool`, which do not depend on G4, still live under `EventGenerator`.

edcallaghan:
The recent set of pushes address two points.

The first is to update the `PhysicsParams` listings for carbon --- the bound muon lifetime and decay fraction were inconsistent with each other, and with the reference from which all other values are pulled.

The second is to remedy a deficiency in the implemented element-sampling functionality for the intended use case of IPA Michels --- namely that what matters for choosing which element the Michel came from is not _just_ which atom the muon captured on, but also the different decay fractions for the different elements available (this is fairly small correction for the IPA, at ~1%). To accommodate this, some of the actual sampling functionality from `AtomicVolumeSamplerTool` was factored out into an intermediate base class `WeightedElementSamplerTool`; a new `BoundMuonDecayFractionTool` then subclasses that base class, and contains an `AtomicVolumeSamplerTool` as a member. This allows `AtomicVolumeSamplerTool` to offer the same original functionality, which is useful in all non-DIO contexts in which muons stop in composite materials, and `BoundMuonDecayFractionSamplerTool` to be used for the specific situation of DIO, without duplicating the relevant code for the relative atomic volumes. The random number engine is still propagated correctly.

brownd1978:
@FNALbuild run build test

kutschke (approved):
Thanks for updating the use of random engines and distributions to conform with our standards.  And thanks for responding to the other requests.

It good to go.

brownd1978:
@FNALbuild run build test

Issue #1272, opened 2024-05-24
StrawDigiBundle members and StrawDigiMC matching


edcallaghan:
This is a requested reconsideration of a data structure implemented in #1245, namely `StrawDigiBundle`, to avoid deep-copies. This prompted discussions about the necessity of deep-copies and dummy or null objects, which here are addressed in the opposite order, after two bits of context.

The first piece of context is that the purpose of that class is to aggregate a `StrawDigi`, `StrawDigiADCWaveform`, and `StrawDigiMC`, each of which contains partial information relevant to a tracker digi, in one place. Doing so allows for cleaner source code (i.e. simplified loops), and ensures that, in situations in which "a digi" needs to be generically considered, all of the available information is present.

The second piece of context is the nature of `StrawDigiMC`. This data product contains MC truth information associated with a tracker digi. The actual association between the MC truth and digital data products is implicit: that is, a user analyzing an art event may encounter one of two scenarios (neglecting the existence of `StrawDigiADCWaveform`s, to simplify the discussion): i) both a `StrawDigiCollection` and `StrawDigiMCCollection` are present, in which case they should be of the same size, and the nth element of the latter contains the MC truth info underlying the digital info contained in the nth element of the former. There is, e.g., no pointer inside the `StrawDigiMC` to its matching `StrawDigi`. This is typical when analyzing the output of simulation. ii) Only a `StrawDigiCollection` is present; this is the expectation of what real data will look like. These two situations are very distinct from a user's perspective: they are either doing some sort of debugging / review study and are intentionally looking at the output of simulation, or are performing an "actual analysis" and only using digital information.

[...]

PR #1291, opened 2024-06-24
Add TrkStrawHitMC provenance and guard against empty StrawDigiMCs

edcallaghan:
This PR addresses fallout from PR #1245. In particular, that PR implemented functionality which can produce dummy `StrawDigiMC` objects which do not actually contain any useful information. Issue #1272 is a discussion place for structural replacements for that situation, but as it stands these dummy objects can be produced and must be accounted for in downstream analysis, which is partially addressed and partially enabled here. This PR enables the use of the `SelectRecoMC` module on events which contain `StrawDigi`s from purely external origin.

Changes to existing data structures:
-`TrkStrawHitMC`: a provenance data member is added, in sync with and acting to the same effect as the newly-introduced provenance data member of `StrawDigiMC`. This allows to guard against inspecting the underlying information of a `TrkStrawHitMC` associated with a `StrawDigiMC` for which no meaningful truth information exists. This is not only desireable for correct analysis, but structurally necessary to short-circuit references of invalid pointers.

-`SelectRecoMC`: Loops over `StrawDigiMC`s which may attempt to inspect the truth information contain therein now check the provenance of the `StrawDigiMC` and do not proceed with such inspection if the provenance is `External`. `TrkStrawHitMC` objects which are instantiated inherit their provenance from the preexisting `StrawDigiMC`.
-`TrkMCTools::findMCTrk`: A loop over `StrawDigiMC`s now checks the provenance field before inspecting truth information, as above. This is a situation where it appears that there one stl-vector is being constructed in sync with another, but such synchronization is not actually assumed or required anywhere.


kutschke:
@edcallaghan Does this need to be tested with your PR in Production?  or are the failures a different issue?

edcallaghan:
The failure is unrelated to the PR in Production. This has something to do with the root dictionaries (I think) for a struct which has been extended in this PR. I didn't have any problems in my testing, so I'll have to look into it.

brownd1978 (commented):
The intent here is clear, I suggest a few minor cleanups.

CommonMC/src/SelectRecoMC_module.cc
```diff
@@ -250,12 +253,21 @@ namespace mu2e {
...
+      limit = 1;
+    }
+    size_t ispmax = _saveallunused? spcc.size() : limit;
     for(size_t isp=0; isp < ispmax; ++isp){
       auto const& spc = spcc[isp];
       for (size_t isdmc=0; isdmc < sdmcc.size(); isdmc++){
         auto const& sdmc = sdmcc[isdmc];
+        // if this contains no MC information, then we cannot inspect it further
+        if (sdmc.provenance() == StrawDigiProvenance::External){
+          continue;
```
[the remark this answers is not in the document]
edcallaghan: Done.

TrkDiag/src/TrkMCTools.cc
```diff
@@ -85,6 +85,12 @@ namespace mu2e {
         // loop over the hits and find the associated steppoints
         bool isactive = tshs.flag().hasAllProperties(active);
         StrawDigiMC const& mcdigi = mcdigis.at(tshs.index());
+        // if mc info is not meaningful, then skip this digi.
+        // this looks sketchy, but nowhere is an implicit association
+        // between the sct and mcdigis collection actually assumed
+        if (mcdigi.provenance() == StrawDigiProvenance::External){
```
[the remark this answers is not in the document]
edcallaghan: Done.

CommonMC/src/SelectRecoMC_module.cc
```diff
@@ -250,12 +253,21 @@ namespace mu2e {
   void SelectRecoMC::fillUnusedTSHMC( SPCC const& spcc, StrawDigiMCCollection const& sdmcc,
       Tracker const& tracker, std::shared_ptr<const StrawResponse>const& srep,
       KalSeedMC& mcseed) {
-    // either keep hits only from the primary or from all contributing
-    size_t ispmax = _saveallunused? spcc.size() : 1;
+    // either keep hits only from the primary or all contributing particles,
+    // assuming any exist. if they do not, then there is no primary.
+    size_t limit = 0;
+    if (0 < spcc.size()){
```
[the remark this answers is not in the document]
edcallaghan: That assumption that there is exactly one primary is hard-coded in the preexisting implementation. The change here addresses the situation in which no primaries exist, which can of course happen if there is no meaningful MC information present. To accommodate more than multiple primaries would require a more substantive update to the downstream logic, to identify which particles are primaries.

CommonMC/src/SelectRecoMC_module.cc
```diff
@@ -278,7 +290,20 @@ namespace mu2e {
 
   void SelectRecoMC::fillTSHMC(TrkStrawHitMC& tshmc, size_t isdmc, size_t isp,StrawDigiMC const& sdmc,
       Tracker const& tracker, std::shared_ptr<const StrawResponse>const& srep ) {
+    // always record the index, to avoid downstream corruptions
     tshmc._sdmcindex = isdmc;
+    // propagate interpretability of StrawDigiMC to TrkStrawHitMC
+    if (sdmc.provenance() == StrawDigiProvenance::External){
```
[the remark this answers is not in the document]
edcallaghan: There is no mismatch --- I've consolidated into a single cast, but there needs be be a check for the `External` case to shirt-circuit initialization of the data fields. In this case, it is an immediate `return`.

MCDataProducts/inc/KalSeedMC.hh
```diff
@@ -95,6 +104,7 @@ namespace mu2e {
     float _wireTau; // threshold cluster distance to the wire along the perpedicular particle path
     float _strawDOCA; // signed doca to straw
     float _strawPhi; // cylindrical phi from -pi to pi with 0 in Z direction
+    TrkStrawHitProvenance _provenance; // TODO default read value == Sim?
```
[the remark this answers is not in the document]
edcallaghan: A default constructor is now explicitly defined for the `TrkStrawHitMC` struct, which initializes the provenance to `External`.

CommonMC/src/SelectRecoMC_module.cc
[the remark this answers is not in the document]
edcallaghan: This is now squashed into a single line, in which one branch of the tertiary operator calls `std::min`.

MCDataProducts/inc/KalSeedMC.hh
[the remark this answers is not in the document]
edcallaghan: The default provenance has been updated to `Simulation`, so that existing simulation will be deserialized correctly.

MCDataProducts/inc/KalSeedMC.hh
```diff
@@ -95,6 +105,7 @@ namespace mu2e {
     float _wireTau; // threshold cluster distance to the wire along the perpedicular particle path
     float _strawDOCA; // signed doca to straw
     float _strawPhi; // cylindrical phi from -pi to pi with 0 in Z direction
+    TrkStrawHitProvenance _provenance; // origin/validity of MC info object
```
kutschke, line 108: If it's possible to do the default initialization here rather than in the c'tor initializer list, that would be better.  I don't know if that's possible.

CommonMC/src/SelectRecoMC_module.cc
```diff
@@ -278,7 +286,14 @@ namespace mu2e {
 
   void SelectRecoMC::fillTSHMC(TrkStrawHitMC& tshmc, size_t isdmc, size_t isp,StrawDigiMC const& sdmc,
       Tracker const& tracker, std::shared_ptr<const StrawResponse>const& srep ) {
+    // always record the index, to avoid downstream corruptions
     tshmc._sdmcindex = isdmc;
+    // propagate interpretability of StrawDigiMC to TrkStrawHitMC
+    tshmc._provenance = static_cast<TrkStrawHitProvenance>(sdmc.provenance(), true);
```
kutschke, line 292: This relies on two independent enums having the same enumerator values meaning the same thing.  It works now and it works so long as we never change one enum without making the matching changing the other.  While I think that the odds are good that we would get a way with this, it just feels dangerous.

At line 80 of KalSeedMC.hh, try the following;
using TrkStrawHitProvenance = StrawDigiProvenance;

Then get rid of all traces of TrackStrawHitProvenenance and add a comment to StrawDigiMC alerting future maintainers that this using is present and they should be aware of it.

If it works this still feels kind of icky because the name of the detail class is now a lie but it's far less icky than where we are now.

kutschke (commented):
A few followups

edcallaghan:
@kutschke Sorry, I may have been using a stale page yesterday, which is why I seemed to have ignored your comments! I agree with you that the multiple provenance classes is starting to look like the wrong pattern (and we haven't even gotten beyond tracker yet), so I've consolidated into a single `DigiProvenance` class. There's an argument that this "sounds" wrong in the `TrkStrawHitMC` context, but the reality is that it is a one-to-one pairing. @brownd1978 chatted a bit about how this will look outside of the tracker, and I think a good option is that we recommend the `DigiProvenance` class be a common structure for this book-keeping, shared between subsystems.

CommonMC/src/SelectRecoMC_module.cc
edcallaghan: See out-of-thread comment.

CommonMC/src/SelectRecoMC_module.cc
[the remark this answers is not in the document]
edcallaghan: This is now a clean `operator=` as both sides are now of common `DigiProvenance` type.

MCDataProducts/inc/KalSeedMC.hh
edcallaghan: AFAIK the default values have to be designated in the constructor (maybe in the definition for argument-supplied fields, but definitely in the initializer list otherwise).

MCDataProducts/inc/KalSeedMC.hh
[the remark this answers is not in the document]
edcallaghan: @kutschke The common `DigiProvenance` now inhabits a standalone src/inc pair.

Re default values: maybe I misunderstand your commentary, but the discussion here is about how to correctly initialize the provenance data member of `TrkStrawHitMC` when deserializing preexisting simulation in which the provenance field did not exist. Since the default value for the actual provenance enum will be `unknown`, as you asy, this needs to be explicitly overridden in the `TrkStrawHitMC` constructor to mark the preexisting simulation as having been produced as such.

TrkDiag/src/TrkMCTools.cc
```diff
@@ -85,17 +85,22 @@ namespace mu2e {
...
-        bool found(false);
-        for(auto& spc : sct ) {
-          if(spc._spp == spp ){
-            found = true;
-            spc.append(spp,isactive);
-            break;
+        // if mc info is not meaningful, then skip this digi.
+        // this looks sketchy, but nowhere is an implicit association
+        // between the sct and mcdigis collection actually assumed
+        if (mcdigi.provenance() != DigiProvenance::External){
```
brownd1978, line 91: this logic would fail to do what I think is intended if provenance() == Unknown.
I suggest adding a 'bool hasSim()' member to DigiProvenance, returning true if provenance == Sim || provenance == Mixed.  Then this line (and similar lines in other codes) becomes: if(mcdigi.provenance().hasSim()){....

brownd1978 (commented):
Thanks a lot for the cleanups the structure now loos really good. I have one last suggestion, then I think this is good for merging.

TrkDiag/src/TrkMCTools.cc
edcallaghan: This requires inheriting from an `EnumToStringSparse<>` specialization, which is more delicate than I'd have guessed (due to `explicit` constructors and the details of assignment operators). @kutschke may want to validate what I came up in my last push --- it defines one extra constructor and sprinkles is in some intermediary `static_cast<>`s. This allows us to construct the subclass and have the functionality you requested, but at the cost of only reimplementing one simple constructor.

MCDataProducts/inc/StrawDigiMC.hh
```diff
@@ -23,18 +23,9 @@
 #include <string>
 
 // Mu2e includes
-#include <Offline/GeneralUtilities/inc/EnumToStringSparse.hh>
+#include <Offline/MCDataProducts/inc/DigiProvenance.hh>
```
kutschke, line 26: I recognize that you are following the mode of existing code here but I would like you to make one more change.

My understanding of the C++ standard is that we should use the angle bracket syntax for system-defined and compiler-defined headers and that we should use the quote syntax for headers to be resolved from the -I command line options.   There is a difference.  While that difference is not likely to matter for our code for our compilers I would like to do it right.

There are several other files with this same issue.

kutschke (commented):
Thanks for making the changes.  It's great that you and Dave are thinking ahead to the other subsystems - I had not got that far yet.

I just noticed on more thing - see the inline comment.  Then i think we are done.

MCDataProducts/inc/StrawDigiMC.hh
edcallaghan: The model I've always used is "quotes for in-directory includes, and brackets for search-path includes." The standard is 200 bucks but I interpret ChatGPT as being on your side, so I've updated all of the files touched by this PR to use the quotes.

brownd1978 (approved):
Thanks for making the requested changes, this looks good.

edcallaghan:
FYI, I snuck in one extra provenance check to make the standard `Digitize` production workflow compatible with mixed samples.

edcallaghan:
...and also a higher-level but similar-purpose interface `KalSeedMC::ContainsSimulation` which is necessary to iron out downstream issues in `TrkAna`. I do not foresee adding any more functionality to this PR.

MCDataProducts/inc/DigiProvenance.hh
```diff
@@ -0,0 +1,32 @@
...
+  // enum equipped with std::string descriptions
+  class DigiProvenanceDetail{
+    public:
+      enum enum_type {unknown=0, Simulation, Mixed, External};
+      static std::string const& typeName();
+      static std::map<enum_type, std::string> const& names();
+  };
+  using StringedDigiProvenance = EnumToStringSparse<DigiProvenanceDetail>;
+
+  class DigiProvenance: public StringedDigiProvenance{
```
kutschke, line 25: Three things. First, EnumToStringSparse does not have a virtual d'tor so we should not inherit from it. That's easy to fix, we can give it a virtual d'tor and also add the other 4 rule of 5 functions as =default.  ( My style is if the compiler will do the right thing for all rule of 5 functions I just leave them out and add a one line comment to that effect  - I see that EnumToStringSparse is so old that the rule of 5 was then the rule of 3.)  Go ahead and edit EnumToStringSparse.

The second issue is the CSAID C++ experts strongly advise against using inheritance when containment will do.  We do violate this in some places.  I think we are OK here.

Third,  I think that if you add
typedef StringedDigiProvenance::enum_type enum_type;

(or try the equivalent using would be better ).

Then downstream code can avoid needing the DigiProvenanceDetail:: and you may also be able to avoid the static casts. I am not 100% sure that I have all details right here but I am pretty sure I am close.

MCDataProducts/src/DigiProvenance.cc
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  DigiProvenance::DigiProvenance(StringedDigiProvenance provenance)
+      : StringedDigiProvenance(provenance){
+        /**/
+    }
+
+  bool DigiProvenance::ContainsSimulation() const{
+    DigiProvenanceDetail::enum_type id = this->id();
+    bool rv = ((id == DigiProvenanceDetail::Simulation) || (id == DigiProvenanceDetail::Simulation));
```
kutschke, line 31: Is this a typo?  Should  one of them be Mixed?

MCDataProducts/inc/StrawDigiMC.hh
kutschke: Thanks for making the change.  While the standard costs real money, the drafts don't !  This is a recent draft of the upcoming 2023 standard.  https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/n4950.pdf  .  In section 15.3 p 466, 
 
> An implementation can provide a mechanism for making arbitrary source files available to the < > search.
>   However, using the < > form for headers provided with the implementation and the " " form for sources outside the  control of the implementation achieves wider portability. For instance:
>   #include <stdio.h>
>   #include <unistd.h>
>   #include "usefullib.h"
>   #include "myprog.h"

( A former member of the international standards committee has an office 2 doors down from me.  ).

TrkDiag/src/TrkMCTools.cc
kutschke: Thanks for giving me the heads up on this - see comments in my most recent review.

kutschke:
... and I like that you added ContainsSimulation() - it should reduce pilot error.

MCDataProducts/inc/KalSeedMC.hh
kutschke: What I was getting at is that for sufficiently simple types, you can initialize it in the same line as the declaration

_provenance = (static_cast<StringedDigiProvenance>(DigiProvenanceDetail::Simulation);

Then it is implicitly present in the initializer list of all c'tors.  What I am not sure about is whether this type is simple enough.  I bet it will be if you use the typedef I suggested in my recent review.

MCDataProducts/inc/DigiProvenance.hh
kutschke: After sleeping on it, there is an different solution that avoids introducing the third type and avoids issue with inheritance, just use a free function:

```
using DigiProvenance = EnumToStringSparse<DigiProvenanceDetail>;

bool containsSimulation(  DigiProvenance::enum_type  id ){
   return (  id == DigiProvenance::Simulation || id == DigiProvenance::Mixed );
}
```

edcallaghan: The free-function approach appears to be the path of least resistance --- it is preferable, IMO, to the extra baggage needed to make another distinct class pass muster. (FWIW, I went with inheritance over containment here because otherwise there are extra implementations to add to the list (i.e. forwarding all of the member functions of `EnumToStringSparse<>`).

That being said, I'm seeing issues with serialization of that simpler implementation (`using DigiProvenance = EnumToStringSparse<DigiProvenanceDetail>`) manifesting as data corruption after deserializing, in which the actual `enum_type` not being set correctly. Since this should be simpler the scenario, I'm not really sure what's going wrong. I'll have another try at figuring it out, but if I can't then I may need to just push updates to `EnumToStringSparse<>` to make it "inheritable."

kutschke: Sorry for the slow response.  

Can you point me to your working code that is failing and tell me what the error message is?

edcallaghan: No worries. I have the changes in a branch at https://github.com/edcallaghan/mu2e-Offline/tree/update_TrkStrawHitMC_provenance_free_functions. If you compile that (should work at head) and run the contents of `~ejc3/public/DigiProvenance`,  you can reproduce for yourself. (This is a pair of jobs exemplified in `.../run.sh`, in which we first ``physics-mix'' [pm] a few NoPrimary [np] events, and then ``digi-mix'' [dm] a few CEs into them [ce]). You can see the error in `.../dm-ce.log`, the relevant bit pasted here:
```
---- EventProcessorFailure BEGIN
  EventProcessor: an exception occurred during current event processing
  ---- ScheduleExecutionFailure BEGIN
    Path: ProcessingStopped.
    ---- ProductNotFound BEGIN
      A request to resolve an Ptr to a product containing items of type: mu2e::StrawGasStep with ProductID 2189702261
      cannot be satisfied because the product cannot be found.
      The productGetter was not set -- are you trying to dereference a Ptr during mixing?
      The above exception was thrown while processing module StrawDigiMCFilter/Triggerable run: 1202 subRun: 0 event: 1
    ---- ProductNotFound END
    Exception going through path TriggerablePath
  ---- ScheduleExecutionFailure END
---- EventProcessorFailure END
```

But: my statement that the problem is in the serialization of the provenance was confused and, I think, incorrect. After another round of less-sloppy testing, the problem presents with the current version in this PR as well.

I suspect that the actual problem is with the construction of a new `StrawDigiMC` at https://github.com/Mu2e/Offline/blob/main/TrackerMC/src/StrawDigiBundleCollection.cc#L223, ultimately stemming from the same "scope constraint" on use of art data products as I mention in https://github.com/Mu2e/Offline/pull/1290. That has nothing to do with the implementation of `DigiProvenance`, so I can push the simpler implementation for that here. If I'm right, then another PR will need to be opened which addresses the invalid `Ptr`s.

kutschke: I am looking at it now.

Just a wild guess and it's from memory.  If your are making two data products, ACollection and BCollection, both in the same module, and if one contains Ptrs into the other, you cannot dereference the Ptrs until the module has returned.  So they only work in downstream modules.  The reason is that the pointee data product does not exist until after the return ( the move and the return are like a database two phase commit ).

kutschke: Please make dm-ce.fcl readable by me. 

ls -l ~ejc3/public/DigiProvenance/dm-ce.fcl 
-rw-------+ 1 ejc3 nobody 4177 Jun 30 15:46 /nashome/e/ejc3/public/DigiProvenance/dm-ce.fcl

kutschke: It looks like my wild guess is irrelevant  The changes to Filters/src/StrawDigiMCFilter_module.cc look simple enough.  So I agree it's likely that the bug is earlier in the workflow.

That's for making the change to the free function and fixing the typo.    My reason for suggesting the free function was the same as in your reply.

MCDataProducts/src/DigiProvenance.cc
edcallaghan: Yes --- thank you for catching this!

MCDataProducts/inc/DigiProvenance.hh
edcallaghan: Sorry, `~ejc3/public/DigiProvenance/dm-ce.fcl` is now public-readable.

But, that being said, I think the issue here is actually much dumber than what I was getting at. When "digi mixing," I am mixing in the `StrawDigiMC`s, each of which contains an `art::Ptr<SimParticle>` into a `SimParticleCollection`. But I don't actually mix the `SimParticleCollection` in, so of course the pointer can't be dereferenced.

kutschke: Are all of the fixes in now?

kutschke:
@FNALbuild run build test

edcallaghan:
@kutschke To checkpoint the status of this, `DigiProvenance` is now implemented an an alias of an `EnumToStringSparse<>` specialization, in the usual pattern, and there are no issues using this for the "intended" workflow of mixing external data with current simulation. So, barring further comments, this is ready to merge.

But something we've learned here is that, because `StrawDigiMC` contains an `art::Ptr<>` into a `SimParticleCollection`, which is itself not considered in the current digi mixing implementation, trying to digi-mix preexisting simulation will not work as expected, because the information in the preloaded `StrawDigiMC`s is incomplete, as `art::Ptr<SimParticle>`s cannot be resolved. We'll need to decide what the preferable way to deal with this: either by allowing to mix the full extent of the MC information necessary for complete MC truth inspection, or by dropping support for the MC truth information when digi-mixing preexisting simulation.

kutschke:
I think you are saying that you plan to deal with the outstanding issues in a new PR.  Is that correct?  Will old workflows work correctly in the mean time?

edcallaghan:
Yes, and yes. No existing workflows are effected, but the projected workflow for mixing simulation onto simulation needs to be adjusted, which can be a separate PR.

kutschke:
> but the projected workflow for mixing simulation onto simulation needs to be adjusted, which can be a separate PR.

That works for me.  Presuming the CI comes in cleanly I will approve and merge

kutschke (approved):

PR #1293, opened 2024-06-25
fix inverted flag to propagate preloaded StrawDigiMCs

edcallaghan:
This PR fixes a bug present in https://github.com/Mu2e/Offline/pull/1245, which introduced the overlay of simulated tracker digis onto a collection of preexisting digis, which may or may not have been produced by earlier simulation. Whether or not the preexisting digis were produced by simulation is maintained by a field in `StrawDigiMC`, which is set in the `StrawDigisFromStrawGasSteps` module, following fcl configuration. The original implementation used the flag in an inverted fashion, marking presimulated digis as non-simulated, and vice versa. This does not effect any tagged datasets.


kutschke (approved):
This looks straightfoward.

edcallaghan:
Thanks!

PR #1297, opened 2024-07-02
Utilize art::PtrRemapper when mixing StrawDigiMCs

edcallaghan:
This PR improves the mixing of `StrawDigiMC`s by remapping internal `art::Ptr<>` data members thereof. Provisions for mixing tracker-related digi data products were introduced in https://github.com/Mu2e/Offline/pull/1245. When following up about the propagation of MC truth info in https://github.com/Mu2e/Offline/pull/1291, it was realized that the mixed `StrawDigiMC`s contained invalid `art::Ptr<>`s, which is solved by the functionality implemented here.

-`Mu2eProductMixer::mixStrawDigiMCs`: A loop over the mixed `StrawDigiMC`s is introduced, which utilizes `art::PtrRemapper` to update internal `art::Ptr<StrawGasStep>` fields to point into the newly-inserted (as of mixing) `StrawGasStepCollection` data product.

Changes to existing data products:
-`StrawDigiMC`: A non-const method `StrawDigiMC::strawGasSteps()` is introduced, which effectively returns a non-const reference to the relevant `art::Ptr<>`s. A mechanism for updating the `StrawDigiMC::_sgspa` field in-place must be made; the implementation of a non-const reference return is meant to mirror the analogous functionality when remapping internal `art::Ptr<>`s of `SimParticle`s during mixing.

### DIFF
--- a/EventMixing/src/Mu2eProductMixer.cc
+++ b/EventMixing/src/Mu2eProductMixer.cc
@@ -366,11 +366,11 @@ namespace mu2e {
                                           StrawGasStepCollection& out,
                                           art::PtrRemapper const& remap)
   {
-    std::vector<StrawGasStepCollection::size_type> stepOffsets;
-    art::flattenCollections(in, out, stepOffsets);
+    //std::vector<StrawGasStepCollection::size_type> stepOffsets;
+    art::flattenCollections(in, out, sgsOffsets_);
 
     for(StrawGasStepCollection::size_type i=0; i<out.size(); ++i) {
-      auto ie = getInputEventIndex(i, stepOffsets);
+      auto ie = getInputEventIndex(i, sgsOffsets_);
       auto& step = out[i];
       step.simParticle() = remap(step.simParticle(), simOffsets_[ie]);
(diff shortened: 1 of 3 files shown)


brownd1978 (approved):
looks good

EventMixing/src/Mu2eProductMixer.cc
```diff
@@ -366,11 +366,11 @@ namespace mu2e {
                                           StrawGasStepCollection& out,
                                           art::PtrRemapper const& remap)
   {
-    std::vector<StrawGasStepCollection::size_type> stepOffsets;
-    art::flattenCollections(in, out, stepOffsets);
+    //std::vector<StrawGasStepCollection::size_type> stepOffsets;
```
kutschke, line 369: I was away on vacation and just saw this now.  With few exceptions , we request that people not level commented out code in files.  The main exceptions is code under development.  It can be Ok to comment out debugging code  but it's generally preferred to wrap debugging code in a verbosity/debug_level  test.

Please keep this in mind moving forward and fix as targets of opportunity.

PR #1303, opened 2024-07-08
Made Ext Mon Collimator Geometry more realistic

fine1-sam:
- added linings
- steel shot
- repositioned ext mon collimators
- changed mother volumes (this required proton beam dump reconstruction)

### DIFF
--- a/Mu2eG4/geom/extmon_fnal_v02.txt
+++ b/Mu2eG4/geom/extmon_fnal_v02.txt
@@ -32,7 +32,21 @@ string extMonFNAL.concreteMaterial = "ShieldingConcrete";
 // may be different):  {upstream size, downstream size}
 vector<double> extMonFNAL.collimator1.channelDiameter  = {50., 50.};
 vector<double> extMonFNAL.collimator1.alignmentPlugDiameter = { 219.1, 355.6 };
-vector<double> extMonFNAL.collimator1.alignmentHoleDiameter = { 247.7, 381.0 };
+vector<double> extMonFNAL.collimator1.alignmentPlugInnerShellThickness = { 5.0, 5.0 };
+vector<double> extMonFNAL.collimator1.alignmentPlugOuterShellThickness = { 8.2, 11.1 };
+
+// Shot Liner Values
+vector<double> extMonFNAL.collimator1.shotLinerInnerDiameter = { 247.7, 381.0 };
+vector<double> extMonFNAL.collimator1.shotLinerInnerThickness = { 12.7, 12.7 };
+double extMonFNAL.collimator1.shotLinerOuterDiameter = 609.6;
+double extMonFNAL.collimator1.shotLinerOuterThickness = 9.7;
+double extMonFNAL.collimator1.length = 4203.7;
+
+// Steel Shot material
+// Density calcuated by multiplying the density of 'normal steel' (7.85g/cm^3)
+// by the densest packing of spheres (63.5%)
+string         extMonFNAL.collimator1.shot.materialName  = "SteelShot";
+
 // Transition between the two radii of the plug and hole happens over dz.
 double extMonFNAL.collimator1.radiusTransitiondZ = 0;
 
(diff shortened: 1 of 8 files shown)


kutschke:
@fine1-sam Is this PR ready for us to rerun the continuous integration tests or do you have more commits planned?

kutschke:
@FNALbuild run build test

gaponenko (approved):
Hi Rob,

I looked at the code with Sam before it was committed; I don't think I will be able to provide a meaningful review at this point.    I am happy to merge it, but maybe you or another reviewer can at least skim the changes.

Andrei

kutschke:
When do you think that Sam will be ready to give a talk on his work?  I think that next natural chance is Wed July 17 but that conflicts with the DOE review so I don't know that we will actually have that meeting.   

We might want to find a time for a special meeting  or add it to the agenda of an existing meeting.  @brownd1978 and @bechenard does this work for you?

kutschke:
> I looked at the code with Sam before it was committed; I don't think I will be able to provide a meaningful review at this point. I am happy to merge it, but maybe you or another reviewer can at least skim the changes.

I will look at it for coding and standards - likely tomorrow.  I don't think that anyone else can review it for physics content.

Mu2eG4/geom/extmon_fnal_v02.txt
```diff
@@ -32,7 +32,21 @@ string extMonFNAL.concreteMaterial = "ShieldingConcrete";
...
+vector<double> extMonFNAL.collimator1.shotLinerInnerDiameter = { 247.7, 381.0 };
+vector<double> extMonFNAL.collimator1.shotLinerInnerThickness = { 12.7, 12.7 };
+double extMonFNAL.collimator1.shotLinerOuterDiameter = 609.6;
+double extMonFNAL.collimator1.shotLinerOuterThickness = 9.7;
+double extMonFNAL.collimator1.length = 4203.7;
+
+// Steel Shot material
+// Density calcuated by multiplying the density of 'normal steel' (7.85g/cm^3)
+// by the densest packing of spheres (63.5%)
+string         extMonFNAL.collimator1.shot.materialName  = "SteelShot";
```
kutschke, line 48: I suggest using a more descriptive name: ClosePackedExtMonSteelShot . 

Also, I would not write the numerical values here - if you ever update the numbers in the .cc file it's hard to remember to fix these.  Although that's not likely, don't create multiple points of maintenance unless you are forced to by external constraints.

Same comment for collimator2.shot.materialName.

Mu2eG4/src/ConstructMaterials.cc
```diff
@@ -1660,7 +1660,14 @@ namespace mu2e {
 
      BP -> AddElement(elB, 5*CLHEP::perCent);
      BP -> AddMaterial(Poly, 95*CLHEP::perCent);
+    }
 
+    mat = uniqueMaterialOrThrow( "SteelShot");
+    {
+      //7.85 g/cm3 is the density of steel and 63.5% is the densest packing of spheres
+      double density = 7.85 * .635;
```
kutschke, line 1668: You followed the example code exactly, which is great.

I would like to start improving on this model as targets of opportunity.  When we first wrote this, constexpr did not exist - we should use it everywhere that we can and this is one place.   No need to change it in other places in this file - that's what I mean by targets of opportunity.

Mu2eG4/src/constructExtMonFNALBuilding.cc
```diff
@@ -68,169 +96,298 @@ namespace mu2e {
...
-                  geomOptions->isSolid("extMonFNAL"+collimator.name()+"alignmentHole"),
-                  forceAuxEdgeVisible,
-                  placePV,
-                  doSurfaceCheck
-                  );
+    vector<double> rOuterInnerShotLinerRadius = rInnerSteelShot;
 
-    //----------------------------------------------------------------
-    // Alignment plug
+    assert( zPlanes.size() == rOuterInnerShotLinerRadius.size() );
```
kutschke, line 223: Do you really want an assert here?  Remember that assert is only effective in debug builds.  It is not present in our normal builds, which we all "prof", because it strips the libraries of most debugger symbols, leaving only those necessary for code profiline.  See https://en.cppreference.com/w/cpp/error/assert .  When we do prof builds we define NDEBUG. 

There are other places where you also use assert.

Andrei may want to contribute to this discusision.

kutschke (changes requested):
Great work.  I have a few comments I would like you to address.  It should be quick. Then it's ready to go.

Mu2eG4/src/constructExtMonFNALBuilding.cc
gaponenko: Hi Rob, I think the assert() use got duplicated from my original code in that file.   I agree that it is better to change it to throwing a cet::exception.

kutschke: I suggest that we make a global change of this a separate PR.   Maybe encapsulate it in a small function that takes __FILE__ and __LINE__ as arguments?  Does that work for you?

gaponenko:
When I wrote about replacing the assert() with an exception I had an impression that the check was meant to detect inconsistencies in user inputs.  Taking a look at the file now I see that the asserts are comparing sizes of hand-coded vectors, so this is really a code check.  An exception for that is an overkill.  (One can introduce more bugs by writing more lines of code while the check itself is trivial.)   One can rely on checking the code by eye, but assert() is an appropriate tool here so there is an automatic check at least in debug running.  So that when a programmer suspects something is wrong with the code, they can run the debug build and be alerted to a problem.

All in all, I am now inclined to keep the asserts() in constructExtMonFNALBuilding.cc as is.
The alternative is to completely delete them, but not add anything else.

Andrei

Mu2eG4/geom/extmon_fnal_v02.txt
kutschke: Thanks.

Mu2eG4/src/ConstructMaterials.cc
kutschke: Thanks for fixing it.

kutschke:
@gaponenko let me know if you agree with the static_assert suggestion.  If so do you want to make it a separate PR and change it everywhere in the ExtMon area?

@FNALbuild run build test

fine1-sam:
@kutschke I am looking at static_assert now and will update you later today.

kutschke:
> @kutschke I am looking at static_assert now and will update you later today.

That works for me.

fine1-sam:
@kutschke I tried to implement static_assert in a few different ways. However, I kept running into the problem that the underlying vectors would need to be rewritten with constexpr, which would require changing many more values that they reference to const. 

For now, I think it makes sense to keep the code as it was (with assert). Let me know what you think.

kutschke:
You are right that static_assert only works for tests that can be evaluated at compile time .  I misunderstood Andrei's comment to mean that they were known at compile time.

I will merge this PR with the changes to date and we can decide if we want do something with this in a separate PR.  Just be aware that this test is currently not exercised as part of CI or validation.

kutschke (approved):

PR #1304, opened 2024-07-08
cmake maintenance

rlcee:


kutschke (approved):
Looks good to me.  Thanks for looking after this

PR #1305, opened 2024-07-09
Add verbose printouts for more STM beamline geometry components

AndrewEdmonds11:
These changes were useful for finding out the location and dimenions of various components in the STM beamline. For future reference, to get all the relevant printouts, set the following geometry parameters in a geometry file:

```
int ds.verbosityLevel = 1; # geom/DSShielding_v##.txt
int stm.verbosityLevel = 1; # geom/STM_v##.txt
int mbs.verbosityLevel = 1;  # geom/muonBeamStop_v##.txt
```

### DIFF
--- a/Mu2eG4/src/constructDS.cc
+++ b/Mu2eG4/src/constructDS.cc
@@ -548,6 +548,9 @@ namespace mu2e {
                 "dsShielding"
                 );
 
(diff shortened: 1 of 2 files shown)


kutschke (approved):
Looks good to me.

PR #1306, opened 2024-07-09
Add SurfaceStep to compression modules

AndrewEdmonds11:
Here is the minimal working version of compression for SurfaceSteps (#1300). Here is a summary of what can be done in each module:

 - CompressDetStepMCs: only ```noCompression``` option is allowed
 - CompressDigiMCs: the same compression is done as for StepPointMCs (i.e. it will only keep SurfaceSteps if we are keeping the corresponding SimParticle with the option to turn the compression off)

To test I have run the Validation scripts ceSteps.fcl, ceDigis.fcl, and reco.fcl. At each stage I kept all SurfaceSteps and then did a by-eye check of the print.fcl output to make sure the SurfaceSteps are kept and their SimParticlePtrs are updated correctly.

I will open a PR in Production for the necessary configuration changes. There may be some discussion needed there about what we want to run in production

### DIFF
--- a/Compression/fcl/prolog.fcl
+++ b/Compression/fcl/prolog.fcl
@@ -8,6 +8,7 @@ DetStepCompression : {
     strawGasStepCompressionLevel : "noCompression" # keep all StrawGasSteps
     caloShowerStepCompressionLevel : "noCompression"
     crvStepCompressionLevel : "noCompression"
+    surfaceStepCompressionLevel : "noCompression"
     simParticleCompressionLevel : "noCompression"
     stepPointMCCompressionLevel : "noCompression"
     keepNGenerations : -1
@@ -18,6 +19,7 @@ DetStepCompression : {
     strawGasStepCompressionLevel : "noCompression" # keep all StrawGasSteps
     caloShowerStepCompressionLevel : "noCompression"
     crvStepCompressionLevel : "noCompression"
+    surfaceStepCompressionLevel : "noCompression"
     simParticleCompressionLevel : "fullCompression" # only keep SimParticles related to other data products we are keeping
     stepPointMCCompressionLevel : "simParticleCompression" # only keep StepPointMCs related to SimParticles we are keeping
     keepNGenerations : -1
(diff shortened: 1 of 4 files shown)


Compression/fcl/prolog.fcl
```diff
@@ -28,6 +30,7 @@ DetStepCompression : {
     strawGasStepCompressionLevel : "noCompression" # keep all StrawGasSteps
     caloShowerStepCompressionLevel : "noCompression"
     crvStepCompressionLevel : "noCompression"
+    surfaceStepCompressionLevel : "noCompression"
```
brownd1978, line 33: I agree this is the only level that makes sense.

Compression/src/CompressDigiMCs_module.cc
```diff
@@ -541,6 +555,30 @@ void mu2e::CompressDigiMCs::produce(art::Event & event)
     }
   }
 
+  // Now compress the SurfaceSteps
+  for (const auto& surfaceStepsTag : _surfaceStepTags) {
```
brownd1978, line 559: the plan is to produce 1 SurfaceStep collection. If having a sequence of tags makes things more consistent this is fine, otherwise a single value would also be fine.

AndrewEdmonds11:
@FNALbuild run build tests with Mu2e/Production#330

AndrewEdmonds11:
@FNALbuild run build tests with Mu2e/Production#330

brownd1978 (approved):
This looks good Andy, thanks for the quick response.
The plan is to update the primary production scripts to save SurfaceSteps instead of StepPointMCs. Then the downstream workflows will need to be updated accordingly.

Compression/src/CompressDigiMCs_module.cc
AndrewEdmonds11: Thanks, Dave. A sequence of tags is more consistent with what StepPointMCs has. It also gives us some flexibility so that we can combine SurfaceSteps from multiple inputs like backround frames (although maybe that should be done in the MixBackgroundFrames module...)

brownd1978:
@FNALbuild run build tests with Mu2e/Production#331

PR #1307, opened 2024-07-11
Cal ad-hoc laser table

rlcee:
The ad-hoc table will allow calo group to record what CID's in their archive tables are associated with a laser run.  @simonagiov please note I've created columns for both the time archive table CID and the energy table CID.  I imagine both will be available when you make this entry.  If not, it should be fine to leave one CID as zero, and make another entry later.  Let me know what you think.


kutschke:
@simonagiov This is a gentle reminder that this PR is on your list of work to review.  Thanks.

simonagiov (commented):
Hi Ray, sorry for the delayed answer, but I was not receiving notifications from GitHub. The energyCid field is not necessary.
What I need at the end is an ad-hoc table linked to CaloCosmicsT0alig done as:

T0aligRunNumber timeCidLaser Comment

Is CalLaserRuns this or an intermediate table to achieve what is needed?

rlcee:
>T0aligRunNumber timeCidLaser Comment

I made the columns agree with this list.  Let me know if you want the table to have a different name.  I can't comment on whether this is logical because I have a shaky understanding of the needs.  If it make sense to you I think we should proceed.

simonagiov (approved):

kutschke:
@simonagiov Do you see this comment ?

kutschke:
@FNALbuild run build test

PR #1310, opened 2024-07-16
Allow ComboHitCollections to have a parent at the same level

bonventre:
Allows modules to provide filtered uniquestraw level ComboHitCollections that can still be unwound correctly for MC truth tracking.

### DIFF
--- a/CosmicReco/src/LineFinder_module.cc
+++ b/CosmicReco/src/LineFinder_module.cc
@@ -77,7 +77,7 @@ namespace mu2e{
 
       ProditionsHandle<Tracker> _alignedTracker_h;
 
-      int findLine(const ComboHitCollection& shC, art::Event const& event, CosmicTrackSeed &tseed);
+      int findLine(const ComboHitCollection& shC, std::vector<StrawHitIndex> const& shiv, art::Event const& event, CosmicTrackSeed& tseed);
   };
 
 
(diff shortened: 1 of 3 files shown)


kutschke:
@bonventre I have had a look at this.  I cannot meaningfully review the physics content without a learning curve.  But technically it is straight forward.  Do you feel that it's important for anyone else to have a look at it before I merge it?  Dave is away for the rest of the week?  If you prefer to wait for Dave's return, it's your call.

kutschke:
Nevermind  - I see it's a draft.  I will leave it for now.  Sorry for the noise

brownd1978 (commented):
This looks technically correct. I would add some comments to explain the purpose of the optional argument as allowing ComboHit subset collections at the same hierarchical level. The code would also be clearer if that nomenclature were used consistently, ie changing the variable name 'stopatfirst' to 'allowsubsets'.

What are the downsides of making 'allowsubsets' to be default true? What bad things would/could happen if the default is 'true'?  If there aren't any downsides I suggest removing 'allowsubsets' and making these code branches default.

Please investigate nesting the new functions to avoid code duplication.

kutschke:
Checking in: I think that this is waiting for Dave to followup on a comment from Richie.  Right?

bonventre:
As implemented parents at the same level are now always allowed, the optional argument determines which is chosen when you ask for the parent at that level. By default it will choose the deepest one, so if you ask for uniquestraw it should get the makeSH collection.

kutschke:
@brownd1978 and @bonventre  I am checking in.   I see that is still in draft state.  Is that correct or is it ready to move forward?

kutschke:
@brownd1978  Richie changed it out of draft status.  As I read the history I think the ball is in your court.

brownd1978 (approved):
Thanks for making the updates.

brownd1978:
@FNALbuild run build test

PR #1320, opened 2024-08-02
Fix CRY generator. Select only one intersection with the world

oksuzian:
Before, we could have events with two intersections, which was unphysical

### DIFF
--- a/EventGenerator/src/CosmicCRY.cc
+++ b/EventGenerator/src/CosmicCRY.cc
@@ -223,7 +223,7 @@ void CosmicCRY::generate(GenParticleCollection& genParts) {
                                           return a.y() < b.y();
                                         });
           //Dereference the iterator to find the projected Hep3Vector position from the world
-          Hep3Vector projectedPos = *highest_worldIntersection;
+          Hep3Vector const& projectedPos = *highest_worldIntersection;
           genParts.push_back(GenParticle(static_cast<PDGCode::type>(secondary->PDGid()),
                                          GenId::cosmicCRY, projectedPos, mom4,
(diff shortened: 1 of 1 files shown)


EventGenerator/src/CosmicCRY.cc
```diff
@@ -211,29 +211,22 @@ void CosmicCRY::generate(GenParticleCollection& genParts) {
...
+          if(_worldIntersections.size()==0)
+            throw cet::exception("Rethrow") << "No intersection with the world. Can't happen. Abort.\n";
+          //Find the iterator that has the highest intersaction with the world
+          //Very rare cases of bottom intersections are below the projection and unphysical
+          auto highest_worldIntersection = std::max_element(_worldIntersections.begin(), _worldIntersections.end(),
+                                        [](const CLHEP::Hep3Vector& a, const CLHEP::Hep3Vector& b) {
+                                          return a.y() < b.y();
+                                        });
+          //Dereference the iterator to find the projected Hep3Vector position from the world
+          Hep3Vector projectedPos = *highest_worldIntersection;
```
kutschke, line 226: I think this should be 

`Hep3Vector const&`

kutschke (changes requested):
Looks OK to me.  Please make the one small change.

EventGenerator/src/CosmicCRY.cc
oksuzian: Thanks for looking into this, @kutschke .
I've implemented your suggestion.

kutschke:
@FNALbuild run build test

PR #1322, opened 2024-08-06
Add normalization printout to generator tool

brownd1978:
This is an intermediate PR, progressing towards adding normalization metadata to generator outputs to facilitate downstream analysis. This current PR just prints the relevant information to cout. Subsequent PRs need to design a class that covers this and other generators. Refactoring of BinnedSpectrum may ultimately be needed. The current change results in this printout (for example) running DIOtail:

Restricted Spectrum min 75.005 max 104.965
Full Spectrum min 1.005 max 104.965
Restricted Spectrum integral 4.15143e-05
Full Spectrum integral 99.1255
Sampled spectrum fraction 4.18805e-07

### DIFF
--- a/CommonMC/fcl/TestSurfaceSteps.fcl
+++ b/CommonMC/fcl/TestSurfaceSteps.fcl
@@ -9,7 +9,12 @@ services : @local::Services.Reco
 physics :
 {
   producers : {
-    MakeSS : @local::CommonMC.MakeSS
+    MakeSS : {
+      @table::CommonMC.MakeSS
+      VDStepPointMCs : "compressRecoMCs:virtualdetector"
+      IPAStepPointMCs : "compressRecoMCs:protonabsorber"
+      TargetStepPointMCs : "compressRecoMCs:stoppingtarget"
+    }
   }
   TrigPath : [MakeSS]
   analyzers : {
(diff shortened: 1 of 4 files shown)


kutschke (approved):

sophiemiddleton (approved):
looks good to me

kutschke:
@rlcee   and   @oksuzian : a gentle reminder that this PR is on your list to review.

rlcee (approved):
I'm approving assuming these questions have answers
- why not do the integral in the BinnedSpectrum class
- what is the adjustment of the first bin in the full integral, and, assuming the answer is some ancient lore, then why does it not count in the cut integral?

kutschke:
@oksuzian  - Hi Yuri - let me know if you see this post.

brownd1978:
> * why not do the integral in the BinnedSpectrum class
BinnedSpectrum has to handle a lot of different physics cases. I can only provide a correct implementation for DIO.
> * what is the adjustment of the first bin in the full integral, and, assuming the answer is some ancient lore, then why does it not count in the cut integral?
The Heeke/Szafron spectrum prediction only goes down to 1 MeV. To get the full integral you must assume something in the range 0-1. The authors suggest a linear extrapolation to 0, which is what I implemented. Technically I should include that when calculating the numerator, but there is no physics case for such a low (<1MeV) threshold.

oksuzian (approved):

kutschke:
@FNALbuild run build test

kutschke:
@oksuzian Yuri, please let me know if you see this.

PR #1324, opened 2024-08-09
Set calo-tracker time offset in extracted data

brownd1978:
We observe that the calo-tracker time offset is different in extracted sims compared to beam sims. The cause of this difference is not understood. This PR just configures the reco to be self-consistent, which aids in making track-calo associations in the fit.


kutschke:
This looks good to me.  I will merge after Richie approves.

FYI.  There are other timing related items in the pipeline from me.  We don't need them for near term work but we will need them before we see data.  They are on hold until after the shielding geometry.

kutschke:
@bonventre a gentle reminder that this PR is on your list of PRs to review

bonventre (approved):

PR #1326, opened 2024-08-19
Add code to convert between DUSAF and Mu2e coordinate systems.

kutschke:
This PR is a pure addition, no changes.

I debated whether to add this in GeometryService or GlobalConstantsService. I settled on GeometryService on the grounds that it is more Geometry related than just-plain-constants.  It also preserves the possibility that it could evolve with a re-survey of the tie-in point, although that is not expected.


kutschke:
Check to see if the clang-tidy limitations worked.

@FNALbuild run build test

rlcee (approved):
Once again I missed the assignment.  This all looks good.  I did not check the numbers vs the notes, let me know if I should do that.

kutschke:
>  I did not check the numbers vs the notes, let me know if I should do that.

No need.

PR #1329, opened 2024-08-22
Fixes for s132 (Don't include GeometryService in rootcling)

eflumerf:
Update CMakeLists.txt for new files.

Make it so check_cmake.sh can be run as `./bin/check_cmake.sh` (no copy to base directory)

### DIFF
--- a/Analyses/CMakeLists.txt
+++ b/Analyses/CMakeLists.txt
@@ -142,6 +142,15 @@ cet_build_plugin(CountPionDecays art::module
       Offline::MCDataProducts
 )
 
+cet_build_plugin(CountVDHits art::module
+    REG_SOURCE src/CountVDHits_module.cc
+    LIBRARIES REG
+      
+      
+      Offline::DataProducts
+      Offline::MCDataProducts
+)
+
 cet_build_plugin(CRYGenPlots art::module
     REG_SOURCE src/CRYGenPlots_module.cc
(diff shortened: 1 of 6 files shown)


kutschke:
@eflumerf is the need to exclude the section of Surface steps drive by the new root in art 3.15 or is it driven by something intrinsic to cmake?  My own opinion is that data products should not use services but we have never been able to get agreement on that.

eflumerf:
Basically, in s132, fhicl-cpp 4.19.00 includes c++20 code. ROOT cling does not understand c++20, so you cannot have a `#include "fhiclcpp/ParameterSet.h"` visible to any files which are processed by `rootcling` (i.e. included in a dictionary).

The change I made hides the GeometryService methods from `rootcling`, as they shouldn't be needed for object persistence in ROOT.

eflumerf:
I think the result of this change to fhiclcpp is that you _cannot_ include services in data products until ROOT is updated to support c++20 in `rootcling`, which I think has an indefinite time-scale.

kutschke:
> I think the result of this change to fhiclcpp is that you _cannot_ include services in data products until ROOT is updated to support c++20 in `rootcling`, which I think has an indefinite time-scale.

Thanks Eric.  I will approve and merge the PR and create an issue to discuss getting services out of data products, which I think is the preferred solution.

PR #1331, opened 2024-08-25
 fixed CMake files

gianipez:


kutschke:
As I read the time stamps, the second commit came after the test.  So redoing the CI.

@FNALbuild run build test

kutschke (approved):

PR #1333, opened 2024-09-06
Seed CentralHelix fit with cosmictrackseed

bonventre:
CentralHelixFit takes cosmictrackseeds instead of building seed from timeclusters. Updates cosmictrackseedfilter and removes duplicate filter module


kutschke:
@FNALbuild run build test with Mu2e/Production#346,Mu2e/mu2e_trig_config#39

kutschke:
This CI run succeeded.   The red X is an artifact because the CI scripts are missing one detail - so it fails and falls back to a a default that happens to be correct.

brownd1978 (approved):

PR #1336, opened 2024-09-12
extrapolation fixes

brownd1978:
Small fixes and optimization for extrapolation. This PR is paired with Production PR 347.


brownd1978:
@FNALbuild run build test with Mu2e/Production#347

PR #1337, opened 2024-09-12
fixed the TriggerInfoMerger module

gianipez:


brownd1978 (commented):
Can you please explain how the copies of tracks etc are intended to be used? Is it  for the merged triggerinfo to reference? You make deep copies, and the triggerinfo references will still point to the old objects, so I'm not sure how this is supposed to work.

gianipez:
Hello Dave, These deep copies are meant to be used ONLY when you want to generate ntuples or running an analyzer job on the collections of the objects that passed the trigger selection. In this way, the I/O interface doesn’t have to be manipulated when instead you want to read the collections from the producer module instances. In Stntuple for instance, this allows to create an ntuple where we have all these blocks together in the same file in a very clean way.Sent from my iPhoneOn Sep 13, 2024, at 1:33 AM, David Brown ***@***.***> wrote:﻿
@brownd1978 commented on this pull request.

Can you please explain how the copies of tracks etc are intended to be used? Is it  for the merged triggerinfo to reference? You make deep copies, and the triggerinfo references will still point to the old objects, so I'm not sure how this is supposed to work.

gianipez:
> Deep copies always make me nervous, but in this case it makes sense. We could use that also for the nightly trigger validation. We might even want to persist the copies instead of (or in addition to?) the originals. Can you change the name of the option to 'DeepCopy' or something like that instead of 'ntuple mode' since it could have a broader set of use cases.
> 
> I don't see any logic to select only objects where the trigger actually passed, is that not necessary? What happens for instance if this is a random trigger? Does it make sense to copy the objects in that case?

Glad to hear that! This module is run as the last block of all the trigger paths. The only paths where this is not run are the minBias and the random trigger. Each trigger path has its own module instance of this module, so if within a trigger path we get to run the MergeTriggerInfo module, it means that the trigger path selected the event.

brownd1978 (approved):

PR #1341, opened 2024-09-20
pions

sophiemiddleton:
pion filters for RPC

### DIFF
--- a/EventGenerator/src/PionFilter_module.cc
+++ b/EventGenerator/src/PionFilter_module.cc
@@ -61,11 +61,11 @@ namespace mu2e {
         for(const auto& aParticle : *ah){
           art::Ptr<SimParticle> pp(ah, aParticle.first.asUint());
           float _endglobaltime = pp->endGlobalTime();
-          if( pp->stoppingCode() == ProcessCode::mu2eKillerVolume and std::abs(part->pdgId()) == PDGCode::pi_plus){
+          if( pp->stoppingCode() == ProcessCode::mu2eKillerVolume and std::abs(pp->pdgId()) == PDGCode::pi_plus){
             const PhysicsParams& gc = *GlobalConstantsHandle<PhysicsParams>();
             totalweight += exp(-1*pp->endProperTime() / gc.getParticleLifetime(pp->pdgId()));
           }
-          if( pp->stoppingCode() == ProcessCode::mu2eKillerVolume and (std::abs(part->pdgId()) == PDGCode::pi_plus and _endglobaltime > tmin_ and _endglobaltime < tmax_ )){
+          if( pp->stoppingCode() == ProcessCode::mu2eKillerVolume and (std::abs(pp->pdgId()) == PDGCode::pi_plus and _endglobaltime > tmin_ and _endglobaltime < tmax_ )){
             passed = true;
             const PhysicsParams& gc = *GlobalConstantsHandle<PhysicsParams>();
             selectedweight += exp(-1*pp->endProperTime() / gc.getParticleLifetime(pp->pdgId()));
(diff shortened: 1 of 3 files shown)


sophiemiddleton:
Still need to do some work on the filter to tidy it up

EventGenerator/src/RPCGun_module.cc
```diff
@@ -152,14 +154,12 @@ namespace mu2e {
       const PhysicsParams& gc = *GlobalConstantsHandle<PhysicsParams>();
       double weight = 0.;
       double tau = part->endProperTime() / gc.getParticleLifetime(part->pdgId());
-
       while(part->parent().isNonnull()) { //while particle has a parent which is not null
         part = part->parent();
-        if ( abs(part->pdgId()) == PDGCode::pi_plus ) { //if pion
+        if ( abs(part->pdgId()) == 211 ) { //if pion
```
kutschke, line 159: Two things:

1. Always use std::abs, never bare abs.  It leaves you vulnerable to changes in behaviour if transitively included headers change.
2.  Always use symbolic names.  I excpect you can use std::abs(PDGCode::pi_plus).

EventGenerator/src/RPCGun_module.cc
```diff
@@ -169,20 +169,18 @@ namespace mu2e {
     auto output{std::make_unique<StageParticleCollection>()};
     const auto simh = event.getValidHandle<SimParticleCollection>(simsToken_);
     const auto pis = stoppedPiMinusList(simh);
-
+    std::cout<<"event: "<<event.id().event()<<std::endl;
```
kutschke, line 172: Please protect this printout with a verbosity flag.

EventGenerator/src/PionFilter_module.cc
```diff
@@ -0,0 +1,99 @@
+// Purpose: Event filter for RPC simulations
```
kutschke, line 1: This module should be split up into a filter module and an analyzer that produces the TTree.

kutschke (changes requested):

EventGenerator/src/PionFilter_module.cc
kutschke: This means that the isNull parameter will not be needed.

EventGenerator/src/RPCGun_module.cc
sophiemiddleton: yes, probably this needs removing.

EventGenerator/src/PionFilter_module.cc
sophiemiddleton: I will remove the ntuple, it was for debugging

EventGenerator/src/RPCGun_module.cc
sophiemiddleton: yes, was part of a test due to wierd result, wanted to make sure it was parsing.....fixed now

EventGenerator/src/PionFilter_module.cc
kutschke: OK

kutschke:
Please look through the code and fix other instances of abs to std::abs.

sophiemiddleton:
thanks, I found two other instances, now fixed

sophiemiddleton:
@FNALbuild run build test

sophiemiddleton:
Any more comments on this, I would like to begin making pion samples this week

kutschke (approved):
For a future PR, please split RPCGun_module into a generator, without histograms, and an analyzer that makes histograms.  In the very early days we encouraged a single module that does both but we changed our minds some years ago and we want to move everything in the new direction.  SInce I was not available to pay attention to PRs for so long, I won't insist on it now.

Added issue #1352 to address this.

kutschke:
@FNALbuild run build test

kutschke:
This PR has no coverage in the nightly validation - as best I can tell it affects only JobConfig/beam/POT_infinitepion.fcl .  The oustanding request will be covered in Issue #1352 .  Merging now.

PR #1342, opened 2024-09-24
First pass at KalSeedCollection sculpting modules

edcallaghan:
This PR implements a system of modules for filtering `KalSeed`s out of events using arbitrary 1-dimensional functional prescales across a few variables. ~~This PR is marked as a draft for now; modification and much more details will be added here before it's necessary to initiate review.~~

New module:
  - `KalSeedFunctionalPrescale`: As the name suggests, this module produces a `KalSeedCollection` which results from applying a functional prescale,(meaning a random prescale which is calculated as a function of the `KalSeed` itself, to an existing `KalSeedCollection`. The prescale calculation itself is deferred to a "tool" implementing the `KalSeedFunctionalPrescaleTool` interface, described below.
  
  New tools:
   - `KalSeedFunctionalPrescaleTool`: Abstract base class which defines the interface for calculating a prescale factor as a function of a `KalSeed`.
   - `UnivariateLookupKalSeedPrecaleTool`: Abstract base class which defines the interface for calculating a prescale using pretabulated samples of a one-dimensional function. Concrete implementations define a method to calculate the observable on which to prescale on from a given `KalSeed`. The prescale tables are assumed to be structured with equal sample spacing, and are supplied via fields in the `fcl` configuration of the implemented tool - this is not great, and probably should be changed to standalone ascii files, such as how input spectra for event generators are supplied. The prescale calculation is achieved via a `boost::cardinal_cubic_b_spline` within the range over which the table is supplied; outside of this range, the prescale is assumed to be zero.
   - `MomentumLookupKalSeedPrescaleTool`: Concrete implementation of the above which prescales on the magnitude of the fit momentum at the "first" downward-going sample of the track in a configurable collection of surfaces.
   - `TrackerExitAzimuthLookupKalSeedPrescaleTool`: Concrete implementation of the above which prescales on the azimuthal coordinate of the track when exiting the tracker.
   - `QuasiImpactParameterKalSeedPrescaleTool`: Concrete implementation of the above which prescales on "rmin," i.e. the approximate impact parameter of the track assuming vacuum. This is a stop-gap meant to fudge a selection of tracks which began or passed through the stopping target.


brownd1978 (commented):
Would it make sense to extend this module to use extrapolated positions at the stopping target instead of the tracker?

edcallaghan:
@FNALbuild run build test

edcallaghan:
@brownd1978 there are three prescales involved here. For the blinding proposal, they are: i) momentum at tracker entrance, ii) azimuth at tracker exit, and iii) ~impact parameter.

I'd rather leave (i) as-is until people have looked into using extrapolated at-target momenta more deeply. But (ii) and (iii) could be changed to do a binary selection on stopping-target origin and at-target azimuth.

brownd1978:
@FNALbuild run build test

brownd1978:
@FNALbuild run build test

PR #1343, opened 2024-09-24
First pass at extracting digis from tracks

edcallaghan:
This PR implements a modules which deep-copies `StrawDigi`s and `StrawDigiWaveform`s which are associated with a reconstructed track via ` KalSeed` into new `*Collection`s. ~~This PR is marked as a draft for now; more detail will be added here before it's necessary to initiate review.~~

New module:
- `TrackDigiExtractor`: Given a `KalSeedCollection`, produces a `StrawDigiCollection` and `StrawDigiADCWaveformCollection` which contain data products pointed to by `KalSeed`s in the specified collection. Future development implement include data products from the calorimeter.


brownd1978 (commented):
I wonder if it makes sense to combine this module with the time-shifting module, as I suspect both will end up needing to reference a track, and both otherwise produce and consume the same products. It seem more efficient to do this in one pass.

edcallaghan:
@FNALbuild run build test

edcallaghan:
> I wonder if it makes sense to combine this module with the time-shifting module, as I suspect both will end up needing to reference a track, and both otherwise produce and consume the same products. It seem more efficient to do this in one pass.

@brownd I envision the two as needing to be distinct via the following reasoning:
  i) The offspill job needs to write an output stream containing the digis which may be used for blinding. To minimize the size on disk, that stream would only consist of digis taken from this module.
 ii) The onspill job performs the actual sculpting, which in this case (time) may depend on the onspill run configuration (we could change the digitization window, or something like that).

When you propose that this part of the chain would reference the track, I assume you mean to use some fit `t0` as the reference time for the track. That makes sense, but we would use the time relevant for the fit after mixing these digis with onspill pileup, not the time of the relatively clean offspill track, so there's no need to propagate the latter from the offspill job.

brownd1978 (approved):

brownd1978:
@FNALbuild run build test

Blinding/src/TrackDigiExtractor_module.cc
```diff
@@ -0,0 +1,122 @@
...
+      throw cet::exception("TrackDigiExtractor") << msg << std::endl;
+    }
+
+    // deep-copy digis underlying tracks
+    auto ext_strawdigis = std::make_unique<StrawDigiCollection>();
+    auto ext_strawadcss = std::make_unique<StrawDigiADCWaveformCollection>();
+    for (const auto& kalseed: *kalseeds){
+      // tracker
+      auto seeds = kalseed.hits();
+      for (const auto& seed: seeds){
```
brownd1978, line 97: Its not clear to me we want to keep the inactive hits in this use case.

Blinding/src/TrackDigiExtractor_module.cc
```diff
@@ -0,0 +1,122 @@
...
+
+  void TrackDigiExtractor::produce(art::Event& event){
+    // data-product lookups
+    auto kalseeds = event.getValidHandle<KalSeedCollection>(_kalseed_tag);
+    auto strawhits = event.getValidHandle<StrawHitCollection>(_strawhit_tag);
+    auto strawdigis = event.getValidHandle<StrawDigiCollection>(_strawdigi_tag);
+    auto strawadcss = event.getValidHandle<StrawDigiADCWaveformCollection>(_strawdigi_tag);
+
+    // sanity checks
+    if (strawhits->size() != strawdigis->size()){
```
brownd1978, line 82: This test relies on this module's config being consistent with what was used when fitting the tracks.
 
A way to ensure consistency is to use the KKTrack instead of KalSeed; KKTrack can have Ptrs back to underlying collections, but it cannot be persisted, due to payload that includes std::Ptrs. We can discuss if that makes sense in this context.

Blinding/src/TrackDigiExtractor_module.cc
```diff
@@ -0,0 +1,122 @@
...
+#include "Offline/RecoDataProducts/inc/StrawHit.hh"
+#include "Offline/RecoDataProducts/inc/StrawDigi.hh"
+#include "Offline/RecoDataProducts/inc/TrkStrawHitSeed.hh"
+
+namespace mu2e{
+  class TrackDigiExtractor: public art::EDProducer{
+    public:
+      struct Config{
+        // track
+        fhicl::Atom<art::InputTag> kalseed_tag{
```
brownd1978, line 31: Is it conceivable this module will need to coherently process multiple KalSeed collections (ie multiple particle hypotheses)?

edcallaghan: It is conceivable, but IMO unlikely for the conversion-blinding use case where we really only want electron-like tracks. Maybe we'll determine that selecting another stream is useful for some book-keeping purpose, but I doubt that we'd want to collate that with the electron-likes at this stage.

Blinding/src/TrackDigiExtractor_module.cc
edcallaghan: The only place I can identify this being set is in the `TrkStrawHitSeed` constructor, where it propagates from a `WireHitState` --- it looks like this is still a fit-level construct (e.g., it contains the decision on left/right ambiguity), so is contextual to the fit. If that's the case, then I think we'd want to keep the inactive hits, as they could become active when fit in the presence of pileup. Is that a reasonable take? 

Or, if I'm wrong about the above, then we have to ask whether only propagating active hits risks a break in blindness. Since these are overlaid onto pileup which could contribute hits to the fit but end up inactive, the answer may be no, but it's worth mentioning.

Blinding/src/TrackDigiExtractor_module.cc
edcallaghan: Hmmm, ok. The current implementation does depend on this check passing, as e.g. `StrawHit`s and `StrawDigi`s do not contain any ptrs to each other. To be sure that both for a given hit are included, we have to assume that the collections in which they are contained are aligned (and the best check we have for that is same-length).

Blinding/src/TrackDigiExtractor_module.cc
brownd1978: In OffSpill cosmic track events, the inactive hits will mostly be poorly measured hits created by the primary particle, so you want to keep them. However in a mixed event, many/most of the inactive hits will be from pileup, so you would not want to keep them if you planned to re-mix the extracted digis.
I think the current code covers 90% of the foreseeable use cases, lets go with this for now. If in future we want to use this code to extract digis from events with pileup we can extend it to include a hit selection algorithm.

Blinding/src/TrackDigiExtractor_module.cc
brownd1978: There is a long-term issue that we need to improve the robustness of the digi->hit->trkhit indexing using ProductPtr, which didn't exist when the original codes were written. That's beyond the scope of this PR. We can deal with this code when we address that issue in general.

Blinding/src/TrackDigiExtractor_module.cc
edcallaghan: Ok, that all makes sense.

Blinding/src/TrackDigiExtractor_module.cc
edcallaghan: Sounds good to me.

PR #1344, opened 2024-09-24
Doctoring StrawDigi timing to shift tracks in time

edcallaghan:
This PR implements a module which coherently shifts the TDC values of a `StrawDigiCollection`, so as to shift the timing of a to-be-reconstructed track. Future development should implement a similar shift for calorimeter data products. ~~This PR is marked as a draft for now; more detail will be added here before it's necessary to initiate review.~~

New modules:
  - `DisplaceDigiTimes`: A  module which coherently shifts the timing of all digis. The shift is calculated by determining a reference time of the digis, and a target time to shift to. The reference time is currently the reconstructed time as reported by a `KalSeed` object, which is assumed to be the only `KalSeed` in its collection. For the blinding use-case, this is currently a reasonable assumption, but we should think about how to ensure it upstream. The target time is determined by random-sampling from a pretabulated distribution, using the existing `BinnedSpectrum` utility and `CLHEP`-assisted pattern.

I think that the main thing to think about is that, for the blinding flow, at some point we have to make an association between between a single `KalSeed` and a `StrawDigiCollection`. Right now, having this check here and an upstream filter to select events with a single `KalSeed` works, but maybe there is a more elegant way (e.g. just storing ``the track'' as a data product in the event, so that it is unique).


edcallaghan:
@bonventre Can you confirm that the tracker waveforms are time-invariant, i.e. if we are shifting the digi tdc values in time, we do not need to also adjust the position of the signal inside the waveforms?

Blinding/src/DisplaceDigiTimes_module.cc
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    // if no valid intersection found, then we cannot use this event
+    if (!adequate){
+      std::string msg = "no valid KalIntersection found";
+      throw cet::exception("DisplaceDigiTimes") << msg << std::endl;
+    }
+    // otherwise, take the time of the intersection as the reference time
+    double reference_time = intersection.time();
+    double target_time = this->sample_target_time();
```
brownd1978, line 159: Does 'target_time' here refer to the physical stopping target? A KalSeed can intersect multiple foils, is the intent here to shift to a generic intersection, ignoring the specific foil? Or will the sampling time somehow take the foil into account?

Blinding/src/DisplaceDigiTimes_module.cc
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  double DisplaceDigiTimes::sample_target_time(){
+    const double scaled = _target_sampler->fire();
+    const double rv = _target_distribution.sample(scaled);
+    return rv;
+  }
+
+  TrkTypes::TDCValue DisplaceDigiTimes::straw_analog_to_digital(double time, const StrawElectronics& electronics){
+    auto rv = electronics.tdcResponse(time);
```
brownd1978, line 201: What's the benefit of wrapping the electronics function here?

Blinding/src/DisplaceDigiTimes_module.cc
```diff
@@ -0,0 +1,211 @@
...
+    return rv;
+  }
+
+  double DisplaceDigiTimes::straw_digital_to_analog(TrkTypes::TDCValue time, const StrawElectronics& electronics){
+    auto rv = static_cast<double>(time) * electronics.tdcLSB();
```
brownd1978, line 206: should this function be moved to StrawElectronics?
Are there edge effects one needs to worry about? How about the channel-specific effects described [here](url)?

brownd1978 (commented):
I like the concept of shifting to an intersection time, but there are some details I'd like to understand, as that returns calibrated times, and the target is an extended object with multiple foils potentially intersected.

edcallaghan:
> I like the concept of shifting to an intersection time, but there are some details I'd like to understand, as that returns calibrated times, and the target is an extended object with multiple foils potentially intersected.

@brownd1978 I think we're better off using e.g. the time of the front of the tracker. (which I realize is not what one would glean from this PR) --- I wrote "target" in the mathematical sense, not the Mu2e meaning. As you say, we won't generally have one well-defined intersection with the target, and I don't think it's worth constructing a way to deal with that when it's at best isomorphic to using a different location which will have a unique intersection.

Blinding/src/DisplaceDigiTimes_module.cc
edcallaghan: This is just for a symmetric and easy-to-parse interface, i.e. a non-expert could pick up "analog-to-digital" and "digital-to-analog" faster than "tdcResponse" and "tdcLSB."

Blinding/src/DisplaceDigiTimes_module.cc
edcallaghan: I don't feel strongly about shifting this into the electronics class or not. If there is any other use case for inverting (IMO, likely) then it would make sense to. Maybe @bonventre can comment on nuance that may be missing --- I just didn't see anything else in the existing electronics utility to make use of.

Blinding/src/DisplaceDigiTimes_module.cc
edcallaghan: As I said in the out-of-thread comment, this was nomenclature clash (target of a mathematical operation vs stopping target). I think we use a location with a unique (at least, we expect unique) intersection. But we do have leeway in choosing where to sample, because the distribution we are shifting to is also within our control (and we can take it from data or MC as we see fit).

brownd1978 (approved):

brownd1978:
@FNALbuild run build test

Blinding/src/DisplaceDigiTimes_module.cc
bonventre: Right now this is missing _electronicsTimeDelay from strawelectronics. I think this is a redundant parameter that is set to zero and can be removed, but for now maybe its easier to put this function in strawelectronics and add this offset back in there.

bonventre (changes requested):

bonventre (approved):

brownd1978 (commented):
One final comment: given my confusion about the term 'target' and given this code will be used as a template for others, can you swap in a different term for where the timing is sampled? Maybe 'reference'?

PR #1345, opened 2024-09-24
First pass at inclusive digi merging module

edcallaghan:
This PR implements a module which, mechanically, flattens multiple `StrawDigiCollection`s and `StrawDigiADCWaveformCollection`s into single collections, and defers to an already-factored-out collision-resolution algorithm. ~~This PR is marked as a draft for now; more detail will be added here before it's necessary to initiate review.~~

New module:
- `MergeDigis`: Given specified data product collections for tracker digis, produce a new collection which is the concatenation of those collections subject to an already-factored-out collision resolution policy. Future development should include an implementation for the calorimeter, and potentially crv, digis.

### DIFF
--- a/Blinding/src/MergeDigis_module.cc
+++ b/Blinding/src/MergeDigis_module.cc
@@ -1,0 +1,88 @@
+// Ed Callaghan
+// Collate digis from multiple sources, and apply collision resolution
+// August 2024
+
+// stl
+#include <string>
+#include <vector>
+
+// art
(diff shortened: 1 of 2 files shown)


edcallaghan:
@FNALbuild run build test

edcallaghan:
@brownd1978 @bonventre Is there a sorting that should be applied here to maintain blindness? e.g. "all digis in a real event are channel- (or time-) ordered."

brownd1978 (approved):

PR #1346, opened 2024-09-24
Simple scheme for a crytopgraphic data product

edcallaghan:
This PR implements a simple data product, a `BigNumber`, which for now does nothing but wrap `std::string`, and two modules which implement a simple encryption scheme for `BigNumbers` via standard `GMP` calls. ~~This PR is marked as a draft for now; more detail will be added here before it's necessary to initiate review.~~

The encryption scheme used is the Goldwasser-Micali cryptosystem applied to a single bit of information. This is a probabilistic encryption scheme, i.e. a key-invertible one-to-many mapping which relies on the computational hardness of i) factoring large composite integers, and ii) deciding quadratic residuosity of composite numbers. References for these topics as they relate to the cryptosystem are included in commentary in the low-level implementation.

The low-level implementation of the cryptosystem relies on the GNU MultiPrecision library (GMP), which provides both a space-efficiency arbitrary-length integer container, and correct/efficient implementations of necessary number-theoretic functions.

New data products:
- `BigNumber`: As stated above, this class wraps a standard lib string, and is meant to act as a transparent, if not efficiency, container for arbitrary-length integers.

New Modules:
- `BigNumberProducer`: Creates and inserts a single `BigNumber`, specified in configuration, into the event. This would be used, for tagging salt events in a blinding scheme, to either label a data stream as tagged (`BigNumber` of `1`) or not (`0`).
- `BigNumberChecker`: Checks two `BigNumber`s of different provenances for (string-based) equality. This is used for testing the encryption/decryption cycle.
- `GoldwasserMicaliEncrypter`: Encrypts a single bit of information (a preexisting `BigNumber` being nonzero or not), and stores the encrypted value as a new `BigNumber` into the event. The configuration includes the cryptosystem public key, which is comprised of both a) the product of two prime numbers, and b) a quadratic nonresidue of that product. The random state utilized when randomizing the encryption is initialized using a random seed from `mu2e::SeedService`.
- `GoldwasserMicaliDecrypter`: Decrypts a single bit of information from a `BigNumber`, and stores the result as a new `BigNumber` in the event. The configuration includes the cryptosystem private key, which is comprised of two unique prime numbers.

New functionality:
- `GMPRoutines`: A collection of relatively high-level number-theoretic routines and the core encryption/decryption routines, performed in GMP.


brownd1978 (commented):
Are there any any citations or credits for the algorithms you're using that should be included in the comments?

edcallaghan:
> Are there any any citations or credits for the algorithms you're using that should be included in the comments?

I've added both a reference to Goldwasser and Micali's primary publication describing the method, as well as pointers to the relevant wikipedia pages, which are a bit easier to parse when getting a grip on the manipulations being done.

brownd1978 (approved):

brownd1978:
@FNALbuild run build test

brownd1978 (approved):

brownd1978:
@FNALbuild run build test

PR #1348, opened 2024-09-30
Add optional longitudinal residual to kinkal fits

bonventre:
Enabled in CentralHelix and KinematicLine fits

### DIFF
--- a/CosmicReco/fcl/prolog.fcl
+++ b/CosmicReco/fcl/prolog.fcl
@@ -42,7 +42,7 @@ CosmicReco : { @table::CosmicReco
 
    SimpleTimeCluster : {
       module_type : SimpleTimeCluster
-      ComboHitCollection : "makePH"
+      ComboHitCollection : "makeSH"
       TestFlag : true
       HitSelectionBits  : ["EnergySelection"]
       HitBackgroundBits : ["Background","Noisy","Dead"]
@@ -59,7 +59,7 @@ CosmicReco : { @table::CosmicReco
 
    LineFinder : {
       module_type : LineFinder
-      ComboHitCollection : "makePH"
+      ComboHitCollection : "makeSH"
       TimeClusterCollection : "SimpleTimeCluster"
       NSteps: 0
       NTSteps: 0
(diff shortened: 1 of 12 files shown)


CosmicReco/fcl/prolog.fcl
```diff
@@ -78,7 +78,7 @@ CosmicReco : { @table::CosmicReco
 
    CosmicTrackFinder : { #split into Loose and tight below for efficieny studies
       module_type                   : CosmicTrackFinder
-      ComboHitCollection     : "makePH"
+      ComboHitCollection     : "makeSH"
```
brownd1978, line 81: If cosmicreco never uses panel hits I suggest overwriting makePH with RecoNullProducer as is done [here](https://github.com/Mu2e/Offline/blob/dd31c82e4e5c4edeea674b8a894c24acffccd39f/TrkHitReco/fcl/prolog.fcl#L127), then no time will be wasted on calling it.

brownd1978 (changes requested):
This looks good, there are just a few small things I'd appreciate you looking at.

CosmicReco/fcl/prolog.fcl
bonventre: Added dummy producers to prolog, will have to change in Production to implement in reco

brownd1978: You could name these instances makePH, makeSTH, then no production change is neeeded

CosmicReco/fcl/prolog.fcl
```diff
@@ -167,7 +174,7 @@ CosmicReco : { @table::CosmicReco
       module_type : CosmicTrackDiag
       StrawHitCollection : "makeSH"
       ComboHitCollection : "makeSH"
-      PanelHitCollection : "makePH"
+      PanelHitCollection : "makeSH"
```
brownd1978, line 177: Does PanelHitCollection still make sense for this module?

brownd1978 (approved):
I suggest a few possible improvements.

brownd1978:
@FNALbuild run build test

PR #1354, opened 2024-10-04
Useful Geometry Macros

AndrewEdmonds11:
Hi Everyone,

I have developed a couple of scripts that have been useful to me when inspecting the STM beamline geometry. After discussing with Rob and Stefano, I've made them more general and put them in a new directory: ```Mu2eG4/geomMacros```. The README.md provides more information. Hopefully these are useful for others

Thanks,
Andy


Mu2eG4/geomMacros/README.md
```diff
@@ -0,0 +1,44 @@
...
+```
+root -l -b -q where_is_volume.C\(\"VirtualDetector\"\)
+root -l -b -q where_is_volume.C\(\"VirtualDetector\"\,\"mu2e.gdml\")
+root -l -b -q where_is_volume.C\(\"VirtualDetector\"\,\"mu2e.gdml\",true)
+```
+
+## ```find_volume_at_point.C```
+This macro takes an x, y, z position (in Mu2e global coordinates) and prints information for the volume at that position. Some warnings:
+* this will return the deepest volume in the hierarchy, and
+* if two volumes are touching at that point, the volume returned is ambiguous
```
kutschke, line 21: Abput bullet 1. If a volume A is fully contained inside volume B and you ask for a point on the boundary of volume A, round off errors could give you volume B instead of the expected volume A.

Instead of "ambiguous" maybe "apparently random".  I think that the real answer is that it will take the first match but the sort order might involve memory addresses so it will have randomness.

Also remind people that many StepPointMCs are on volume boundaries.

These ambiguities are why we invented the PhysicalVolumeInfo system but we never completed it properly.

Mu2eG4/geomMacros/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+There are two functions defined in this macro. One can be used on the command line and takes the following arguments:
+* ```x, y, z``` is the position (required)
+* ```gdmlname``` is the name of the GDML file you want to inspect (optional, default = ```mu2e.gdml```)
+
+Examples:
+```
+root -l -b -q find_volume_at_point.C\(-3904,0,4000\)
+root -l -b -q find_volume_at_point.C\(-3904,0,4000,\"mu2e.gdml\"\)
+````
```
kutschke, line 31: There is another syntax that is a little simpler:

```
root -l -b -q 'find_volume_at_point.C(-3904,0,4000)'
root -l -b -q 'find_volume_at_point.C(-3904,0,4000,"mu2e.gdml")'
```

Mu2eG4/geomMacros/clean_volume_names.C
```diff
@@ -0,0 +1,18 @@
+//
+// The volume names in the CSV can contain memory address (i.e. 0xAAAAAAA).
+// For comparing geometries these need to be removed
+//
+
+std::string clean_volume_names(std::string name) {
+  bool all_found = false;
```
kutschke, line 7: I like this idea but there is a hook.  We have some distinct physical volumes with identical names.  This is why gdml decorates names with the memory address of the objects.  The right answer is for us to eliminate identical names and add a check when we close the geoemetry.  I will create an issue.  Until this is fixed it's possible that people could get confused if you strip off the trailing memory addresses.

Does it make sense to warn about this?

Mu2eG4/geomMacros/clean_volume_names.C
```diff
@@ -0,0 +1,18 @@
...
+  while (!all_found) {
+    auto pos_0x = name.find("0x");
+    if (pos_0x == std::string::npos) {
+      all_found = true;
+    }
+    else {
+      name = name.erase(pos_0x, 9); // always 9 chars
+    }
+  }
```
kutschke, line 16: Wouldn't it be simpler to just find the first instance of 0x and delete from there to the end of the string?  I think that "always 9" might be an accident of the machine/OS/compiler we are using now.  You could also strength the test to 0x followed by all digits?    C has an isdigit function.  I am not sure if there is something better in C++.

Mu2eG4/geomMacros/find_volume_at_point.C
```diff
@@ -0,0 +1,30 @@
+//
+// Useful functions to find a volume at a specific point
+//
```
kutschke, line 3: Add the url to the README.md and say that it has important caveats about what happens when a point is on a volume boundary.

kutschke (changes requested):
Thanks Andy.  This looks great.  A have requests for some minor changes.

Mu2eG4/geomMacros/README.md
AndrewEdmonds11: Thanks, Rob. Done in f540dd1ef

Mu2eG4/geomMacros/README.md
AndrewEdmonds11: Done in 7b7b6f338

Mu2eG4/geomMacros/clean_volume_names.C
AndrewEdmonds11: Hi Rob - do you have a specific example to hand? The cleaning is only done when outputting to csv so is not the default, so I think it will be OK. However, in 7e06db41f I've added a warning to the README about using that option

Mu2eG4/geomMacros/clean_volume_names.C
AndrewEdmonds11: Thanks, Rob. The reason I had it this way was because I was also passing the volume's path (since I wanted to do a diff of two geometries). But you raise a good point so in 4adcb16d9, I've made a separate ```clean_path_names``` function, which calls the ```clean_volume_names``` function that now just deletes to the end of the string

Mu2eG4/geomMacros/find_volume_at_point.C
AndrewEdmonds11: Done in b52104d39

Mu2eG4/geomMacros/README.md
kutschke: Thanks Andy.

Mu2eG4/geomMacros/clean_volume_names.C
kutschke: Thanks for addding the comments

There  multiple instances of physical volumes named caloPipePV* for *=1,2,3,4,5,6,7,8,9,10. PanelEBKey, StrawPanelEnvelope.

Internally G4 does not use the volume names.  It estabishes identity with memory addresses.  The names are for people.   So it's possible to have a working G4 and ambiguous human readable identities. 

The one operational place that this is bad is the following.  If you ask the volume store to find a volume by name, it gives the first match and ignores the others.  Internally G4 never does this.  Some of our code does - so we need to loop over all volumes and match names to be sure we get them all.  We only do this at startup so the cost is small.

I think it's Ok to keep your code the way it is - few people will be interested in the volumes that need to be distinguished by their addresses.

Mu2eG4/geomMacros/clean_volume_names.C
kutschke: Thanks.

kutschke (approved):

kutschke:
This PR has no coverage in the CI.  Merging it now without rerunning CI.

PR #1355, opened 2024-10-07
Add summary CosmicLivetime print

brownd1978:


brownd1978:
@FNALbuild run build test

rlcee (approved):
- exceptions should have a category, since, I think, that is the only way to refer to them in the configuration. What's written here is the message as the category(?). But I also see that we are not using this system thoughtfully in general.
- ideally the printer doesn't throw.  It can print that the product is malformed, but I wouldn't stop the exe because of that.  A malformed product is not a fatal error to the purpose of the module and throwing prevents further printing which is the purpose.  Other places we have protected the print from malformed products, like empty ptrs.
- the print modules are not intended to have counters, which, in some sense, promotes them from printers to analyzers

I'm going ahead and approving since I am usually the outlier in these opinions.

brownd1978:
I've removed the throw.
I'm confused when you say PrintModule isnt an analyzer: it inherits from EDAnalyzer, no? 
Isn't its job to print (to the screen) useful information? In that context I fail to see how a summary counter printout at endjob doesn't qualify. In this particular case CosmicLivetime of a single subrun is almost meaningless, what the end user needs is the integral, which this module now provides.

brownd1978:
@FNALbuild run build test

kutschke:
Thanks for making the changes Ray requested.  Merging now.

rlcee:
changing printMetadataSim.fcl might break metacat tools, let me take a look.  Do you regularly use this fcl?

rlcee:
metacat routine didn't break, but I'd prefer to leave this as a dedicated fcl

PR #1356, opened 2024-10-08
Fix the ArtBinaryPacketsFromDigis_module, introducing better logging and

eflumerf:
generation of empty SubEvents when no data is present, as required by the packet protocol.


kutschke:
We still use art 3.14.03,  artdaq-core v3.09.16 and artdaq-core-mu2e v2.01.03 .

There have been on-going issues integrating the prebuilt art v3.15 series as-delivered by CSAID.  

If you have newer artdaq-core etc tags that bulid against art v3.14.03 we can ask @rlcee how much work it is to prepare a view that uses these.

rlcee:
Incorporating the 3.15 build has been in trouble all summer long, even with extensive input from the experts.   We might be seeing the light at the end of the tunnel, though there are still a few things we need to do that are not well-tested.  I could try to "hurry this up" if this is urgent.  
To the other suggestion, simply replacing artdaq-core-mu2e in the 3.14 build, I suspect that is simple, we have been able to update kinkal several times.

kutschke:
@eflumerf any thoughts on the options that @rlcee presented?

eflumerf:
We do not use any advanced art features in artdaq-core-mu2e, so a simple upgrade should be possible.

rlcee:
Which version of artdaq-core-mu2e should I build?  Latest is 3.2.0.  Is artdaq-core 3.9.16 OK for it?

kutschke:
@rlcee does it look like you can build without editing package.py ?

rlcee:
This build with artdaq-core-mu2e updated to the latest, v3_02_00, was just published.  I also created p062.  Let me know if I should put in that PR too.
Rob - Eric keeps his package.py up to date, so I only need to pull it, but it is no problem if I have to update one of these.

kutschke:
@rlcee Thanks Ray.  

@eflumerf Please update your PR to advance the envset line in Offline/.muse to p062 .  Then rerun the CI.

eflumerf:
@FNALbuild run build test

kutschke:
@FNALbuild run build test

kutschke:
It found another white space error.  No idea why it missed it the first time.  Please fix.  Sorry for the serial requests.

eflumerf:
@FNALbuild run build test

PR #1357, opened 2024-10-09
added new module to CRVReco/CMakeLists.txt

ehrlich-uva:


kutschke:
Thanks Ralf.  I verified that the cmake build works and does produce this module.

PR #1358, opened 2024-10-10
fcl for metadata only

rlcee:
Keep this fcl single-purpose, for data-handling metadata.


kutschke:
@rlcee I think the spirit of this PR is that you would like to keep this fcl file limited to extracting information that will be stored as SAM ( Metacat) data.  Is that right?

@brownd1978 Where do you expect the cosmic live time to live?  In a DB?  If so which one?

brownd1978 (approved):

PR #1359, opened 2024-10-10
adding helix propagation direction info into HelixSeed/HelixRecoDir, and implementing it into AgnosticHelixFinder

matthewstortini:
Now in AgnosticHelixFinder you have a fhicl::Sequence(std::string) validHelixDirections parameter, as well as a fhicl::Atom(float) tzSlopeSigThresh parameter.

In the Offline I've set validHelixDirections : ["downstream", "upstream", "ambiguous"] ... could imagine in the trigger we have ["downstream", "ambiguous"]. 

For tz slope significance (std::abs(tzSlope/tzSlopeErr) < tzSlopeSigThresh we consider the helix direction "ambiguous".

Other helix modules can follow the same process that AgnosticHelixFinder is following by using the new tools implemented. The HelixRecoDir::PropDir _propDir data member in the HelixSeed can be set with:

    HelixTool helTool(&hseed,_tracker);
    float tzSlope = 0.0;
    float tzSlopeErr = 0.0;
    float tzSlopeChi2 = 0.0;
    helTool.dirOfProp(tzSlope, tzSlopeErr, tzSlopeChi2);
    HelixRecoDir helDir(tzSlope, tzSlopeErr, tzSlopeChi2);
    hseed._recoDir = helDir;
    hseed._propDir = helDir.computeDirection(_tzSlopeSigThresh);


gianipez:
Thank you, @matthewstortini . tagging @brownd1978

gianipez:
I can take care of CalHelixFinder once this PR gets merged

CalPatRec/src/AgnosticHelixFinder_module.cc
```diff
@@ -1388,11 +1402,41 @@ namespace mu2e {
...
+  bool AgnosticHelixFinder::validHelixDirection(HelixRecoDir::PropDir direction) {
+
+    std::string hDir;
+
+    if (direction == HelixRecoDir::downstream) hDir = "downstream";
+    if (direction == HelixRecoDir::upstream) hDir = "upstream";
+    if (direction == HelixRecoDir::ambiguous) hDir = "ambiguous";
+
+    for (size_t i=0; i<_validHelixDirections.size(); i++) {
+      if (_validHelixDirections.at(i) == hDir) return true;
```
brownd1978, line 1433: Using strings for this comparison is inefficient. Why not just test the value (as an int) is within the enum range?
What situation is this logic trying to protect against? Normally, an enum should never have  a value outside its pre-defined range, unless you have hard-cast it from an int. If you can avoid that use case you can avoid even needing this function.

RecoDataProducts/inc/HelixRecoDir.hh
```diff
@@ -18,6 +18,28 @@ namespace mu2e {
...
 
+    // Enum declaration for direction
+    enum PropDir {
+      upstream,
+      downstream,
+      ambiguous
+    };
+
+    // Method to compute direction based on slope and slopeSig
+    PropDir computeDirection(float sigThreshold) const {
```
brownd1978, line 29: This function name is misleading, the direction (slope) isn't being computed, just compared. slopeSignificance?

RecoDataProducts/inc/HelixRecoDir.hh
```diff
@@ -18,6 +18,28 @@ namespace mu2e {
     float slopeSig() const { return std::fabs(_slope/_slopeErr); }
     float chi2ndof() const { return _chi2ndof; }
 
+    // Enum declaration for direction
+    enum PropDir {
+      upstream,
```
brownd1978, line 23: Please specify at least the initial value. It would be intuitive if you set ambiguous=0, downstream=1, upstream=-1

brownd1978 (changes requested):
I request some changes to enum handling, otherwise this looks good.

matthewstortini:
I think all the comments have been addressed. Let me know if anything else is desired.

brownd1978 (approved):

brownd1978:
@FNALbuild run build test

PR #1364, opened 2024-10-22
flatten trackerstatus map

bonventre:
Decreases median time per event in TTmakeSH on NoPrimaryMix from 2 ms to <0.1ms


brownd1978 (approved):
I presume this addresses the trigger timing issue?

rlcee (approved):
Looks good

PR #1366, opened 2024-11-09
explicit dts InputTags for tracker digi maker configuration

edcallaghan:
This PR changes the configuration interface for `StrawDigisFromStrawGasSteps` module to select an explicit set of `StrawGasStepCollection`s created by specific modules, as opposed to the current implementation of selecting either a single module or accepting all. This is necessary to accommodate improvements to the mixing simulations (https://github.com/Mu2e/Production/pull/361), which add new producers of such collections and break the existing logic. A sibling PR for `Production` is https://github.com/Mu2e/Production/pull/360.


brownd1978 (approved):
looks good, thanks for cleaning this up.

edcallaghan:
@FNALbuild run build test with Mu2e/Production#360

PR #1368, opened 2024-11-11
changed CRV digis from array to vector and added NZS data

ehrlich-uva:
Changed CrvDigis and CrvDigiMCs from (fixed-length) array of samples to (variable-length) vector of samples.
Previously, all digis used 8 samples. The new FEB II firmware will use 12 samples for zero suppressed (ZS) data and 134 samples of non-zero suppressed (NZS) data. These numbers may change in the future.
This code change introduced the production and reconstruction of NZS data. The NZS data will be handled by the same modules as the ZS data, and also get the same collection names except for the use of the "NZS" instance label. The ZS collection will keep the "" as instance label. All modules have a fcl parameter to turn on/off the production of NZS data. The default setting is off.   

This change is not backward compatible. This means older CrvDigis cannot be read anymore, which will require a redigitization of older simulations - including the digi files used for the nightly validation. So all validations that rely on previously generated digis will fail.

Since we first need to come up with a plan on how to make this transition, I will make this a draft pull request.

### DIFF
--- a/CRVReco/src/CrvRecoPulsesFinder_module.cc
+++ b/CRVReco/src/CrvRecoPulsesFinder_module.cc
@@ -51,6 +51,7 @@ namespace mu2e
       using Name=fhicl::Name;
       using Comment=fhicl::Comment;
       fhicl::Atom<std::string> crvDigiModuleLabel{Name("crvDigiModuleLabel"), Comment("module label for CrvDigis")};
+      fhicl::Atom<bool> NZSdata{Name("NZSdata"), Comment("use digis with the NZS instance label")};  //false
       fhicl::Atom<float> minADCdifference{Name("minADCdifference"), Comment("minimum ADC difference above pedestal to be considered for reconstruction")};  //40
       fhicl::Atom<float> defaultBeta{Name("defaultBeta"), Comment("initialization value for fit and default value for invalid fits (regular pulses: 19.0ns, dark counts for calibration: 12.0ns)")};
       fhicl::Atom<float> minBeta{Name("minBeta"), Comment("smallest accepted beta for valid fit [ns]")};  //5.0ns
@@ -83,6 +84,7 @@ namespace mu2e
     boost::shared_ptr<mu2eCrv::MakeCrvRecoPulses> _makeCrvRecoPulses;
 
     std::string _crvDigiModuleLabel;
+    bool        _NZSdata;
     art::InputTag _eventWindowMarkerTag;
     art::InputTag _protonBunchTimeTag;
 
@@ -101,6 +103,7 @@ namespace mu2e
   CrvRecoPulsesFinder::CrvRecoPulsesFinder(const Parameters& conf) :
     art::EDProducer(conf),
     _crvDigiModuleLabel(conf().crvDigiModuleLabel()),
+    _NZSdata(conf().NZSdata()),
     _eventWindowMarkerTag(conf().eventWindowMarkerTag()),
     _protonBunchTimeTag(conf().protonBunchTimeTag()),
     _timeOffsetScale(conf().timeOffsetScale()),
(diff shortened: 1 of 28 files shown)


CRVResponse/fcl/prolog_v11.fcl
brownd1978, line 1: Do we need a special version of this file for Extracted data (test modules)?

brownd1978 (commented):
Overall this looks good. If it's possible to retain backwards-compatibility without too much effort (custom streamer) we should do that, otherwise we will branch Offline. Either way it sounds like we need to plan for a re-digitization campaign at some point to propagate these changes.
Are there any implications from these changes to Online/DAQ?

CRVResponse/fcl/prolog_v11.fcl
ehrlich-uva: We are using this epilog Offline/CRVResponse/fcl/epilog_extracted_v03.fcl for the extracted position. It overwrites some of the default values in the prolog.

ehrlich-uva:
@brownd1978 I will talk with Ray about the custom streamer.

RecoDataProducts/src/classes_def.xml
```diff
@@ -341,6 +341,19 @@
...
+        checksum="[3596926525]"
+        targetClass="mu2e::CrvDigi"
+        target="_ADCs, _NZS"
+        include="array, vector">
+<![CDATA[
+  _ADCs.resize(onfile._ADCs.size());
+  for(size_t i=0; i<onfile._ADCs.size(); ++i) _ADCs[i]=onfile._ADCs[i];
+  _NZS=false;
+]]>
+</ioread>
```
ehrlich-uva, line 355: This portion in the classes_def.xml allows us to read older digis with the updated CrvDigi class.

CRVReco/src/CrvRecoPulsesFinder_module.cc
```diff
@@ -167,9 +170,8 @@ namespace mu2e
       const CRSScintillatorBarIndex &barIndex = digi.GetScintillatorBarIndex();
       int SiPM = digi.GetSiPMNumber();
       uint16_t startTDC = digi.GetStartTDC();
-      std::vector<int16_t> ADCs;
+      std::vector<int16_t> ADCs=digi.GetADCs();
```
brownd1978, line 173: Can you access the ADCs from the digi as const&? That would avoid an unnecessary copy.

brownd1978 (approved):
It's great to see the structures evolving to better represent our expected real data. Great job getting backwards-compatibility with MDC2020 data. There's a minor point about using const&, I suggest you look for other similar instances in your code.
I would like someone more knowledgable about CRV to review the content, can you suggest someone?

CRVReco/src/CrvRecoPulsesFinder_module.cc
ehrlich-uva: Yes, GetADCs() returns the digis as const&. I'm going to change this part of the code and also look for other places.

ehrlich-uva: It turns out that this vector can't be a const, because subsequent digis may be added for longer pulses.

brownd1978:
@FNALbuild run build test

brownd1978:
clang-tidy found some errors, Ralf could you please look at those? You can ignore the warnings.

ehrlich-uva:
I think it's ready to merge now. Could you please merge it separately from other PRs to make it is easier to interpret the plots of the nightly validation jobs?

brownd1978:
Ok I’ll merge it tomorrow alone. Thanks for making the changes.

Office Phone (510) 486-7261

brownd1978:
@FNALbuild run build test

ehrlich-uva:
It looks like my latest change increased the number of clang-tidy errors instead of reducing it. I will work on it.

ehrlich-uva:
@brownd1978 How can I check clang_tidy myself before pushing it to github?

rlcee:
> @brownd1978 How can I check clang_tidy myself before pushing it to github?

try [this page](https://mu2ewiki.fnal.gov/wiki/CodeTools) and let me know if it doesn't work

ehrlich-uva:
@rlcee Thank you. I will try it.

brownd1978:
@FNALbuild run build test

PR #1369, opened 2024-11-11
Switch ExtMon filter to the commissioning config

gaponenko:
We do not plan to install the steel shot for the initial commissioning.

This PR modifies material in some existing G4 volumes and may affect random number sequences in G4 jobs where particles cross into those volumes.


brownd1978 (approved):

brownd1978:
Will the shot be there for Run 1?
It sounds like this change needs to be coordinated with the shielding radiation dose studies. Perhaps we need a 'commissioning' geometry as distinct from the running geometry.

gaponenko:
Hi Dave,

I want the no-shot geometry to be used for the shielding assessment, and making this the default is a good way to proceed.  We will be able to add more shielding (install the shot) without re-doing the studies, but not the other way around.

Andrei

PR #1374, opened 2024-11-20
Adding Berms

sridhar130:
Adding Berm layers to the G4 geometry over and surrounding the mu2e building.


sridhar130:
@FNALbuild run build test with #1374

sdifalco:
I have checked that both Offline/bin/overlapCheck.sh and Offline/Mu2eG4/fcl/surfaceCheck.fcl show no overlaps

sdifalco (approved):

Mu2eG4/geom/mu2eHall_v03.txt
```diff
@@ -661,17 +691,15 @@ vector<string> dirt.prefix.list = {
...
-#include "Offline/Mu2eG4/geom/bldg/dirtBeamlineBermWest.txt"
-#include "Offline/Mu2eG4/geom/bldg/dirtBeamlineBermSW.txt"
-
-vector<string> dirt.trap.prefix.list = {
-  "dirt.retaining.Wall.berm.east",
-  "dirt.retaining.Wall.berm.west",
-  "dirt.beamline.berm.west",
-  "dirt.beamline.berm.sw"
-};
+//#include "Offline/Mu2eG4/geom/bldg/dirtRetainingWallBermEast.txt"
```
brownd1978, line 694: please remove the commented lines, GitHub will keep track of the past for us

Mu2eG4/geom/mu2eHall_v03.txt
```diff
@@ -661,17 +691,15 @@ vector<string> dirt.prefix.list = {
...
+//#include "Offline/Mu2eG4/geom/bldg/dirtRetainingWallBermWest.txt"
+//#include "Offline/Mu2eG4/geom/bldg/dirtBeamlineBermWest.txt"
+//#include "Offline/Mu2eG4/geom/bldg/dirtBeamlineBermSW.txt"
+//  "dirt.retaining.Wall.berm.east",
+//  "dirt.retaining.Wall.berm.west",
+//  "dirt.beamline.berm.west",
+//  "dirt.beamline.berm.sw"
+vector<string> dirt.trap.prefix.list = {}; // All the dirt trap geometries are removed. The function makeTrapDirt can be found in Mu2eHallMaker.cc
```
brownd1978, line 702: Should dirt.trap be removed from the code? That seems cleaner than keeping an empty list.

brownd1978 (commented):
Overall this looks good, thanks Sridhar. I'd like to understand whether the corresponding code should be cleaned as well.

Mu2eG4/geom/mu2eHall_v03.txt
sridhar130: Hi Dave

I am removing the commented lines.
about the vector vector<string> dirt.trap.prefix.list = {};
This is called by the method "makeTrapDirt" of the class Offline/GeometryService/src/Mu2eHallMaker.cc
I have tried removing this array and building it; however, this is connected to "mu2eenv" which are in turn the dynamic padding surrounding the mu2e geometry (these were added from the very beginning by Kyle). Therefore, this breaks a lot of things.

Moreover, there is a chance that we will add trap solids (trapezoidal volumes) in the next iterations, most probably in the next PR.

In think, we should remove the comment and keep the vector declared as it is. 
Sridhar

sridhar130:
@FNALbuild run build test with #1374

Mu2eG4/geom/mu2eHall_v03.txt
brownd1978: OK sounds good.

brownd1978 (approved):

PR #1375, opened 2024-11-20
Add RMC as a muon stop pileup product option

michaelmackenzie:
Add logic for RMC to be included as a muon stop pileup product. This mostly entailed adding a flag to the RMC tool to use the RMC rate per muon to generate a random number of photons (usually 0), and add the corresponding global constant. The global constant RMC rate for aluminum is 1.4e-5 for E > 57 MeV over the closure approximation spectrum fraction for E > 57 MeV (~1/10,000 muon captures). I will make a corresponding PR in Production adding this product to the MuStopPileup configuration.

### DIFF
--- a/EventGenerator/src/RMCGenerator_tool.cc
+++ b/EventGenerator/src/RMCGenerator_tool.cc
@@ -20,6 +20,7 @@
 #include "Offline/GlobalConstantsService/inc/ParticleDataList.hh"
 #include "Offline/GlobalConstantsService/inc/PhysicsParams.hh"
 
+#include "fhiclcpp/types/Atom.h"
 #include "fhiclcpp/types/DelegatedParameter.h"
 
 // ROOT includes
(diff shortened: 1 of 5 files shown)


brownd1978 (changes requested):
I just realized that this tool is using unvalidated fhicl. We have the policy that all (legacy) codes using unvalidated fhicl must upgrade when they are modified. Validated fhicl examples can be found throughout our code base, look for ‘Config’ structs.

EventGenerator/src/RMCGenerator_tool.cc
```diff
@@ -45,14 +45,16 @@ namespace mu2e {
       _czmin(conf().spectrum.get<fhicl::ParameterSet>().get<double>("czmin", -1.)),
       _czmax(conf().spectrum.get<fhicl::ParameterSet>().get<double>("czmax",  1.)),
       _spectrum(BinnedSpectrum(conf().spectrum.get<fhicl::ParameterSet>())),
+      _useRate(conf().spectrum.get<fhicl::ParameterSet>().get<bool>("useRate",  false)),
+      _internalRate((_external) ? 0. : 1.),
```
brownd1978, line 49: I’m struggling to follow the settings needed to get the following configurations
- external (photons) only
- internal conversions only
- physical admixture of internal and external
Some combination of the ‘external’ and ‘userate’ bools seems to provide these, but I can’t quite follow.
It would be clearer if the valid configuration options were explicitly enumerated (enum) and set in the config.

I also find the term ‘external’ for producing RMC photons unintuitive, as those photons might be detected directly in the calorimeter, in which case there’s nothing ‘external’ about them.

michaelmackenzie: Only one of the `external` or `useRate` flags are necessary. I updated the configuration to use validated fhicl and in the update I set the `external` flag to only be required if `useRate` is false. I could update this to use strings or enums to check for something like "natural", "external", or "internal" if this is a preferred logic. I also don't prefer `external` as I term (I typically write real photons or virtual photons/internal conversions in talks), but it better matches the process IDs used and alternate generators.

GlobalConstantsService/data/globalConstants_01.txt
```diff
@@ -125,6 +125,7 @@ double physicsParams.Al.capture.protonRate = 0.05;
 double physicsParams.Al.capture.deuteronRate = 0.025;
 double physicsParams.Al.capture.neutronRate = 1.2;
 double physicsParams.Al.capture.photonRate = 2.0;
+double physicsParams.Al.capture.RMCRate = 9.460e-5;
```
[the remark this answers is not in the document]
brownd1978: Please add it as a comment  after this line. We need  to document the provenance of every physics number in this table, which is beyond the scope of this PR. I will open an issue once the PR is merged

EventGenerator/src/RMCGenerator_tool.cc
brownd1978: I prefer the conceptually more clear config to choose between the 3 sensible options. The names you give above are clear. If the process names are unclear we can change them, but lets do that in a separate PR.

michaelmackenzie:
I've updated the `RMCGenerator_tool` to use three modes: `external`, `internal`, or `physical`. I've updated the corresponding Production PR. This will require additional updates to the Production RMC primary fcl, but unfortunately I branched off Production before the RMC primary fcl were merged as I didn't anticipate changing how the tool is configured for primary productions. Is it best to pull in the Production updates (or rebase), or should I make a new PR for this fix after these updates are merged?

brownd1978:
Hi Michael,
thanks for changing the config options.
Merging (or rebasing) to the current head in your work branch is the
suggested workflow for this situation. The PR will follow those changes
once you commit and push.

GlobalConstantsService/data/globalConstants_01.txt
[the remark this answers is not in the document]
michaelmackenzie: I added a note here and for the internal RMC branching fraction to reference where it comes from.

brownd1978:
@FNALbuild run build test

brownd1978 (approved):

PR #1379, opened 2024-12-01
Berm add3

sridhar130:
Correct berm elevations.

Mu2e exterior wall and below-grade wall heights adjusted to berm levels.


sdifalco (approved):
I have checked that the new berms height is correct and that there are no overlaps

sridhar130:
@FNALbuild run build test with #1379

brownd1978 (approved):

PR #1383, opened 2024-12-07
Fix config and reduce severity of intersection test

brownd1978:


edcallaghan (approved):
Passes validation and looks good to me. Thanks for patching this.

PR #1385, opened 2024-12-09
 Fragments to hits/digi with new calo decoders (replaces PR #1372 ) 

giro94:


giro94:
Ok

giro94:
CaloRecoFromFragments prepared to accomodate the new CaloDecoder and be able to interpret all types of calo hits.

(replaces PR #1372 )

brownd1978 (approved):
I can't comment on the algorithm but the code looks fine.

gianipez:
@giro94 tested this module with emulated data recorded from the test stand at IERC. The additional decoding function introduced is used for debugging purposes. I asked Paolo to leave untouched the older functionality that was designed for the "default" calorimeter data-packet. 
A question for @rlcee is if at some point we can import TRACE in the Offline, so that for instance all the `std::cout` in modules used in the online reco can be converted to `TLOG` messages controlled via TRACE more consistently together with all the other modules involved in the online processing.

gianipez (approved):
Approving this PR, BUT @giro94 I think the editor you are using is suboptimal and creates a ton of fake changes that show up in the git diff. This could be a major problem when reviewing larger PRs. Please, consider switching to a different editor or fix the configuration of the one you are currently using.

giro94:
I see that the first commit shows strange diffs. I'll investigate.

Note that I will improve on this module as soon as we'll be able to test it online.

rlcee:
>  A question for @rlcee is if at some point we can import TRACE in the Offline, so that for instance all the `std::cout` in modules used in the online reco can be converted to `TLOG` messages controlled via TRACE more consistently together with all the other modules involved in the online processing.

I think the place to ask this question is in a sim/reco meeting or make a new Issue in Offline.  I think it is an interesting question and there's lots to talk about.

PR #1387, opened 2024-12-12
Explicit fcl for trigger hitreco and prescale now checks for spilltype

bonventre:
Added optional fcl MaximumTimeOffSpill which overrides MaximumTime in offspill only. To be used in the case where we want to run the OnSpill trigger exactly in both On/Off spill using the exact same sequence. Other Offspill triggers can run on a separate instance of makeSH

For trigger prescale - eventually EventMode should be a separate object to do the string compare but I'm not sure exactly what the format of eventmode is in the header so hardcoded for now.

Goes with mu2e/production#375 and mu2e/mu2e_trig_config#47

### DIFF
--- a/TrkHitReco/src/CombineStrawHits_module.cc
+++ b/TrkHitReco/src/CombineStrawHits_module.cc
@@ -7,6 +7,7 @@
 // Dave Brown confirmed this is the case
 
 #include "fhiclcpp/types/Atom.h"
+#include "fhiclcpp/types/OptionalAtom.h"
 #include "fhiclcpp/types/Sequence.h"
 #include "fhiclcpp/types/Table.h"
 #include "canvas/Persistency/Common/Ptr.h"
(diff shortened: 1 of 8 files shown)


TrkHitReco/src/CombineStrawHits_module.cc
```diff
@@ -98,6 +102,8 @@ namespace mu2e {
     _maxR2(     config().maxR()*config().maxR()),
     _minT(      config().minT()),
     _maxT(      config().maxT()),
+    _overrideminTOff(      config().minTOff(_minTOff)),
```
brownd1978, line 105: Because mintoff is optional, I  believe _minTOff will be uninitialized in case it isn't set. It shouldn't be used, but it might be cleaner to initialize it to _minT in that case.

brownd1978 (commented):
Where do you expect mint, maxt, mintoff and maxtoff to get set, since they are not in prolog?

PR #1390, opened 2024-12-16
Add a simple antiproton generator for annihilation studies

michaelmackenzie:
Add a simple antiproton generator for annihilation studies. The generator simple places antiprotons in the stopping target foils with a flat distribution in r and phi. The foil number is an exponential distribution roughly taken from SU2020 antiproton stops. The timing distribution is also exponential starting from 600 ns, which is a very rough approximation as the antiproton arrival time is a fairly complex distribution that peaks at ~500-600 ns region.

The point of this module is for simple testing and debugging of multi-track fits motivated by antiproton annihilations, but this should generally be replaced in the future by a Production workflow that produces the dataset of antiproton stops that can be resampled as we do with other primary process datasets.

### DIFF
--- a/EventGenerator/src/SimpleAntiprotonGun_module.cc
+++ b/EventGenerator/src/SimpleAntiprotonGun_module.cc
@@ -3,6 +3,7 @@
 
 
 #include <iostream>
+#include <string>
 #include <cmath>
 
 #include "CLHEP/Vector/ThreeVector.h"
@@ -25,8 +26,9 @@
 #include "Offline/GlobalConstantsService/inc/ParticleDataList.hh"
 #include "Offline/GlobalConstantsService/inc/PhysicsParams.hh"
 #include "Offline/DataProducts/inc/PDGCode.hh"
-#include "Offline/MCDataProducts/inc/GenId.hh"
-#include "Offline/MCDataProducts/inc/GenParticle.hh"
+#include "Offline/MCDataProducts/inc/ProcessCode.hh"
+#include "Offline/MCDataProducts/inc/SimParticle.hh"
+#include "Offline/MCDataProducts/inc/StageParticle.hh"
 #include "Offline/GeometryService/inc/GeomHandle.hh"
 #include "Offline/StoppingTargetGeom/inc/StoppingTarget.hh"
 #include "Offline/StoppingTargetGeom/inc/TargetFoil.hh"
@@ -43,6 +45,8 @@ namespace mu2e {
       fhicl::Atom<double> foilSurvivalRate{Name("foilSurvivalRate"), Comment("Rate of antiprotons surviving each foil"), 0.01};
       fhicl::Atom<double> timeArrivalRate{Name("timeArrivalRate"), Comment("Exponential rate for antiproton arrival"), 130.};
       fhicl::Atom<double> t0{Name("t0"), Comment("Start time for the antiproton distribution"), 600.}; //time distribution falls again below 600 ns
+      fhicl::Atom<art::InputTag> inputSimParticles{Name("inputSimParticles"), Comment("Input sim particle collection")};
+      fhicl::Atom<bool> ignoreInput{Name("ignoreInput"), Comment("Ignore the input sim particle information, using random stops"), false};
       fhicl::Atom<int> verbosity{Name("verbosity"), Comment("Verbosity level"), 0};
       fhicl::Atom<bool> makeHistograms{Name("makeHistograms"), Comment("Make histograms of the conversion kinematics"), false};
     };
@@ -51,18 +55,22 @@ namespace mu2e {
       float y;
       float z;
       float t;
+      Stop_t(): x(0.), y(0.), z(0.), t(0.) {}
+      Stop_t(float x_in, float y_in, float z_in, float t_in) : x(x_in), y(y_in), z(z_in), t(t_in) {}
     };
 
 
     PDGCode::type       pdgId_;
-    GenId               genId_;
+    ProcessCode         processCode_;
     double              mass_;
 
     double              foilSurvivalRate_;
     double              timeArrivalRate_;
     double              t0_;
+    art::ProductToken<SimParticleCollection> const simsToken_;
+    bool                ignoreInput_;
 
-    int               verbosity_;
+    int                 verbosity_;
 
     art::RandomNumberGenerator::base_engine_t& eng_;
     CLHEP::RandExponential randExp_;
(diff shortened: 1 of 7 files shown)


michaelmackenzie:
I've updated the generation to use sim particle inputs rather than creating gen particle collections. The sim particle information is used by default, but an option allows to ignore the sim particle positions and instead use the approximate stopping distribution generation. This option can likely be removed in the future when stopped antiproton datasets are available, but it may be useful for testing until then.

EventGenerator/src/SimpleAntiprotonGun_module.cc
```diff
@@ -0,0 +1,215 @@
...
+
+namespace mu2e {
+
+  //================================================================
+  class SimpleAntiprotonGun : public art::EDProducer {
+    struct Config {
+      using Name=fhicl::Name;
+      using Comment=fhicl::Comment;
+      fhicl::Atom<double> timeArrivalRate{Name("timeArrivalRate"), Comment("Exponential rate for antiproton arrival"), 130.};
+      fhicl::Atom<double> t0{Name("t0"), Comment("Start time for the antiproton distribution"), 600.}; //time distribution falls again below 600 ns
```
brownd1978, line 45: It looks like this gun fires AP into the stopping target 'point blank'. I was rather expecting it would fire from the production target, to allow the beamline to naturally provide timing transport and momentum selection. In that case, the initial timing distribution could be the primary proton timing, and the dependence on the initially assumed momentum distribution would be muted.

brownd1978 (commented):
This is getting closer. I'd also like to have Stefano take a look.
I think the main issue now is the philosophy of the AP gun, lets continue the discussion in GitHub.

EventGenerator/src/SimpleAntiprotonGun_module.cc
sdifalco: in the antiproton studies for SU2020 we have found that Geant4 overestimates the probability to emit in the backward direction the antiprotons produced by proton interactions in the stopping target. A patch using  a correction obtained comparing Geant4 and MCNP has been introduced using event weights. The generation at the production target is also important to produced the delayed RPC produced by antiproton annihilation in the TS, these give a non negligible contribution to the antiproton related background. I think the goal of Michael's generator is limited to the study of antiproton annihilations in the production target. If it's just a tool to test the multiple track reconstruction the distribution of the anitproton annihilation position and time should not be so relevant. Perhaps a flat distribution could be more easily reweighted with the wanted distribution in case we try for example to change the production target

michaelmackenzie: The idea here was to make a simple sample of annihilation events in the stopping target for tracking studies. I believe the survival of the antiprotons are very low along the beamline, so resampling stages were originally setup along the path (I believe before the TS, before the TS window, etc.). This workflow would need to be implemented and debugged by an expert in order to produce a useful antiproton dataset at the ST. The goal of this generator was to approximate the final result at the ST. I can update the arrival time to be flat, so studies can be performed easily be reweighting the time and momentum distributions. Overall, the goal here was to have a quick workflow for annihilations until someone more expert implements a proper workflow and produces a large dataset of stopping target annihilations.

brownd1978:
In the AP production job we don’t care about AP annihilation daughters.
Can we then improve the efficiency by turning off all daughter generation
in that job?
As Michael says, the AP generator should produce stopped AP that are
similar to ‘real’ stopped AP, just for tracking studies. It’s not intended
to calculate rates (that’s a separate question). The AP daughter generator
though should be physically accurate.

Office Phone (510) 486-7261

michaelmackenzie:
> In the AP production job we don’t care about AP annihilation daughters. Can we then improve the efficiency by turning off all daughter generation in that job? As Michael says, the AP generator should produce stopped AP that are similar to ‘real’ stopped AP, just for tracking studies. It’s not intended to calculate rates (that’s a separate question). The AP daughter generator though should be physically accurate.

I agree that killing the antiproton daughters in the simple antiproton generator make sense as we will simulate/resample these annihilations in the resampling stage . Is there a simple way to do this? For reference, it takes ~0.2 s / event currently with 98% stopping efficiency, which means to produce a dataset the size of the SU2020 antiproton stop dataset (1k events) would take about 4 minutes and we can produce 20k stops in 1.5 hours. Given we then resample these stops and there's a lot of randomness in the final state (and we shouldn't be using this generator for detailed physics studies as it's just a placeholder for the real samples), I think if turning off the antiproton daughters isn't trivial we should leave it as is rather than optimize a workflow we hope to quickly retire. Please let me know what you think when you have a chance!

michaelmackenzie:
I updated to a flat time spectrum as recommended by Stefano, as then reweighting the time, momentum, or radial distributions become more consistent and straightforward if one wanted to match to a real antiproton dataset.

brownd1978 (approved):
This looks good. Could you please verify in your test that [FIndMCPrimary](https://github.com/Mu2e/Offline/blob/main/CommonMC/src/FindMCPrimary_module.cc) is compatible with this generator? I think it is, if you specify genID. You should try adding it to a sequence and verify that [print](https://github.com/Mu2e/Offline/blob/main/Print/fcl/printPrimaryParticle.fcl)  produces sensible output (it should find all the AP daughters)

michaelmackenzie:
Hi Dave, I generated events with the simple antiproton gun and saved antiproton stops, and then used this as input to the antiproton stop resampling generator with the same fcl paths as used in the normal primary stopped particle paths. This included the `PrimaryFilter` and `FindMCPrimary` filters with the `PrimaryProcess` set to `mu2eAntiproton`, which accepted 33% of events (100% after `PrimaryFilter`). Does this cover what you had in mind, or would these filters both accept the event even if the MC primary wasn't found?

brownd1978:
@FNALbuild run build test

sdifalco (approved):
ok for me

brownd1978:
> accepted 33% of events (100% after PrimaryFilter).
I think this is reasonable for a 'point-blank' source.
I think it would be useful to allow the location of the isotropic AP source to be configurable. You could define a few preset locations (front of ST, middle of TS, front of PT, ...) either by enum in the code or (more flexibly) by having the XYZ as a sequence, and defining the specific points in prolog.

michaelmackenzie:
Hi Dave,

Thank you for response and review. I can edit the gun to optionally take in (x, y, z) coordinates in place of the stopping target location, but I think the stopping target is a good default. If we start making updates like this, we'll likely want to add more flexibility as well (what plane are we generating in? Is this user provided or assumed to be the transverse plane to the beamline for any given z position? Which direction should they move in? What size/shape plane should they be generated in? Should a unit sphere generation direction with optional cos(z) restrictions be a generating option? etc.). The only current use case I can think of for this generator is for trigger/reconstruction studies, where it should be replaced by the real antiproton sample as soon as this is available.

If we want to move forward with more generalizing this generator, rather than it being a proof of principal for multi-track reconstruction, can we decouple this from the APR trigger sequence updates? Please let me know when you have a chance!

Best,
Michael

brownd1978:
All valid points Michael. I still worry the point-blank source is too simplistic for trigger studies, but if the AP group think's its adequate and will accept the results for algorithm tuning and selection fI will defer to your judgement.
I will merge this and the other AP-related PRs now.
The next step is to integrate this generator into Production and produce an official sample. Please work with Yuri to figure out the best way to do that.

michaelmackenzie:
Thank you for merging this in and the helpful review! I'll check in with the antiproton people about this, but I believe their plan is to generate a real dataset of antiprotons from the production target and only use the resampling generator component of this PR. It's possible they would use the simple antiproton gun for some unfolding/efficiency/systematics studies, but I'm not positive. I'll let you know what they say and open an issue/new PR for feature updates if they have any requests. Please let me know if there's anything else needed for the APR/DeltaFinder PRs as well!

PR #1391, opened 2024-12-30
start runTool

rlcee:
a command line interface to print information from the run configuration database

### DIFF
--- a/DbService/CMakeLists.txt
+++ b/DbService/CMakeLists.txt
@@ -8,6 +8,8 @@ cet_make_library(
       src/DbTool.cc
       src/DbValTool.cc
       src/EpicsTool.cc
+      src/RunInfo.cc
+      src/RunTool.cc
     LIBRARIES PUBLIC
       PostgreSQL::PostgreSQL
       CURL::libcurl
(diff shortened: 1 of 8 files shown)


brownd1978 (approved):
Can you provide a brief statement of what RunTool is intended to do? A presentation at the next tools/infra meeting might be appropriate.

rlcee:
> Can you provide a brief statement of what RunTool is intended to do? A presentation at the next tools/infra meeting might be appropriate.

As otsdaq creates new runs online, the run configuration is stored in the run_info database, which is streamed to offline.  This code is a user-friendly way to access that database - a "run discovery" tool.  It could be used to answer questions like "what was the run number of the last calo laser run?", "what subdetectors were part of run x", or "what is a recent large cosmic run?".  It might be useful in some automated calibration scripts. Without this tool, a user would have to search through the log book or run SQL in the db itself. It becomes part of the family with dbTool, epicsTool, and dqmTool.  I expect there will also be a grlTool (good runs) and a subrunToool, when we have those databases, and maybe a lumTool.

>A presentation at the next tools/infra meeting

already in progress

brownd1978:
I'm somewhat concerned we might be fragmenting our db interface across several tools. I'm also not clear how this tool is intended to work with the data file catalog (currently sam), which we currently use to search for runs by description, payload type, etc. I suggest we discuss this at an upcoming meeting.

PR #1394, opened 2025-01-11
temp PR to test CI

rlcee:


rlcee:
There is a place in the code to say "if this test fails, don't report it to the PR as a failure, but still show it as a failure in the report" and the new trigger check is installed this way.  It is easy to reset it so that the error will cause a PR check failure but a warning won't.

PR #1395, opened 2025-01-13
Loosen test of direction consistency

brownd1978:
Addresses problem reported by M. Mackenzie and Y. Oksuzian.


michaelmackenzie (approved):
This fixes the RECO error I ran into and I don't see any issues with this update, thank you for the quick fix!

PR #1325, opened 2024-08-16
Ext mon detector restructure

fine1-sam:
Work in progress

There are overlapping geometries in the ext mon detector modules when the stackRotation flag is set to "true."

PR #1396, opened 2025-01-13
Extinction updates

caohc:
This pull request includes request #1325 , the overlapping geometries are fixed.


rlcee:
New trigger test failed (it looks like the timing test actually improved too much!?)  The test is set to never block merging, which looks like it worked as expected (all checks passed).

michaelmackenzie:
The printout is accidentally reversed in the trigger job timing test, this job ran slower than the reference. It's not obvious this is due to the updates though and not due to the node it ran on or the node's load. Where are these tests run?

Mu2eG4/src/constructExtMonFNALBuilding.cc
```diff
@@ -155,7 +155,7 @@ namespace mu2e {
     TubsParams shotLinerOuterParams(collimator.shotLinerOuterRadius()
                                     -collimator.shotLinerOuterThickness(),
                                     collimator.shotLinerOuterRadius(),
-                                    0.5*collimator.length()
+                                    collimator.length()/2
```
gaponenko, line 158: Hi Haichuan,

Please avoid unnecessary changes to the existing code.  The old code was not wrong, and it is debatable which version is "better".   A C programmer sees a single operation (multiplication of two floating point numbers) in 0.5*length, but two operations (conversion of an integer to a double and then division) in length/2.    Of course, the final optimized code is the same in both cases.   Similar for 90*degrees vs M_PI/2.

Sometimes we do implement style changes that don't change the result (notably, whitespace fixes), but this is best done via a separate PR.   Otherwise, it just adds noise that makes substantive changes harder to review.

gaponenko: To make it clear, I am not asking you to revert the changes you have already added to this PR.  The message above is a request for the future.

caohc: Hi Professor Gaponenko,
    Sorry for the inconvenience. The only reason I did that is 0.5 or 90 are magic numbers. I wanted to remove the warnings. I will not do this later.

gaponenko:
Haichuan, thank you for your reply.  I did not realize that the floating point numbers triggered clang warnings.   It is nice you cleaned them up, thank you.

Andrei

gaponenko (changes requested):
Hi Haichuan,

Please add the new extMonFNAL.chipGapX and extMonFNAL.chipOffsetY parameters to the ExtMonFNALModule geometry class.
Also, please take a look at the extMonFNAL.stackRotation = false case, I still see overlaps there.

Andrei

gaponenko:
@FNALbuild run build tests

gaponenko:
@FNALbuild run code checks

ExtinctionMonitorFNAL/Geometry/inc/ExtMonFNALModule.hh
```diff
@@ -31,8 +33,8 @@ namespace mu2e {
     unsigned int nxChips() const;
     unsigned int nyChips() const;
 
-    ExtMonFNALModule(const ExtMonFNALPixelChip& chip, const std::vector<double>& hs)
-      : chip_(chip), sensorHalfSize_(hs)
+    ExtMonFNALModule(const ExtMonFNALPixelChip& chip, const std::vector<double>& hs, const double& gapX, const double& offsetY)
+      : chip_(chip), sensorHalfSize_(hs), chipGapX_(gapX), chipOffsetY_(offsetY)
     {}
     // Required by genreflex persistency
     ExtMonFNALModule() {}
```
gaponenko, line 40: Please initialize the doubles:

ExtMonFNALModule() : chipGapX_{}, chipOffsetY_{} {}

Mu2eG4/src/constructExtMonFNALDetector.cc
```diff
@@ -0,0 +1,814 @@
...
+      CLHEP::HepRotation* mRotInPlane = reg.add(new G4RotationMatrix);
+
+      mRotInPlane->rotateZ(stack.planes()[iplane].module_rotation()[imodule]);
+
+      if( stack.planes()[iplane].module_zoffset()[imodule] < 0.0 ) {
+        mRotInPlane->rotateY(M_PI);
+      }
+
+      CLHEP::Hep3Vector soffset;
+      CLHEP::HepRotation* mRot;
```
gaponenko, line 411: Please initialize the pointer.   Perhaps just say "mRot = nullptr", then the rest of the code can be left as is.   The alternative would be to use "mRot = stackRotation ? A : B" construct.

Mu2eG4/src/constructExtMonFNALDetector.cc
```diff
@@ -0,0 +1,814 @@
...
+                                   forceAuxEdgeVisible,
+                                   placePV,
+                                   doSurfaceCheck
+                                   );
+
+
+
+      CLHEP::Hep3Vector coffset0;
+      CLHEP::Hep3Vector coffset1;
+      CLHEP::HepRotation* pRot;
```
gaponenko, line 460: Please initialize the pRot pointer, similar to mRot.

gaponenko (changes requested):
Hi Haichuan,

Thank you for the updates.  I spotted a few uninitialized variables.  They should not affect the code behavior, but we would like to avoid them as a matter of style.  Can you please fix them per the attached comments?

Andrei

gaponenko (approved):
Haichuan, thank you for the updates.   @brownd1978, please merge.   This PR does not directly affect validation, but it is a geometry change, so expect statistical differences in G4 jobs.

brownd1978:
@FNALbuild run build test

PR #1398, opened 2025-01-21
Changes to implement the delta t0 cut between helices

iffatzarif:
To implement a multi-helix trigger, we added a cut on delta t0 between pairs of helices in an event. 

Tagging: @gianipez @matthewstortini @michaelmackenzie

### DIFF
--- a/TrkFilters/src/HelixFilter_module.cc
+++ b/TrkFilters/src/HelixFilter_module.cc
@@ -226,8 +226,8 @@ namespace mu2e
     _maxDt0            (config().maxDt0())
     {
       produces<TriggerInfo>();
+      if(_minNHelices < 2 && _maxDt0 >= 0.) throw cet::exception("BADCONFIG") << "Requested a timing difference cut of " << _maxDt0 << " ns between helices but only " << _minNHelices << " helices required";
     }
-  if(_minNHelices < 2 && _maxDt0 >= 0.) throw cet::exception("BADCONFIG") << "Requested a timing difference cut of " << _maxDt0 << " ns between helices but only " << _minNHelices << " helices required";
 
   void HelixFilter::beginJob() {
     if ( (!_posHelCuts._configured) && (!_negHelCuts._configured)) {
@@ -326,8 +326,8 @@ namespace mu2e
   evt.put(std::move(triginfo));
     
     
-  if(!_noFilter) {return (nGoodHelices >= _minNHelices) && dt_range};
-    else {return true}; //filtering is turned off
+  if(!_noFilter) {return (nGoodHelices >= _minNHelices) && dt_range;}
+    else {return true;} //filtering is turned off
 
     
   
--- a/TrkFilters/src/HelixFilter_module.cc
+++ b/TrkFilters/src/HelixFilter_module.cc
@@ -325,10 +325,10 @@ namespace mu2e
     evt.put(std::move(triginfo));
     if (!_noFilter){
       if( _minNHelices > 1){
-        return ((nGoodHelices >= _minNHelices) && (dt_range = true));
+        return ((nGoodHelices >= _minNHelices) && (dt_range == true));
       }
       else {
-        return (nGoodHelices > 0);
+        return (nGoodHelices >= _minNHelices);
       }
 
     }else {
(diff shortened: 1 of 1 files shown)


TrkFilters/src/HelixFilter_module.cc
```diff
@@ -292,9 +296,35 @@ namespace mu2e
...
+    bool dt_range = false;
+    if (_maxDt0 < 0){
+      dt_range = true;
+    }
+    else{
+      for(auto ihs = hscol->begin();ihs != hscol->end(); ++ihs) {
+        auto const&  hel0 = ihs;
+        float  hel0_t0 =  hel0->t0()._t0;
+
+        for(auto ihs = hscol->begin();ihs != hscol->end(); ++ihs) {
```
michaelmackenzie, line 309: This loops over the entire collection again (reusing the `ihs` variable name, perhaps use `jhs` or something?), can this find the same helix again which would have `dt == 0` (i.e. `hel0 == hel1`). Perhaps this loop could start at the next iteration, avoiding this collision as well as rechecking comparisons you've already done, something like:
`for(auto jhs = std::next(ihs); jhs != hscol->end(); ++jhs) {`

TrkFilters/src/HelixFilter_module.cc
```diff
@@ -292,9 +296,35 @@ namespace mu2e
...
+          auto const&  hel1 =  ihs;
+          float hel1_t0 = hel1->t0()._t0;
+          float dt      = hel1_t0 - hel0_t0;
+          if (dt < _maxDt0 && dt > -_maxDt0) {dt_range = true; }
```
michaelmackenzie, line 313: You could simpy put (but this is just preference, not necessary):
`dt_range |= std::fabs(dt) < _maxDt0`
You could also break out of both `for` loops if `dt_range` is ever true to avoid unnecessary checks.

TrkFilters/src/HelixFilter_module.cc
```diff
@@ -292,9 +296,35 @@ namespace mu2e
...
+        }
+      }
+    }
+
+
     evt.put(std::move(triginfo));
     if (!_noFilter){
-      return (nGoodHelices >= _minNHelices);
+      if( _minNHelices > 1){
+        return ((nGoodHelices >= _minNHelices) && (dt_range = true));
```
michaelmackenzie, line 322: I believe this line assigns `dt_range = true`, perhaps:
`return (nGoodHelices >= _minNHelices) && dt_range`

TrkFilters/src/HelixFilter_module.cc
```diff
@@ -292,9 +296,35 @@ namespace mu2e
...
+      }
+      else {
+        return (nGoodHelices > 0);
```
michaelmackenzie, line 325: If `_maxDt0` is non-zero and `nGoodHelices` is less than two, perhaps this should be an error at module construction. I'm not sure what the use case is that we accept single helix events but reject two helix events without a time match. Also, this should be `nGoodHelices >= _minNHelices` still, as `_minNHelices` could be 0. As `dt_range` defaults to `true` when no range cut is applied, this could be reduced to a single check of `(nGoodHelices >= _minNHelices) && dt_range`.

michaelmackenzie (commented):
This looks great! I've added a couple of comments and questions to better understand the filter as well as make some suggestions.

TrkFilters/src/HelixFilter_module.cc
iffatzarif: I think the mistake here is that it is setting dt_range to true instead of checking. I corrected it to dt_range == true. Please let me know if you think any more errors remain.

michaelmackenzie: I think the only other comment I would have is to add to the constructor (perhaps after `produces<TriggerInfo>()`) a check that the configuration makes sense, that if a timing check is requested, the minimum number of helices must be greater than one. Here's an example of what I mean:
```
if(_minNHelices < 2 && _maxDt0 >= 0.) throw cet::exception("BADCONFIG") << "Requested a timing difference cut of " << _maxDt0 << " ns between helices but only " << _minNHelices << " helices required";
```

With this, your return check can be simplified since you set `dt_range` to `true` in the case `_maxDt0 < 0`:
```
return (_noFilter || ((nGoodHelices >= _minNHelices) && dt_range);
```

michaelmackenzie:
@FNALbuild run build test

michaelmackenzie (approved):
Everything looks good to me, thanks!

PR #1399, opened 2025-01-22
Fix to TrackQuality module bug related to missing tracker entrance segment

AndrewEdmonds11:
@michaelmackenzie found that when a track is missing the segment at the tracker entrance, the TrackQuality module uses some input features from the previous track.

This PR fixes the bug by setting the trkqual value to 0 for tracks with missing segments. It also includes Michael's debug printout, which was very useful.

I'm open to suggestions as to whether this is the best way to handle this situation.

### DIFF
--- a/TrkDiag/src/TrackQuality_module.cc
+++ b/TrkDiag/src/TrackQuality_module.cc
@@ -136,16 +136,25 @@ namespace mu2e
       features[6] = (double)nmatactive / nactive;
 
       // Now get the features that are for the entrance of the trackre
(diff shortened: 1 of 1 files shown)


brownd1978 (approved):
Hi Andy, I think this is fine for now. I found and fixed the underlying problem.
In general I think we should migrate our analysis tools to work with tracker_mid: that sampling has the highest accuracy and precision. It's also more robust to sample a loop fit at TT_Mid.

PR #1400, opened 2025-01-23
macro to visualize volumes and their 2D projections

sdifalco:
These are a set of macros useful to visualize volumes in a given area or close to a target volume. They take the volume information from a user defined gdml file. In case of extruded solids the 2D projections are also available.


brownd1978 (approved):
Thanks for adding these!

PR #1402, opened 2025-01-24
straw digis from artdaq fragments

pavel1murat:
- add module reconstructing straw digis from artdaq fragments
- known missing pieces: 
- correct assignment of strawID's : 
  - need to have that uploaded to FPGAs 
  - need to decide how to assign the straw IDs to the existing panels
- lower bits of the TDC words - to be fixed in firmware 
- bit ordering in ADC samples - to be fixed in the firmware


edcallaghan:
@pavel1murat Thanks for pushing this, Pasha. I can confirm that this runs and produces digis. I will take a look at the code a bit later, but one question to clarify:

>     * correct assignment of strawID's :
>       
>       * need to have that uploaded to FPGAs
>       * need to decide how to assign the straw IDs to the existing panels

what does this mean for data we have already taken? I thought there was an effectively-fixed straw <-> id mapping, but this implies (to me) that we don't currently know which straws are being hit. Am I misunderstanding?

rlcee:
> >   * need to decide how to assign the straw IDs to the existing panels

I want to add there is only one map (~ROC number to ~panel number) in offline.  Data with only small parts of the detector can be handled in different way for things like calibrations.  Please keep me (and I'd also suggest Richie) in the loop for any channel numbering discussions, if it might effect offline.  Thanks

brownd1978:
Is there anyone besides Ed who should review this?

DAQ/src/StrawRecoFromArtdaqFragments_module.cc
```diff
@@ -0,0 +1,297 @@
...
+                wf[++idx] = adc_packet->ADC4();
+                wf[++idx] = adc_packet->ADC5;
+                wf[++idx] = adc_packet->ADC6;
+                wf[++idx] = adc_packet->ADC7();
+                wf[++idx] = adc_packet->ADC8;
+                wf[++idx] = adc_packet->ADC9;
+                wf[++idx] = adc_packet->ADC10();
+                wf[++idx] = adc_packet->ADC11;
+                npackets++;
+                adc_packet++;
```
edcallaghan, line 248: The only general concern I have is that I don't see any integrity checks. It is an efficient deserialization assuming the fragments are error-free, but it may produce uncaught exceptions, or even terminate ungracefully, if there is an error in the payload. The canonical example being the missing-packet problem we've encountered, which L248 for example assumes cannot happen. Without considering that, then this will eventually read an out-of-bounds address, which will either cause process termination or silent data corruption.

DAQ/src/StrawRecoFromArtdaqFragments_module.cc
```diff
@@ -0,0 +1,297 @@
...
+//-----------------------------------------------------------------------------
+// this is a tracker DTC fragment, loop over the ROCs
+//-----------------------------------------------------------------------------
+        ushort*  buf          = (ushort*) fdata;
+        int      nbytes       = buf[0]; // frag.dataSizeBytes() includes extra 0x20
+        uint8_t* roc_data     = fdata+sizeof(*seh);
+        uint8_t* last_address = fdata+nbytes;     //
+
+        while (roc_data < last_address) {
+          RocDataHeaderPacket_t* rdh = (RocDataHeaderPacket_t*) roc_data;
```
edcallaghan, line 188: Related to my other comment: is there a magic byte, or something, in this header that we can use a flag that there was an error at some point and our we've gone out-of-sequence?

edcallaghan (commented):
A couple of comments trying to think ahead to dealing with a nonzero error rate in the payload. I'm inclined to say that the `ProtonBunchTime` and `IntensityInfoTrackerHits` aren't necessary in this module, but if they fit into a larger pattern then it's ok.

edcallaghan:
@rlcee Following a discussion with Richie, we think the channel-numbering business amounts to applying the mapping you reference to a per-panel DCS write during configuration, so I don't foresee needing any to tweak anything in Offline.

DAQ/src/StrawRecoFromArtdaqFragments_module.cc
rlcee: DJ will bring this up at the pass1 workshop.

DAQ/src/StrawRecoFromArtdaqFragments_module.cc
```diff
@@ -0,0 +1,297 @@
...
+
+#include "TRACE/tracemf.h"
+#define TRACE_NAME "StrawRecoFromArtdaqFragments"
+
+namespace art {
+  class StrawRecoFromArtdaqFragments;
+}
+// ======================================================================
+
+class art::StrawRecoFromArtdaqFragments : public EDProducer {
```
rlcee, line 40: My preference would be to call this "StrawDigiFrom" instead of "StrawRecoFrom" since there is no reco in the way I'd use the word.

DAQ/src/StrawRecoFromArtdaqFragments_module.cc
```diff
@@ -0,0 +1,297 @@
...
+
+public:
+
+  struct RocDataHeaderPacket_t {            // 8 16-byte words in total
```
rlcee, line 44: isn't this the sort of thing that goes in artdaq_core_mu2e?

rlcee (commented):
How does this module relate to the existing StrawRecoFromFragments?

DAQ/src/StrawRecoFromArtdaqFragments_module.cc
```diff
@@ -0,0 +1,297 @@
...
+//-----------------------------------------------------------------------------
+            mu2e::TrackerDataDecoder::TrackerDataPacket* hit ;
+            int offset          = (ihit*np_per_hit+1)*packet_size;   // in bytes
+            hit = (mu2e::TrackerDataDecoder::TrackerDataPacket*) (roc_data+offset);
+//-----------------------------------------------------------------------------
+// unpack hit and fill the StrawDigiCollection
+// HV side has an extra bit in teh straw ID
+//-----------------------------------------------------------------------------
+            uint16_t straw_id = hit->StrawIndex;
+            if (straw_id >= 0x80) straw_id = straw_id - 0x80;
```
bonventre, line 209: This comes from just how the panel id registers are set on powerup right now - if those are written to correctly by the DAQ before taking data this fix is not needed

bonventre:
One thing that seems to be missing here is the ability to deal with different packet formats - the artdaq_core_mu2e decoder has code to automatically upgrade older formats, how would that work here?

DAQ/src/StrawRecoFromArtdaqFragments_module.cc
```diff
@@ -0,0 +1,297 @@
...
+#include <iostream>
+
+#include <string>
+
+#include <memory>
+
+#define TRACEMF_USE_VERBATIM 1
+
+#include "TRACE/tracemf.h"
+#define TRACE_NAME "StrawRecoFromArtdaqFragments"
```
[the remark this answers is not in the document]
rlcee: Pasha agree to comment out trace for now, there is no specific need

DAQ/fcl/prolog.fcl
```diff
@@ -0,0 +1,11 @@
+# -*- mode:tcl -*-
+
+DAQ : {
+    producers : {
+        StrawRecoFromArtdaqFragments: { module_type: StrawRecoFromArtdaqFragments
+            nADCPackets   : 1
```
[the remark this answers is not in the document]
rlcee: This setting will be correct for all data given to pass1 and only change for expert-only runs.  So setting it here is OK.

DAQ/src/StrawRecoFromArtdaqFragments_module.cc
edcallaghan: Following an offline discussion, I had misunderstood the nature of the errors we currently expect to encounter. We may go out-of-sequence when parsing the packed data, but there is not a mechanism by which we foresee encountering an out-of-bounds read like I suggested. Pasha will implement a simple check for having gone out-of-sequence, and utilize the `StrawDigiFlag` data product to mark digis in such an event as having been part of problematic event.

DAQ/src/StrawRecoFromArtdaqFragments_module.cc
edcallaghan: Following offline discussion, a consistency check will be implemented at the hit-block level.

DAQ/src/StrawRecoFromArtdaqFragments_module.cc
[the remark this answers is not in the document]
rlcee: I see others are embracing trace even more, so it look like trace will be in offline for things that run online.

edcallaghan:
@FNALbuild run build test

edcallaghan (approved):

DAQ/fcl/prolog.fcl
gianipez, line 1: This module instance belongs to [mu2e-trig-config/core/](https://github.com/Mu2e/mu2e-trig-config/blob/main/core/producers/trigDAQProducers.fcl) as it was agreed upon a long time ago. could you move it there?

gianipez (changes requested):
Everything looks fine but I'd like to keep no module instance for online-reco modules outside mu2e-trig-config.

rlcee (approved):

DAQ/fcl/prolog.fcl
brownd1978: This request cannot be resolved within the scope of this PR, as it refers to a different repo. There are 2 ways forward:
- merge this PR and move the content to mu2e-trig-config at a later date
- close this PR and re-open a new PR in mu2e-trig-config
The latter is the cleanest but will result in some delays. I leave it to the author to decide.

DAQ/CMakeLists.txt
```diff
@@ -129,6 +129,14 @@ cet_build_plugin(STMWaveformDigisFromFragments art::module
       BTrk_difAlgebra
 )
 
+cet_build_plugin(StrawRecoFromArtdaqFragments art::module
```
gianipez, line 132: @pavel1murat I think you made a typo here. It should be `StrawDigisFromArtdaqFragments`

DAQ/fcl/prolog.fcl
pavel1murat: Giani, when saying 'This module instance', are you referring to the default FHICL configuration of the module or the module itself ?

DAQ/CMakeLists.txt
pavel1murat: right !- thanks for catching that

DAQ/fcl/prolog.fcl
gianipez: Just the FHICL instance of the module.

DAQ/CMakeLists.txt
pavel1murat: just fixed - thought that due to the triviality of the matter it has been already done

DAQ/fcl/prolog.fcl
gianipez: This PR is not the best place for re-opening a conversation on this subject. I'm not against re-discussing this but not here please.

DAQ/src/StrawRecoFromArtdaqFragments_module.cc
pavel1murat: fixed

DAQ/fcl/prolog.fcl
pavel1murat: - I agree - it is not the right place, but this is where issues like this pop up and you start thinking of them :) 
- lets move this one to where you proposed and schedule a discussion about handling the FCL defaults

gianipez (approved):

DAQ/src/StrawDigisFromArtdaqFragments_module.cc
```diff
@@ -0,0 +1,437 @@
...
+    diagLevel_    (config().diagLevel    ()),
+    debugLevel_   (config().debugLevel   ()),
+    saveWaveforms_(config().saveWaveforms())
+{
+  produces<mu2e::StrawDigiCollection>();
+  if (saveWaveforms_) produces<mu2e::StrawDigiADCWaveformCollection>();
+
+  produces<mu2e::IntensityInfoTrackerHits>();
+  // FIXME!
+  //  produces<mu2e::ProtonBunchTime>();
```
brownd1978, line 98: Is this comment still relevant? ProtonBunchTime is a high-level object coming from analyzing (potentially) multiple data products and applying physics corrections, why would it be produced by a low-level module such as this?

DAQ/src/StrawDigisFromArtdaqFragments_module.cc
```diff
@@ -0,0 +1,437 @@
...
+  print_("-- START",1);
+
+   // Collection of StrawDigis for the event
+  std::unique_ptr<mu2e::StrawDigiCollection> straw_digis(new mu2e::StrawDigiCollection);
+  std::unique_ptr<mu2e::StrawDigiADCWaveformCollection> straw_digi_adcs(new mu2e::StrawDigiADCWaveformCollection);
+
+  // IntensityInfoTrackerHits
+  std::unique_ptr<mu2e::IntensityInfoTrackerHits> intInfo(new mu2e::IntensityInfoTrackerHits);
+
+  // FIXME! this is temporary
```
brownd1978, line 208: Is there a reason to keep this commented code?

DAQ/src/StrawDigisFromArtdaqFragments_module.cc
```diff
@@ -0,0 +1,437 @@
...
+                                        // unknown DTC, return an error
+    print_(std::format("ERROR: unknown DtcID:{}",DtcID));
+    return -1;
+  }
+//-----------------------------------------------------------------------------
+// for the moment, assign unique panel ID's just to get through,
+//-----------------------------------------------------------------------------
+  panel_id = idtc*6+LinkID;
+
+//   if (MnID == panel_map_[idtc][LinkID]) return MnID;
```
brownd1978, line 160: Are the commented stanzas obsolete, or under development?

DAQ/src/StrawDigisFromArtdaqFragments_module.cc
```diff
@@ -0,0 +1,437 @@
...
+
+                wf[0] = hit_data->ADC00;
+                wf[1] = hit_data->ADC01();
+                wf[2] = hit_data->ADC02;
+
+                auto npackets   = 0;
+                auto idx        = 2;
+                auto adc_packet = (mu2e::TrackerDataDecoder::TrackerADCPacket*)((char*) hit_data + 16); // the packet size is 16 bytes
+                while (npackets < nADCPackets_) {
+                  wf[++idx] = adc_packet->ADC0;
```
brownd1978, line 363: I am confused that some adc_packet members are data members and some are functions. This is outside scope of this PR, but shouldn't we consolidate to a single access pattern?

DAQ/src/StrawDigisFromArtdaqFragments_module.cc
```diff
@@ -0,0 +1,437 @@
...
+
+//-----------------------------------------------------------------------------
+// to be written - needs DB access
+// for now - all good
+// return panel ID part of the geographical (offline) channel ID - first channel ID of the panel
+//-----------------------------------------------------------------------------
+int art::StrawDigisFromArtdaqFragments::panelID(uint16_t MnID, int DtcID, int LinkID) {
+  int panel_id(-1), idtc(-1);
+
+  if      (DtcID == 18) idtc = 0;   // daq09
```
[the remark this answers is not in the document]
edcallaghan: Richie, Pasha, and I just talked about a few minutes ago. I will open a PR with a fcl-based mapping; once we are satisfied, based on experience with data, with what maps to what, we will work with Ray to transition to a database table and compiled interface.

DAQ/src/StrawDigisFromArtdaqFragments_module.cc
```diff
@@ -0,0 +1,437 @@
...
+// ======================================================================
+
+class art::StrawDigisFromArtdaqFragments : public EDProducer {
+
+public:
+
+  struct Config {
+    fhicl::Atom<int> diagLevel    {fhicl::Name("diagLevel"    ), fhicl::Comment("diagnostic severity level, default = 0" ), 0};
+    fhicl::Atom<int> debugLevel   {fhicl::Name("debugLevel"   ), fhicl::Comment("debug level, default = 0"               ), 0};
+    fhicl::Atom<int> saveWaveforms{fhicl::Name("saveWaveforms"), fhicl::Comment("save StrawDigiADCWaveforms, default = 1"), 1};
```
brownd1978, line 52: Is this an option that might be changed in normal production? If so, it should not have a hard-coded default.

brownd1978: Also, using 'int' as 'bool' is deprecated. Please change to bool so that values 'true' and 'false' can be set in fcl.

DAQ/src/StrawDigisFromArtdaqFragments_module.cc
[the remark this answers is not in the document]
rlcee: Pasha and talked a week ago.  I would fairly strongly recommend not starting a database table for swaps.  Pasha made the case he wanted to swap fibers in his tests and have reco just work.  By not scripting the swap in the download now, the cost is that you are maintaining the database swap table forever.   In either case, you are managing list of swaps.  I would want to manage that as early as possible in the processing chain to get it out of the way.  A temp fcl setup to make the teststand easier, and but no permanent table, is fine with me.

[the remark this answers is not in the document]
rlcee: We do have static channel maps based on text files for CRV and calo.  If this situation is effectively a single permanent map, this pattern should work well.  Ultimately it is up to you.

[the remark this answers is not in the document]
edcallaghan: Ah, ok. If there's precedent for that, then it will probably work for us --- thanks. We don't really know yet if something will change when we are extracted, but certainly once inside the solenoid it will be frozen for Run I.

eflumerf (approved):

eflumerf (changes requested):
Magic numbers are present in this code, that should be resolved.

eflumerf (approved):
#1476 resolves my comment. If that PR fully contains the commits from this one, I would advise simply closing this one and merging that one. (In that case, if #1476 is merged, Github should recognize this and automatically mark this PR as merged)

PR #1405, opened 2025-01-28
Add compile option to workaround GCC optimizer bug

eflumerf:
Also update CMakeLists.txt for added modules/libraries.


rlcee (commented):
@eflumerf can you note what code triggered the stringop-overflow error?  Can it be rewritten to avoid the error?  I think I have to ask since this is a potentially useful switch to leave on. 
Al the rest looks good.  Thanks

eflumerf:
So it was a standard vector.insert, the problem isn't the code but a bug in the GCC optimizer: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=106199

rlcee (approved):

PR #1410, opened 2025-02-03
Make GMP a target so that the exported dependencies can find the library

eflumerf:
on their own (instead of putting an absolute path in)


rlcee:
If I'm reading this correctly, the library will be built or not based on whether an rpm is installed.  This would be a new technique for offline.  One idea would be to remove this dependence, if possible.  But in general, we can control the rpm content of the interactive machines and grid containers, so if faced with this issue, we would update the rpm.  This happened many times during the commissioning of al9, for example.

eflumerf:
In the Spack world, this is a non-issue, since `gmp` is available as a Spack package, and the Offline recipe can add `depends_on(gmp)`.

rlcee:
I dug into this further and understand better, this is OK with me.  Thanks

PR #1413, opened 2025-02-08
Updates for CRV Decoder and CRV Calibration

ehrlich-uva:


ehrlich-uva:
This PR does not interfere with any simulations or reconstructions.

RecoDataProducts/inc/CrvDAQerror.hh
```diff
@@ -0,0 +1,38 @@
...
+               _errorCode(errorCode), _subEvent(subEvent), _dataBlock(dataBlock), _packetCount(packetCount) {}
+
+    int  GetErrorCode() const     {return _errorCode;}
+    int  GetSubEvent() const      {return _subEvent;}
+    int  GetDataBlock() const     {return _dataBlock;}
+    int  GetPacketCount() const   {return _packetCount;}
+
+    private:
+
+    int    _errorCode{0};
```
brownd1978, line 30: How do you propose to deal with interpreting these values?
Could you use https://github.com/Mu2e/Offline/blob/b4e9f4fd506f1f4d57b4fdbca2184eb93e468eac/GeneralUtilities/inc/EnumToStringSparse.hh
to give values names?

CRVConditions/data/calib_widebandExtracted.txt
brownd1978, line 1: Why does this table need so many non-physical entries?

brownd1978 (commented):
I have some generic question, plus I'd like to hear from the DAQ people.

RecoDataProducts/inc/CrvDAQerror.hh
```diff
@@ -0,0 +1,38 @@
...
+    public:
+
+    enum errorCode{unableToGetDataBlock=0, invalidPacket=1, wrongSubsystemID=2, errorUnpackingStatusPacket=3, errorUnpackingCrvHits=4};
+
+    CrvDAQerror() {}
+
+    CrvDAQerror(int errorCode, int subEvent, int dataBlock, int packetCount) :
+               _errorCode(errorCode), _subEvent(subEvent), _dataBlock(dataBlock), _packetCount(packetCount) {}
+
+    int  GetErrorCode() const     {return _errorCode;}
```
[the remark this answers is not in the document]
ehrlich-uva: As far as I know, they are being used by all subsystems. I think it will require a discussion with experts of all subsystems to find a permanent solution. 
For now, I just wanted to catch the errors we get from our Wideband CRV data. So this class more as starting point until we decide on a more permanent solution (perhaps some base class, with specific implementations for each subsystem).

RecoDataProducts/inc/CrvDAQerror.hh
ehrlich-uva: Yes, it could be done this way. Otherwise, we can just check the enums/ints to see if/which errors occurred.

CRVConditions/data/calib_widebandExtracted.txt
ehrlich-uva: Currently, only a small number of channels are connected. And it's likely going to stay this way for a while until we get our new FEBs. In the mean time, we still want to be able to work with this subset of channels and need their calibration constants.

RecoDataProducts/inc/CrvDAQerror.hh
brownd1978: The strings are useful to communicate with humans (print).

RecoDataProducts/inc/CrvDAQerror.hh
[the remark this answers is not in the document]
brownd1978: I'm OK approving a prototype class serving an immediate urgent need. I'd still like a DAQ person to review, and comment on the general situation.

RecoDataProducts/inc/CrvDAQerror.hh
ehrlich-uva: I updated this class using EnumToStringSparse.

brownd1978 (approved):
Thanks for making this a named enum. I'd like someone from Online to review as well.

DAQ/src/CrvDigisFromFragments_module.cc
```diff
@@ -103,41 +106,66 @@ void CrvDigisFromFragments::produce(Event& event)
 
     for(size_t iDataBlock = 0; iDataBlock < CRVDataDecoder.block_count(); ++iDataBlock)
     {
-      if(_diagLevel>0) std::cout << "iSubEvent/iDataBlock: " << iSubEvent << "/" << iDataBlock << std::endl;
-
       auto block = CRVDataDecoder.dataAtBlockIndex(iDataBlock);
       if(block == nullptr)
       {
-        std::cerr << "Unable to retrieve block in " << std::endl;
+        std::cerr << "iSubEvent/iDataBlock: " << iSubEvent << "/" << iDataBlock << std::endl;
```
eflumerf, line 112: If this module is to be used in the TDAQ, `TLOG(TLVL_ERROR)` is preferred for compatibility with messaging systems. For purely Offline use, I'd still recommend using at least mf::LogError (though TLOG should be fully compatible)

eflumerf: Using TLOG also obviates the need for _diagLevel-style constructs, as TRACE messages can be enabled/disabled dynamically at runtime without altering module configuration.

RecoDataProducts/inc/CrvDAQerror.hh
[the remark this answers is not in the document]
eflumerf: Data from the ROCs follow a structure of Event/SubEvent/DataBlock/Packet#, this is documented in DocDB #4097 and #4914.

DAQ/src/CrvDigisFromFragments_module.cc
ehrlich-uva: This module will only be used in Offline. Most of the code in Offline uses cout/cerr/diagLevel while TLOG and mf::LogError are used only at a few places in Offline.

ehrlich-uva:
Merge needs to wait until we get the new artdaq-core-mu2e tag in Offline.

brownd1978 (approved):

PR #1416, opened 2025-02-14
update adcm with p068

rlcee:
artdaq-core v3_13_00
artdaq-core-mu2e  v3_04_00


rlcee:
@FNALbuild run build test with #1413

brownd1978 (approved):

PR #1433, opened 2025-03-06
Update CMake targets with their Spack-compatible names

eflumerf:


eflumerf:
(Requires change in artdaq-core-mu2e, pending)

brownd1978:
what tag of artdaq-core-mu2e is needed for compatibility with this PR?

DAQ/src/CrvDigisFromFragments_module.cc
ehrlich-uva, line 1: This change is already part of my PR 1424, but it's Ok to do it here.

eflumerf:
Not tagged yet. I linked the relevant PR in artdaq-core-mu2e. We want to be very careful about this one, as it is a breaking change due to the CMake target names changing up and down the product stack.

brownd1978:
@michaelmackenzie the trigger CI is failing in this PR. I checked the log file, the art job ran successfully and exited status 0, but the timing has changed. Can you please comment if this was an expected change?

ehrlich-uva (approved):

PR #1436, opened 2025-03-15
Add example module to examine the trigger path results

michaelmackenzie:
Add example module that exercises the trigger results navigator and plots trigger path quantities. Also import changes to an example trigger analysis module used by students. I added an example fcl that can be used to plot the trigger path information for files that include the trigger results.


Trigger/CMakeLists.txt
```diff
@@ -38,4 +38,19 @@ cet_build_plugin(ReadTriggerInfo art::module
       Offline::TrackerGeom
 )
```
brownd1978, line 39: Given that ReadTriggerInfo now depends on KinKal libraries shouldn't those be in the CMakeList as well?

brownd1978 (commented):
I'm not sure the CMake dependencies are correct.

Trigger/CMakeLists.txt
michaelmackenzie: I updated the CMakeList to include these, but I'm not positive if I've used the correct syntax. Does it look right now?

michaelmackenzie:
> Does the ReadTriggerInfo_module now include the plot of the counts sequence within a given trigger_path? All the changes look good to me

This is included in the ReadTriggerPaths module, which plots the fired trigger path names, bits, and indices as well as the module cut-flow for each trigger path.

michaelmackenzie:
@FNALbuild run build test

brownd1978:
@FNALbuild run build test

PR #1437, opened 2025-03-16
good run list

rlcee:
Start the good run list as described (mostly) in doc 46571.  See also [wiki page](https://mu2ewiki.fnal.gov/wiki/Goodruns).  This should have no effect on validation.

### DIFF
--- a/DbService/inc/GrlBit.hh
+++ b/DbService/inc/GrlBit.hh
@@ -1,0 +1,42 @@
+#ifndef DbService_GrlBit_hh
+#define DbService_GrlBit_hh
+
+// definition of a bits that can be set within a given word
+
+#include <iostream>
+#include <string>
+#include <vector>
+
+namespace mu2e {
+
+class GrlBit {
+ public:
+  typedef std::vector<GrlBit> BitVec;
+
+  GrlBit(std::string const& name, std::string const& bitname, int bitnumber,
+         std::string const& description = "",
+         std::string const& create_time = "",
+         std::string const& create_user = "") :
+      _name(name),
+      _bitname(bitname), _bitnumber(bitnumber), _description(description),
+      _create_time(create_time), _create_user(create_user) {}
+
+  std::string const& name() const { return _name; }
+  std::string const& bitname() const { return _bitname; }
+  int bitnumber() const { return _bitnumber; }
+  std::string const& description() const { return _description; }
+  std::string const& createTime() const { return _create_time; }
+  std::string const& createUser() const { return _create_user; }
+
+ private:
+  std::string _name;
+  std::string _bitname;
+  int _bitnumber;
+  std::string _description;
+  std::string _create_time;
+  std::string _create_user;
+};
+
+}  // namespace mu2e
+
+#endif
--- a/DbService/inc/GrlHeader.hh
+++ b/DbService/inc/GrlHeader.hh
@@ -1,0 +1,39 @@
+#ifndef DbService_GrlHeader_hh
+#define DbService_GrlHeader_hh
+
+// holds a defintion of a goodrun list - name, etc
+// locked==0 is unlocked, locked >0 is locked
+
+#include <iostream>
+#include <string>
+#include <vector>
+
+namespace mu2e {
+
+class GrlHeader {
+ public:
+  typedef std::vector<GrlHeader> HeaderVec;
+  GrlHeader(const std::string& name, int locked = 0, int lid = -1,
+            std::string const& create_time = "",
+            std::string const& create_user = "") :
+      _name(name),
+      _locked(locked), _lid(lid), _create_time(create_time),
+      _create_user(create_user) {}
+  const std::string& name() const { return _name; }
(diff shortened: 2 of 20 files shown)


DbService/inc/GrlBit.hh
```diff
@@ -0,0 +1,42 @@
...
+  std::string const& bitname() const { return _bitname; }
+  int bitnumber() const { return _bitnumber; }
+  std::string const& description() const { return _description; }
+  std::string const& createTime() const { return _create_time; }
+  std::string const& createUser() const { return _create_user; }
+
+ private:
+  std::string _name;
+  std::string _bitname;
+  int _bitnumber;
```
brownd1978, line 34: Is bitnumber intended to be an index into a bitfield? If so, what would a negative value mean?

brownd1978 (commented):
There seems to be a lot of overlap in functionality between good run bits and the existing [bitfield class](https://github.com/Mu2e/Offline/blob/main/GeneralUtilities/inc/BitMap.hh). Is there a reason that couldn't be reused here?

rlcee:
> There seems to be a lot of overlap in functionality between good run bits and the existing [bitfield class](https://github.com/Mu2e/Offline/blob/main/GeneralUtilities/inc/BitMap.hh). Is there a reason that couldn't be reused here?

I admit I forgot about this class, but the GRL bits needs to be dynamic, you can create one then use it immediately, but the bit class requires a static, compiled list of bit definitions.
>The detail class must contain an enum named bit_type

DbService/inc/GrlBit.hh
rlcee: The database does not have a unsigned int, only an int. I prefer to let the code reflect that.  I throw an error if the high bit might be set

DbService/inc/GrlHeader.hh
```diff
@@ -0,0 +1,39 @@
...
+#include <iostream>
+#include <string>
+#include <vector>
+
+namespace mu2e {
+
+class GrlHeader {
+ public:
+  typedef std::vector<GrlHeader> HeaderVec;
+  GrlHeader(const std::string& name, int locked = 0, int lid = -1,
```
brownd1978, line 16: what use case do you imagine requires everything except name have default (non-functional) values? Would it be better to have a default (no argument) constructor, and one that requires all fields be set?

DbService/inc/GrlHeader.hh
```diff
@@ -0,0 +1,39 @@
...
+            std::string const& create_time = "",
+            std::string const& create_user = "") :
+      _name(name),
+      _locked(locked), _lid(lid), _create_time(create_time),
+      _create_user(create_user) {}
+  const std::string& name() const { return _name; }
+  int locked() const { return _locked; }
```
brownd1978, line 23: What is the use case for locking a good run list? Is the implication that grl are 'mutable' unless locked? Does that refer to anything besides extension (ie is actual definition intended to be mutable)?  Who will be allowed to lock/unlock a grl and what mechanism do you forsee for controlling that?

DbService/inc/GrlTool.hh
```diff
@@ -0,0 +1,69 @@
...
+ public:
+  GrlTool();
+
+  // define a quality word
+  int defineWord(const std::string& name, const std::string& desc = "");
+  // define bits within the word
+  int defineBit(const std::string& name, const std::string& bitname,
+                const std::string& desc = "");
+  // set quality bits for a word, if bits is empty, then no bits set, all good
+  int setBits(const std::string& name, const DbIoV& iov,
```
brownd1978, line 29: Do you intend this interface allowing creation/definition of grl bits to be open to all? If not, how do you forsee administering access?

brownd1978 (commented):
more questions inline

DbService/inc/GrlHeader.hh
rlcee: The idea is that an unlocked GRL can be extended, but not contracted, for the use case of incoming data, for example "good for pass1" or "good for calibration".   An example for locking a GRL is to fix the list permanently for APS 2026, for example, or 10% e- publication.  There is no unlocking function. (trivial to create new version tho).

DbService/inc/GrlTool.hh
rlcee: Database write access is determined by the user, who has to have a kerberos identity and a database account with the right privilege.  Currently, only mu2epro can write, but I imagine an online shift or ops account would be another obvious candidate.  Maybe certain people from analysis groups.  As it is now, anyone with write access can write to any GRL, but if there is a need to narrow privileges, we can do that. Most errors could be repaired by an admin too.

DbService/inc/GrlHeader.hh
rlcee: you are right, there are no general use cases so I reduced it to more logical pair

tedeschi (approved):
Code base looks fine to me to start.  I checked my expected usage via the wiki page descriptions more than the code details.

PR #1442, opened 2025-03-19
Remove otsdaq-mu2e dependency

eflumerf:


matthewstortini:
@rlcee @brownd1978 we need this merged before we make the new tag we discussed in slack.

PR #1455, opened 2025-04-03
p075 for new adcm and kinkal

rlcee:
New dependencies:
artdaq-core-mu2e v5_00_00
kinkal v3.1.7
tested for basic functionality


brownd1978:
@FNALbuild run build test with Mu2e/Production#417

PR #1456, opened 2025-04-04
new CRV geometry and prolog

ehrlich-uva:
1. Added new CRV geometry (v10) and prologs (v12). 
2. New geometry still needs to be tested for overlaps.
3. Default geometry is still v09. Default prologs are still v11.
4. New geometry is incompatible with old art files, since it has a different number of channels.
5. Based on latest crv_parameter.xlsx.
6. Needs completely new set of database entries for channel map, light yields, calibration constants, channel status, since the number of channels has changed.


brownd1978 (approved):

ehrlich-uva:
@brownd1978 Could you please add @AndrewEdmonds11 as a reviewer, because I needed to make some changes to the STM geometry (to remove the dependency of the STM location from the location of the CRV-D).

ehrlich-uva:
@brownd1978 Could you please also add @oksuzian  as a reviewer?

Mu2eG4/geom/STM_v08.txt
```diff
@@ -22,6 +22,8 @@ bool   stm.SScollimator.liner.build   = false;
 bool   stm.detector.stand.build       = false;
 bool   stm.shield.build               = true;
 
+//STM reference point that was previously calculated automatically based on the CRV-D location
+double stm.referenceZ = 18766.81;
```
AndrewEdmonds11, line 26: Just to ask how you got the number: did you just print out the value of ```z_crv_max``` in ```STMMaker```, or something else?

AndrewEdmonds11 (approved):
Thanks for making the change to the STM reference point, Ralf! I just have one question in-line but otherwise it all looks good to me

Mu2eG4/geom/STM_v08.txt
ehrlich-uva: Yes, that's exactly what I did. I printed out the value and used it in the txt file.

ehrlich-uva:
The recent commits did the following
1. Removed all overlaps.
2. Added channel map for new geometry.
3. Added text files for the database entries (light yields, calibration constants, channel status) for the new geometry.
4. Added a new geom_common_crv_v10.txt that uses the current geom_common, but updates the CRV geometry.

The remaining tasks are
1. Switching to the new geometry (which will break backward compatibility).
2. Switching to the new prologs.
3. Adding the database text files to the official database and change the default settings so that the new database entries will be used.

Mu2eG4/src/constructSTM.cc
```diff
@@ -124,20 +124,6 @@ namespace mu2e {
...
-    CLHEP::Hep3Vector   crvd_position    = CRS->getSectorPosition("D");
-    const double z_crv_max = crvd_position.z() + crvd_halflengths[2];
-
-    // Create a reference position (most things in the STM geometry will be defined w.r.t. this position)
-    // Our reference z is the downstream edge of the CRV
-    const G4ThreeVector stmReferencePositionInMu2e(dsP.x(),
-                                                    0.0,
-                                                    z_crv_max );
-    const G4ThreeVector stmReferencePositionInParent = stmReferencePositionInMu2e - parentCenterInMu2e;
-
```
ehrlich-uva, line 140: This section of the code was used to fill a variable that was not used anywhere in the code.

brownd1978:
Can you check whether default Offline can still read MDC2020 data with these changes? If so we can merge it now.

ehrlich-uva:
I double checked the changes and also tested several art file. I found that this PR does not prevent us from reading older art files.

PR #1460, opened 2025-04-09
Remake of all the STM simulation study changes

PawelPlesniak:
# STM Normalization toolkit
To be used in parallel with the parallel PR in Production [here](https://github.com/Mu2e/Production/pull/416)
Details of how these piece together are shown in doc-db [51487](https://mu2e-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=51487)

PR contains the following
 - Photon gun for development of the digitization and analysis readout chain
 - HPGe digitization model 
 - Fixing the STMReco tools with simulated data passed through
 - Includes all the tools to plot and analyze the results 
 
TLDR - this is the same PR that I made [previously](https://github.com/Mu2e/Offline/pull/1449), but without all the `out/` history. The other PR became very messy with my attempts to remake the history. I checked that the main fcl file `STMMC/fcl/HPGeReco.fcl` still works as it uses most of the modules that were changed. All the updates I wrote to make this analysis happen have been copied over from the other PR.


PawelPlesniak:
@AndrewEdmonds11 agreed to review this PR, @brownd1978 can he be added as the reviewer?

PawelPlesniak:
For the position of the absober, I fired photons at the absorber and looked at the particle stepping logs. I validated the position is centred on the SSC aperture by adding a print to the stm geometry and recompiling, and it was where expected, but I didn't see any steps in the absorber until this y co-ordinate. I didn't complete this study due to time constraints, but the effects of the absorber were seen in the VD101 spectrum at the position it is expected. I also viewed the geometry with root and it is in the correct location, this may have been a misconfiguration of the geometry that I was using at the time

Analyses/CMakeLists.txt
```diff
@@ -136,18 +136,13 @@ cet_build_plugin(CosmicFilter art::module
 cet_build_plugin(CountPionDecays art::module
     REG_SOURCE src/CountPionDecays_module.cc
     LIBRARIES REG
-
```
AndrewEdmonds11, line 139: If you could correct whitespace changes, that would be good. It will help future merges

STMMC/src/HPGeTree_module.cc
```diff
@@ -0,0 +1,229 @@
...
+      parent = parent->parent();
+
+    // Return the ID and PDG ID
+    return std::make_tuple(parent->id(), parent->pdgId());
+  };
+
+  double HPGeTree::parentTime(const art::Event& event, key_type parentId) {
+    // Get the data products from the event
+    auto const& StepPointMCs = event.getProduct(StepPointMCsToken);
+    auto const& SimParticles = event.getProduct(SimParticlemvToken);
```
AndrewEdmonds11, line 131: Is there a reason you are grabbing the SimParticles directly here? Usually we just get to them through the StepPointMC

AndrewEdmonds11 (commented):
Thanks, Pawel. Just one request to fix whitespace differences and then one question. Otherwise this looks good to me

Analyses/CMakeLists.txt
PawelPlesniak: Sorry I don't understand what the requested whitespace changes are - could you clarify? The only whitespace convention I know of (because it's enforced at the commit level) is end of line whitespace, but I do not know about the others.

STMMC/src/HPGeTree_module.cc
PawelPlesniak: I wasn't sure how to implement this. I am happy to change it, but will need some help to implement it. Either way, I suspect that this module will be used quite rarely as it generates the toy reconstruction of the HPGe detector

AndrewEdmonds11: I think it's not worth implementing now. Could you add a ```// TODO:``` comment on this line so that we remember to do it when we next look at the file?

AndrewEdmonds11:
I've just checked the failing cmake test and these are the problems:
 - modules not defined in a CMakeLists.txt: PhotonGun, ConcatenateDigitizedWaveforms, ConcatenationFilter, DetectorEfficiency
 - modules defined in a CMakeLists.txt but no module found: ConcatenationFilter, HPGeEfficiency

The "?" trigger test just reports: ```trigger.log: No such file or directory```. This PR doesn't touch the trigger so I don't think this should prevent the PR from being merged when it's ready

STMMC/src/HPGeTree_module.cc
PawelPlesniak: Done

Analyses/CMakeLists.txt
PawelPlesniak: Done

PawelPlesniak:
@AndrewEdmonds11 Let me know if there's anything else to change

AndrewEdmonds11 (approved):
Looks good to me. Thanks, Pawel!

PawelPlesniak:
@brownd1978 can this be merged?

PR #1466, opened 2025-04-14
p074 envset

rlcee:
This will change the envset to p074, which points to art 3.1.5.  This should fix the problem with xrootd not accepting tokens.
A validation test with 2k reco events shows no difference. A test with 1k ceSimReco shows compatible but not identical results, and is shown [here](https://mu2e.fnal.gov/atwork/computing/ops/val/plots/temp/ce/result.html)

### DIFF
--- a/.muse
+++ b/.muse
@@ -1,5 +1,5 @@
 # prefer to build with this environment
-ENVSET p074
+ENVSET p076
 # add Offline/bin to path
 PATH bin
 # recent commits can take enforce these flags
--- a/.muse
+++ b/.muse
@@ -1,5 +1,5 @@
 # prefer to build with this environment
-ENVSET p075
+ENVSET p074
 # add Offline/bin to path
 PATH bin
 # recent commits can take enforce these flags


brownd1978 (commented):
Why go backwards?

rlcee:
> Why go backwards?

For the last several envset changes, I'd make two envsets, one in art 3.1.5 and one as a tweak to current 3.1.4 .  So p074 is the same as p075, except for this and few other minor differences.  I did it this way to keep the current head as the highest number envset, which muse defaults to in certain rare-ish cases.  One could argue that I should copy p074 to p076 for the same reason.  This would trigger more spack work, depending on your tolerances for misaligned labels.  To me it was a toss-up.  One could also consider that adcm was just tagged, so we are heading to a new envset soon anyway.  Let me know what you want to do.

brownd1978:
I am probably missing something in this pattern. Please give me a number to call you at so we can discuss.

brownd1978 (approved):

PR #1469, opened 2025-04-16
open intervals

rlcee:
By request, add capability for managing db conditions content with open intervals.  This is for convenience in supplemental work, and will not be used in production. Documentation will be added in the [wiki](https://mu2ewiki.fnal.gov/wiki/OpenIntervals)


rlcee:
@FNALbuild run build test

brownd1978 (approved):

PR #1472, opened 2025-04-17
Add SurfaceSteps to product mixer

brownd1978:
This won't be needed for regular mixing but could be useful with cosmic resampling. coupled with [Production PR #422](https://github.com/Mu2e/Production/pull/422)


brownd1978:
@FNALbuild run build test with Mu2e/Production#422

PR #1475, opened 2025-04-25
Add additional compile-time flag for updated spack-mpd build

eflumerf:


gianipez (approved):

gianipez:
Hi @brownd1978 , this PR seems quite smooth... shall we merge it?

rlcee (approved):
I'd be curious why this is needed now.
I see this gets turned on by "VIGILANT", but so are some other items.
Eric and I synced the flags between Muse and cmake a while ago, but I think maybe we missed this.  In any case, this change brings the two a bit closer together.  
This makes some of the other flags obsolete.
I believe it will reduce warning coverage a little.

PR #1476, opened 2025-04-29
fcl-based mapping of tracker panel identification

edcallaghan:
Offline data structures and conditions databases have historically labeled individual panels using pair of integers, enumerating planes in the tracker and panels within each plane, constituting a "mostly geographic" system; this is based on the nomenclature of DocDB-888. Each panel is read out by an individual "DTC link," which is a pair identifying DTCs within the computing system, and a channel within each DTC. The data payload from the tracker will label each hit by the panel's so-called ``Minnesota number,'' which is uncorrelated with both the geometry and readout channeling of the detector. Thus, data read out from the digi contain redundant panel-level information (the Minnesota number and DTC channeling), which must be translated into geographic coordinates for downstream analysis.

PR #1402 implements a deserialization routine for tracker digis and contains a fixed mapping from a few DTC channels to geographic coordinates. This PR generalizes that translation into a two-stage mapping derived from fcl: i) nominally, Minnesota numbers are translated into geographic coordinates, but; ii) an invalid Minnesota number in the data causes a second lookup in an optional table mapping DTC channeling and geographic coordinates (retaining a flag that the payload containing this digi is known to contain a mistake, i.e. the bad MN number). The Minnesota number is used as the primary key because it will be stored in persistent ROM on the frontend, and hence will not be susceptible to e.g. errors introduced by rearranging fibers/computers.

Try as I might, I could not construct a fhicl schema which would tolerate tables with keys not known at runtime, so all maps are written in fcl as tuples, from which maps are constructed in the deserialization module.

This PR should be merged after #1402 to preserve authorship and allow for a cleaner diff.


rlcee:
CRV and cal have this functionality implemented in the conditions service framework, like [here](https://github.com/Mu2e/Offline/blob/3c0bbbc2bcd695ec806bb0533436b0aa8e24d82b/DAQ/src/CRVDigisFromDataDecoders_module.cc#L164).  This code currently reads text files to define the conversion.

edcallaghan:
I didn't make a point of it in this PR, but the fcl interface, specifically, is only meant as a temporary measure to give us flexibility in decoding data from different test stands / tracker stations "today." We definitely will want something else as we approach having a single, static detector, and I don't currently have a preference between text-/db-based and/or moving the functionality into the data format codespace.

brownd1978:
The PR addresses shortcomings of the original PR (1402). It is intended as a stopgap to allow detector work to continue while code development proceeds. I agree with the comments about long-term solutions, but those are outside the scope of this PR, which is to provide code which meets Offline standards. I would appreciate your review from that perspective.

brownd1978:
Note 2 tests failed in the CI. cmake is failing with:
File /scratch/workspace/GitHubPRTests/mu2e-offline-build-test/Offline/DAQ/src/StrawDigisFromArtdaqFragments_module.cc is not defined in CMakeLists.txt!
File /scratch/workspace/GitHubPRTests/mu2e-offline-build-test/Offline/DAQ/src/StrawRecoFromArtdaqFragments_module.cc defined in CMakeLists.txt but not found in directory!
Job completed with failure.

There are also 20+ clang-tidy errors associated with the use of Trace. Addressing those are beyond the scope of this PR, but  it needs to be addressed eventually.

edcallaghan:
@FNALbuild run build test

eflumerf (approved):
This is fine for now, but an Issue should be opened to make sure that the channel-mapping is done in a way that satisfies all parties, and consistently across subsystems.

PR #1477, opened 2025-05-02
moved the borated polyethylene back to its original position

ehrlich-uva:
This fixes an overlap problem at the CRV-TS-Ext


brownd1978:
Is this change to fix to a G4 overlap, or just for backwards-compatibility with MDC2020 datasets?

ehrlich-uva:
This change fixes the G4 overlap. It doesn't affect the backwards-compatibility.

brownd1978 (approved):

brownd1978:
What is the long-term fix? If the moved poly results in a volume overlap, what else need to change with it?

ehrlich-uva:
The new geometry (crv_counters_v10) removes the FEBs for the CRV-TS-Ext modules, because they don't belong there (they were placed to a default location above the CRV modules, which is the wrong location). So the moved poly won't create any overlaps.

PR #1479, opened 2025-05-02
Update accept-reject filter module

michaelmackenzie:
Update the accept-reject filter module to use validated fcl and include some debug histograms. This is useful in producing physical spectra from weighted events, such as is needed for mock data samples. I will also put in a corresponding PR in Production to add RPC fcls that have the physical pion stop time distribution for use in MDS2.


sophiemiddleton (approved):
Hi Michael,

This looks like it does what we want for the RPC at least. Would be good to see the Production changes too to see how its implemented

michaelmackenzie:
Thank you for the prompt review! Here is the associated Production update to add these generation fcl:
https://github.com/Mu2e/Production/pull/427

PR #1482, opened 2025-05-04
Pion beam fix

sophiemiddleton:
Fix to pion beam campaign to select pions stopping at target


brownd1978 (approved):

EventGenerator/src/PionFilter_module.cc
```diff
@@ -61,23 +66,30 @@ namespace mu2e {
         for(const auto& aParticle : *ah){
           art::Ptr<SimParticle> pp(ah, aParticle.first.asUint());
           float _endglobaltime = pp->endGlobalTime();
-          if( pp->stoppingCode() == ProcessCode::mu2eKillerVolume and std::abs(pp->pdgId()) == PDGCode::pi_plus){
+          if( pp->stoppingCode() == ProcessCode::hBertiniCaptureAtRest and std::abs(pp->pdgId()) == PDGCode::pi_plus){
```
brownd1978, line 69: Is this selection dependent on physics list? Ie ci
Outdoor the pion stop process change? If so this should be a configurable parameter.

sophiemiddleton: I dont think it should be

EventGenerator/src/PionFilter_module.cc
```diff
@@ -17,7 +18,6 @@
 #include "Offline/TrackerGeom/inc/Tracker.hh"
```
michaelmackenzie, line 18: I think some of these includes aren't needed here and can be removed (e.g. geometry ones)

EventGenerator/src/PionFilter_module.cc
```diff
@@ -26,22 +26,26 @@ namespace mu2e {
       struct Config {
         using Name=fhicl::Name;
         using Comment=fhicl::Comment;
-        fhicl::Atom<int> diagLevel{Name("diagLevel"),0};
+        fhicl::Atom<int> diagLevel{Name("diagLevel"),1};
         fhicl::Atom<double> tmin{Name("tmin"),0};
         fhicl::Atom<double> tmax{Name("tmax"),1e6};
         fhicl::Atom<bool> isNull{Name("isNull"),true};
```
michaelmackenzie, line 32: Is it normal to have a filter module default to off even if the filtering parameters are configured?

EventGenerator/src/PionFilter_module.cc
```diff
@@ -61,23 +66,30 @@ namespace mu2e {
...
           float _endglobaltime = pp->endGlobalTime();
-          if( pp->stoppingCode() == ProcessCode::mu2eKillerVolume and std::abs(pp->pdgId()) == PDGCode::pi_plus){
+          if( pp->stoppingCode() == ProcessCode::hBertiniCaptureAtRest and std::abs(pp->pdgId()) == PDGCode::pi_plus){
             const PhysicsParams& gc = *GlobalConstantsHandle<PhysicsParams>();
-            totalweight += exp(-1*pp->endProperTime() / gc.getParticleLifetime(pp->pdgId()));
+            _totalweight += exp(-1*pp->endProperTime() / gc.getParticleLifetime(pp->pdgId()));
+            _ntot += 1;
           }
-          if( pp->stoppingCode() == ProcessCode::mu2eKillerVolume and (std::abs(pp->pdgId()) == PDGCode::pi_plus and _endglobaltime > tmin_ and _endglobaltime < tmax_ )){
+          if( pp->stoppingCode() == ProcessCode::hBertiniCaptureAtRest and (std::abs(pp->pdgId()) == PDGCode::pi_plus and _endglobaltime > tmin_ and _endglobaltime < tmax_ )){
```
michaelmackenzie, line 74: Was this used before to select pions output from a previous beam stage and then resample the beamline transport simulation? Does this correspond to a change in the production campaign flow? (I genuinely don't know much about how these campaigns are structured, so I'm really just asking out of curiosity!)

michaelmackenzie (approved):
This looks good to me, I've just added a few minor comments that shouldn't hold up the merging of the PR as they're just suggestions/questions.

sophiemiddleton:
Thanks Michael. Tbh the pion campaign has always been a little unfinished because we never had any physics users until now. This is the first real RPC production and analysis (except the small sample for MDS1g). We are catching a lot of small bugs thay werent apparent with previous simple tests

EventGenerator/src/PionFilter_module.cc
sophiemiddleton: will fix

EventGenerator/src/PionFilter_module.cc
sophiemiddleton: will fix

PR #1483, opened 2025-05-05
Add protection for poor maximum weight configuration

michaelmackenzie:
Add an event weight output from the accept/reject filter module. This will be a flat weight in the normal operating mode, but if the maximum event weight is set too low, it will correct the output spectrum shape.


brownd1978 (approved):
Looks great Michael, thanks!

PR #1484, opened 2025-05-07
Beam dump move

gaponenko:
The changes:
- move the beam dump core to its nominal position
- more realistic steel on top of the beam dump
- more realistic air gap around the dump core
-  ExtMon filter geometry switch to the absolute positioning, the numbers use Metrology inputs
- an option to visualize ExtMon area wall scans done by the Metrology group
- cleanups and updates of the beam dump-related code


GeometryService/src/ProtonBeamDumpMaker.cc
```diff
@@ -20,224 +21,308 @@
...
+    const Hep2Vector& psWoff2 = toOutlinePlane(psWall.getOffsetFromMu2eOrigin());
+
+    dump->_dumpConcreteOutline.emplace_back(pswVertices[14]+psWoff2 + ovlgap*(-nx+nz));
+    dump->_dumpConcreteOutline.emplace_back(pswVertices[15]+psWoff2 + ovlgap*( nx+nz));
+    dump->_dumpConcreteOutline.emplace_back(intersection(faceLine, Line(pswVertices[15] + psWoff2, pswVertices[16] + psWoff2)) + ovlgap*( nx));
+    dump->_dumpConcreteOutline.emplace_back(intersection(faceLine, Line(pswVertices[13] + psWoff2, pswVertices[14] + psWoff2)) + ovlgap*(-nx));
+
+    // The ExtMon room subtraction.
+    // The the back is the same as the concrete back.  Add a margin for the G4 boolean.
+    const double booleanVolumeMargin = 50; // mm
```
bechenard, line 169: Is that a value that should be in the config file?

ProtonBeamDumpGeom/inc/ProtonBeamDump.hh
```diff
@@ -23,52 +23,60 @@ namespace mu2e {
...
+    // the ExtMon room cutout.  Another subtraction is based on the
+    // Entrace collimator for the Extinction Monitor, its parameters
+    // are not provided by this class.
+    //
+    // The parameters for the mouth and neutron cave subtraction
+    // volumes here are nominal.  The margins to ensure proper
+    // overlaps for G4 booleans are added by the code talking to G4.
+    // OTOH extMonSubtractionOutline is tweaked here, as postponing
+    // the extruded solid tweak would make the overall code too entangled.
+
```
bechenard, line 49: Technically, you don't need a const accessor if you return by value in this function. Is there a reason you return by value and not reference?

bechenard (commented):
Nice, you one or two little things.

ProtonBeamDumpGeom/inc/ProtonBeamDump.hh
gaponenko: Good catch.  I will look at the header and clean it up, changing to const ref returns.

GeometryService/src/ProtonBeamDumpMaker.cc
gaponenko: No, this is a non-physical parameter that is only needed to deal with G4 subtleties. Adding it to a config file would only create confusion.

gaponenko:
@FNALbuild run build tests

bechenard (approved):
All good

PR #1486, opened 2025-05-08
Add track overlap clustering

michaelmackenzie:
Add a module that produces clusters of tracks with overlaps. Additionally add an example module that will resolve these overlaps, but this should generally be done at the analysis level.

The example track resolution algorithm is effective at identifying the correct track trajectory for the correct particle hypothesis, but it is not as effective at resolving PID and trajectory together (e.g. separating downstream electron from upstream muon). PID is likely an analysis-level decision, where, for example, pion and muon track separation will require subtle handling and will vary with the goals of the analysis. Given the track matching collections, analyzers will be able to check if any resolution is needed at the analysis/ntuple level.

### DIFF
--- a/RecoDataProducts/inc/KalSeed.hh
+++ b/RecoDataProducts/inc/KalSeed.hh
@@ -96,5 +96,6 @@ namespace mu2e {
   typedef std::vector<mu2e::KalSeed> KalSeedCollection;
   typedef art::Ptr<mu2e::KalSeed> KalSeedPtr;
   typedef std::vector<mu2e::KalSeedPtr> KalSeedPtrCollection;
+  typedef std::vector<KalSeedPtrCollection> KalSeedPtrCollections;
 }
 #endif
--- a/RecoDataProducts/inc/KalSeed.hh
+++ b/RecoDataProducts/inc/KalSeed.hh
@@ -96,6 +96,7 @@ namespace mu2e {
   typedef std::vector<mu2e::KalSeed> KalSeedCollection;
   typedef art::Ptr<mu2e::KalSeed> KalSeedPtr;
   typedef std::vector<mu2e::KalSeedPtr> KalSeedPtrCollection;
-  typedef std::vector<KalSeedPtrCollection> KalSeedPtrCollections;
+  typedef std::vector<mu2e::KalSeedPtr> KalSeedCluster;
+  typedef std::vector<mu2e::KalSeedCluster> KalSeedClusterCollection;
 }
 #endif
--- a/TrkDiag/src/TrackMatching_module.cc
+++ b/TrkDiag/src/TrackMatching_module.cc
@@ -50,7 +50,7 @@ namespace mu2e
 
   private:
     void produce(art::Event& event) override;
-    bool match(const KalSeedPtr k_1, const KalSeedPtr k_2);
+    bool match(const KalSeedPtr& k_1, const KalSeedPtr& k_2);
     void merge_clusters(KalSeedCluster& cluster, KalSeedCluster& cluster_j);
     void create_clusters(std::vector<KalSeedCluster>& clusters);
 
@@ -92,8 +92,9 @@ namespace mu2e
 
   //--------------------------------------------------------------------------------------------------------
   // Perform the matching
-  bool TrackMatching::match(const KalSeedPtr k_1, const KalSeedPtr k_2) {
-    if(!k_1 || !k_2)
+  bool TrackMatching::match(const KalSeedPtr& k_1, const KalSeedPtr& k_2) {
+    if(_debug > 3) printf("  Starting track overlap check\n");
+    if(k_1.isNull() || k_2.isNull())
       throw cet::exception("RECO") << "mu2e::TrackMatching::" << __func__ << ": Null input track seeds!";
 
     // Retrieve the hit lists
@@ -143,7 +144,6 @@ namespace mu2e
       if(cluster.size() > 1) // input clusters should always be 1 track, as matches are clustered into lower index clusters
         throw cet::exception("RECO") << "mu2e::TrackMatching::" << __func__ << ": Input track cluster has already been clustered! Size = " << cluster.size();
       if(_debug > 2) printf("  Checking track cluster %zu for overlaps\n", index);
-      KalSeedPtr& k_i = cluster.front();
 
       // Check each following cluster (of size 1) for overlapping tracks
       for(size_t jtrk = index + 1; jtrk < ntrks; ++jtrk) {
@@ -151,8 +151,14 @@ namespace mu2e
         if(cluster_j.empty()) continue;
         if(cluster_j.size() > 1) // input clusters should always be 1 track, as matches are clustered into lower index clusters
           throw cet::exception("RECO") << "mu2e::TrackMatching::" << __func__ << ": Track cluster matching against has already been clustered! Size = " << cluster_j.size();
-        if(_debug > 2) printf("    Checking against track cluster %zu for overlaps\n", jtrk);
+
+        // Get the tracks to compare
+        KalSeedPtr& k_i = cluster.front();
         KalSeedPtr& k_j = cluster_j.front();
+        if(k_i.isNull() || k_j.isNull())
+          throw cet::exception("RECO") << "mu2e::TrackMatching::" << __func__ << ": Bad track in clusters";
+
+        if(_debug > 2) printf("    Checking against track cluster %zu for overlaps\n", jtrk);
 
         // If the tracks overlap, add the tracks to the match lists
         if(match(k_i, k_j)) {
@@ -187,7 +193,7 @@ namespace mu2e
     // Start by creating a track cluster for each track, then merge collections as overlaps are found
 
     std::vector<KalSeedCluster> clusters;
-    for(auto handle : handles) {
+    for(auto& handle : handles) {
       const size_t ntrks = handle->size();
       for(size_t itrk = 0; itrk < ntrks; ++itrk) {
         KalSeedPtr ptr(handle, itrk);
--- a/TrkDiag/src/TrackMatching_module.cc
+++ b/TrkDiag/src/TrackMatching_module.cc
@@ -50,6 +50,14 @@ namespace mu2e
   private:
     void produce(art::Event& event) override;
     bool match(const KalSeed* k_1, const KalSeed* k_2);
+    void add_tracks(std::vector<std::vector<const KalSeed*>>& clusters,
+                    std::map<const KalSeed*, size_t>& cluster_index,
+                    const KalSeed* itrkptr, const KalSeed* jtrkptr);
+    void create_clusters(std::vector<const KalSeedCollection*>& trackColls,
+                         std::vector<art::ValidHandle<KalSeedCollection>>& handles,
+                         std::map<const KalSeed*, KalSeedPtr>& ptr_map,
+                         std::vector<std::vector<const KalSeed*>>& clusters,
+                         std::map<const KalSeed*, size_t>& cluster_index);
 
     std::vector<art::InputTag> _seedTags;
     double _threshold;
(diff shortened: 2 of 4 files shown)


michaelmackenzie:
@FNALbuild run build test

michaelmackenzie:
I believe the track clustering now covers all the expected cases, now checking within the track collection for overlaps as well as merging track collections if two tracks overlap but have already been clustered in separate collections. Here is an example fcl that can be used for testing if needed: `/exp/mu2e/app/users/mmackenz/Yale/main/fcl/test_trk_matching.fcl`

I don't know if having one master track overlap collection will be sufficient for all analyses though. For example, in an RMC pair conversion we often see a bad track that merges hits from the two helices into one large track. If this sufficiently overlaps with both good tracks, this will result in the two separate conversion tracks being marked as overlapping even if they don't share hits. The only way to cover this case, I think, would be to have a list of overlapping tracks per track, rather than a single cluster of tracks, or a track filter stage before this overlap checking that removes these bad tracks.

 I think this clustering is sufficient for most track analysis cases, and analyses it doesn't cover can modify this algorithm in the future (such as having veto criteria on tracks included in the clustering) or use upstream track filtering.

TrkDiag/src/TrackMatching_module.cc
```diff
@@ -0,0 +1,304 @@
...
+      KalSeedPtrCollection ptrs;
+      ptrs.push_back(entry.second);
+      if(_makeHists) _hists.cluster_sizes->Fill(1);
+      ptr_clusters->push_back(ptrs);
+    }
+
+    if(_debug > 0) printf("[TrackMatching::%s::%s] Found %i track clusters and %i single tracks\n",
+                          __func__, moduleDescription().moduleLabel().c_str(), nclusters, nsingle_clusters);
+    if(_makeHists) _hists.nclusters->Fill(nclusters + nsingle_clusters);
+
```
brownd1978, line 298: I suggest adding a test that every input track ends up in exactly 1 cluster.

I believe you could simplify the flow by starting by putting every input track into its own cluster. Then the clustering would proceed purely as merges, using a double-nested pairing loop ,instead of quadruple-nesting. This has the advantage of working directly with the output product, and should allow you to remove some (all?) of the intermediate bookkeeping structures.

TrkDiag/src/TrackResolution_module.cc
```diff
@@ -0,0 +1,166 @@
...
+      if(_debug > 1) printf("  Checking cluster of %zu tracks\n", cluster.size());
+      const KalSeedPtr* seedPtr = &(cluster.front());
+      if(_makeHists) {
+        _hists.fitcon->Fill((*seedPtr)->fitConsistency());
+      }
+
+      // Resolve the overlap for each track in the cluster
+      const size_t ntrks = cluster.size();
+      for(size_t itrk = 1; itrk < ntrks; ++itrk) {
+        const KalSeedPtr testPtr = cluster.at(itrk);
```
brownd1978, line 125: Usual practice is to skip bounds checking when the indices clearly cannot be out-of-bounds (as in this case). I also don't see why you need indices at all: an expression loop should do the same thing more compactly.

TrkDiag/src/TrackResolution_module.cc
```diff
@@ -0,0 +1,166 @@
...
+    // get the KalSeedPtrs
+    auto handle = event.getValidHandle<KalSeedPtrCollections>(_seedTag);
+    const auto& clusters = *handle;
+    if(_debug > 0) printf("[TrackResolution::%s::%s] Collection of %zu track clusters retrieved\n", __func__, moduleDescription().moduleLabel().c_str(), clusters.size());
+
+    // Go through the tracks and check for alternate track fit hypotheses using the same helix seed
+    for (const auto& cluster : clusters) {
+      if(cluster.empty()) continue;
+      if(_debug > 1) printf("  Checking cluster of %zu tracks\n", cluster.size());
+      const KalSeedPtr* seedPtr = &(cluster.front());
```
brownd1978, line 117: Pointers to ptrs is a convoluted data type I suggest to avoid. Since you eventually need a deep copy Ptr for your output collection, I suggest to start with that (ie push_back cluster.front() to your output collection), and then re-assign it as necessary. That simplifies the memory manipulation, and avoids the final translation.

brownd1978 (changes requested):
Overall the algorithms look good. The main request is to create an explicit type for KalSeed clusters. The other suggestions are to improve the code clarity and avoid direct memory manipulations.

TrkDiag/src/TrackMatching_module.cc
```diff
@@ -0,0 +1,304 @@
...
+  {
+  public:
+    struct Config {
+      using Name=fhicl::Name;
+      using Comment=fhicl::Comment;
+
+      fhicl::Sequence<art::InputTag> seedTags{Name("TrackCollections"), Comment("Input tags for track collections")};
+      fhicl::Atom<double> threshold{Name("overlapThreshold"), Comment("Fractional threshold in hits to match two tracks"), 0.5};
+      fhicl::Atom<bool> useIndices{Name("useIndices"), Comment("Use digi indices when matching, otherwise use straw ID + time"), true};
+      fhicl::Atom<double> timeWindow{Name("timeWindow"), Comment("Hit time window in matching, if not using indices"), 50.};
```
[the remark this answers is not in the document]
michaelmackenzie: I've made this optional, but I'm not sure how to best set a default value with OptionalAtom. For now I've set the default value in the constructor.

TrkDiag/src/TrackMatching_module.cc
```diff
@@ -0,0 +1,304 @@
...
+  }
+
+  //--------------------------------------------------------------------------------------------------------
+  // Perform the matching
+  bool TrackMatching::match(const KalSeed* k_1, const KalSeed* k_2) {
+    if(!k_1 || !k_2)
+      throw cet::exception("RECO") << "mu2e::TrackMatching::" << __func__ << ": Null input track seeds!";
+
+    // Retrieve the hit lists
+    const auto hits_1 = k_1->hits();
```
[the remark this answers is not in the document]
michaelmackenzie: Good catch, fixed!

TrkDiag/src/TrackMatching_module.cc
```diff
@@ -0,0 +1,304 @@
...
+  private:
+    void produce(art::Event& event) override;
+    bool match(const KalSeed* k_1, const KalSeed* k_2);
+    void add_tracks(std::vector<std::vector<const KalSeed*>>& clusters,
+                    std::map<const KalSeed*, size_t>& cluster_index,
+                    const KalSeed* itrkptr, const KalSeed* jtrkptr);
+    void create_clusters(std::vector<const KalSeedCollection*>& trackColls,
+                         std::vector<art::ValidHandle<KalSeedCollection>>& handles,
+                         std::map<const KalSeed*, KalSeedPtr>& ptr_map,
+                         std::vector<std::vector<const KalSeed*>>& clusters,
```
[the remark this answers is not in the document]
michaelmackenzie: I've significantly cleaned this up with your suggestions.

TrkDiag/src/TrackMatching_module.cc
```diff
@@ -0,0 +1,304 @@
...
+          if(_debug > 2) printf("    Checking against track collection %s (size = %zu)\n", _seedTags[j].encode().c_str(), n_jtrks);
+
+          // Check against each track
+          for(size_t j_index = 0; j_index < n_jtrks; ++j_index) {
+            if(_debug > 3) printf("      Checking track %zu of collection %s\n", j_index, _seedTags[j].encode().c_str());
+            const auto& jtrk = jtrks->at(j_index);
+            const auto* jtrkptr = &jtrk; // pointer to the track
+
+            // Check if both tracks have already been clustered together by another track
+            const bool i_clustered = cluster_index.count(itrkptr);
```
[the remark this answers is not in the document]
michaelmackenzie: This has been deleted.

RecoDataProducts/inc/KalSeed.hh
```diff
@@ -96,5 +96,6 @@ namespace mu2e {
   typedef std::vector<mu2e::KalSeed> KalSeedCollection;
   typedef art::Ptr<mu2e::KalSeed> KalSeedPtr;
   typedef std::vector<mu2e::KalSeedPtr> KalSeedPtrCollection;
+  typedef std::vector<KalSeedPtrCollection> KalSeedPtrCollections;
```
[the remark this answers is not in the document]
michaelmackenzie: I think this is a good idea, I've made this update.

TrkDiag/src/TrackMatching_module.cc
michaelmackenzie: This is definitely a cleaner logic, I've cleaned up a lot of the code by making this simplification. I've also added a test that N(in tracks) == N(out tracks), but I haven't yet added a test that every input KalSeedPtr is in the end clustered. Should I add this test as well?

TrkDiag/src/TrackResolution_module.cc
michaelmackenzie: I'm only using indices in the track collection to create KalSeedPtr from the handle and the index and in the track matching loop to check tracks j > i after these updates. In the loop I can switch to an iterators and `jtrk = itrk; ++jtrk` style if this is better/cleaner. Is there a better way to create the KalSeedPtrs? I'm not very familiar with making these.

TrkDiag/src/TrackResolution_module.cc
michaelmackenzie: I was definitely not satisfied with this implementation, but I forgot to clean this up after implementing the matching module. I've cleaned this up with your suggestion, but I'm sure it could use more cleaning.

TrkDiag/src/TrackResolution_module.cc
brownd1978: You need indices to create Ptrs.

TrkDiag/src/TrackMatching_module.cc
```diff
@@ -0,0 +1,235 @@
...
+
+      // Check each following cluster (of size 1) for overlapping tracks
+      for(size_t jtrk = index + 1; jtrk < ntrks; ++jtrk) {
+        KalSeedCluster& cluster_j = clusters[jtrk];
+        if(cluster_j.empty()) continue;
+        if(cluster_j.size() > 1) // input clusters should always be 1 track, as matches are clustered into lower index clusters
+          throw cet::exception("RECO") << "mu2e::TrackMatching::" << __func__ << ": Track cluster matching against has already been clustered! Size = " << cluster_j.size();
+
+        // Get the tracks to compare
+        KalSeedPtr& k_i = cluster.front();
```
brownd1978, line 156: You test only against the 1st KalSeed in the cluster. An alternative is to test against all current cluster members and merge if any match. That's a more aggressive algorithm and in some cases can lead to over-clustering, but I think it would be safe in your case. Neither algorithm is robust against changes in input order: that would require testing all possible pairings and merging the 'best' matches first. That's more complicated to implement and is probably overkill for our case. Just options to consider.

TrkDiag/src/TrackMatching_module.cc
brownd1978: Your test should be adequate.

brownd1978 (approved):
This is good enough for a first implementaiton. I make some suggestions for possible improvements, which may not be necessary.

brownd1978:
@FNALbuild run build test

TrkDiag/src/TrackMatching_module.cc
michaelmackenzie: Each track cluster on input has size 1, and every later cluster only have 1 track so I was thinking it's always 1 size comparisons. I forgot that tracks that get clustered then don't get this comparison loop done as they've been clustered and moved. I've fixed this oversight, where I now keep moving through the track cluster performing the matching as tracks are added to the cluster until all have been checked, then I move to the next non-clustered track.

michaelmackenzie:
@FNALbuild run build test

PR #1487, opened 2025-05-09
STM Shielding House Geometry Updates

caohc:
The shielding house of the STM detector is updated matching the engineer drawing.

### DIFF
--- a/GeometryService/src/STMMaker.cc
+++ b/GeometryService/src/STMMaker.cc
@@ -354,9 +354,44 @@ namespace mu2e {
                                _STM_SSCoffset_Spot,
                                _STM_SSCleak,
                                _STM_SSCFrontToWall,
+                               _STM_SSCZGap,
+                               _STM_SSCZGapBack,
                                _STM_SSCOffsetInMu2e,
                                _STM_SSCRotation,
                                _STM_SSCMaterial));
+
+   ////////////////////////////////////////////////////////////////
+   //Spot-Size Collimator Support
+
+    const CLHEP::HepRotation _SSCSupportRotation     = CLHEP::HepRotation::IDENTITY;
+    const CLHEP::Hep3Vector  _SSCSupportOffsetInMu2e = _STMShieldingRef + CLHEP::Hep3Vector(0, 0, _STM_SSCZGap/2);
+
+          stm._pSSCSupportParams = std::unique_ptr<SSCSupport>
+                  (new SSCSupport(_SSCSupportBuild,
+                               _SSCSupporttable_L,
+                               _SSCSupporttable_H,
+                               _SSCSupporttable_T,
+                               _SSCSupportleg_L,
+                               _SSCSupportleg_H,
+                               _SSCSupportleg_T,
+                               _SSCSupportbase_L,
+                               _SSCSupportbase_H,
+                               _SSCSupportbase_T,
+                               _SSCSupportwall_L,
+                               _SSCSupportwall_H,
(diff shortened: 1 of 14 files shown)


GeometryService/src/STMMaker.cc
```diff
@@ -349,10 +349,45 @@ namespace mu2e {
...
                                _STM_SSCOffsetInMu2e,
                                _STM_SSCRotation,
                                _STM_SSCMaterial));
 
+   ////////////////////////////////////////////////////////////////
+   //Spot-Size Collimator Support
+
+    const CLHEP::HepRotation _SSCSupportRotation     = CLHEP::HepRotation::IDENTITY;
+    const CLHEP::Hep3Vector  _SSCSupportOffsetInMu2e = _STMShieldingRef + CLHEP::Hep3Vector(0, 0, _STM_SSCZGap/2);
+
```
bechenard, line 363: A simple data member could be used instead of a unique_ptr in this class, I would consider this possibility in future development work

GeometryService/src/VirtualDetectorMaker.cc
```diff
@@ -635,9 +653,10 @@ namespace mu2e {
       }
 
       if ( c.getBool("vd.STMDet1UpStr.build", false) ) { // VD89, just upstream of LaBr
+        STM_SSC const & pSTM_SSCParams = *stmgh.getSTM_SSCPtr();
         LaBrDetector const & pLaBrDetectorParams = *stmgh.getLaBrDetectorPtr();
         double VD89_offset = 1.0;
-        double xOffset = pLaBrDetectorParams.originInMu2e().x()-stmDnStrEnvPositionInMu2e.x(); // Add collimator hole and offset shifts
+        double xOffset = -3904 + pSTM_SSCParams.offset_Spot() - stmDnStrEnvPositionInMu2e.x();
```
bechenard, line 659: It would be better to avoid hard-coding this number (-3904) and use "_config.getDouble("mu2e.solenoidOffset");" from the geometry files to access this value.

STMGeom/inc/SSCSupport.hh
```diff
@@ -0,0 +1,112 @@
...
+      _wall_L(wall_L), _wall_H(wall_H), _wall_T(wall_T),
+      _hole_H(hole_H), _hole_T(hole_T),
+      _FLeadStand_L(FLeadStand_L), _FLeadStand_H(FLeadStand_H), _FLeadStand_T(FLeadStand_T),
+      _FLeadShim_H(FLeadShim_H), _FLeadShim_T(FLeadShim_T),
+      _FAluminumShim_T(FAluminumShim_T), _FAluminumExtra_L(FAluminumExtra_L), _FAluminumExtra_H(FAluminumExtra_H),
+      _originInMu2e(originInMu2e),
+      _rotation(rotation)
+    {
+    }
+
```
bechenard, line 44: This class only contains plain old data types (POD). Consider using a struct in future work, this will simplify the coding

bechenard (commented):
Nice work, I only have one comments that needs to be addressed (use geometry accessor to avoid hard-coded number) and a few suggestions for future work.

sdifalco (approved):
I have rebuilt the executable with this PR and found no errors or overlaps. The geometry shown by TGeoManager looks correct.

caohc:
@bechenard I added a new commit, which I used geometry accessor to avoid hard-coded number.

caohc:
@FNALbuild run build test

PR #1490, opened 2025-05-12
Update CalCosmicEnergyCalib.hh table length

giro94:
Changed table length to only include crystal channels


giro94:
@rlcee

brownd1978 (approved):

rlcee (approved):
Sorry for the delay - I always miss when a PR is assigned me..

simonagiov (approved):

PR #1492, opened 2025-05-12
p078

rlcee:
advance kinkal from 3.1.7 to 3.2.0 
advance adcm from v5_00_00 to v6_00_00


brownd1978:
@FNALbuild run build test

PR #1497, opened 2025-05-14
Update track clustering to properly check each track combination

michaelmackenzie:
Fix an oversight where clustered tracks weren't check for additional overlaps.


brownd1978 (commented):
Something I missed in your first PR: this module is a producer, so it doesn't belong in TrkDiag. A better location is TrkReco.

michaelmackenzie:
I forgot I was going to ask the best place to put this, I'll put in a PR today to move this to TrkReco.

PR #1498, opened 2025-05-15
Add CID to CalCosmicEnergyCalibInfo

giro94:
Commit ID added as first column.
This number helps matching the info content with the calibration constants


rlcee (approved):
looks good, thanks

PR #1504, opened 2025-05-20
Optional skip of DTC-level headers in tracker fragments

edcallaghan:
PR #1402 implemented efficiency deserialization of artdaq fragments for the tracker. It has since been realized that the data produced by the current tracker daq does not include a DTC-level header which should be present to maintain alignment of the data structure whether produced using hardware- or software-event-building. The data produced by the tracker daq will soon transition to including this header. This PR introduces a configuration flag for the absence of the header, which defaults to false; future data will decode properly by default, and previous datasets may be processed by manually overriding the flag as true, indicating that the relevant headers are not present.

### DIFF
--- a/DAQ/fcl/prolog.fcl
+++ b/DAQ/fcl/prolog.fcl
@@ -6,7 +6,7 @@ DAQ: {
             module_type: StrawDigisFromArtdaqFragments
             diagLevel: 0
             saveWaveforms: false
-            containsDTCHeaders: true
+            missingDTCHeaders: false
         }
     }
 }
--- a/DAQ/fcl/prolog.fcl
+++ b/DAQ/fcl/prolog.fcl
@@ -1,10 +1,12 @@
 # -*- mode:tcl -*-
 BEGIN_PROLOG
-DAQ : {
-    producers : {
-        StrawDigisFromArtdaqFragments: { module_type: StrawDigisFromArtdaqFragments
-            saveWaveforms : 0
-            diagLevel     : 0  # not used yet
+DAQ: {
+    producers: {
+        StrawDigisFromArtdaqFragments: {
+            module_type: StrawDigisFromArtdaqFragments
+            diagLevel: 0
+            saveWaveforms: false
+            containsDTCHeaders: true
         }
     }
 }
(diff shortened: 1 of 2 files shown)


edcallaghan:
I've requested review from @gianipez and @michaelmackenzie as they were the first to encounter / work around this issue.

michaelmackenzie:
Hi Ed,

Thank you for making this update! Is there a more proper way to handle this by checking the fragment types to retrieve the relevant tracker subevent data (this is probably and @eflumerf  question)? And is there any issue associated with the following casting when the header is there:
```
        DTCLib::DTC_SubEventHeader* seh = (DTCLib::DTC_SubEventHeader*) fdata;
        if ((seh->link0_subsystem != DTCLib::DTC_Subsystem::DTC_Subsystem_Tracker) and
            (seh->link1_subsystem != DTCLib::DTC_Subsystem::DTC_Subsystem_Tracker) and
            (seh->link2_subsystem != DTCLib::DTC_Subsystem::DTC_Subsystem_Tracker) and
            (seh->link3_subsystem != DTCLib::DTC_Subsystem::DTC_Subsystem_Tracker) and
            (seh->link4_subsystem != DTCLib::DTC_Subsystem::DTC_Subsystem_Tracker) and
            (seh->link5_subsystem != DTCLib::DTC_Subsystem::DTC_Subsystem_Tracker)     )
                                                            continue;
        int dtc_id = seh->source_dtc_id;
```
And just to check, have you verified this works as expected on data with the DTC_Event header included?

edcallaghan:
@michaelmackenzie good catch --- I would guess that the SubEventHeader needs to be offset as well (i.e. the EventHeader, precedes the SubEventHeader precedes the payload), but Eric should say if that's wrong.

AFAIK the fragment type is just assigned when the fragment is instantiated, but it is not e.g. coupled to specific constructors (at least not that are being used to produce tracker VST data), so it's value would not correlate with the presence/absence of this header. But, again, I should be corrected if that's off.

michaelmackenzie (approved):
Looks good, thank you!

brownd1978:
@FNALbuild run build test

PR #1506, opened 2025-05-24
DetStepFilter fix

brownd1978:
Fix an old bug that fortunately hasn't been exposed.


brownd1978:
@FNALbuild run build test with Mu2e/mu2e-trig-config#85

PR #1509, opened 2025-06-01
New updates

caohc:


STMGeom/inc/BackShielding.hh
```diff
@@ -30,7 +30,7 @@ namespace mu2e {
    double   BPHeight()                 const {return _BPHeight;}
    double   Back_dX()                  const {return _Back_dX;}
    double   Back_dY()                  const {return _Back_dY;}
-
+   double   PipeGap()                  const {return _PipeGap;}
```
AndrewEdmonds11, line 33: Can we find a more descriptive name for this? I thought this was for the shield pipe (close to the CRV) for a second but this is for the downstream shield house.

AndrewEdmonds11 (commented):
Thanks, Haichuan. This looks good to me. Just one request to find a more descriptive name since we have at least one other "pipe" in the STM

caohc:
@FNALbuild run build tests

AndrewEdmonds11 (approved):

PR #1510, opened 2025-06-02
Energy Deposition Threshold Flag

nikita-mazotov:
Flag for triplet seeding at helix finding level. Prevents hits with high energy from being used for initial triplet, but they may still be added onto helix in subsequent steps. This significantly improves helix finder search time without noticeably diminishing efficiency.

### DIFF
--- a/CalPatRec/src/AgnosticHelixFinderDiag_tool.cc
+++ b/CalPatRec/src/AgnosticHelixFinderDiag_tool.cc
@@ -22,8 +22,7 @@ namespace mu2e {
            kNTimeClusterHistsSets = 1,
            kNHelixSeedHistsSets = 1,
            kNLineInfoHistsSets = 1,
-           kNLineSegmentHistsSets = 1
-    };
+           kNLineSegmentHistsSets = 1 };
 
     struct EventHists {
       TH1F* nHelices;
--- a/CalPatRec/src/AgnosticHelixFinder_module.cc
+++ b/CalPatRec/src/AgnosticHelixFinder_module.cc
@@ -426,7 +426,6 @@ namespace mu2e {
       _diagInfo.timeClusterData.clear();
       _diagInfo.helixSeedData.clear();
       _diagInfo.lineSegmentData.clear();
-
     }
 
     // flag whether event is intense or not
@@ -632,7 +631,6 @@ namespace mu2e {
       return _chColl->at(a.hitIndice).pos().z() > _chColl->at(b.hitIndice).pos().z();
     });
   }
-
 
   //-----------------------------------------------------------------------------
   // logic for setting certain flags on hits


nikita-mazotov:
Added line for fillAllHitsData() and edited the description of pull request.

michaelmackenzie (approved):
This looks great, thank you!

michaelmackenzie:
@dnbrow01 I believe this PR is ready for merging, but please let us know if you have any recommendations before this. The default behavior is unchanged, but we intend to turn this triplet seed selection on for the trigger processing with the associated https://github.com/Mu2e/mu2e-trig-config/pull/86 PR

CalPatRec/inc/AgnosticHelixFinder_types.hh
```diff
@@ -11,6 +11,11 @@ namespace mu2e {
       fhicl::Atom<std::string> tool_type{fhicl::Name("tool_type"), fhicl::Comment("tool type: AgnosticHelixFinderDiag")};
     };
 
+    struct hitInfo {
```
brownd1978, line 14: What is the value of a struct containing a single float?

CalPatRec/inc/AgnosticHelixFinder_types.hh
```diff
@@ -37,6 +42,7 @@ namespace mu2e {
       std::vector<hsInfo>            helixSeedData;
```
brownd1978, line 42: The design pattern where several vectors are assumed to be in-sync is error-prone. A stronger design is to have single struct with multiple fields; ie tcInfo contains hsInfo, LineSegmentInfo, etc, and then create a vector of those.

brownd1978 (commented):
I don't have strong objections to this PR, but improvements are possible, as noted.
My main question goes back to the earlier discussion of generalizing the edep StrawHitFlag bit, to avoid re-implementing this cut in multiple places (like the cosmic track reco). It seems you've given up on that.

matthewstortini:
> I don't have strong objections to this PR, but improvements are possible, as noted. My main question goes back to the earlier discussion of generalizing the edep StrawHitFlag bit, to avoid re-implementing this cut in multiple places (like the cosmic track reco). It seems you've given up on that.

@brownd1978 I think there is some confusion relating this PR to the PR I recently made for Pasha. That PR was trying to limit noise in the VST for Pasha. This PR we are making now is unrelated. This is not to limit noise. This is adding a eDep threshold for hits to be used during the triplet stage of the helix finder since, as of now, hits that more consistent with protons than e- are being used to make triplets, and it is slowing the module down. Nikita showed that adding this cut in the trigger improved the timing by ~15% with zero effect on the efficiency. He found that putting this eDep cut for the points to seed the helix search (i.e. the triplets) got rid many of the proton hits, while keeping enough of the CE hits to still have good search seeds. This is very algorithm specific. I don't see anywhere else where this value would want to be used.

** Also let me note that this edep threshold is only applied to the triplets considered when created a seed for the helix search. The hits above the threshold can still be added to the helix post-seed.

CalPatRec/inc/AgnosticHelixFinder_types.hh
matthewstortini: I am not sure what you mean by "in-sync" here. The vectors are going to be different lengths. It is never assumed in the diag tool (where these vectors are used) that the vectors are the same length / aligned.

brownd1978: If these vectors are not intended to have the same length then my comment is irrelevant.

brownd1978:
I believe the comment is relevant, as the edep bit in the ComboHit flag is used for removing proton hits in other algorithms. Having APR define its own proton hit filter creates multiple points of maintenance. Is there a reason APR can't use the edep bit that's in the ComboHit StrawHitFlag?

matthewstortini:
> I believe the comment is relevant, as the edep bit in the ComboHit flag is used for removing proton hits in other algorithms. Having APR define its own proton hit filter creates multiple points of maintenance. Is there a reason APR can't use the edep bit that's in the ComboHit StrawHitFlag?

Fair point. But I think the value we are using in the trigger could be a bit lower than I would be comfortable using for an _official _edep flag. What we are flagging in AgnosticHelixFinder here are not "likely protons". They are "not very likely electrons". We are pushing it down quite low, significantly cutting into to CE tail. Every function in AgnosticHelixFinder is individually fast. The thing that slows is down the sheer number of triplet seeds. We've found you can cut out a lot of the CE eDep tail and still get a good seed though. But I wouldn't be comfortable with this low value we are setting being used for an official eDep flag.

PR #1515, opened 2025-06-13
Add safeguard to compression for CRV dark counts in CrvCoincClusterMCs

AndrewEdmonds11:
@oksuzian was seeing exceptions when going through compression at the reco stage. @ehrlich-uva helped us solve the problem. There are some pulses in CrvCoincidenceClusterMCs that come from dark counts that have an MCDigi but not a SimParticle. This PR adds a check for these


oksuzian (approved):
Thank you for fixing this!

PR #1516, opened 2025-06-16
Fixes for tracker station reco

bonventre:
Adds options to run selectreco and kkline without calorimeter or crv objects
Adds option to require hit passing esel in time cluster
Bug fix in strawresponse for time offsets


bonventre:
Requires mu2e-trig-config#91

PR #1517, opened 2025-06-17
p080 for adcm v8

rlcee:


rlcee:
@FNALbuild run build test with #1512

matthewstortini:
@FNALbuild run build test with https://github.com/Mu2e/Offline/pull/1512

matthewstortini:
> I added you as reviewer, but don't you have to retag adcm, then I will make p081?  I will probably close this and make a new PR, but I guess I could reuse this one.

I just deleted v8_00_00, made a commit, and re-tagged. My logic was that 1) there is really no motivation to keep the old unpatched v8 in the history and 2) it would allow me to just keep this PR open and re-trigger a build test overnight without hassling you. I of course made this decision not realizing I would need to be made a reviewer to trigger a build test, though this requirement seems appropriate.

matthewstortini:
Ray, we now have v8_00_01 and the old tag (v8_00_00) has been reverted with the correct commit hash so as not to cause issues with your spack recipes. We can now close this and open a new PR with p081 pointing to ADCM v8_00_01.

PR #1520, opened 2025-06-24
STM Upstream Geometry Changes

AndrewEdmonds11:
This PR makes changes to the STM upstream geometry to more closely match what we expect. It:
* removes the magnet geometry (the magnetic field was already turned off)
* makes the shield pipe longer
* adds the air gaps between the shield pipe and the end cap shielding, and the shield pipe and the FOV collimator

It does not:
* replace the table with the trolley

There are before and after pictures in DocDB-53208.

This PR should probably be merged after Haichuan's #1509. We touch some of the same files but I think any merge conflicts would be easy to resolve.

### DIFF
--- a/GeometryService/inc/STMMaker.hh
+++ b/GeometryService/inc/STMMaker.hh
@@ -196,6 +196,7 @@ namespace mu2e {
     double       _shieldDnStrWallGap;
     double       _shieldUpStrWallGap;
     std::string  _shieldDnStrWallMaterial;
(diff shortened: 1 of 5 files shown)


AndrewEdmonds11:
Converted to draft because there are couple more changes still to make

AndrewEdmonds11:
@FNALbuild run build test

caohc (approved):

caohc (approved):

PR #1528, opened 2025-07-04
Adding missing pointers to db table

giro94:


rlcee (commented):
ptr looks good.  The length change is fine at this point (no real db tables exist, so no backwards compatibility considerations) but I noticed the check [here](https://github.com/Mu2e/Offline/blob/abfb7de4f202f3c9c479c713575bcbf8d9ac17db/CaloConditions/src/CalCalibMaker.cc#L51) should probably be removed.  When the table arrives here, it will already be checked against nRowFix

giro94:
Adjusted both energy and time calib to 2656 rows. (no pin diodes). I kept that check.
These tables are indexed by CaloSiPMId, which goes from 0 to 2655 + 9999 for pin and empty channels.
Pin diodes will probably have an id different from 9999, but they are still of different nature.

brownd1978 (approved):

PR #1529, opened 2025-07-04
adding functionality to generate flat muons

matthewstortini:
preparing to make FCLs for Yuri to generate a flat mu- dataset (will make corresponding PR in Production) -- will be used for PID training

### DIFF
--- a/EventGenerator/src/FlatGeneratorFromStoppedMuon_module.cc
+++ b/EventGenerator/src/FlatGeneratorFromStoppedMuon_module.cc
@@ -100,6 +100,7 @@ namespace mu2e {
     else if (pid == PDGCode::e_plus) { process = ProcessCode::mu2eFlatePlus; }
     else if (pid == PDGCode::gamma) { process = ProcessCode::mu2eFlatPhoton; }
     else if (pid == PDGCode::mu_minus) { process = ProcessCode::mu2eFlatMuMinus; }
+    else if (pid == PDGCode::mu_plus) { process = ProcessCode::mu2eFlatMuPlus; }
     else {
       throw   cet::exception("BADINPUT")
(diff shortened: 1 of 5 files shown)


brownd1978 (commented):
Is this to model stopped pion decay? Since 'flat' isn't a physical spectrum it probably doesn't matter.
We should rename the generator to avoid confusion since the intent of this module has evolved. I suggest FlatStoppedParticleDaughterGenerator. And remove references to muons being the parent, as that isn't required.

matthewstortini:
This is just so we can generate a flat muon sample for the sake of training a PID model. I agree with changing the name of the module. Let me make a commit to update the PR.

michaelmackenzie:
Hi @brownd1978, I just wanted to clarify this isn't connected to my RPC updates. This is for generating flat muons from muon stops, so the electron vs. muon PID training isn't biased by the kinematics of the muon production and is only learning to separate electron from muon tracks.

EventGenerator/src/FlatGeneratorFromStoppedMuon_module.cc
```diff
@@ -102,21 +102,21 @@ namespace mu2e {
     else if (pid == PDGCode::mu_minus) { process = ProcessCode::mu2eFlatMuMinus; }
```
michaelmackenzie, line 102: Should we also allow for mu+?

michaelmackenzie (approved):
I think this looks good, I added one comment but it's just a suggestion.

michaelmackenzie:
@FNALbuild run build test with https://github.com/Mu2e/Production/pull/439

brownd1978:
@FNALbuild run build test with Mu2e/Production#439

brownd1978 (approved):

PR #1530, opened 2025-07-10
Beam dump concrete update

gaponenko:
The opening in the poured concrete is larger than what we had in the Offline model.   This PR resolves the difference.  The effect is a reduction in the amount of the concrete shielding.


brownd1978 (approved):
Thanks for the descriptive comment.

PR #1531, opened 2025-07-13
Address small inconsistency in how closestapproach is used

brownd1978:
This makes sure the updaters use drift information calculated from the same ClosestApproach value as is used by the ANNs.


bonventre:
Is this just setup for a future change or is something missing? Looks like fillDriftInfo would be called with the same ca as before in all cases here

brownd1978:
Previously FillDriftInfo was being called with the cached CA from the previous iteration. Now it's being called with the current (this iteration) CA.

bonventre (approved):

PR #1535, opened 2025-07-18
Scoring

bechenard:
Code to add internal Geant4 scoring capabilities with different scorer for measuring the effective dose. Support for variance reduction radioactive decay is also included


MCDataProducts/inc/ProcessCode.hh
```diff
@@ -88,9 +88,9 @@ namespace mu2e {
       mu2eExternalRMC,         mu2eFlateMinus,      mu2eFlatePlus, mu2eFlatPhoton, // 175
       mu2eCePlusLeadingLog, mu2ePionCaptureAtRest, mu2eExternalRPC, mu2eInternalRPC, // 179
       mu2eCaloCalib, mu2ePienu, mu2eunused7, mu2eunused8, // 183
-      uninitialized, NoProcess, GammaGeneralProc, // 186
-      mu2eGammaConversion, Radioactivation, nCaptureHP, nFissionHP, mu2eAntiproton, // 191
-      mu2eFlatMuMinus, mu2eFlatMuPlus, // 193
+      uninitialized, NoProcess, GammaGeneralProc, mu2eGammaConversion, // 187
+      Radioactivation, nCaptureHP, nFissionHP, mu2eAntiproton, // 191
+      mu2eFlatMuMinus, mu2eFlatMuPlus, CoupledTransportation, // 194
```
brownd1978, line 93: This block of codes typically is for mu2e physics processes. What is CoupledTransportation? Is it mu2e specific?
Instead of extending the process code list please overwrite one of the existing ' unused' or mu2eunused' codes.

brownd1978 (changes requested):
Please consider renaming the new process 'mu2eCoupledTransport' (if it's mu2e specific). Please reuse an existing unused process number (we believe in recycling!)

MCDataProducts/inc/ProcessCode.hh
bechenard: CoupleTransportation is used for tracking in parallel (scoring) world, it is a generic Geant4 code. I'll admit I didn't read the full list of codes, I'll try to find an unused slot (ah, California's obsession with recycling....)

bechenard:
@FNALbuild run build test

Mu2eG4/src/Mu2eG4_module.cc
```diff
@@ -322,6 +328,7 @@ namespace mu2e {
 #endif
     _runManager->SetUserInitialization(physicsList_);
 
+    _scorer->initialize();
```
resnegfk, line 331: Given the potential impact on CPU and memory, could you make the use of the scorer conditional and controlled by a fhicl parameter in the prolog? Both here in the sequential mode and in the MT one?

brownd1978 (approved):

Mu2eG4/src/Mu2eG4_module.cc
bechenard: There is a parameter in fhicl that controls the instantiation of the scoring manager (off by default). The Mu2eG4ScoringManager is only initialized if this turned on (see Mu2eG4ScoringManager::initialize)

resnegfk: ok

PR #1536, opened 2025-07-24
Introduce new top-level geometry configs for run 1

brownd1978:
geom_run1.txt is for the as-built geometry we expect assuming the full detector and all shielding are installed. geom_run1a.txt is for the reduced shielding/CRV scenario. Note there's no need for geom_run1b.txt: that's the same as geom_run1a.txt, just with a different BField file.


Mu2eG4/src/ConstructMaterials.cc
```diff
@@ -605,6 +604,26 @@ namespace mu2e {
...
+    {
+      G4Material* Al6061 = new G4Material(mat.name, 2.70*CLHEP::g/CLHEP::cm3, 4);
+      Al6061->AddMaterial(findMaterialOrThrow("G4_Al"), 0.980);
+      Al6061->AddMaterial(findMaterialOrThrow("G4_Mg"), 0.010);
+      Al6061->AddMaterial(findMaterialOrThrow("G4_Si"), 0.006);
+      Al6061->AddMaterial(findMaterialOrThrow("G4_Fe"), 0.004);
+    }
+
+    // Aluminium alloy 7075
+    mat = uniqueMaterialOrThrow("A7075");
```
bonventre, line 618: Should be Al7075? (same for Al6061)

bonventre (changes requested):

Mu2eG4/src/ConstructMaterials.cc
bonventre: I also get an error about components sum to 1 here

brownd1978: good catch. I had only tested the reco side. Fixed with the latest push.

brownd1978:
@FNALbuild run build test

brownd1978:
@FNALbuild run build test

brownd1978:
@FNALbuild run build test

PR #1541, opened 2025-07-24
Fix panels and planes in alignedTracker

bonventre:
Aligned tracker had an issue where only the straws were updated, but the planes and panels still contained pointers to nominal straws. This impacted straw xing finding in kk fits when misalignments were large.

### DIFF
--- a/Mu2eKinKal/inc/KKFit.hh
+++ b/Mu2eKinKal/inc/KKFit.hh
@@ -128,7 +128,7 @@ namespace mu2e {
       double tprec_; // TPOCA calculation nominal precision
       StrawHitFlag addsel_, addrej_; // selection and rejection flags when adding hits
       // parameters controlling adding hits
-      float maxStrawHitDoca_, maxStrawHitDt_, maxStrawDoca_, maxStrawDocaCon_;
+      float maxStrawHitDoca_, maxStrawHitDt_, maxStrawDoca_, maxStrawDocaCon_, maxStrawUpos_;
       int maxDStraw_; // maximum distance from the track a strawhit can be to consider it for adding.
       // cached info computed from the tracker, used in hit adding; these must be lazy-evaluated as the tracker doesn't exist on construction
       mutable double strawradius_;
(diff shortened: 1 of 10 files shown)


TrackerGeom/src/Panel.cc
```diff
@@ -50,13 +50,13 @@ namespace mu2e {
...
     auto vdir = DStoUVW.rotation()*_vdir;
     auto wdir = DStoUVW.rotation()*_wdir;
-    if( fabs(1.0 - udir.dot(Hep3Vector(1.0,0.0,0.0))) > 1e-6 ||
-        fabs(1.0 - vdir.dot(Hep3Vector(0.0,1.0,0.0))) > 1e-6 ||
-        fabs(1.0 - wdir.dot(Hep3Vector(0.0,0.0,1.0))) > 1e-6 )
-      throw cet::exception("Geom") << "Panel direction error: id " << _id << " udir " << udir << " vdir " << vdir << " wdir " << wdir << std::endl;
-    auto po = DStoUVW*origin;
-    if( fabs(po.r()) > 1e-6 )
-      throw cet::exception("Geom") << "Panel origin error: id " << _id << " po " << po << " transform " << DStoUVW << std::endl;
+//    if( fabs(1.0 - udir.dot(Hep3Vector(1.0,0.0,0.0))) > 1e-6 ||
```
brownd1978, line 53: Are these tests no longer valid? Or do they need a future update?

Mu2eKinKal/inc/KKFit.hh
```diff
@@ -356,7 +356,7 @@ namespace mu2e {
                     double du = (pca.sensorPoca().Vect()-smid).R();
                     double doca = fabs(pca.doca());
                     double dsig = std::max(0.0,doca-strawradius_)/sqrt(pca.docaVar());
-                    if(doca < maxStrawDoca_ && dsig < maxStrawDocaCon_ && du < straw.halfLength()){
+                    if(doca < maxStrawDoca_ && dsig < maxStrawDocaCon_ && du < straw.halfLength() * maxStrawUpos_){
```
brownd1978, line 359: So the test is relative to the straw length? Is that better than adding a fixed buffer?
What are you using for maxStrawUpos_ in VST reco?

brownd1978 (commented):
I'm guessing the TrackerGeom changes fit in with corresponding changes in TrackerAlignment. The KKFit changes look fine. Do we want to introduce a VST.epilog to set values as needed for reconstructing VST data? That way production could invoke it without having to know details.

TrackerGeom/src/Panel.cc
bonventre: its not valid for alignment not built from straws. I moved to a separate constructor for aligned version

PR #1545, opened 2025-07-26
start ParseCLI command interpreter

rlcee:
This utility class provides a simple interface to create and interpret command lines like:
`command <global options> subcommand <subcommand options>`
I have several tools which use this format and I've found the getopts and boost::program_options are not well-suited to this style, which they are not designed for.  I haven't been able to find a good alternative. This core code is pretty simple so I'd like to switch to this class to simplify and unify the tools.  For example, dqmTools is getting out of hand, with several times more lines of code than this class would require.

### DIFF
--- a/GeneralUtilities/CMakeLists.txt
+++ b/GeneralUtilities/CMakeLists.txt
@@ -13,6 +13,7 @@ cet_make_library(
       src/OrientationResolver.cc
       src/ParameterSetFromFile.cc
       src/ParameterSetHelpers.cc
+      src/ParseCLI.cc
       src/PathnameWithNextVersion.cc
       src/RMS.cc
       src/RootNameTitleHelper.cc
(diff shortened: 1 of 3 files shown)


brownd1978 (commented):
Since the intent of this new tool is to unify others, I'd like to get feedback from the concerned developers. What is the domain for this tool? Is it just pass1? or do you see it expanding into grid job control, poms, ...?

rlcee:
If you look into dbTool, runTool, epicsTool, grlTool, openTool, dqmTool, and valCompare, they all have this command pattern, and all handle it differently according to what I could get away with, since I was lacking a good system.  For example dbTool is ad hoc code, grlTool can't re-use a switch in different subcommands, etc. Currently, I am holding off committing dqmTool because it has gotten so out of hand using boost::po (I can reduce this code section x5 or more).  I'd like to simplify and unify these tools.  I don't expect this to have anything in particular to do with pass1.  I do expect to need maybe one more tool (luminosity database), but that's all I foresee. Based on experience, I don't expect anyone else will use this code, though they are certainly welcome to.

brownd1978 (approved):

Issue #1544, opened 2025-07-25
KKStrawXing computes material interactions using the wire position not straw


brownd1978:
When a straw is associated with a hit, its material effect (KKStrawXing) is computed using the closest approach from that hit. That CA is relative to the wire, which in general is not the same as the center of the straw. The KKStrawXing CA should be changed to use the straw axis, with the hit CA as hint.

brownd1978:
This is resolved in PR #1571

PR #1547, opened 2025-07-28
Use straw axis to compute straw material intersections

brownd1978:
Since straws and wires aren't co-linear, the material effects from the straw can't use the POCA computed from the wire. This PR adds computation of POCA from the straw central axis, assuming it's straight. This addresses issue #1544


bonventre (approved):
Looks like this has to go after #1536 since it has those commits as well

brownd1978:
@FNALbuild run build test

PR #1548, opened 2025-07-30
Add THMu2eCaloDisk library

giro94:
This ROOT-like library allows the creation of calo disk plots.
Bins contain full crystal and channel information.
Values displayed can be single, sum, difference, average, asymmetry of left/right channels.
The type of content displayed can be chosen even after filling the hist.
Calo map is currently read from file. Will transition to online or offline database.

This PR depends on https://github.com/Mu2e/Offline/pull/1533


brownd1978:
no problem about 1533: it will close automatically once this PR is merged.

rlcee:
Sorry to raise this question after you made the PR - how do you plan to supply values to the display?  Do you want to embed this  in an art module?  Or read from an EventNtuple?  Or is it just for displaying calibration data, and therefore driven only by text or database?  Do you plan to load the library at the root command line?  Thanks

giro94:
Hi @rlcee , the Draw function displays the original TH2Poly->fContent value, so no change to the Draw() function is needed. The trick is that when you FillOffline(channel,value) you increase the fContentL and fContentR values. Then, the automatic Update() function will compute the displayed fContent depending on the fCombineMode (left, right, sum, difference, average, asymmetry). The SetCombineMode() function will update the values too. I find this very powerful for our purposes.

This library is already fully functioning in art modules, root files, and root macros (already tested them all!).
It is designed to be a general purpose tool, the same way that a TH1 is.
We are going to use it for calibration, for DQM, and for physics.

rlcee:
Typically a TH1 in an art module would be run in the background and not displayed.  Did you run this in an art module and display, and step though event by event?  Do you run TApplication?  Or when you display, you are at the root prompt?  Do you load the library or do some kind of discovery via a library path?  Can you send an example?  Thanks
I'd also like to find a way remove the LinkDef, since we don't have that anywhere else in Offline.  It is just better to not have one-off techniques.  I can try to figure that out, but we can proceed if I can't..  Also I want to make scons work, but I understand you don't use that.

giro94:
@rlcee I use THMu2eCaloDisk in the same way I use any other TH1 or TH2.
For example, my art analyzer does this:
```
	h1_means     = tfs->make<TH1D>("h1_means", "All channels baselines;ADC", int(0.5 * (xmax - xmin)), xmin, xmax);
	g_fitsigmas = tfs->makeAndRegister<TGraph>("g_fitsigmas", "Gaussian fit sigmas;BoardID*100 + ChannelID;ADC");
	h2_disk0 = tfs->makeAndRegister<mu2e::THMu2eCaloDisk>("h2_disk0", "Disk 0", "h2_disk0", "Disk 0", 0);
```
About that last line: I don't understand why the classic `tfs->make(name,title,args)` doesn't write the object to the file. So I need the `tfs->makeAndRegister` and repeat name and title twice. Maybe I'm missing something somewhere.

When I open the output root file, no need to load the THMu2eCaloDisk library if I'm in the same mu2e environment.
Otherwise, I simply `#include "Offline/CaloVisualizer/THMu2eCaloDisk.h"`.

giro94:
Example of a simple macro to create and show a disk:

```
#include "srcs/Offline/CaloVisualizer/THMu2eCaloDisk.h"

void mymacro(){

    mu2e::THMu2eCaloDisk* disk0 = new mu2e::THMu2eCaloDisk("disk0","Disk 0",0);

    while (...){
        disk0->FillOffline(offlineId,calib);
    }
    
    disk0->SetCombineMode(mu2e::ECombineMode::kAverage);
    disk0->Draw("colzl");
}
```

giro94:
Example of opening a root file:

```
source setup_ots.sh //Setup environment
root -l myfile.root
[0] .ls
 KEY: mu2e::THMu2eCaloDisk	h2_disk0;1	Disk 0
 KEY: mu2e::THMu2eCaloDisk	h2_disk1;1	Disk 1
[1] h2_disk1->Draw("colzl")
```

giro94:
@FNALbuild build

giro94:
@FNALbuild build

giro94:
@FNALbuild build

giro94:
Ok, all done.

rlcee:
Thank you for the examples.  Here are a few things I'd like to ask for to bring it into alignment with the Offline conventions  
- the dictionaries can be made like the [existing examples](https://github.com/Mu2e/Offline/blob/main/DataProducts/CMakeLists.txt) I think this is the same functionality and would not introduce a new build pattern
- I'd prefer if the code were put in src and inc subdirectories, like other the other subdirectories
- I can see the case for a ".cxx" source file but this again would be the first example of cxx file in Offline, so it is just a little one-off
- I don't see a SConscript, so muse builds (and CI) will skip this subdir.  
I can help with any of these, if you'd like

CaloVisualizer/THMu2eCaloDisk.cxx
```diff
@@ -0,0 +1,391 @@
...
+  fmap.ignore(256, '\n');
+  while (!fmap.eof()) {
+    double temp, x, y;
+    int row, column;
+    int board, channel, cryid, roid;
+    std::string type;
+    fmap >> row >> column >> temp >> temp >> temp >> temp >> temp;
+    fmap >> board >> temp >> temp >> channel;
+    fmap >> temp >> x >> y >> cryid >> roid >> type;
+    if (fmap.eof())
```
bechenard, line 247: Is that needed even if you do the check in the while() condition

CaloVisualizer/THMu2eCaloDisk.cxx
```diff
@@ -0,0 +1,391 @@
...
+  } else {
+    fOverflow[-bin - 1] = content;
+  }
+
+  SetBinContentChanged(kTRUE);
+  fEntries++;
+}
+
+void mu2e::THMu2eCaloDisk::SetBinContentR(Int_t bin, Double_t content) {
+  if (bin > GetNumberOfBins() || bin == 0 || bin < -9)
```
bechenard, line 367: It would be better to have a constant instead of the "-9"

bechenard (approved):
Ray made comments regarding the code structure in Offline so I won;t repeat this. For the rest, the code is very cleanly written, I only have minor comments. I'll approve and let you work the detail to be "Offline compatible" with Ray.

CaloVisualizer/THMu2eCaloDisk.cxx
giro94: The reason is that eof() becomes true after a read operation fails.
The "proper" way would be to do something like:
```
while (fmap >> row >> column >> ...){
  /* code */
}
```
but I don't like it when reading many columns.
I could use `getline()`. But I guess it's matter of personal taste.
I've got used to `if (file.eof()) break;`

CaloVisualizer/THMu2eCaloDisk.cxx
giro94: I would agree, but it's a re-implementation of `TH2Poly::SetBinContent()`:
https://root.cern.ch/doc/master/TH2Poly_8cxx_source.html#l01380
so I thought I would copy 1-1. Th2poly has 9 overflow bins.

giro94:
I've tried `muse build` with a SConscript file but I get the same errors I get with `cet_make_library` and `art_dictionary` in a CMakeLists.txt.

Examples:
```
undefined reference to `vtable for mu2e::THMu2eCaloDiskBin'
undefined reference to `mu2e::THMu2eCaloDisk::Class()'
undefined reference to `mu2e::THMu2eCaloDisk::Streamer(TBuffer&)'
undefined reference to `non-virtual thunk to mu2e::THMu2eCaloDisk::Streamer(TBuffer&)'
```
Any help would be appreciated

giro94:
@FNALbuild build

brownd1978 (approved):

PR #1552, opened 2025-08-02
remove ambiguity in clockwise rotation

sdifalco:
Some volumes, as dirtFoundationExtMonPSareaR, were defined as degenerate trapezoids with only 2 vertices in one of the two  faces to create a wedge shape. This creates an ambiguity in the clockwise orientation of the vertices. Two vertices have been shifted by 10 micron to remove this ambiguity.
In addition the collimator at the end of M4 line has been removed: it was a rough approximation overlapping with the beam line model in MARS


sdifalco:
@FNALbuild run build test

PR #1554, opened 2025-08-04
N(POT) reco and flattened trigger updates

michaelmackenzie:
This is work done by @happlegate1 to add a reconstructed estimate of N(POT) from observables, as well as a filter to select a flattened PBI distribution from data. I moved these to a new subdirectory for this luminosity stream analysis. There are still updates needed, but I believe this is a good point to save the work done so we can continue to build upon it.
Additionally, should I move `ProtonBunchIntensity` to `DataProducts` since it's both an MC value and a reco value? Or to `RecoDataProducts`?

Main updates:
- `RecoNPOTMaker` producer module to reconstruct nPOT from Observables. Currently only reconstructing from Calorimeter Energy.

- `RecoNPOTFilter` filter module to sample even counts from all intensities of the nPOT range. Around 3% of events pass currently, assuming a 2-batch mode intensity profile.
  - We will need to update this to take the input POT distribution assumption as a table or ROOT histogram

- `NPOTAnalysis` plots histograms of Luminosity Stream Observables (1D, nPOTvObservable,nProtonsvObservable). NPOTAnalysis plots recoNPOT, calculates and plots normalized residual ((npot-reconstructed)/npot), plots residuals v NPOT and vObservable. `NPOTAnalysis` also plots the recoNPOT that passed the filter. This was used to derive the observable --> N(POT) functions. 

- Remove flag from `TZClusterFinder` and replace with energy selection, to allow complementary proton time cluster counting as the energy selection flag is currently being set by `DeltaFinder`'s proton counting


michaelmackenzie:
@FNALbuild run build test

michaelmackenzie:
@FNALbuild run build test

matthewstortini (approved):

brownd1978 (approved):

PR #1557, opened 2025-08-06
Add reco version of PBI

michaelmackenzie:
Add a reco version of the PBI, produce this instead of MC data product in reco modules

### DIFF
--- a/Lumi/src/RecoNPOTMaker_module.cc
+++ b/Lumi/src/RecoNPOTMaker_module.cc
@@ -9,7 +9,7 @@
 #include "fhiclcpp/ParameterSet.h"
 
 //Reco DataProducts
-#include "Offline/MCDataProducts/inc/ProtonBunchIntensity.hh"
+#include "Offline/RecoDataProducts/inc/RecoProtonBunchIntensity.hh"
 #include "Offline/RecoDataProducts/inc/IntensityInfoTimeCluster.hh"
 #include "Offline/RecoDataProducts/inc/IntensityInfoTrackerHits.hh"
 #include "Offline/RecoDataProducts/inc/IntensityInfoCalo.hh"
@@ -76,7 +76,7 @@ namespace mu2e {
     const IntensityInfoTimeCluster*  _tcIntInfoDF;
 
 
-    ProtonBunchIntensity*            recoPBI; //Pointer to collection module produces
+    RecoProtonBunchIntensity*            recoPBI; //Pointer to collection module produces
 
   public:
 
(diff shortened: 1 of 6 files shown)


Lumi/src/RecoNPOTMaker_module.cc
```diff
@@ -212,8 +212,8 @@ namespace mu2e {
     }
     else if(_debugLevel > 0) {std:: cout <<"[RecoNPOTMaker::produce] Did not find DeltaFinder IntensityInfoTimeCluster data" << std::endl;}
 
-    std::unique_ptr<ProtonBunchIntensity> recoPBI (new ProtonBunchIntensity);
-    recoPBI->set(POT_caloEnergy); // only using calo energy to make the estimate for now
+    // only using calo energy to make the estimate for now, with a fixed 10% uncertainty
+    std::unique_ptr<RecoProtonBunchIntensity> recoPBI (new RecoProtonBunchIntensity(POT_caloEnergy, 0.1*POT_caloEnergy));
```
brownd1978, line 216: A cleaner line to do the smae thing:
`    auto recoPBI = std::make_unique<ProtonBunchIntensity>(POT_caloEnergy); `

RecoDataProducts/inc/RecoProtonBunchIntensity.hh
```diff
@@ -0,0 +1,25 @@
...
+  class RecoProtonBunchIntensity {
+    public:
+    explicit RecoProtonBunchIntensity(unsigned long long intensity = 0, unsigned long long uncertainty = 0) :
+      _intensity(intensity), _uncertainty(uncertainty) {}
+      unsigned long long intensity() const { return _intensity; }
+      unsigned long long uncertainty() const { return _uncertainty; }
+      bool operator == (RecoProtonBunchIntensity const& other ) const {
+        return _intensity == other.intensity() && _uncertainty == other.uncertainty(); }
+      bool operator != (RecoProtonBunchIntensity const& other ) const { return !(operator ==(other)); }
+    void add(RecoProtonBunchIntensity const& other) { _intensity += other.intensity(); _uncertainty = std::sqrt(std::pow(_uncertainty, 2) + std::pow(other.uncertainty(), 2)); }
```
brownd1978, line 18: Why hard-code the uncertainty calculation?

brownd1978 (commented):
A few things to think about.

RecoDataProducts/inc/RecoProtonBunchIntensity.hh
michaelmackenzie: I was just adding the same `add` function as the MC PBI object. I could have it pass in a function that returns a new uncertainty given the two uncertainties to let the user define the method, would that be a better strategy? I could also remove this function and add setter functions, so a user could do the addition as defined by their analysis and just update the `_intensity` and `_uncertainty` fields

Lumi/src/RecoNPOTMaker_module.cc
michaelmackenzie: I updated to this style, thank you for the suggestion!

PR #1562, opened 2025-08-12
Initial prototype for mu2eX generator tool

sophiemiddleton:
Generator tool to make binned spectra for Mu2eX simualtions.

Also includes tabulated values from two example models: 

https://journals.aps.org/prd/abstract/10.1103/PhysRevD.84.113010 and https://journals.aps.org/prd/abstract/10.1103/PhysRevD.102.095007

Validation done a few years ago https://mu2e-docdb.fnal.gov/cgi-bin/sso/RetrieveFile?docid=38423&filename=Mu2eX%284%29.pdf&version=1 but cross-checked for current version of Offline and Production.

tabulared values are for lightest case, but users can use this infrastructure to input spectra of other types.

I also made a BinnedSpectrum generator, that would do the math for the generator itself (no need to tabulate) but it doesnt work in current offline. If people want that I can try to make it work, but I think the tabled values are fine, its the same as we do for DIO.

Also happy to work with others to add in other Mu2eX models. I just wanted to put this in Offline in case others are wanting to take a look. I know Mu2eX has lots of possible search strategies.


brownd1978 (commented):
Could the data files for this be put in cvmfs instead of GitHub?

sophiemiddleton:
good point, defintiely the ones from the paper with the three possible types could be columns in the same data file. I would need to figure out how the .tbl would need to be interpreted though in the source code

sophiemiddleton:
yes, they are three models

Issue #98, opened 2020-01-09
Scrub Mu2eUtilities for inappropriate #includes 


kutschke:
There are a lot of files in Mu2eUtilities that include art headers for no reason.  Remove them.

ryuwd:
I ran a tool called IWYU (https://github.com/include-what-you-use/include-what-you-use) over Mu2eUtilities. It produced these sets of proposed changes to cut out unnecessary includes (see link): 
https://gist.github.com/ryuwd/ddbcf324322b96d50aedd1b74bcb7d36

gaponenko:
Hi Ryun,  I like the spirit, but one should be careful with the implementation.   I took at quick glance at the proposed changes, and some of them seem wrong.  For example, per the C++ standard we should use the exception header.   The existence of bits/exception.h is an implementation detail of our C++ environment, and it would be wrong for us to include that header directly.

rlcee:
Years ago, we estimated that up to 30% of compile time is due to unnecessary includes.  We did not address it due to these comments:
- tool will not detect if a #define statement changed (will this tool catch that?) Since that time, we have greatly advanced our validation, so maybe this is moot?
- some includes are for clarity, future development, or commented out code.  Also my impression was that no one liked the idea of a tool editing their code.
Personally, I'm all for running this tool (module Andrei's comment)

ryuwd:
Hi Andrei, Ray,

> Hi Ryun, I like the spirit, but one should be careful with the implementation. I took at quick glance at the proposed changes, and some of them seem wrong. For example, per the C++ standard we should use the exception header. The existence of bits/exception.h is an implementation detail of our C++ environment, and it would be wrong for us to include that header directly.

I agree, as IWYU has some nuances (running it blindly would be bad). It might still be worth taking the output with a grain of salt, such that e.g. headers such as `<exception>` are left untouched to respect C++ standards.  At the very worst those changes/suggestions can be used just as a guide to resolve this issue

In this specific case, [a iwyu 'pragma' can be used to make sure `<exception>` is included rather than bits/exception by adding a comment to the latter file](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUPragmas.md). Or, just remove the suggested change. [I revised the gist to remove the changes to `<exception>` includes](https://gist.github.com/ryuwd/ddbcf324322b96d50aedd1b74bcb7d36/revisions).


[...]

PR #1563, opened 2025-08-12
Update luminosity stream reconstruction/analysis

michaelmackenzie:
This is connected to https://github.com/Mu2e/mu2e-trig-config/pull/98

This updates the intensity stream observables to include the CAPHRI hit energies and the N(Calo hits) by disk. I also did some reorganization of the example reco N(POT) analyzer. I've also added the lumi stream filter here, adding a compiler flag to check if this is an Online build or not.


michaelmackenzie:
@FNALbuild run build test with https://github.com/Mu2e/mu2e-trig-config/pull/98

CaloReco/src/CaloHitMakerFast_module.cc
```diff
@@ -70,7 +77,7 @@ namespace mu2e {
     private:
         using pulseMapType = std::unordered_map<unsigned, std::vector<HitInfo>>;
 
-    void extractHits(const CaloDigiCollection& caloDigis, CaloHitCollection& caloHitsColl, CaloHitCollection& caphriHitsColl, IntensityInfoCalo&intCalo, double pbtOffset);
+        void extractHits(const CaloDigiCollection& caloDigis, CaloHitCollection& caloHitsColl, CaloHitCollection& caphriHitsColl, IntensityInfoCalo&intCalo, double pbtOffset);
```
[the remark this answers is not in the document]
michaelmackenzie: I'm not exactly sure what the suggestion is here. I added a space after the `&` for the intensity object.

CaloReco/src/CaloHitMakerFast_module.cc
```diff
@@ -132,29 +140,53 @@ namespace mu2e {
...
+                 else                                ++nhits_d1;
+               }
+
+               // CAPHRI info
+               if(isCaphri) {
+                 if(energy > caphriEDepMin_ && energy < caphriEDepMax_) {
+                   const unsigned short e_short = IntensityInfoCalo::encodeCaphriEnergy(energy);
+                   const unsigned short caphri_ID = IntensityInfoCalo::encodeCaphriIndex(crID);
+                   if(caphri_ID > 3) printf("[CaloHitMakerFast::%s] Unknown CAPHRI index of %i from crystal ID %i\n", __func__, caphri_ID, crID);
+                   else {
```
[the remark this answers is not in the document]
michaelmackenzie: Good catch, this was leftover from debugging. I made this an error instead.

DAQ/src/DummyLumiInfoProducer_module.cc
```diff
@@ -60,9 +60,17 @@ namespace mu2e
     // Assign non-zero entries for harder compression
     if(_simMode == 1) {
       const static int prime_1(17), prime_2(251), prime_3(503), prime_4(1523); //for distributing values somewhat evenly
```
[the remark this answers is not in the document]
michaelmackenzie: This module is just to produce somewhat random data in Online luminosity stream testing. The prime number don't hold any real meaning and aren't used anywhere else, it's just to make the data harder to compress.

DataProducts/inc/CaloConst.hh
```diff
@@ -33,6 +33,11 @@ namespace mu2e {
 
     constexpr static uint16_t _invalid      = 9999;
 
+    // For storing compact CAPHRI hit info
+    constexpr static double _caphriEnergyUnits       = 0.01; // Store CAPHRI hit energies in units of 0.01 MeV
+    constexpr static int    _caphriIndexBits         =   14; // Store the CAPHRI hit index 14 bits into the 16-bit word
```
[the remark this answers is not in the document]
michaelmackenzie: I agree this is a bit different as it describes how we're storing the data in the IntensityInfoCalo object rather than anything about the detector really. I can move these fields to the IntensityInfo object where they're relevant? This packing needs to be done to store the <index | reco energy> variable in the IntensityInfoCalo data product, so it's after the reconstruction sequence as opposed to decoding input detector data. I can move the functions and variables to another location if there's a better place for them.

RecoDataProducts/src/IntensityInfoCalo.cc
```diff
@@ -0,0 +1,41 @@
...
+    // add it to the hit vector
+    caphriHits_.push_back(hit_info);
+
+    return true;
+  }
+
+  // Decode CAPHRI hit information
+  void IntensityInfoCalo::getCaphriHit(const size_t ihit, double& energy, int& ID) const {
+    ID = -1; energy = 0.;
+    if(ihit >= caphriHits_.size()) return;
```
brownd1978, line 30: Is there a legitimate use case for requesting a non-existing hit? If not I suggest treating OOB as an error (exception). That also avoids having to return a negative ID, which otherwise is unsigned.

DAQ/src/CaloHitsFromDTCEvents_module.cc
```diff
@@ -320,16 +320,27 @@ void art::CaloHitsFromDataDTCEvents::analyze_calorimeter_(
       float time = thisHitPacket.Time + peakIndex * digiSampling_ + timeCalib_[SiPMID];
 
       bool isCaphri = offlineId.crystal().isCaphri();
+      if(!isCaphri) {
+        if(offlineId.crystal().disk() == 0) ++nhits_d0;
+        else                                ++nhits_d1;
+      }
 
       // FIX ME! WE NEED TO CHECK IF TEH PULSE IS SATURATED HERE
       if (((eDep >= hitEDepMin_) || (isCaphri && (eDep >= caphriEDepMin_))) &&
```
brownd1978, line 329: This logic would be clearer split fully into isCaphri and !isCaphri blocks. As is that flag is tested multiple times. I understand that pre-exists this PR, but adding the intensity info branch makes it worse.

DAQ/src/CaloHitsFromDTCEvents_module.cc
```diff
@@ -320,16 +320,27 @@ void art::CaloHitsFromDataDTCEvents::analyze_calorimeter_(
       float time = thisHitPacket.Time + peakIndex * digiSampling_ + timeCalib_[SiPMID];
 
       bool isCaphri = offlineId.crystal().isCaphri();
+      if(!isCaphri) {
+        if(offlineId.crystal().disk() == 0) ++nhits_d0;
```
brownd1978, line 324: if nhits were an array this line would be simply  ++nhits[offlineId.crystal().disk()]

brownd1978 (commented):
This is much improved, thanks Michael. There are still some issues I'd like you to review, some are just suggestions.

RecoDataProducts/inc/IntensityInfoCalo.hh
```diff
@@ -14,23 +16,37 @@ namespace mu2e {
...
+    size_t         nCaphriHits  () const { return caphriHits_.size(); }
+
+    // Methods to store/retrieve caphri hit info
+    bool addCaphriHit(const double energy, const int ID);
+    void getCaphriHit(const size_t ihit, double& energy, int& ID) const;
 
   private:
-    unsigned short  nCaloHits_    = 0;
+    // unsigned short  nCaloHits_    = 0;
+    unsigned short  nCaloHitsD0_  = 0;
```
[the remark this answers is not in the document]
michaelmackenzie: I updated to use an array instead here

RecoDataProducts/src/IntensityInfoCalo.cc
michaelmackenzie: I was unsure of where to put the error throwing initially. I've updated the product to throw errors now, and removed this error checking from the modules

DAQ/src/CaloHitsFromDTCEvents_module.cc
michaelmackenzie: I've simplified this as suggested, and also in `CaloHitMakerFast`

RecoDataProducts/src/IntensityInfoCalo.cc
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  // Decode CAPHRI hit information
+  void IntensityInfoCalo::getCaphriHit(const size_t ihit, double& energy, int& ID) const {
+    ID = -1; energy = 0.;
+    if(ihit >= caphriHits_.size()) throw cet::exception("RECO") << "Hit index " << ihit << " is >= N(hits) = " << caphriHits_.size();
+
+    const auto hit_info = caphriHits_[ihit];
+
+    constexpr static unsigned short caphriIndexMask = 0x3 << caphriIndexBits_; // mask for where the hit index is stored
```
brownd1978, line 33: Is there a reason this is defined static here, while the companion bitshift constexpr is defined in the .hh file?

brownd1978 (commented):
Looks better, I still have 1 question.

michaelmackenzie:
It wasn't very intentional the separation, just noting that it was only needed in one place. I've moved the mask definition to be with the bit definition. The index overflow is checked when adding a hit, throwing an error if it would overflow 2 bits.

brownd1978 (approved):
Thanks for your patience.

michaelmackenzie:
@FNALbuild run build test with https://github.com/Mu2e/mu2e-trig-config/pull/98

brownd1978:
Please let me know when mu2e-trig-config #98 is merged so I know when to merge this PR.

PR #1566, opened 2025-08-13
Optimized the FindPixel algorithm for digitization of extinction moni…

caohc:
…tors


gaponenko (approved):
Haichuan, thank you for updating the code to implement the non-uniform pixel size.

PR #1567, opened 2025-08-13
minor tweaks for alternate compilers

marcmengel:
I was testing builds with different compilers and ran across these

The code is currently relying on some side-effects of one C++ header including another for you, which the newer compilers are getting better at avoiding.

These are all added standard includes, except for one place where we're replacing size_t with std::size_t, and another place where we're casting to std::string.

### DIFF
--- a/CaloMC/inc/CaloWFExtractor.hh
+++ b/CaloMC/inc/CaloWFExtractor.hh
@@ -4,6 +4,7 @@
 // Utility to simulate waveform hit extraction in FPGA
 //
 #include <vector>
+#include <cstddef>
 
 namespace mu2e {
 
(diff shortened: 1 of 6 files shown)


rlcee:
@marcmengel this is a pleasant surprise, thanks!  I'm also surprised at how short the list is.  Two questions 
- what was the alternate compiler, and will it be of interest to us?
- why Mu2e? Was it just a convenient test sample?

marcmengel:
It was gcc 12.5.0, although I think you would get similar issues with other later gcc's.
I am trying out our proprosed process for a  Spack Grand Unified Environment build to get releases of Art, Larsoft, boost, etc. that work for everybody; thus I'm trying to build dunesw, mu2e's Offline, the fife tools, etc. all in one Grand Unified environment.  
This PR is me trying not to lose the tweaks I made to get it to build.

rlcee:
I'll go ahead and merge this, since I don't see why not, but I remain curious why you see errors when we don't.

PR #1570, opened 2025-08-18
Drift calibration studies

GonzaloDB13:
I have modified 5 files according to the latest results of the drift calibration studies that I presented last July.

- DriftCalibPDF has included the "extended" range of calibration up to 3 mm (instead of 2.5 mm) to account for the correction of the tail of the rdrift distribution, looking now more similar to the MC DOCA.

- The changes in the files committed to RecoDataProducts and Mu2eKinKal reflect the new reconstruction of the angle that we use to calculate the Lorentz force over a drift electron. Previously, the angle was wrongly reconstructed, which allowed a bad calculation of the drift time for high MC DOCA. That has been fixed.


brownd1978 (commented):
Thanks for putting in the PR Gonzalo! I have some questions and few suggestions.

bonventre (approved):

PR #1572, opened 2025-08-20
Use t0 segment for validation

brownd1978:


brownd1978:
@FNALbuild run build test

PR #1577, opened 2025-08-22
Geometry changes for crvmuontagger

joshgreaves332:
No code changes. Will only affect geometry for CRV Wideband Muon Tagger setup. 
Please add Ralf Ehrlich as reviewer.


brownd1978:
@FNALbuild run build test

PR #1579, opened 2025-08-27
Refactor straw modeling to be more physically consistent

brownd1978:
This reduces the number of parameters and makes their interpretation simpler. The results for path lengths and active straws are now consistent with expectations, though the momentum at TT_Front is shifted.


brownd1978:
@FNALbuild run build test with Mu2e/mu2e-trig-config#109

brownd1978:
@FNALbuild run build test

PR #1581, opened 2025-08-29
Fixes to run1_a geometry.

brownd1978:
These got lost somehow, now restored.

### DIFF
--- a/Mu2eG4/geom/geom_run1_a.txt
+++ b/Mu2eG4/geom/geom_run1_a.txt
@@ -48,8 +48,8 @@ double mu2e.detectorSystemZ0 = 10171.;   // mm  G4BL: (17730-7292=9801 mm)
 #include "Offline/Mu2eG4/geom/TransportSolenoid_v08.txt"
 
 // External Shielding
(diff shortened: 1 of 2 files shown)


Mu2eG4/geom/geom_run1_a.txt
```diff
@@ -100,7 +100,9 @@ bool   vd.solid          = false;
 #include "Offline/Mu2eG4/geom/visualization_regions.txt"
 
 
-
+// patch to change East DS hatch block height
+double building.dsArea.hatchblock.E.offsetFromFloorSurface.y = 6400.80;
+double building.dsArea.hatchblock.E.yHalfThickness           = 457.20;
```
ehrlich-uva, line 105: I remember the 3ft thickness (=457.2mm halfthickness), but I don't know the other value.

ehrlich-uva: @oksuzian Could you please check these values?

ehrlich-uva (approved):

Mu2eG4/geom/geom_run1_a.txt
oksuzian: Looks right to me.

PR #1588, opened 2025-09-05
Update TrackPID module for TMVA-SOFIE

lborrel:
Prepare the code in Offline to use TMVA-SOFIE for TrackPID. It has not been tested as we need to wait for an update of TMVA-SOFIE to support sigmoid function.


lborrel:
Will need review from @AndrewEdmonds11 and @brownd1978

brownd1978 (approved):
Looks good, thanks for making the updates.

PR #1589, opened 2025-09-05
skip collimator at PS entrance

sdifalco:
Disable collimator at PS entrance to avoid conflicts with MARS geometry


caohc:
If we use Offline/Mu2eG4/geom/geom_common.txt, the default file related to External Shielding is Offline/Mu2eG4/geom/ExtShieldUpstream_v06.txt, not Offline/Mu2eG4/geom/reduced_ExtShieldUpstream_v06.txt

caohc (approved):

PR #1598, opened 2025-09-18
Modified Radioactive VR mode switch behavior

bechenard:
Following the discussion with Krzysztof, I implemented the solution we agreed on


brownd1978:
Has this development converged, or is it still in progress?

resnegfk (approved):
I think it's okay now

PR #1600, opened 2025-09-23
Add module to find tracks in different lists that are 'the same' particle (based on the same hits).

brownd1978:
This module is useful to compare tracks made by different reconstruction versions or fit options.


brownd1978:
Thanks for the comments, I believe I've addressed them all.

gianipez (approved):

gianipez:
@FNALbuild run build test

brownd1978:
@FNALbuild run build test

PR #1601, opened 2025-09-24
TrkQual ANN1 v2

AndrewEdmonds11:
This PR adds the retrained TrkQual ANN1 to Offline. ~Note that although the ANN structure didn't change, we still ended up with a different .hxx file. For this reason, I have added the ability for the ```TrackQuality``` module to use either the old or "new" models.~ The code to train is in Mu2e/MLTrain#1

Edit: after the meeting today, Dave pointed me to a script that reformats the .dat file so that we don't need two different .hxx files for the same model. I have removed this functionality from TrackQuality. The script to convert the .dat file is now in Mu2e/MLTrain#1


brownd1978 (commented):
Andy, did the set of variables used change between v1 and v2? Or is it just that the training + SOPHIE produced a different data struct that forces a 2nd version?

AndrewEdmonds11:
After the meeting today, Dave pointed me to a script that reformats the .dat file so that we don't need two different .hxx files for the same model. I have removed this functionality from TrackQuality. The script to convert the .dat file is now in https://github.com/Mu2e/MLTrain/pull/1. I have updated the lead comment of this PR to make this clear to anyone reading in the future

This PR is now ready to merge

AndrewEdmonds11:
I forgot to add that I tested that the converted .dat with the old .hxx and the original .dat with the new .hxx produce the same results. In the plot below trkqual3 is the new .hxx with the unconverted .dat, and trkqual4 is the old .hxx with the converted .dat. The difference between all values is 0

image

TrkDiag/inc/TrkQual_ANN1_v2.hxx
```diff
@@ -0,0 +1,262 @@
+//Code generated automatically by TMVA for Inference of Model file [TrkQual_ANN1_v2.onnx] at [Wed Sep 24 16:57:16 2025]
```
brownd1978, line 1: Is this file still needed?

brownd1978 (commented):
I maybe misunderstood the effect of your last update WRT *v2.hxx

PR #1602, opened 2025-09-26
fix overlap for nw blocks

sdifalco:
The two placeholder blocks that will be inserted to support the pipe were extruding in the volume beneath: fixed


sdifalco:
@FNALbuild run build test

brownd1978 (approved):

PR #1604, opened 2025-09-30
Update to MDC20205, add legacy MDC2020 geometry

brownd1978:
This is a breaking, non-backwards-compatible change advancing the default simulation geometry from that used in MDC2020 datasets to that expected from run1 (run1_a scenario). New geometry configurations corresponding to the MDC2020 configuration are added, and must be used to analyze MDC2020 datasets, or to create new MDC2020 datasets.


brownd1978:
@FNALbuild run build test

oksuzian (approved):

PR #1605, opened 2025-10-02
Ovlp fix2

sdifalco:
Fix tracker endcap ring extrusion in tracker mother by extending it by 400 microns in each direction along z.
Disable by default the upstream STM virtual detector waiting to make it compatible with run1a CR geometry.
Add a txt file to be used to check the surfaces with run1a geometry.


brownd1978 (commented):
I have done a preliminary test of the mother volume change and found no major issues. I'm still a bit worried about it. I will do a few further tests before merging.

PR #1606, opened 2025-10-02
Solve the  of mis-position of VirtualDetector_STM_UpStr in crv_counte…

caohc:
…rs_run1a_v01.txt


sdifalco (changes requested):
If possible I would prefer to correct crv_counters_run1a_v01.txt instead of creating a new file. Please check with Iuri Oksusian that the changes you introduce don't have undesired effects on other CRV volumes.

ehrlich-uva:
There shouldn't be any CRV-D modules in the run1a geometry. We only have CRV-T modules in run1a.

PR #1619, opened 2025-10-17
Adding table for channel statuses

giro94:
Adding the table to hold channel statuses.
Cat-3 table, this is typically filled with online values (MongoDB) and then propagated offline.

Status is a string, with typical values of "good", "dead", "hot", and open for other options.

Reconstruction code should discard hits from channels where (status != "good")

### DIFF
--- a/DbTables/inc/CalChannelStatus.hh
+++ b/DbTables/inc/CalChannelStatus.hh
@@ -1,0 +1,68 @@
+#ifndef DbTables_CalChannelStatus_hh
+#define DbTables_CalChannelStatus_hh
+
+#include <string>
+#include <iomanip>
+#include <sstream>
+#include <map>
+#include "cetlib_except/exception.h"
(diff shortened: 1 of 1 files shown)


rlcee:
Did you consider using a bitmap of bad conditions, and then "good" is bitmap=0.  Trk uses [this](https://github.com/Mu2e/Offline/blob/main/DataProducts/inc/StrawStatus.hh) [here](https://github.com/Mu2e/Offline/blob/b15175cf01e2a76d7f68406c04d72aa38315e64f/DbTables/inc/TrkElementStatus.hh#L28)

rlcee:
Another considerations is whether you want this table to be fixed length.   If you expect only a handful of bad channels, it might be easier to list only those channels - a little more "human friendly".
After you reply with your final decisions on these two points, I will quickly follow with the technical review (should be minor).

PR #1621, opened 2025-10-20
Use CMake Target notation for KinKal (depends on KFTrack/KinKal#212)

eflumerf:
While these changes only affect CMake, they do rely on a new build of KinKal, so should not be merged until the referenced KinKal PR is merged and a new version distributed.


brownd1978 (approved):
Looks good

PR #1635, opened 2025-10-30
Fix trigger navigator printout, add example print call

michaelmackenzie:


brownd1978 (approved):

brownd1978:
@FNALbuild run build test

PR #1640, opened 2025-11-04
Rename the Extinction Monitoer's readout chips 

caohc:
Rename the Extinction Monitoer's readout chips  so that they will not be recognized as sensitive detectors in the simulation


gaponenko (approved):
Hi Dave,

Sam Fine is not on Mu2e now (he had an internship with me last summer), so he won't be able to review this PR.   I discussed this with Haichuan and agreed with the fix.

A brief summary:  the ExtMon simulation code originally did not care about logical volume names.     When Sam was updating simulation geometry, he added and modified volumes, and we were not concerned about the details of their naming.  What we did not know was that simulation code changes to support multithreading modified parts of our sensitive detector code and started using string matches on volume names to decide what needs to be made sensitive.   These two sets of changes interacted to produce a bug - the creation of unintended sim hits.

The current PR fixes the immediate issue.   In the long term, we should fix the simulation to avoid the fragile string matching approach.  I think I know how to do this for ExtMon, but the same string matching is used for other detectors as well.

Andrei

PR #1645, opened 2025-11-13
Time optimization: APR

michaelmackenzie:
I'm reviewing Online trigger modules that have the largest impact on event processing time, and attempting to improve the performance through operation reduction, caching data, reducing object creation, reserving space to avoid copies during vector expansion, etc. I have also included a `StopWatch` utility that is useful for these tests, and I intend to deploy it in the Online modules that consume the most processing time for detailed evaluations.

These changes primarily only affect these elements, without changes to logic. One exception is in the `findSeedPhiLines` function, where the negative dphi/dz slope hypothesis was skipped in some conditions in the positive hypothesis test. This didn't change the behavior much at all, but it is a change. I also added a new parameter to stop adding hits to the seed circle if we have enough already, but it's off by default. Otherwise, there is potential for many numerical effects, but logical steps shouldn't change.

The APR timing performance is significantly improved through these updates, reducing from ~1.8 ms / event to 1.2 ms / event in the Online configuration using 1BB no primary simulations.

### DIFF
--- a/Mu2eUtilities/CMakeLists.txt
+++ b/Mu2eUtilities/CMakeLists.txt
@@ -44,6 +44,7 @@ cet_make_library(
       src/SimpleSpectrum.cc
       src/SortedStepPoints.cc
       src/STMUtils.cc
+      src/StopWatch.cc
       src/Table.cc
       src/TrackCuts.cc
       src/TrackerBFieldInfo.cc


michaelmackenzie:
One quick correction: The times I put weren't normalized properly (I was using a certain N(events) benchmark in my notes). It's actually ~0.22 ms / event --> ~0.15 ms / event. The same ratio, just less worrying of a total time (the total time / event estimate from this summer was 0.9 ms / event).

PR #1646, opened 2025-11-14
Add missing reference to artdaq-core-mu2e::Overlays

eflumerf:


rlcee:
This is a bit mysterious since we build the head of Offline, in spack/cmake only,  fresh checkout, every night, and don't see this error.  I don't know why we would see a difference here.

PR #1648, opened 2025-11-19
Time optimization: DeltaFinder

michaelmackenzie:
This builds off of PR https://github.com/Mu2e/Offline/pull/1645, looking at the DeltaFinder module. A significant time is spent resetting the time bins and sorting the hits. This updates the time bin resetting to only reset the ranges that will be used in the event type, and sorts using a count-sort algorithm. This improves speed by ~10-20% on background-only simulation.

### DIFF
--- a/CalPatRec/src/DeltaFinder_types.cc
+++ b/CalPatRec/src/DeltaFinder_types.cc
@@ -16,10 +16,10 @@ namespace mu2e {
 
 //-----------------------------------------------------------------------------
     FaceZ_t::FaceZ_t() {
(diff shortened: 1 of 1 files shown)


brownd1978:
@FNALbuild run build test

michaelmackenzie:
Hi Pasha, I agree that the compiler is likely already handling this level of optimization. I tested this earlier in the optimization investigation just to see if it would have an effect since we call the `addPoint` function so many times, and then I left it in as I assumed there was no harm in manually reducing the number of operations.

PR #1652, opened 2025-11-22
some AI cleanup suggestions

rlcee:
Some suggestions from an AI code review - should have no effect on validation.  One is a bug which worked OK by accident, the rest are from a cleanup suggestion.


brownd1978 (approved):
snprintf is superior to sprintf so this looks good to me.

PR #1653, opened 2025-11-24
Addition of TwoDPoint and other associated classes to classes_def

NamithaChitrazee:
Added TwoDPoint, CombineTwoDPoint etc classes to the classes_def.xml and assigned some default value to the variables in struct CWT, after encountering the following errors while trying to keep the BkgClusters data product in the reconstructed art file:

1) checkDictionaries: Retrieving list of base classes for type 'mu2e::TwoDPoint' returned a nullptr.

2)---- DictionaryNotFound BEGIN No dictionary found for the following classes: map<unsigned long,mu2e::CombineTwoDPoints::CWT> mu2e::CombineTwoDPoints::CWT Most likely they were never generated, but it may be that they were generated in the wrong package.

3) candidate: 'template<class _U1, class _U2> requires _S_constructible<_U1, _U2>() && !_S_dangles<_U1, _U2>() constexpr std::pair<_T1, _T2>::pair(std::pair<_U1, _U2>&&) [with _U2 = _U1; _T1 = const long unsigned int; _T2 = mu2e::CombineTwoDPoints::CWT]' 345 | pair(pair<_U1, _U2>&& __p) | ^~~~

..TCollectionProxyInfo.h:343:13: note: candidate expects 1 argument, 0 provided 343 | ::new(m) Value_t(); | ^~~~~~~~~~~~~~~~~

### DIFF
--- a/RecoDataProducts/CMakeLists.txt
+++ b/RecoDataProducts/CMakeLists.txt
@@ -2,8 +2,6 @@ cet_make_library(
     SOURCE
       src/AlgorithmID.cc
       src/BkgClusterFlag.cc
-      src/BkgCluster.cc
-      src/BkgClusterHit.cc
       src/BkgQual.cc
       src/ComboHit.cc
       src/CosmicTrack.cc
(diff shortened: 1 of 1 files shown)


brownd1978:
@FNALbuild run build test

RecoDataProducts/CMakeLists.txt
```diff
@@ -2,6 +2,8 @@ cet_make_library(
     SOURCE
       src/AlgorithmID.cc
       src/BkgClusterFlag.cc
+      src/BkgCluster.cc
```
brownd1978: This .cc files don't exist, please remove these lines from CMakeLists

brownd1978:
@FNALbuild run build test

brownd1978 (approved):
looks good, thanks for the quick response

PR #1655, opened 2025-11-25
initialize electron mass

sdifalco:
correct a bug in CE leading log generator: electron mass was not initialized


brownd1978:
Have you tested this?

PR #1659, opened 2025-11-30
Convert trigger info to art::Ptr collections

michaelmackenzie:
Create collections of art::Ptrs from the trigger info objects. This will more easily allow for analysis of objects that fired a trigger of interest (e.g. adding a collection of tracks that fired the high momentum APR trigger).


michaelmackenzie:
@FNALbuild run build test

michaelmackenzie:
> Please add these new types to classes_def.xml

I added these to the classes_def.xml and removed the non-named versions that were there before these typedef lines were added.

PR #1663, opened 2025-12-04
Included mean z position of a background cluster and a vector to store the hit positions 

NamithaChitrazee:
1. The mean z position of a background cluster was set to 0 previously. I have updated it to take the weighted average of the z position of the hits belonging to a background cluster.
2. Added a vector to store the 3-D positions of the hits belonging to each cluster.

The changes are necessary to visualize the background clusters and see which hits belong to which cluster using the EventDisplay.

### DIFF
--- a/TrkHitReco/src/FlagBkgHits_module.cc
+++ b/TrkHitReco/src/FlagBkgHits_module.cc
@@ -240,17 +240,6 @@ namespace mu2e
       for (const auto& chit : cluster.hits())
         chfcol[chit].merge(flag);
     }
-    /*for (auto& cluster : bkgccol) {
-        std::cout << "Cluster has " << cluster.hitposition().size() << " hit positions\n";
-        for (size_t i = 0; i < cluster.hitposition().size(); ++i) {
-          auto const& p = cluster.hitposition()[i];
-          std::cout << "  hit " << i
-                  << " pos = (" << p.x()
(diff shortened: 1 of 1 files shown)


TrkHitReco/src/FlagBkgHits_module.cc
```diff
@@ -230,16 +231,26 @@ namespace mu2e
...
         flag.merge(StrawHitFlag(StrawHitFlag::bkg));
         cluster._flag.merge(BkgClusterFlag::bkg);
       }
 
-      for (const auto& chit : cluster.hits()) chfcol[chit].merge(flag);
-
+      for (const auto& chit : cluster.hits())
+        chfcol[chit].merge(flag);
     }
+    /*for (auto& cluster : bkgccol) {
```
brownd1978: If this is a useful debug printout I suggest to uncomment it but put it behind a if(_debug>X) condition

NamithaChitrazee: No, I added it at the time to make sure that the new function "addHitPosition" is functioning correctly. 

But now, we can use the EventDisplay to visually discern which hits belong to which BkgCluster, so we can use that as the debugging tool.

oksuzian:
@FNALbuild run build test

brownd1978 (approved):

PR #1667, opened 2025-12-05
Update trigger tools

michaelmackenzie:
Add a method to retrieve a trigger bit by trigger path name. Also fix a bug in ReadTriggerPaths in jobs where no events are processed.

### DIFF
--- a/Mu2eUtilities/inc/TriggerResultsNavigator.hh
+++ b/Mu2eUtilities/inc/TriggerResultsNavigator.hh
@@ -32,9 +32,15 @@ namespace mu2e {
     size_t    find(std::map<std::string, unsigned int> const& posmap, std::string const& name) const;
     size_t    findTrigPathID(std::string const& name) const;
 
+    bool      validPath(std::string const& name) const {
+      size_t path_index = findTrigPath(name);
+      return path_index < _trigPathsNames.size();
+    }
+
     size_t    getTrigBit(std::string const& name) const {
-      size_t pathID = findTrigPathID(name);
-      return getTrigBit(pathID);
+      if(!validPath(name))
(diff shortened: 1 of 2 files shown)


oksuzian:
1. Why trigger test failed?

2. if (i>_trigPathsNames.size()) {
  throw cet::exception("TRIGGER") << "TRIG PATHID " << i << " NOT FOUND";
  return 0;  // This line is unreachable after throw
}
The return 0; statement after the throw is dead code and should be removed.

3. size_t getTrigBit(std::string const& name) const {
  size_t pathID = findTrigPathID(name);
  return getTrigBit(pathID);
}
Consider what happens if findTrigPathID(name) returns an invalid value. You might want to add error handling here or ensure findTrigPathID throws appropriately.

michaelmackenzie:
> 1. Why trigger test failed?

This was due to an out-of-date CI log file in mu2e-trig-config, I updated this log file to have the bit-reassignment changes.

> 2. if (i>_trigPathsNames.size()) {
>    throw cet::exception("TRIGGER") << "TRIG PATHID " << i << " NOT FOUND";
>    return 0;  // This line is unreachable after throw
>    }
>    The return 0; statement after the throw is dead code and should be removed.

I agree, I've removed this line.

> 3. size_t getTrigBit(std::string const& name) const {
>    size_t pathID = findTrigPathID(name);
>    return getTrigBit(pathID);
>    }
>    Consider what happens if findTrigPathID(name) returns an invalid value. You might want to add error handling here or ensure findTrigPathID throws appropriately.

Previously it would throw an error on the getTrigBit call if the name isn't defined. I've updated this to have a more explicit `validPath` function and added a few more error throws.

PR #1673, opened 2025-12-17
Run1B

oksuzian:
First PR of Run1B configuration


rlcee (approved):
I didn't study the geometry itself, so only looked briefly for any issues.
Generally we wouldn't leave large blocks of code commented, but this could be counted as a special case, so OK.

PR #1678, opened 2025-12-22
Calorimeter digi mixing

edcallaghan:
This PR implements digi mixing for the calorimeter. The mechanics largely follow the pattern set by the tracker: the relevant data products are mixed using standard technology, and digis from the same channel which overlap in time are resolved into a single digi. For calorimeter digis, which are principally waveforms, this is conceptually straightforward: the waveforms are summed. It will, however, require extra effort to retain MC info for the calorimeter: unlike the tracker or CRV, which instantiate all MC info at the time of digi creation, it seems that some MC info for the calorimeter is reconstructed, making use of primary `SimParticle` info, after digis have already been made and interpreted as hits. That should be fine, but it require extra book-keeping/massaging to work in situations where there is MC info (i.e. detector data) or some digis were digitized independently of the simulated signal (e.g. digi mixing with conventional pileup).

New classes:
  - `CaloDigiWrapper`: A thin wrapper for calorimeter digis which is meant to collapse multiple loops, i.e. over synchronized digi + mc info collections, into a single loop. MC info is yet not included, for reasons described above.
  - `CaloDigiWrapperCollection`: stl-vector of the above, with additional logic for resolving a set of overlapping digis into one.

  - `Mu2eProductMixer`: Include provisions for mixing `CaloDigi`s and `CaloShowerSim`s, which (it seems to me) is the principal data product, not already mixable, which will allow to retain MC info.
  - `TimeBasedBuckets`: Generalized from a sequence of fixed-length timing windows to a sequence of variable-length timing windows. This is necessary as waveforms from the calorimeter of of variable length.

Changes to existing modules:
  - `MergeDigis`: Resolve collisions between overlapping calorimeter digis using `CaloDigiWrapperCollection` functionality.

Thoughts looking forward: The simple sum of multiple waveforms into a single waveform admits two subtleties: the saturation of the ADC must be reevaluated, and any analog noise present in the waveform is now "counted" multiple times in the sum.
  The former is accounted for in this PR, albeit with the ADC bit depth configuration now duplicate, as it is supplied directly to the `CaloDigiMaker` module, and not factored into some kind of conditions class. It would be good for that to be done, or maybe move the definition to fcl-prolog, to avoid the possibility of them getting out of sync.
  The latter is not accounted for, as currently an implemented waveform is just that: a waveform; it does not know "what" it contains. Conceptually, one can imagine ways of keeping the noise contribution to waveforms distinct until it absolutely needs to be added to any signal present, but that is outside of the scope of this PR and I'm not in a position to recommend something right now. But it's something we'll want to get right, since potential applications include mixing digis from the detector with both detector data and MC, both in situations with overlapping digis and not, so artefacts associated with different levels of noise are possible.

### DIFF
--- a/CaloMC/src/CaloDigiWrapperCollection.cc
+++ b/CaloMC/src/CaloDigiWrapperCollection.cc
@@ -112,7 +112,7 @@ namespace mu2e{
     std::map<SiPMID_t, CaloDigiWrapperCollection> wrappers_map;
     for (const auto& pair: unsorted_map){
       const auto& id = pair.first;
-      auto& wrappers = pair.second;
+      const auto& wrappers = pair.second;
       std::vector<CaloDigiWrapper*> sortable(wrappers.size());
       for (size_t i = 0 ; i < wrappers.size() ; i++){
         sortable[i] = const_cast<CaloDigiWrapper*>(&wrappers[i]);


edcallaghan:
@bechenard I've requested your review based on authorship of existing code, but obviously please delegate/defer as appropriate. The real calorimeter action is in `CaloDigiWrapperCollection::ResolveCollision` and `ResolveCollisions`. The latter may look a bit opaque, but is just identifying which digis have waveforms which overlap in time. The former performs the waveform sum.

rlcee:
In the CI report, check_cmake is reporting a problem, please take a look at that.  Thanks

edcallaghan:
@rlcee thanks, I forgot to propagate new classes into `CMakeLists`.

CaloMC/inc/CaloDigiWrapper.hh
```diff
@@ -0,0 +1,31 @@
...
+      using pos_t = size_t;
+      CaloDigiWrapper(const CaloDigi&);
+      const CaloDigi& Digi() const;
+
+      // interface for sorting into buckets of overlapping digitization windows
+      const double time() const;
+
+    protected:
+      const CaloDigi _digi;
+
```
bechenard: Why do you need to have this as a protected member instead of private? (I'll admit I;m a bit lazy and don;t want to read all the mixing code...)

edcallaghan: It wasn't a particularly motivated choice, but since private is a stronger restriction than protected, I usually default to members being protected unless I am confident that someone making use of the member in a subclass (whatever the reason that subclass ends up appearing) is inappropriate. Since this isn't stateful to some internal algorithm, or anything, I don't see a reason to further restrict it.

CaloMC/src/CaloDigiWrapperCollection.cc
```diff
@@ -0,0 +1,208 @@
...
+      std::vector<CaloDigiWrapper*> sortable(wrappers.size());
+      for (size_t i = 0 ; i < wrappers.size() ; i++){
+        sortable[i] = const_cast<CaloDigiWrapper*>(&wrappers[i]);
+      }
+      std::sort(sortable.begin(), sortable.end(), compare_t0s);
+      CaloDigiWrapperCollection sorted;
+      for (size_t i = 0 ; i < wrappers.size() ; i++){
+        sorted.Append(*sortable[i]);
+      }
+      wrappers_map.emplace(id, sorted);
```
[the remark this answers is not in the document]
edcallaghan: You are right on both points: there is a pointer-stability issue from pointing into an stl-vector, but this was chosen to avoid a performance penalty of deep-copying the objects during sorting. Maybe we could get away with a deep-copy for the calorimeter, but we have to do the same conceptual loop for the tracker, and in that case we have fixed waveforms and potentially thousands of digis from pileup, so deep-copying won't be tolerable (have run the tracker version periodically going back over a year, and never ran into an invaliated pointer).

Would you settle for marking the vector being pointed-into as `const`? :slightly_smiling_face: That compiles and would preclude any reallocations inside this block of code. If preferable, we could also factor that block out (may want to anyway, to share the implementation with the tracker) and explicitly comment that we only accept it due to the `const` condition.

edcallaghan:
> If you have several sources of CaloShowerSim objects, you need to merge them into a single collection before the truth association is performed (incidentally, you need to save the relevant SimParticles, CaloShowerSteps, and CaloShowerRO). I don't recall exactly when your salting is performed, but if this is done before the MC truth association then all should be fine.
> 
> Is there a point regarding MC that I misunderstood from your code?

Concatenating those collections into one for each type is easy enough, so if that's all if takes, then it's no big deal. But one thing that caught my attention is that the reconstruction of the MC info involves the assumption of a single `PrimaryParticle`: https://github.com/Mu2e/Offline/blob/main/CaloMC/src/CaloHitTruthMatch_module.cc#L140-L141. But I guess your point is that if we already have the single collections that you mentioned, and one `PrimaryParticle` for whatever it is we are simulating (e.g. a conversion), then it doesn't matter if the signals were overlayed as energy depositions or as digis. That would be great.

edcallaghan:
@FNALbuild run build test

oksuzian:
@FNALbuild run build test

PR #1680, opened 2026-01-09
Added StrawGasStepFilter

bonventre:
For pruning extracted sims for single stations etc.


brownd1978 (approved):
This is fine.
I'd encourage using a computed for loop to make it easier to read:
for(auto const& step : steps){
...

PR #1684, opened 2026-01-15
Updates to the STM simulation framework

PawelPlesniak:
This adds a set of updates to the STM simulation infrastructure. Details will be provided once complete.
Note to self - run tests on all written modules before uploading - there is a substantial rewrite included here.


oksuzian:
@PawelPlesniak , do you want to implement Ray's suggestions before merging?

PawelPlesniak:
Hi all, thank you so much reviewing, this is still a work in progress and would rather it be left as is until it is complete. Yongi and I will discuss time lines on Slack

PR #1691, opened 2026-01-24
Add data product to record filtering fraction

brownd1978:
This is intended to record prescale factors, but is more generalized in case there are other uses.

### DIFF
--- a/Print/src/CosmicLivetimePrinter.cc
+++ b/Print/src/CosmicLivetimePrinter.cc
@@ -7,8 +7,7 @@
 
 void mu2e::CosmicLivetimePrinter::Print(art::Event const& event, std::ostream& os) {}
 
-void mu2e::CosmicLivetimePrinter::PrintSubRun(art::SubRun const& subrun,
-                                              std::ostream& os) {
+void mu2e::CosmicLivetimePrinter::PrintSubRun(art::SubRun const& subrun, std::ostream& os) {
   if (verbose() < 1) return;
   if (tags().empty()) {
     // if a list of instances not specified, print all instances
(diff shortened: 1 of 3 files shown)


Trigger/src/PrescaleEvent_module.cc
```diff
@@ -40,9 +41,10 @@ namespace mu2e
     };
 
     struct EventMode {
-      EventWindowMarker::SpillType _type;
```
michaelmackenzie: Do you not prefer this style of field labeling? Andrew had recommended prepending `_` rather than appending to avoid naming collisions with fortan functions, but I have no opinion otherwise

Trigger/src/PrescaleEvent_module.cc
```diff
@@ -63,68 +65,66 @@ namespace mu2e
...
-    bool retval(false);
-
-    // Apply the prescale, if requested for this event type
-    if(ps > 0) {
-      retval = e.event() % ps == 0;
-      if (retval){
-        ++_npass;
+        ++nevt_[imode];
+    // Apply the prescale
+        bool retval = e.event() % mode.prescale_ == 0;
```
michaelmackenzie: We use `ps <= 0` as a special case to entirely prescale a path while still loading the modules and having the path in the `art::TriggerResults`. Therefore, we need to check this before apply the modulo.

michaelmackenzie (changes requested):
This looks good and I appreciate you adding the ability to add the filter fraction options. I only requested minor edits to preserve the 100% prescale mode used in Online processing

Trigger/src/PrescaleEvent_module.cc
```diff
@@ -63,68 +65,66 @@ namespace mu2e
...
   }
 
-  bool PrescaleEvent::endRun( art::Run& run ) {
-    if(_debug > 0){
-      std::cout << moduleDescription().moduleLabel() << " passed " << _npass << " events out of " << _nevt
-                << " for a ratio of " << ((_nevt > 0) ? float(_npass)/float(_nevt) : 0.f) << std::endl;
+  bool PrescaleEvent::endSubRun( art::SubRun& subrun ) {
+    for (size_t imode = 0; imode < eventMode_.size(); imode++){
+      auto const& mode = eventMode_[imode];
+      double frac = 1.0/double(mode.prescale_);
```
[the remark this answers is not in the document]
brownd1978: I added protection against prescale == 0, I agree that's a legitimate use case.

oksuzian:
@FNALbuild run build test

brownd1978:
Most of the AI points are knockon effects from combining result of general filtering and prescale filtering in a single output product. The best solution is to create a general class and prescale subclass. I'm not sure it's worth it. 
AI point 3 is a downstream effect of the fact that RandomPrescale is configured to use a float, not an int. To fix this I'd need to change the module and the data product. That would restrict the use cases to nominal filter fractions of form 1/N. In principle we could allow rational fractions.

PR #1694, opened 2026-01-24
Add a filter to veto events with high occupancies

michaelmackenzie:
This can kill an event based on:
- Combo hits
- Straw hits
- Straw digis
- Calo hits
- Calo digis
- Fragments

The intent is to filter events that could be too difficult for Online to process.


brownd1978 (approved):
This is an elegant design reusing common concepts across multiple data types through templating. I am ambivalent about the module allowing filtering on multiple types of data at once.

PR #1701, opened 2026-01-29
Reduce printout

brownd1978:
Per @rlcee request


rlcee (approved):
This is the same as #1700 except that one also changes line 279.  This particular flag was not driving the print I was looking at (I think), so I don't feel strongly if this flag should stay on.

PR #1703, opened 2026-01-30
Adding prodition service to CaloRecoDigiMaker

giro94:
Also, added fcl flag to be able to use proton bunch time or not (in case no beam is present)


giro94:
@FNALbuild build

giro94:
@FNALbuild build

PR #1707, opened 2026-01-31
move files out of ConditionsService

rlcee:
Move config files out of ConditionsService to EventGenerator 
This will need to be tested with the PR in Production, then PR's in EventNtuple and Tutorial will also have to be merged


rlcee:
@FNALbuild run build test with Mu2e/Production#501

rlcee:
@FNALbuild run build test with Mu2e/Production#501

PR #1712, opened 2026-02-02
Add calo tables

rlcee:
CalChannels for raw to offline channel number conversion
CalChannelStatus for a status word per channel


giro94 (approved):
Thanks!

PR #1728, opened 2026-02-14
Remerge main into Run1B branch

oksuzian:
Remerge main into Run1B


rlcee (approved):
While this can't be reviewed in detail, but it is almost all due to the BTrk removal, so I expect it is OK.  I scanned the diff for anything unexpected and didn't see anything.

PR #1730, opened 2026-02-14
Allow for restricted cos(theta_z) in DIO generation

michaelmackenzie:


oksuzian:
I'll review this pull request by first loading the PR details and its changes.Now let me look at the full file and the `RandomUnitSphere` class to understand the broader context:Now I have a complete picture of the PR. Let me also check the `RMCGenerator_tool.cc` to see how it handles the same pattern (for comparison with validation logic):Here is my review of [PR #1730](https://github.com/Mu2e/Offline/pull/1730):

## Summary

This PR adds two new FHiCL configuration parameters (`czmin`, `czmax`) to `DIOGenerator` to allow restricting the cos(θ_z) range when generating DIO (Decay-In-Orbit) events, mirroring the same feature that already exists in `RMCGenerator_tool.cc`.

The change is small (6 additions, 1 deletion) and modifies only one file: `EventGenerator/src/DIOGenerator_tool.cc`.

---

## Issues Found

### 1. 🔴 Missing input validation for `czmin`/`czmax`

The `RMCGenerator` (which this PR follows as a pattern) validates the cos(θ_z) range in its constructor:

```cpp name=EventGenerator/src/RMCGenerator_tool.cc url=https://github.com/Mu2e/Offline/blob/10120d9a62e5541e4ae1a3c0696492801c5f6d18/EventGenerator/src/RMCGenerator_tool.cc#L88
if(_czmin > _czmax || _czmin < -1. || _czmax > 1.) throw cet::exception("BADCONFIG") << "RMCGenerator cos(theta_z) range is not defined\n";
```

The `DIOGenerator` in this PR **does not** validate the new `czmin`/`czmax` inputs at all. Invalid configurations such as `czmin > czmax`, `czmin < -1`, or `czmax > 1` would be silently accepted. While the `fhicl::Atom` defaults are valid (`-1.` and `1.`), user-supplied values could be nonsensical and would produce incorrect physics results without any error message.

**Recommendation:** Add a validation check in the `DIOGenerator` constructor body, similar to `RMCGenerator`:

```cpp
if(_czmin > _czmax || _czmin < -1. || _czmax > 1.)
  throw cet::exception("BADCONFIG") << "DIOGenerator cos(theta_z) range is not defined\n";
```

### 2. 🟡 No accounting for the angular restriction in the normalization

The constructor already computes a "sampled spectrum fraction" to account for the restricted energy range, printing lines like:

```cpp name=DIOGenerator_tool.cc url=https://github.com/michaelmackenzie/Offline/blob/dc4139cd3efceb2c9f158a060997af8fb3696341/EventGenerator/src/DIOGenerator_tool.cc#L63
std::cout << "Sampled spectrum fraction " << integral/fullintegral << std::endl;
```

However, restricting `cos(��_z)` also reduces the solid angle coverage. The fraction of the unit sphere sampled is `(czmax - czmin) / 2`. This geometric factor is **not** included in the printed "sampled spectrum fraction." If downstream code relies on this fraction to correctly weight the DIO rate, the normalization will be wrong when `czmin`/`czmax` are restricted.

**Recommendation:** Either:
- Multiply the printed fraction by `(czmax - czmin) / 2.0` so the total efficiency accounts for both the energy and angular restrictions, or
- Print the angular fraction separately so users are aware.

### 3. 🟡 Declaration style: two variables on one line

```cpp name=DIOGenerator_tool.cc url=https://github.com/michaelmackenzie/Offline/blob/dc4139cd3efceb2c9f158a060997af8fb3696341/EventGenerator/src/DIOGenerator_tool.cc#L79
double _czmin, _czmax;
```

Declaring multiple member variables on a single line can reduce readability and makes diffs less precise. The `RMCGenerator` declares them on separate lines:

```cpp name=RMCGenerator_tool.cc url=https://github.com/Mu2e/Offline/blob/10120d9a62e5541e4ae1a3c0696492801c5f6d18/EventGenerator/src/RMCGenerator_tool.cc#L120-L121
    const double _czmin; //range of cos(theta_z) generated
    const double _czmax;
```

**Recommendation:** Declare on separate lines and consider adding comments and/or making them `const`, consistent with `RMCGenerator`.

### 4. 🟢 Minor: Missing `const` qualifier on `_czmin` and `_czmax`

In `RMCGenerator`, the equivalent members are declared `const`:

```cpp name=RMCGenerator_tool.cc url=https://github.com/Mu2e/Offline/blob/10120d9a62e5541e4ae1a3c0696492801c5f6d18/EventGenerator/src/RMCGenerator_tool.cc#L120-L121
    const double _czmin; //range of cos(theta_z) generated
    const double _czmax;
```

Since these values are set in the initializer list and never modified, they should be `const` for consistency and safety.

---

## What Looks Good

- ✅ The `RandomUnitSphere` class already supports `czmin`/`czmax` parameters with appropriate defaults, so the API usage is correct.
- ✅ FHiCL parameters have sensible defaults (`-1.` and `1.`) that preserve backward compatibility — existing configurations are unaffected.
- ✅ The pattern follows `RMCGenerator_tool.cc` (authored by the same developer), showing consistency in approach.
- ✅ Member initialization order in the constructor matches the declaration order (after fixing), avoiding undefined behavior.

PR #1739, opened 2026-02-20
mark event as const in utility function

giro94:
Minor edit in order to be able to use the function in all modules with no "const" qualification drop:
```
void analyze(art::Event const& event){
  caloDAQUtil_.getFragments(event);
}
```


eflumerf (approved):
Adding const-ness is almost never wrong

PR #1746, opened 2026-02-26
Add cos(theta) restriction to RPC gun

michaelmackenzie:


oksuzian:
Issues Found

1. 🔴 Validation occurs after RandomUnitSphere is already constructed with potentially invalid values

The czmin_/czmax_ validation logic (lines 143–146) runs inside the constructor body, but randomUnitSphere_ is initialized in the member initializer list (line 130) — before the body executes. If a user passes czmin > czmax or out-of-range values, RandomUnitSphere will already be constructed with those bad values. While the exception will still be thrown before produce() is ever called, the RandomUnitSphere object briefly exists in an invalid state.

michaelmackenzie / Offline / EventGenerator / src / RPCGun_module.cc
v1
    , randomUnitSphere_ {eng_, czmin_, czmax_}

michaelmackenzie / Offline / EventGenerator / src / RPCGun_module.cc
v2
    // Validate the input cz values
    if(czmin_ > czmax_ || czmin_ < -1. || czmax_ > 1.) throw cet::exception("BADINPUT") << "RPCGun cos(theta_z) range is not defined\n";
    if(process_ == ProcessCode::mu2eInternalRPC && (std::abs(czmin_ + 1.) > 1.e-3 || std::abs(czmax_ - 1.) > 1.e-3))
      throw cet::exception("BADINPUT") << "RPCGun cos(theta_z) restriction can't be used with internal conversion\n";

Suggestion: Move the validation before randomUnitSphere_ initialization (which isn't easily possible in a member init list), or validate the conf() values at the top of the constructor body before anything can use them. Since the order of member initialization is fixed by declaration order, the pragmatic fix would be to add a small helper/validation function that throws before the constructor body, or accept that this is fine because no fire() is called until produce().

Severity: Low — the module will still throw before any event processing, but it's a code hygiene concern.

Issue #1608, opened 2025-10-05
NaN in reco


michaelmackenzie:
There can be NaN issues in reconstruction. An example case is in `Mu2eKinKal/inc/KKFit.hh:makeStrawHits` where
```c++
      auto ppos = ptraj.position3(htime);
      auto vel = ptraj.velocity(htime);
```
return undefined values. This leads to `ptime` being NaN and this can be passed into downstream reco, sometimes causing errors or segmentation faults such as in `TrackerConditions/src/StrawDrift.cc:T2D` due to undefined bin values.

An example instance of a NaN time (but no errors/faults):
```tcl
# Process on-spill reco
#include "Production/JobConfig/reco/OnSpill.fcl"

source : {
  module_type : RootInput
  fileNames: [
    "xroot://fndca1.fnal.gov/pnfs/fnal.gov/usr/mu2e/tape/phy-sim/dig/mu2e/ensembleMDS2cMix1BBTriggerable/MDC2020ba_best_v1_3/art/0f/f3/dig.mu2e.ensembleMDS2cMix1BBTriggerable.MDC2020ba_best_v1_3.001201_00000001.art"
  ]
  firstEvent: 7660
  firstRun: 1201
  firstSubRun: 1
  maxEvents: 1
}

services.DbService.purpose: MDC2020_best
services.DbService.version: v1_3
services.DbService.verbose : 2

outputs.Output.fileName   : "mcs.user.test_hit_error.version.sequencer.art"
```
Hit index 566/568 isn't well defined:
```
        KKFit::makeStrawHits: Non-finite ptime for hit index 566 for collection size 568, hpos.z() = 1514.39, ppos.z() = -nan, vel.z() = -nan
```
(can be seen by adding this to above `CAHint hint(ptime,htime);` in KKFit.hh:
```c++
      if(!std::isfinite(ptime)) throw cet::exception("RECO") << "KKFit::" << __func__
                                                             << ": Non-finite ptime for hit index " << strawidx << " for collection size " << chcol.size()
                                                             << ", hpos.z() = " << hpos.Z() << ", ppos.z() = " << ppos.Z() << ", vel.z() = " << vel.Z()
                                                             << "\n";

```
)

PR #1749, opened 2026-03-04
Further checks in MHF

bechenard:
Additional (and maybe redundant) checks to fix issue 1608

### DIFF
--- a/TrkPatRec/src/RobustMultiHelixFinder_module.cc
+++ b/TrkPatRec/src/RobustMultiHelixFinder_module.cc
@@ -22,6 +22,7 @@
 
 #include <numeric>
 #include <stdint.h>
+#include <cmath>
 
 
 namespace {
@@ -437,6 +438,7 @@ chi2dXY = bestHelix.fita_zt_;
           float denominator = 2*(x2*y3 - x3*y2);
           float numeratorX  = x12x12y12y12*y3 - x13x13y13y13*y2;
           float numeratorY  = x13x13y13y13*x2 - x12x12y12y12*x3;
+          if (std::abs(denominator)<1e-9) continue;
 
           float centerX = numeratorX/denominator+x1;
           float centerY = numeratorY/denominator+y1;
@@ -477,7 +479,7 @@ chi2dXY = bestHelix.fita_zt_;
              else if (chisq<val_prev) {val_prev = chisq; ilast=i;}
           }
           if (ilast<chcol.size()) {stubby.push_back(hits[ilast]);nsh += chcol[hits[ilast]].nStrawHits(); sumChisq+=val_prev;}
-          sumChisq /= nsh;
+          if (nsh>0) sumChisq /= nsh;
 
           art::Ptr<CaloCluster> thisCaloPtr{};
           if (caloPtr && caloPtr->energyDep()>ccMinEnergy_) {
@@ -557,7 +559,7 @@ chi2dXY = bestHelix.fita_zt_;
     //perform dz/dphi fit and filter hits based on the fit result
     init_dzdp(helix,chcol,helicity);
     fit_dzdp(helix,chcol,helicity);
-    if (abs(helix.fita_zp_)<1e-3) {helix.hitIndices_.clear(); helix.nStrawHits_=0; return;}
+    if (std::abs(helix.fita_zp_)<1e-3) {helix.hitIndices_.clear(); helix.nStrawHits_=0; return;}
     filterZPhi(helix,chcol,maxDPhiHelFit_);
 
     //Perform a dz/dt fit and filter hits based on the fit result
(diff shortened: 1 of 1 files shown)


oksuzian:
I'll review this pull request by first loading the relevant ability, then examining the PR details and code changes.Now let me look at the full file for broader context around these changes:Now I have the full picture. Here's my review of PR #1749 — "Further checks in MHF":

## PR Summary

This PR adds **4 lines** (0 deletions) to `TrkPatRec/src/RobustMultiHelixFinder_module.cc` to fix [issue #1608](https://github.com/Mu2e/Offline/issues/1608). The changes add additional validation checks on fit parameters in two locations.

---

## Changes Reviewed

### Change 1: In `findHelixCandidate()` — lines 592–593

```cpp name=TrkPatRec/src/RobustMultiHelixFinder_module.cc url=https://github.com/Mu2e/Offline/blob/2b9ae36543698ef0ea5b6595b47760a37cef4cde/TrkPatRec/src/RobustMultiHelixFinder_module.cc#L589-L594
    //final dz/dphi and dz/dt refit
    fit_dzdp(helix,chcol,helicity);
    fit_dzdt(helix,chcol);
    if (abs(helix.fita_zp_)<1e-3 || abs(helix.fita_zt_)<1e-3) {helix.hitIndices_.clear(); helix.nStrawHits_=0; return;}
    if (!std::isnormal(helix.fita_zp_)  || !std::isnormal(helix.fita_zt_))  {helix.hitIndices_.clear(); helix.nStrawHits_=0; return;}
```

### Change 2: In `fit_dzdp()` — line 697

```cpp name=TrkPatRec/src/RobustMultiHelixFinder_module.cc url=https://github.com/Mu2e/Offline/blob/2b9ae36543698ef0ea5b6595b47760a37cef4cde/TrkPatRec/src/RobustMultiHelixFinder_module.cc#L694-L699
    float fa = zphiFitter.fa();
    float fb = zphiFitter.fb();

    if (fabs(fa) < 1e-3) return;
    if ((helicity==Helicity::poshel && fa<1e-3) || (helicity==Helicity::neghel  && fa>-1e-3)) return;
```

---

## Issues Found

### 1. 🔴 Inconsistent use of `abs()` vs `fabs()` vs `std::abs()`

- **Line 592**: Uses `abs()` (unqualified) for `float` values
- **Line 697**: Uses `fabs()` for `float` values
- **Line 560** (pre-existing): Uses `abs()` for `float` values

The unqualified `abs()` on a `float` argument can resolve to the C integer `abs(int)` depending on which headers are included and compiler settings, silently truncating the float to an integer. This would make the check meaningless (e.g., `abs(0.5f)` → `0`). **Use `std::abs()` or `std::fabs()` consistently for float types.** This is a pre-existing issue throughout the file (lines 450, 544, 560, 641, 649, 679, 734, 735, 753, 815), but the new code at line 592 propagates it.

### 2. 🟡 Redundancy between the new `fabs(fa) < 1e-3` check and the existing helicity check in `fit_dzdp()`

At line 697, the new guard `if (fabs(fa) < 1e-3) return;` is added before the existing check:
```cpp
if ((helicity==Helicity::poshel && fa<1e-3) || (helicity==Helicity::neghel && fa>-1e-3)) return;
```

The existing check already catches `fa` values near zero for both helicities:
- `poshel`: returns if `fa < 1e-3` (covers `fabs(fa) < 1e-3` when `fa ≥ 0`)
- `neghel`: returns if `fa > -1e-3` (covers `fabs(fa) < 1e-3` when `fa ≤ 0`)

The new check is truly redundant *except* for the edge case where no data was added to the fitter and `fa == 0.0` exactly (since `LSFitter::fa()` returns `0.0` as a degenerate case). In that scenario, both checks already catch it. The PR description itself acknowledges this may be redundant. **This is fine as defensive programming but worth noting.**

### 3. 🟡 `std::isnormal()` rejects subnormals but also rejects `0.0`

At line 593:
```cpp
if (!std::isnormal(helix.fita_zp_) || !std::isnormal(helix.fita_zt_)) { ... }
```

`std::isnormal()` returns `false` for `0.0`, subnormals, infinity, and NaN. This is probably the intended behavior here (zero slopes are already rejected by the preceding line). However, if `fit_dzdp` returned early without updating the slopes (leaving them at their pre-existing values from `init_dzdp`), the slopes would retain their previous values — those could potentially be subnormal. This guard is good defensive coding for catching NaN/Inf propagation from degenerate fits.

### 4. 🟡 Missing `<cmath>` include for `std::isnormal`

The file includes `<numeric>` and `<stdint.h>`, but `std::isnormal` is defined in `<cmath>`. It likely compiles because `<cmath>` is transitively included by one of the other headers, but for correctness and portability, it should be explicitly included.

### 5. 🟠 Pre-existing: Typo in `filterZPhi` — `6.293185` instead of `6.283185` (2π)

This is **not introduced by this PR**, but is worth flagging. In `filterZPhi()` at lines 734–735:
```cpp
int   n      = round((phi-(chcol[ich].pos().z()-circle.fitb_zp_)/circle.fita_zp_)/6.293185);
float delta  = abs((chcol[ich].pos().z()-circle.fitb_zp_)/circle.fita_zp_-phi+n*6.293185);
```

The constant `6.293185` should be `6.283185` (2π). This appears to be a pre-existing typo that introduces a ~0.16% systematic error in the phi wrapping calculation.

### 6. 🟠 Pre-existing: Division by zero not guarded in `findCircleCandidate`

At line 437, `denominator = 2*(x2*y3 - x3*y2)` can be zero when the three points are collinear. The code divides by `denominator` on lines 441–442 without checking. Similarly, `nsh` at line 480 can be zero leading to division by zero in `sumChisq /= nsh`. These are pre-existing issues.

---

## Verdict

The changes are **small, focused, and defensive** — they guard against degenerate/pathological fit results (near-zero slopes, NaN, Inf) that could cause downstream issues (likely the crash in issue #1608). The logic is correct and appropriate.

**Suggested improvements before merging:**
1. Replace `abs()` with `std::abs()` on line 592 (and ideally throughout the file) to avoid potential integer `abs()` resolution
2. Consider adding `#include <cmath>` explicitly for `std::isnormal`

PR #1762, opened 2026-03-13
Updates to DBSClusterer and FlgBkgHits

NamithaChitrazee:
1. In DBSClusterer module, added the functions to classify clusters based on the keras quailty  
2. In flgBkgHits, added a function to flag proton hits as background as well. 
3. Added the newly trained .dat and .hxx files (using CE1BB and pbar datasets)  

More small refinements will follow for this update.


oksuzian:
@FNALbuild run build test

PR #1768, opened 2026-03-14
add a list of panels to skip while filtering

pavel1murat:
a few panels have eDep distributions significantly wider that the majority. Skip them in the simplest single hit eDep-based trigger


bonventre (changes requested):
Presumably mnid checking instead of straw id so you don't have to worry about the map changing, but I think that's already practically a nonissue, so we should keep mnid out of offline. I'd prefer to have this all using status conditions

pavel1murat:
this PR implements a triggering solution deemed to be a quick but temporary one. The long term solution will involve identifying and masking off individual straws which create the problem - see , for example [this plot](https://mu2e-elog.physics.northwestern.edu/elog/Tracker/3985). Given that this solution will shortly go away, investing an effort into polishing the user interface is a bit of an overkill.

oksuzian:
@FNALbuild run build test

PR #1778, opened 2026-04-02
Calo filter

bechenard:
Updated calo filter without energy and time variables. The NN evaluation and the filtering are also separated. Need a mu2e-trig-config PR to work


CaloDiag/src/CaloNNDiag_module.cc
```diff
@@ -322,90 +321,73 @@ namespace mu2e {
...
-          {
-              const CaloHit& hit    = CaloHits.at(cryList[1]);
-              CLHEP::Hep3Vector hitPos = cal.geomUtil().mu2eToDiskFF(cluster.diskID(),cal.crystal(hit.crystalID()).position());
-              double dx = hitPos.x()-seedPos.x();
-              double dy = hitPos.y()-seedPos.y();
-              dr = sqrt(dx*dx+dy*dy);
+          bool isConv(false);
+          if (itMC != caloClusterTruth.end()){
+             for (auto& edep : itMC->second->energyDeposits()){
+                if (edep.sim()->creationCode() == ProcessCode::mu2eCeMinusEndpoint ||
```
michaelmackenzie: Please add CE+ to this as well, or make the signal process code be a fhicl config parameter so it can test multiple potential signals

CaloDiag/src/CaloNNTrain_module.cc
```diff
@@ -0,0 +1,187 @@
...
+
+     const Calorimeter& cal = *(GeomHandle<Calorimeter>());
+     const CaloClusterCollection&  caloClusters(*caloClustersHandle);
+     const CaloClusterMCTruthAssn& caloClustersMC(*caloClustersMCHandle);
+
+     for (const auto& cluster : caloClusters)
+     {
+        if (cluster.energyDep() < minEtoTest_) continue;
+
+        const auto& hits          = cluster.caloHitsPtrVector();
```
michaelmackenzie: Why not use the util class here instead of re-implementing the logic?

CaloFilters/src/CaloNNEval_module.cc
```diff
@@ -0,0 +1,178 @@
...
+          continue;
+        }
+
+        const auto& hits         = clusterIt->caloHitsPtrVector();
+        const auto& neighborsId  = cal.crystal(hits[0]->crystalID()).neighbors();
+        const auto& nneighborsId = cal.crystal(hits[0]->crystalID()).nextNeighbors();
+
+        float e1(hits[0]->energyDep());
+        float e2(hits[0]->energyDep());
+        if (hits.size()>1) e2 += hits[1]->energyDep();
```
michaelmackenzie: Again, why not use the utils? This adds more opportunity for accidental divergence in the train vs. test logic implementation

michaelmackenzie (approved):
This looks good to me. The trigger results change, but this is due to the updated ordering of calo digis in the Online calo hit maker. I've made a couple of small requests, but I don't think any are required

CaloDiag/src/CaloNNDiag_module.cc
```diff
@@ -322,90 +321,73 @@ namespace mu2e {
...
+       cluSimMom_.clear();cluSimMom2_.clear();cluSimStartX_.clear();cluSimPosY_.clear();cluSimPosZ_.clear();
+       cluSimStartY_.clear();cluSimStartZ_.clear();
 
        for (unsigned  ic=0; ic<caloClusters.size();++ic)
        {
           const CaloCluster& cluster = caloClusters.at(ic);
           std::vector<int> cryList;
           for (auto cryPtr : cluster.caloHitsPtrVector()) cryList.push_back(std::distance(&CaloHits.at(0),cryPtr.get()));
 
+          if (cluster.energyDep() < minCluEnergy_) continue;
```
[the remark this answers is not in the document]
bechenard: IT is a diagnosis module so it doesn't really matter....

CaloFilters/src/CaloNNEval_module.cc
bechenard: It was a bit faster and more contained this way, but I agree it is less maintenance with the Utils. Might be a tad bit slower but that's ok

bechenard:
@FNALbuild run build test with https://github.com/Mu2e/mu2e-trig-config/pull/152

bechenard:
Ok, I have fixed some of the AI stuff, most were actually moot.

oksuzian:
I'll review this pull request by examining its details and changes.Let me get the actual diff to review the changes:Now let me look at the new `CaloNNEval_module.cc` and the modified `FilterEcalNNTrigger_module.cc` more closely, and also check for the `MVAResult` data product:Now let me look at the new `CaloNNEval_module.cc` and `FilterEcalNNTrigger_module.cc` more carefully, and also get the existing files for context:Now I have a thorough view of the entire PR. Let me review the new `CaloNNEval_module.cc` more closely and check the existing `FilterEcalNNTrigger_module.cc` for correctness.

Here is my detailed review of **PR #1778 – Calo filter** in `Mu2e/Offline`:

---

## Summary

This PR refactors the calorimeter NN trigger filter by:
1. Separating NN evaluation (`CaloNNEval`) from filtering (`FilterEcalNNTrigger`)
2. Adding a new `CaloNNTrain` diagnostic analyzer
3. Adding new "fast" hit/cluster makers to the reco chain
4. Updating `CaloNNDiag` to use `vector<vector<>>` for MC truth (better ROOT branch layout) and adding validity checks
5. Fixing the `secondMoment()` calculation (normalizing by `sw`)
6. Updating NN weights (adding `cluTime`, `cluSec`, and ratio-based energy features)
7. Refactoring `CaloHitMakerFast` (sorted digis, changed map type, new pulse-merging logic)

---

## Issues Found

### 1. 🐛 **Typo in exception message** — `FilterEcalNNTrigger_module.cc`

```cpp name=CaloFilters/src/FilterEcalNNTrigger_module.cc url=https://github.com/Mu2e/Offline/pull/1778/files#diff-82
throw cet::exception("FILTER")<< "FilterEcalNNTrigger: Clusters and MVA collection sizes incpmpatible\n";
```

`"incpmpatible"` → should be `"incompatible"`.

---

### 2. 🐛 **Missing space in diagnostic output** — `CaloNNEval_module.cc` (line ~128)

```cpp name=CaloFilters/src/CaloNNEval_module.cc
if (diagLevel_>0) std::cout<<"CaloNNEval cluster with e="<<clusterIt->energyDep()
                           <<" and r="<< clusterIt->cog3Vector().perp()
                           <<"has mvaout="<<mvaout<<"\n";
```

`"has mvaout="` is missing a leading space — the output will read `"...r=450.2has mvaout=0.8"`. Should be `" has mvaout="`.

---

### 3. ⚠️ **Hard-coded MVA vector size of 7 doesn't match all weight files** — `CaloNNEval_module.cc`

The `CaloNNEval` module always builds 7 features:

```cpp name=CaloFilters/src/CaloNNEval_module.cc
std::vector<float> mvavars(7,0.0);
mvavars[0] = clusterIt->cog3Vector().perp();
mvavars[1] = clusterIt->size();
mvavars[2] = cluUtil.e1()/clusterIt->energyDep();
mvavars[3] = cluUtil.e2()/clusterIt->energyDep();
mvavars[4] = cluUtil.e9()/clusterIt->energyDep();
mvavars[5] = cluUtil.e25()/clusterIt->energyDep();
mvavars[6] = cluUtil.secondMoment();
```

But the updated `CE_NN_ReLU.weights.xml` now has **9 variables** (adds `cluEnergy`, `cluTime`). If `CaloNNEval` is ever configured to use `CE_NN_ReLU.weights.xml` instead of `CE_NN_ReLU_noET.weights.xml`, the variable count mismatch will cause incorrect results or a crash. Consider either:
- Making the feature vector construction configurable / driven by the MVA metadata, or
- Adding a runtime check that the number of variables matches what the weights file expects.

---

### 4. ⚠️ **Potential division by zero in `secondMoment()`** — `ClusterUtils.cc` (line 62)

```cpp name=CaloCluster/src/ClusterUtils.cc url=https://github.com/Mu2e/Offline/blob/49d2d9d0bcceff7ebc0249f541c318dac7b1b59d/CaloCluster/src/ClusterUtils.cc#L58-L63
double ClusterUtils::secondMoment() const
{
    double sx(0),sy(0),sz(0),sx2(0),sy2(0),sw(0);
    fill(sx,sy,sz,sx2,sy2,sw);
    return (sx2-sx*sx/sw + sy2-sy*sy/sw)/sw;
}
```

If all hits are on a different disk from `hits_[0]`, the `fill()` method skips them all and `sw` remains 0, causing division by zero. The same issue exists for `cog3Vector()`. While this may be unlikely in practice, it would be prudent to guard against `sw == 0`.

---

### 5. ⚠️ **`addPulse` logic change may silently discard SiPM matches** — `CaloHitMakerFast_module.cc`

The new logic:
```cpp name=CaloReco/src/CaloHitMakerFast_module.cc
addNewHit    = false;
float ratio  = (eDep-pulse.eDep_)/(eDep+pulse.eDep_);
float sigmaR = 0.707*sqrt(1.0/eMean/nPEperMeV_ + noise2_/eMean/eMean);

if (fabs(ratio) < nSigmaNoise_*sigmaR) {pulse.add(time,eDep);}
else if (eDep>pulse.eDep_)             {pulse.time_=time; pulse.eDep_=eDep;}
break;
```

When a digi is within `deltaTPulses_` but fails the energy consistency check **and** has less energy (`eDep <= pulse.eDep_`), the digi is silently dropped — `addNewHit` is `false` and the `break` exits the loop. This is a deliberate design choice, but it differs from the previous behavior (which would continue looking for another matching pulse). Worth confirming this is intentional.

---

### 6. ⚠️ **`unordered_map` → `map` changes ordering behavior** — `CaloHitMakerFast_module.cc`

```cpp name=CaloReco/src/CaloHitMakerFast_module.cc
using pulseMapType = std::map<unsigned, std::vector<HitInfo>>;
```

Changing from `unordered_map` to `std::map` ensures deterministic output ordering (good for reproducibility) but introduces O(log n) lookup overhead. For the digi volumes in this experiment, this is likely fine, but it's a physics-relevant change since output order may affect downstream consumers. Ensure this doesn't cause validation issues.

---

### 7. 💡 **`CaloNNDiag` early-returns before `Ntup_->Fill()` on missing handles**

The new validity checks are good defensive coding:
```cpp
if (!CaloHitsHandle.isValid()) return;
if (!caloClustersHandle.isValid()) return;
```

However, be aware that the tree will simply not get filled for events where handles are missing — no counter or warning is emitted. If this module is used in analysis and handles are unexpectedly missing, it may be confusing to debug. Consider adding a warning message.

---

### 8. 💡 **`secondMoment()` formula change is a physics-relevant correction**

The original formula:
```
sx2 - sx*sx/sw + sy2 - sy*sy/sw
```
was changed to:
```
(sx2 - sx*sx/sw + sy2 - sy*sy/sw) / sw
```

This is a fix to normalize the second moment properly (variance = (Σw·x² − (Σw·x)²/Σw) / Σw). This changes the numeric value of `cluSecMom` and any downstream cuts or MVA features that depend on it. The new NN weights are presumably trained with this corrected definition, so consistency is maintained — but any **existing** downstream analysis code or cuts using the old `secondMoment()` will break.

---

### 9. 💡 **New modules added to default `Reco` sequences**

`CaloHitFastMaker` and `CaloClusterFastMaker` are added to the default `Reco` paths in the prolog files. This means **all** reconstruction jobs will now run these fast reco modules even if the NN trigger isn't used. This adds processing time to every reco job. Consider whether these should be in a separate optional sequence.

---

### Summary Table

| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | 🐛 Bug | `FilterEcalNNTrigger_module.cc` | Typo: `"incpmpatible"` |
| 2 | 🐛 Bug | `CaloNNEval_module.cc` | Missing space in diagnostic output |
| 3 | ⚠️ Risk | `CaloNNEval_module.cc` | Hard-coded 7-feature vector incompatible with 9-var weight file |
| 4 | ⚠️ Risk | `ClusterUtils.cc` | Division by zero if `sw == 0` |
| 5 | ⚠️ Risk | `CaloHitMakerFast_module.cc` | Digis silently dropped when energy mismatch and lower energy |
| 6 | ⚠️ Minor | `CaloHitMakerFast_module.cc` | `map` vs `unordered_map` changes output ordering |
| 7 | 💡 Note | `CaloNNDiag_module.cc` | Silent early returns with no logging |
| 8 | 💡 Note | `ClusterUtils.cc` | `secondMoment()` normalization is a physics-breaking change |
| 9 | 💡 Note | Prolog fcl files | Fast reco modules added to default `Reco` sequence |

bechenard:
I had a loing conversation with my inner self, and we're fine with two typos. Just merge this PR

michaelmackenzie:
@bechenard please see the requested trigger config changes here:
https://github.com/Mu2e/mu2e-trig-config/pull/152

The head of the repos and the CI will be broken until we can merge this

PR #1787, opened 2026-04-07
Muon stop pileup cz restriction

michaelmackenzie:
This restricts the mu stop pileup cos(theta_z), and accordingly adjusts the rates for the processes so the pileup module will produce sensible distributions when only looking in this cos(theta_z) region. This significantly speeds up Run 1B pileup production for rough estimates that don't need exact precision.


michaelmackenzie:
@FNALbuild run build test

EventGenerator/src/DIOGenerator_tool.cc
```diff
@@ -25,6 +25,7 @@ namespace mu2e {
 
       fhicl::Atom<double>       czmin   {Name("czmin")   , Comment("Restrict cos(theta_z) minimum"), -1.};
       fhicl::Atom<double>       czmax   {Name("czmax")   , Comment("Restrict cos(theta_z) maximum"),  1.};
+      fhicl::Atom<bool>         fireAll {Name("fireAll") , Comment("Always add a particle to the event"), false};
```
brownd1978: What is the use case for fireAll = false? Shouldn't a generator produce an output ever event? Without that, downstream workflows that rely on MCPrimary will fail. You could add a filter in this case, but how would that be different from fireAll = true?

brownd1978 (commented):
I'm curious what prompted the addition of fireAll option.

michaelmackenzie:
@brownd1978 this `fireAll` is necessary as the module is both used in the muon stop pileup generator and as a standalone DIO generator. In the case of pileup, it would need to produce a DIO only the (czmax - czmin)/2 fraction of the time to keep rates physically consistent. The other pileup tool inputs similarly only produce outputs some fraction of the time, returning an empty collection if they don't fire. In the case where you're specifically simulating this process and not pileup, fireAll should be set (similar to the 1.8 MeV photon line generator) to ensure you always generate your primary of interest.

I can set this to be true for the DIO tail generator config in Production, or I can set it to false in the pileup tool config and have it default to true, whichever is preferred. I can also remove the default and make it explicit in all cases.

brownd1978:
@michaelmackenzie thanks for the explanation. As it's a formal error for a primary generator to not produce a primary, a way of insuring that in code would be better than a flag that can be misconfigured. How about passing this as a constructor argument, true when the generator is primary, false as part of pileup?

I also think some solution to recording the cz  acceptance is required; we should not go back to harvesting numbers from log files to get correct downstream job configurations.

michaelmackenzie:
I updated to pass the flag for `isPrimary_` at `finishInitialization` when the physical rates are often configured. I will also develop the `SpectrumConfig` data product in a separate PR to ensure we don't need to harvest these parameters/corrections from fcl and log files. Please let me know if there are any other updates I should make to this PR!

michaelmackenzie:
@FNALbuild run build test

EventGenerator/src/MuCap1809keVGammaGenerator_tool.cc
```diff
@@ -39,7 +37,8 @@ namespace mu2e {
     std::vector<ParticleGeneratorTool::Kinematic> generate() override;
     void generate(std::unique_ptr<GenParticleCollection>& out, const IO::StoppedParticleF& stop) override;
 
-    void finishInitialization(art::RandomNumberGenerator::base_engine_t& eng, const std::string& material) override {
+    void finishInitialization(art::RandomNumberGenerator::base_engine_t& eng, const std::string& material, const bool isPrimary) override {
+      isPrimary_ = isPrimary;
```
brownd1978: It's confusing that the underscore postpends vs prepends on the other variables

brownd1978: Are there use cases where fireall and isprimary can be different? If not, I suggest to consolidate them.

EventGenerator/src/MuCapNeutronGenerator_tool.cc
```diff
@@ -25,23 +26,31 @@ namespace mu2e {
...
+    {
+      if(_czMin < -1. || _czMax > 1. || _czMin > _czMax) throw cet::exception("BADCONFIG") << "Cos(theta_z) range unphysical: " << _czMin << " - " << _czMax;
+    }
 
     std::vector<ParticleGeneratorTool::Kinematic> generate() override;
     void generate(std::unique_ptr<GenParticleCollection>& out, const IO::StoppedParticleF& stop) override;
 
-    void finishInitialization(art::RandomNumberGenerator::base_engine_t& eng, const std::string& material) override {
+    void finishInitialization(art::RandomNumberGenerator::base_engine_t& eng, const std::string& material, const bool isPrimary) override {
+      isPrimary_ = isPrimary;
```
brownd1978: Same comment about the underscore as above

brownd1978 (approved):
Minor stylistic comments, and a question if isprimary and fireall can be consolidated

EventGenerator/src/MuCap1809keVGammaGenerator_tool.cc
michaelmackenzie: I switched from `isPrimary_` --> `_isPrimary` as suggested since this better matches the style used by the tools.

I also believe all cases of `fireAll` have been removed with this `isPrimary` flag update, as this new flag takes the role of the `fireAll` (as you point out). I've similarly updated the other capture product tools to always produce a single particle if configured a a primary generator instead of the Poisson sampling

oksuzian:
@FNALbuild run build test

Issue #825, opened 2022-06-30
Read the geometry file earlier?


kutschke:
When we created Offline, the Geometry service was configured to read the geometry file at run-start.  This was in anticipation that the geometry would be run dependent.   We have since moved to the design that Geometry service contains only  nominal geometry, which is not run dependent.

Dave asks if we can move the initialization of the reading of the geometry file to the service c'tor.  This would all the initialization of some of the KinKal nominal geometry at module c'tor time.   Currently he passes it around using the KinKal event object.

Should we proceed with this?  if so, what other use cases do we need to think through?

gaponenko:
Making geometry available to module constructors would be an improvements.   I had to invent workarounds in the past because it was not accessible.

brownd1978:
I've fixed this for KinKal in PR #1792 by moving KinKalGeom into GeometryService. This is a better model as KinKalGeom becomes a singleton.

Issue #456, opened 2021-05-19
BField Information will be time dependent and so belongs in Proditions


kutschke:
The BField information will be time dependent and should be removed from the Geometry Service and added to Proditions.   Presuming that we stay with field maps stored in files, the Proditions information will probably be the list of filenames that are needed.

When designing the Proditions configuration remember that there are several distinct use cases.

1. Some simulations need all of the field maps ( ie cosmic sims )
2. Some simulations may need only a subset of the maps.
3. Reconstruction requires only the DSMap - actually it needs only a subset of the DSMap.
4. Triggering requires only a very small subset of the DSMap

Is the Proditions "Purpose" field the right way to manage these cases?

In the above I have assumed that we do not have a good reason to keep a nominal bfield in the GeometryService, in the same spirit as we do for the Tracker.  Does anyone thing otherwise?

PR #1792, opened 2026-04-09
Create KinKalGeom from GeometryService objects

brownd1978:
This PR addresses the issue that the geometry used for KinKal intersections was hard-coded, not tied to the GeometryService. Fixing this required deferring the creation of several geometry objects from module construction time to first access, as the GeometryService itself isn't accessible during module construction. 
Problems were found in the consistency of GeometryService with the actual G4 model. For instance, the positions of StepPointMCs recorded in the virtual detector TT_Front, TT_Mid, and TT_Back were found to be different from what the GeometryService calculates for those objects. Some of the difference might be attributable to the VD thickness (half-length) of 0.01mm, but that doesn't account for all the differences. For now, the KKGeom positions are adjusted with kludge factors to agree with the observed StepPointMC positions, as otherwise there will be differences comparing (say) momentum at VDs between reco and MC.

The final change is to move the creation of the KinKalGeom surfaces used in reco into GeometryService, and serve KinKalGeom itself as a singleton from that service. That removes all the duplication, cleans up the memory model, and avoids some spurious dependencies.

This PR doesn't address the fact that GeometryService exposes bare c++ pointers moving objects from makers to the service, which uses boost::shared_ptr. A better model is to use std::shared_ptr everywhere, but that's beyond the scope of this PR.


oksuzian:
I'll review this pull request by examining its details and changes.Now let me fetch the full files to review the changes in more detail:Here's my review of PR #1792, which refactors KinKalGeom objects to be created from GeometryService instead of hard-coded values, using a lazy-initialization pattern.

---

## Issues Found

### 1. 🔴 Thread Safety: Lazy Initialization Without Synchronization

The `check_init()` pattern used in both `Tracker` and `SurfaceMap` is **not thread-safe**. The `mutable bool initialized_` flag is read and written without any mutex or `std::call_once` protection. In a multi-threaded art framework, concurrent const-method calls from different threads could race on initialization.

```cpp name=KinKalGeom/src/Tracker.cc url=https://github.com/Mu2e/Offline/blob/b21611dbbdabf40f092e4084bceec78cf5141fb8/KinKalGeom/src/Tracker.cc#L17-L23
    void Tracker::check_init() const {
      if(!initialized_){
        // cast off const
        (const_cast<Tracker*>(this))->initialize();
        initialized_ = true;
      }
    }
```

The same pattern appears identically in `SurfaceMap::check_init()`. Consider using `std::once_flag` + `std::call_once`, or `std::atomic<bool>` with proper memory ordering, to guard initialization.

### 2. 🔴 `SurfaceMap::surface()` Bypasses Lazy Initialization

```cpp name=KinKalGeom/inc/SurfaceMap.hh url=https://github.com/Mu2e/Offline/blob/b21611dbbdabf40f092e4084bceec78cf5141fb8/KinKalGeom/inc/SurfaceMap.hh#L25-L28
      auto const& map() const { check_init(); return map_; }
      // find a surface by its Id.  Return value is an iterator, which may be null.  Note that if
      // a generic index (-1) is given for surfaces with >1 value, this will return a valid but unspecified matching surface
      auto surface(SurfaceId const& sid) const { return map_.find(sid); }
```

`map()` correctly calls `check_init()`, but `surface()` does **not**. If `surface()` is called before `map()` or any other initializing accessor, it will search an empty map and silently return `map_.end()`. The same issue exists for the `surfaces()` method, `DS()`, `ST()`, `tracker()`, and `TCRV()` — none of them call `check_init()`.

### 3. 🟡 Redundant Lazy Init: `SurfaceMap` Triggers `Tracker` Init, Creating Double-Init Path

`SurfaceMap::initialize()` calls `tracker_.frontPtr()`, `tracker_.middlePtr()`, etc., which each call `Tracker::check_init()`. This means `Tracker` lazy-initializes itself during `SurfaceMap`'s own initialization. While functionally correct, it's fragile and confusing — the initialization ordering is implicit. If `SurfaceMap::initialize()` were refactored to access `Tracker` members directly (bypassing accessors), it would break silently.

### 4. 🟡 Unexplained Hard-Coded "Kludge" Corrections

```cpp name=KinKalGeom/src/Tracker.cc url=https://github.com/Mu2e/Offline/blob/b21611dbbdabf40f092e4084bceec78cf5141fb8/KinKalGeom/src/Tracker.cc#L37-L40
      // I don't understand the origin of the corrections below.
      double zFrontLocal  = zFrontGlobal - tracker.g4Tracker()->z0() + 0.39;
      double zBackLocal   = zBackGlobal  - tracker.g4Tracker()->z0() - 0.41;
      double zMidLocal = 10.1 - 0.01; // 10.1 is hard-coded in VirtualDetectorMaker line 224. The 0.01 is another hack.
```

The author explicitly notes they don't understand the origin of the `+0.39` and `-0.41` corrections. These magic numbers are a maintenance risk — if VirtualDetectorMaker changes, these will silently become wrong. At minimum, these should be documented with their derivation or extracted as named constants with comments linking to the source geometry.

### 5. 🟡 Unused Variable `ttOffset_det`

```cpp name=KinKalGeom/src/Tracker.cc url=https://github.com/Mu2e/Offline/blob/b21611dbbdabf40f092e4084bceec78cf5141fb8/KinKalGeom/src/Tracker.cc#L42
      CLHEP::Hep3Vector ttOffset_det = det->toDetector(ttOffset);
```

This variable is computed but never used. The `DetectorSystem` `GeomHandle` is obtained and the offset converted, but `ttOffset_det` is not referenced anywhere. This is likely a leftover from development; it should be removed or used.

### 6. 🟡 Commented-Out Debug Code Left In

```cpp name=KinKalGeom/src/Tracker.cc url=https://github.com/Mu2e/Offline/blob/b21611dbbdabf40f092e4084bceec78cf5141fb8/KinKalGeom/src/Tracker.cc#L46
     // std::cout << " zfronloc " << zFrontLocal << " zmidloc " << zMidLocal << " zbackloc " << zBackLocal << " orvd " << orvd << " irvd " << irvd << std::endl;
```

This should be removed before merging, or converted to a proper debug logging mechanism.

### 7. 🟡 Performance: Geometry Objects Reconstructed on Every Call in `KKExtrap`

Every template method in `KKExtrap` now creates fresh `ExtrapolateToZ`, `ExtrapolateIPA`, and `ExtrapolateST` objects on each invocation. For example, `extrapolate()` calls `toTrackerEnds()` which creates its own `trackerFront` and `trackerBack` — and `extrapolate()` also creates a separate `trackerFront`. This means `smap_.tracker().front().center().Z()` is queried multiple times per track, and the lazy-init `check_init()` is hit repeatedly.

```cpp name=Mu2eKinKal/inc/KKExtrap.hh url=https://github.com/Mu2e/Offline/blob/b21611dbbdabf40f092e4084bceec78cf5141fb8/Mu2eKinKal/inc/KKExtrap.hh#L67-L68
  template <class KTRAJ> void KKExtrap::extrapolate(KKTrack<KTRAJ>& ktrk) const {
    ExtrapolateToZ trackerFront(maxdt_,btol_,smap_.tracker().front().center().Z(),debug_);
```

```cpp name=Mu2eKinKal/inc/KKExtrap.hh url=https://github.com/Mu2e/Offline/blob/b21611dbbdabf40f092e4084bceec78cf5141fb8/Mu2eKinKal/inc/KKExtrap.hh#L105-L107
  template <class KTRAJ> void KKExtrap::toTrackerEnds(KKTrack<KTRAJ>& ktrk) const {
    ExtrapolateToZ trackerFront(maxdt_,btol_,smap_.tracker().front().center().Z(),debug_);
    ExtrapolateToZ trackerBack(maxdt_,btol_,smap_.tracker().back().center().Z(),debug_);
```

While the old member-variable approach had the constraint of being constructed too early (before GeometryService), the current per-call reconstruction is wasteful. These could be cached after first use with the same lazy-init pattern, or initialized once when `smap_` is first accessed.

### 8. 🟡 Copy of Full Tracker Geometry Object

```cpp name=KinKalGeom/src/Tracker.cc url=https://github.com/Mu2e/Offline/blob/b21611dbbdabf40f092e4084bceec78cf5141fb8/KinKalGeom/src/Tracker.cc#L28
      auto tracker = *(GeomHandle<mu2e::Tracker>());
```

This dereferences the `GeomHandle` and copies the entire `mu2e::Tracker` object. This should be a const reference instead:
```cpp
auto const& tracker = *(GeomHandle<mu2e::Tracker>());
```

### 9. 🟢 PR Description Acknowledges Singleton Problem Not Yet Fixed

The description mentions: *"multiple instances of KinKalGeom objects are being created in the course of normal module execution, while it should be a singleton."* Both `KKExtrap` and `KKFit` now hold their own `SurfaceMap smap_` member, each of which contains a full copy of `Tracker`, `StoppingTarget`, `DetectorSolenoid`, and `TestCRV`. This is acknowledged as deferred to a follow-up PR, but worth flagging as technical debt.

### 10. 🟢 Minor: Typo in Comment

```cpp name=Mu2eKinKal/inc/KKFit.hh url=https://github.com/Mu2e/Offline/blob/b21611dbbdabf40f092e4084bceec78cf5141fb8/Mu2eKinKal/inc/KKFit.hh#L210
    // Lookup surfaces to sample; this interface is deprecatecd and should be replaced with extrapolation TODO
```

`deprecatecd` → `deprecated`

---

## Summary

| Severity | Count | Key Areas |
|----------|-------|-----------|
| 🔴 High | 2 | Thread-unsafe lazy init; `surface()`/`surfaces()`/`DS()`/`ST()`/`tracker()`/`TCRV()` skip `check_init()` |
| 🟡 Medium | 5 | Magic kludge numbers, unused variable, debug leftovers, per-call object construction, full tracker copy |
| 🟢 Low | 2 | Acknowledged singleton debt, typo |

The most critical issues are the **missing `check_init()` calls** in several `SurfaceMap` accessors (which will silently return wrong results) and the **thread-unsafe lazy initialization** pattern. I'd recommend addressing at least the missing `check_init()` calls before merging.

rlcee:
If I am scanning the code correctly, the reason you have to lazy init is because GeometryService is designed run-dependent, but the design here avoids that.  I see this is captured in Issue #825 and #456.  If a final decision is made there, then a path to resolve the conflict would be clear.

oksuzian:
Let me look at the PR details and comments to understand the inconsistencies mentioned.Now let me look at the changed files to understand the specific inconsistencies between GeometryService and the G4 model:Now let me look at the VirtualDetectorMaker to understand the source of those inconsistencies:Now I have a thorough understanding of the inconsistencies. Let me look at the specific coordinate calculation in VirtualDetectorMaker to compare with the PR's Tracker.cc:Now I have a thorough understanding of the inconsistencies. Let me provide the analysis.

## Sources of the Geometry Inconsistencies

The PR description mentions that **the positions of StepPointMCs recorded at virtual detectors TT_Front, TT_Mid, and TT_Back differ from what the GeometryService calculates**. After tracing through both the `VirtualDetectorMaker` and the new `KinKalGeom::Tracker::initialize()` code, here are the specific sources of inconsistency:

### 1. Two-step placement vs. single-step calculation

The VDs are placed in a **two-step process**: `VirtualDetectorMaker` computes an offset, then `constructVirtualDetectors.cc` in Mu2eG4 places the actual G4 volume using `vdg->getGlobal(vdId)` relative to a *parent volume*. The `addVirtualDetector` call combines a `ttOffset` (global position of the tracker origin) with a local offset. The actual G4 placement then repositions the VD relative to a different parent (`DS3Vacuum` for Front/Back, `TrackerMother` for Mid):

```c++ name=Mu2eG4/src/constructVirtualDetectors.cc
G4ThreeVector vdLocalOffset = vdg->getGlobal(vdId) - parent.centerInMu2e();
```

This re-referencing to a different parent volume's center can introduce small offsets if the parent volume center doesn't perfectly cancel out. The KinKalGeom code tries to reproduce this from `GeometryService` alone, but the G4 placement has its own chain of coordinate transforms.

### 2. TT_Front and TT_Back: The `+0.39` / `-0.41` mm corrections

In `VirtualDetectorMaker` (lines 241-244), the local z-positions are computed cleanly:

```c++ name=GeometryService/src/VirtualDetectorMaker.cc url=https://github.com/Mu2e/Offline/blob/1d1b9f02f77764136ebbcebc6a54bfd008c3b924/GeometryService/src/VirtualDetectorMaker.cc#L241-L244
double zFrontGlobal = tracker.g4Tracker()->mother().position().z()-tracker.g4Tracker()->mother().tubsParams().zHalfLength()-vdHL;
double zBackGlobal  = tracker.g4Tracker()->mother().position().z()+tracker.g4Tracker()->mother().tubsParams().zHalfLength()+vdHL;
double zFrontLocal  = zFrontGlobal - tracker.g4Tracker()->z0();
double zBackLocal   = zBackGlobal  - tracker.g4Tracker()->z0();
```

But in G4, these VDs are placed in `DS3Vacuum` (not TrackerMother), using `vdg->getGlobal(vdId) - parent.centerInMu2e()`. The `DS3Vacuum` center is **not the same** as the tracker z0 used to compute the local offset. The discrepancy of ~0.39–0.41 mm likely arises from:

- **The difference between `tracker.g4Tracker()->z0()` and the DS3Vacuum parent center** used during G4 placement. The `addVirtualDetector` stores the global position as `ttOffset + localOffset`, but the actual G4 placement subtracts `DS3Vacuum.centerInMu2e()`, which doesn't equal `ttOffset`. The residual is roughly the solenoid-offset effect on the z-component, combined with rounding in the geometry description.
- **The VD half-length (0.01 mm)** accounts for a small part, but not the full ~0.4 mm. The remaining offset suggests the `mother().position().z()` used in VirtualDetectorMaker doesn't exactly match the actual G4 volume center used when placing the TT_FrontHollow and TT_Back volumes.

### 3. TT_Mid: Hard-coded `10.1` with a brass-ring conditional

In `VirtualDetectorMaker` (line 224), TT_Mid's z-offset depends on a config flag:

```c++ name=GeometryService/src/VirtualDetectorMaker.cc url=https://github.com/Mu2e/Offline/blob/1d1b9f02f77764136ebbcebc6a54bfd008c3b924/GeometryService/src/VirtualDetectorMaker.cc#L222-L224
Hep3Vector vdTTMidOffset(0.,0.,0.);
// Version 4 adds brass rings in Tracker, have to move vd to the side
if ( c.getBool("TrackerHasBrassRings",false) ) vdTTMidOffset.setZ(10.1);
```

The KinKalGeom code hard-codes `10.1` (assuming brass rings are present), but then subtracts `0.01`:

```c++ name=KinKalGeom/src/Tracker.cc url=https://github.com/Mu2e/Offline/blob/b21611dbbdabf40f092e4084bceec78cf5141fb8/KinKalGeom/src/Tracker.cc#L40
double zMidLocal = 10.1 - 0.01; // 10.1 is hard-coded in VirtualDetectorMaker line 224. The 0.01 is another hack.
```

The `- 0.01` is the VD half-thickness correction (to go from VD center to the entry surface), but this only applies in one direction depending on particle trajectory. The VD in G4 is a slab centered at 10.1 mm, and the StepPointMC is recorded at the entry face, which is 10.1 ± 0.01 depending on particle direction.

### 4. VD thickness and step recording position

VDs have a half-thickness of 0.01 mm. In G4, StepPointMCs are recorded **where the particle enters the VD volume**, not at the center. So:
- For a particle traveling in +z, TT_Front records at `zFrontCenter - 0.01`
- For a particle traveling in +z, TT_Back records at `zBackCenter - 0.01`

But the KinKal surfaces are infinitely thin planes at a single z. This introduces a ±0.01 mm systematic depending on track direction, which the code acknowledges in its comments.

### 5. Coordinate system mismatch

The computed `ttOffset_det = det->toDetector(ttOffset)` on line 42 of Tracker.cc converts to the DetectorSystem coordinates but is **never actually used** — the surfaces are built in tracker-local coordinates. This suggests the author may have been investigating whether a DetectorSystem transform was needed to resolve the discrepancy but hadn't finished.

### Bottom Line

The root cause is that **GeometryService computes VD positions in one coordinate path** (`ttOffset` + local offset), while **G4 places them via a different parent-volume-relative path** (`getGlobal - parent.centerInMu2e()`). These two paths don't perfectly agree because the parent volumes (DS3Vacuum, TrackerMother) have centers that don't exactly cancel the ttOffset. The ~0.4 mm kludge corrections are compensating for this coordinate system mismatch. A clean fix would require either making G4 placement use exactly the same calculation as GeometryService, or computing the KinKalGeom positions by going through the same `getGlobal()` → parent subtraction path that G4 uses.

brownd1978:
@FNALbuild run build test

brownd1978:
@FNALbuild run build test

brownd1978:
@FNALbuild run build test

rlcee (approved):
After spending an hour with this, I'd say I didn't see anything unreasonable or unexpected.

PR #1796, opened 2026-04-11
Add optional proton time cluster output, update printouts

michaelmackenzie:
Optionally produce proton time clusters for proton triggering and proton tracking. Also update some printouts for debugging.

### DIFF
--- a/CalPatRec/inc/TZClusterFinder_types.hh
+++ b/CalPatRec/inc/TZClusterFinder_types.hh
@@ -71,6 +71,7 @@ namespace mu2e {
       const CaloClusterCollection*    _ccColl;
       TimeClusterCollection*          _tcColl; // 'tcColl': time cluster collection
       IntensityInfoTimeCluster*       _iiTC;
+      TimeClusterCollection*          _protonTCColl; // proton time clusters, optionally filled
       int                             _nTZClusters;
 
       // diagnostic data members used in TZ tool
--- a/CalPatRec/inc/TZClusterFinder_types.hh
+++ b/CalPatRec/inc/TZClusterFinder_types.hh
@@ -71,7 +71,7 @@ namespace mu2e {
       const CaloClusterCollection*    _ccColl;
       TimeClusterCollection*          _tcColl; // 'tcColl': time cluster collection
       IntensityInfoTimeCluster*       _iiTC;
-      TimeClusterCollection*          _protonTCColl; // proton time clusters, optionally filled
+      TimeClusterCollection*          _protonTCColl = nullptr; // proton time clusters, optionally filled
       int                             _nTZClusters;
 
       // diagnostic data members used in TZ tool
--- a/CalPatRec/src/AgnosticHelixFinderDiagDisplay.cc
+++ b/CalPatRec/src/AgnosticHelixFinderDiagDisplay.cc
@@ -774,9 +774,13 @@ void AgnosticHelixFinderDiag::plotHelixStageXY(int stage) {
   } else { // All helices
     const bool use_tc = stage == kTimeCluster;
     if(use_tc) title = "Helices (time cluster)";
+    if(_debugLevel > 1) printf("  %12s: Plotting %zu helices, stage = %i\n",
+                               __func__, _data->helixSeedData.size(), stage);
     plotHitsXY(use_tc); // draw (relevant) hits
     plotMC(kHelix);
     for(size_t index = 0; index < _data->helixSeedData.size(); ++index) {
+      if(_debugLevel > 1) printf("  %12s: Plotting the helix XY circle %zu, stage = %i\n",
+                                 __func__, index, stage);
       plotHelixXY(_data->helixSeedData.at(index).seed, index);
     }
   }
--- a/CalPatRec/src/AgnosticHelixFinderDiag_tool.cc
+++ b/CalPatRec/src/AgnosticHelixFinderDiag_tool.cc
@@ -515,7 +515,7 @@ int AgnosticHelixFinderDiag::fillHistograms(void* Data, int Mode) {
     if(_data->hseed) {
       // only display at this level, where per-helix wasn't displayed but accepted helices are each plotted
       if(_display && _data->diagLevel == 3) {
-        std::cout << "Helix stage: Loop condition " << ConditionName(_data->loopCondition)
+        std::cout << "Final stage: Loop condition " << ConditionName(_data->loopCondition)
                   << std::endl;
         plotTripletStage();
         plotHelixStageXY(kHelix);
@@ -535,6 +535,8 @@ int AgnosticHelixFinderDiag::fillHistograms(void* Data, int Mode) {
 
   if(Mode == DIAG::kTimeCluster) {
     if(_display && _data->diagLevel > 1) {
+      std::cout << "TimeCluster stage:"
+                << std::endl;
       // plotTripletStage(true);
       plotCircleStage(true);
       plotHelixStageXY(kTimeCluster);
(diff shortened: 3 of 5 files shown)


michaelmackenzie:
@FNALbuild run build test

matthewstortini (approved):

oksuzian:
I'll review this pull request by examining its details and changes.Now let me look at the full context of the key files to better understand the changes:Here's my review of PR #1796. I identified several issues across the changed files:

---

## Issues Found

### 1. 🔴 Potential null pointer dereference in `countProtons` (critical)

```cpp name=CalPatRec/src/TZClusterFinder_module.cc url=https://github.com/Mu2e/Offline/blob/3e3e6c4febe4e038c22a81ba6c51243acd1fa0e6/CalPatRec/src/TZClusterFinder_module.cc#L772-L783
      if(_protonTCs) {
        auto tcs = _data._protonTCColl;
        TimeCluster tc;
        for (size_t j=0; j<_f.chunks[i].hIndices.size(); j++) {
          tc._strawHitIdxs.push_back(StrawHitIndex(_f.chunks[i].hIndices[j]));
        }
        tc._t0 = TrkT0(_f.chunks[i].fitter.y0(), 0.);
        const int caloIdx = _f.chunks[i].caloIndex;
        if (caloIdx != -1) tc._caloCluster = art::Ptr<mu2e::CaloCluster>(_ccHandle, caloIdx);
        tc._nsh = _f.chunks[i].nStrawHits;
        tcs->push_back(std::move(tc));
      }
```

The `_ccHandle` may not be valid when `_useCaloClusters != 1`. The `findData` method at line 259 only populates `_ccHandle` when `_useCaloClusters == 1`. If `_protonTCs` is true but `_useCaloClusters` is not enabled, and a chunk happens to have a `caloIndex != -1` (which shouldn't normally happen, but is not defensively prevented), constructing `art::Ptr<mu2e::CaloCluster>(_ccHandle, caloIdx)` with an invalid handle would crash. While the current logic makes this unlikely since `caloIndex` defaults to `-1` and is only set in `checkCaloClusters` (which is gated on `_useCaloClusters == 1`), a defensive guard would be safer.

### 2. 🟡 Unnecessary local alias `auto tcs = _data._protonTCColl`

```cpp name=CalPatRec/src/TZClusterFinder_module.cc url=https://github.com/Mu2e/Offline/blob/3e3e6c4febe4e038c22a81ba6c51243acd1fa0e6/CalPatRec/src/TZClusterFinder_module.cc#L773
        auto tcs = _data._protonTCColl;
```

The variable `tcs` is a raw pointer copy of `_data._protonTCColl`. This is declared **inside the loop body** at line 766 (`for (size_t i=0; ...)`), meaning a new copy of the pointer is created on every iteration. While functionally correct, it would be cleaner and more efficient to hoist this above the loop:

```cpp
auto tcs = _data._protonTCColl;
for (size_t i=0; i<_f.chunks.size(); i++) {
  ...
}
```

### 3. 🟡 `_data._protonTCColl` is not initialized when `_protonTCs` is false

```cpp name=CalPatRec/src/TZClusterFinder_module.cc url=https://github.com/Mu2e/Offline/blob/3e3e6c4febe4e038c22a81ba6c51243acd1fa0e6/CalPatRec/src/TZClusterFinder_module.cc#L301
    _data._protonTCColl = (_protonTCs) ? protonTCColl.get() : nullptr;
```

This is fine in `produce()`, but `_data._protonTCColl` is **never initialized in the constructor or `Data_t` struct definition**. The `Data_t` struct (in `TZClusterFinder_types.hh`) has no default member initializers, so `_protonTCColl` starts as an indeterminate value. If any diagnostic plugin or tool accesses `_data._protonTCColl` before `produce()` runs, it would be undefined behavior. Consider adding a default initializer:

```cpp name=CalPatRec/inc/TZClusterFinder_types.hh
      TimeClusterCollection*          _protonTCColl = nullptr;
```

### 4. 🟡 Debug printouts use mixed `printf` and `std::cout` styles

The new debug output in `AgnosticHelixFinder_module.cc` and `AgnosticHelixFinderDiagDisplay.cc` uses `printf`:

```cpp name=CalPatRec/src/AgnosticHelixFinder_module.cc url=https://github.com/Mu2e/Offline/blob/3e3e6c4febe4e038c22a81ba6c51243acd1fa0e6/CalPatRec/src/AgnosticHelixFinder_module.cc#L501-L502
        if(_diagLevel > 1)
          printf("  Time cluster %zu has %i helices\n", i, _diagInfo.nHelices - nHelicesInitial);
```

While `AgnosticHelixFinderDiag_tool.cc` uses `std::cout`:

```cpp name=CalPatRec/src/AgnosticHelixFinderDiag_tool.cc url=https://github.com/Mu2e/Offline/blob/3e3e6c4febe4e038c22a81ba6c51243acd1fa0e6/CalPatRec/src/AgnosticHelixFinderDiag_tool.cc#L536-L537
      std::cout << "TimeCluster stage:"
                << std::endl;
```

Mixing I/O styles can cause interleaving issues with output buffering. The existing codebase uses both styles, but the new additions should ideally be consistent. The new `std::cout` messages in the diag tool also have **no guard on `_debugLevel`** — they print whenever `_data->diagLevel > 1` and `_display` is true, which is a different condition than the `_debugLevel > 1` used in the other files.

### 5. 🟢 Minor: `_diagLevel` vs `_debugLevel` guard inconsistency

In `AgnosticHelixFinder_module.cc`, the new printouts are guarded by `_diagLevel > 1`:

```cpp name=CalPatRec/src/AgnosticHelixFinder_module.cc url=https://github.com/Mu2e/Offline/blob/3e3e6c4febe4e038c22a81ba6c51243acd1fa0e6/CalPatRec/src/AgnosticHelixFinder_module.cc#L501
        if(_diagLevel > 1)
```

But in `AgnosticHelixFinderDiagDisplay.cc`, the new printouts are guarded by `_debugLevel > 1`:

```cpp name=CalPatRec/src/AgnosticHelixFinderDiagDisplay.cc url=https://github.com/Mu2e/Offline/blob/3e3e6c4febe4e038c22a81ba6c51243acd1fa0e6/CalPatRec/src/AgnosticHelixFinderDiagDisplay.cc#L777
    if(_debugLevel > 1) printf("  %12s: Plotting %zu helices, stage = %i\n",
```

These are different variables with potentially different values. Confirm this is intentional — using `_debugLevel` for display-related debug and `_diagLevel` for event-processing debug is a reasonable pattern, but it should be documented or noted for future maintainers.

---

## Summary

| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | 🔴 Medium-High | `TZClusterFinder_module.cc` | `_ccHandle` may be invalid when building proton TC calo pointers |
| 2 | 🟡 Low | `TZClusterFinder_module.cc` | `auto tcs` alias recreated inside loop on every iteration |
| 3 | 🟡 Medium | `TZClusterFinder_types.hh` | `_protonTCColl` has no default initializer — UB risk |
| 4 | 🟡 Low | Multiple files | Mixed `printf`/`std::cout` and inconsistent stage label guards |
| 5 | 🟢 Info | Multiple files | `_diagLevel` vs `_debugLevel` guard inconsistency |

The most actionable fix is **issue #3** — adding `= nullptr` to the `_protonTCColl` declaration in `Data_t` to avoid any uninitialized pointer risk.

michaelmackenzie:
1. I've added this guard, but it's not possible to have non-negative indices without enabling calo clusters and then finding one.
2. I've moved this definition out of the loop.
3. I've initialized this, but nothing tries to use these variables before the first produce call.
4. Why does it matter if std::cout and printf are both used?
5. These are different modules with different debug/diagnostic output variables and values.

oksuzian:
@FNALbuild run build test

PR #1799, opened 2026-04-13
Allow non-zero starting event or subrun values

michaelmackenzie:
This will allow making the current PBISequence detector steps file work properly, splitting a single subrun into multiple dts files without duplicate event numbers

### DIFF
--- a/Sources/src/PBISequence_source.cc
+++ b/Sources/src/PBISequence_source.cc
@@ -73,7 +73,7 @@ namespace mu2e {
     fhicl::Atom<std::string> instance{Name("deprecatedInstanceName"),Comment("Instance name for deprecated subrun data product ProtonBunchIntensity"),""};
     fhicl::Atom<unsigned int> runNumber{Name("runNumber"), Comment("First run number")};
     fhicl::Atom<unsigned int> firstSubRun{Name("firstSubRunNumber"), Comment("First subRun number, incrementing by one for each input file"), 0};
-    fhicl::Atom<unsigned int> firstEvent{Name("firstEventNumber"), Comment("First event number of first subRun (0 in the subsequent subRuns)"), 0};
+    fhicl::Atom<unsigned int> firstEvent{Name("firstEventNumber"), Comment("First event number offset of first subRun (0 in the subsequent subRuns)"), 0};
     fhicl::Atom<bool>  integratedSummary{Name("integratedSummary"), Comment("If true, then PBI summary is summed over all input files; else per input file.")};
     fhicl::Atom<unsigned int> verbosity{Name("verbosity"), Comment("Verbosity level; larger means more printout"), 0};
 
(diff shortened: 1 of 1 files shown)


sophiemiddleton (approved):

oksuzian:
Here's my review of PR #1799. I found a significant issue:

## 🐛 Unsigned Integer Underflow When `firstSubRunNumber_` Is 0

On **line 144**, `subRunNumber_` is initialized as:

```c++ name=Sources/src/PBISequence_source.cc url=https://github.com/Mu2e/Offline/blob/37942ebe30acf6f78b463e8eddcc5fce859f8eee/Sources/src/PBISequence_source.cc#L144
      , subRunNumber_(firstSubRunNumber_ - 1U)
```

When `firstSubRunNumber_` is `0` (the default), this computes `0U - 1U`, which wraps around to `4294967295` (`UINT_MAX`) due to unsigned integer underflow. This is the same behavior as the old code (`-1U`), so it's **intentionally relying on wrap-around** — the value wraps back to `0` when `subRunNumber_++` is called in `readFile()` at line 167.

While this works *by accident* of unsigned wrap-around semantics, this is still fragile and worth noting. The **real issue** is more subtle:

### The Actual Bug: `firstEventNumber_` Is Ignored for the First Sub-Run's Event Numbering

Look at the event numbering flow:

1. **Constructor** (line 145): `currentEventNumber_` is initialized to `firstEventNumber_`.
2. **`readFile()`** (line 168): `currentEventNumber_` is reset to `0` for every sub-run **except** the first one — good so far.
3. **`readNext()`** (line 210): Events use `++currentEventNumber_`, which is a **pre-increment**.

This means:
- For the **first sub-run**, the first event number will be `firstEventNumber_ + 1`, not `firstEventNumber_`. If a user sets `firstEventNumber: 5`, the first event will actually be **event 6**.
- For **subsequent sub-runs**, events start at `0 + 1 = 1` (via pre-increment), meaning event numbering starts at 1, not 0.

If the intent is that `firstEventNumber` specifies the actual first event number, then the initialization should be:

```c++ name=fix.cc
      , currentEventNumber_(firstEventNumber_ > 0 ? firstEventNumber_ - 1 : 0)
```

Or alternatively, change the comment/documentation to clarify that `firstEventNumber` is an offset *before* the first event.

### Summary of Issues

| Issue | Severity | Location |
|---|---|---|
| **Off-by-one in `firstEventNumber_`**: First event is `firstEventNumber + 1` due to pre-increment `++currentEventNumber_` | 🔴 Bug | Lines 145, 210 |
| Unsigned wrap-around for `subRunNumber_` initialization is intentional but could benefit from a comment | 🟡 Style | Line 144 |

The off-by-one is the most important finding — if a user configures `firstEventNumber: 100`, they'll get events starting at 101, which would silently produce incorrect event numbering when splitting subruns across files.

michaelmackenzie:
1. The unsigned overflow is intentionally used here and has been used already before this PR
2. I've updated the comment to indicate the first event number is an offset

PR #1805, opened 2026-04-22
Removed exception for invalid offlineID; Added functions to print sipm type name

giro94:


rlcee (approved):
It looks like you didn't add a mf call, so please remove the include, thanks

oksuzian:
_AI review — findings from an automated static pass. Please sanity-check each item against intent before acting._

SHA reviewed: `96d3cbbc425a3c504761073f007fc62c4de689d2`.

## Bugs / Semantic risks

**1. Silent-invalid return is now the caller's problem — needs a callsite audit.**
The old `CaloDAQMap::offlineId()` threw twice: once on invalid *input*, once on invalid *lookup result*. This PR removes only the second throw:
\`\`\`cpp
// new
return _raw2Offline[rawId.id()];   // may be an invalid CaloSiPMId
\`\`\`
Every caller of `offlineId()` must now check `isValid()` on the return. The PR body is empty, so it's not clear which callers were audited. Until that's confirmed (e.g. `grep -rn 'offlineId(' --include='*.cc'`), this is a latent source of invalid IDs flowing into downstream code that previously assumed the map always returned valid. Either list the audited callsites in the PR description, or add a short `// caller must check isValid()` comment to the header declaration.

**2. Header uses `std::ostream` but doesn't include any ostream header.**
`DataProducts/inc/CaloConst.hh` adds:
\`\`\`cpp
std::ostream& operator<<(std::ostream& ost, const CaloConst::detType& type);
\`\`\`
but the only new include is `<string>`. `<string>` does not pull in `<iosfwd>` on all libstdc++ / libc++ versions. Any TU that includes `CaloConst.hh` without having already included an ostream header will fail to compile. Fix: add `#include <iosfwd>` (cheapest) or `#include <ostream>` in the header.

**3. Dead include.**
`CaloDAQMap.cc` adds `#include \"messagefacility/MessageLogger/MessageLogger.h\"` but the remaining code uses no `mf::` / `LOG_` symbols. Drop it.

## Code quality

**4. `static const std::string detTypeName(detType type)` — the top-level `const` is meaningless** on a by-value return and will trigger `-Wignored-qualifiers` on gcc. Change to `static std::string detTypeName(detType type)`. Even better: `static std::string_view detTypeName(detType type)` returning string-literal views — avoids a per-call heap allocation in the `operator<<` path.

**5. Style reformat bundled with the behavior change.**
Most of the diff (both `CaloDAQMap.cc` and `CaloConst.hh`) is pure re-indentation / brace-style. Fine per se, but it buries the one real semantic change (#1) inside a large \"looks like noise\" diff. Easier to review as a separate commit.

**6. `default:` branch in `detTypeName` is unreachable.**
The `detType` enum has exactly 4 values (`CsI, CAPHRI, PINDiode, Invalid`) and all are cased. The `default: return \"UNKNOWN\";` never fires under defined behavior. Either remove it (and let `-Wswitch` catch future additions), or keep it with a comment that it exists to tolerate out-of-range integer casts.

## Nits

**7. Pre-existing typo next door:** `_nPINDiodPerDisk` (missing the \"e\") sits next to `_nPINDiodeLaserBox`. Not this PR's fault, but since the file is already being touched, a rename to `_nPINDiodePerDisk` would tidy it up — leave if you want the PR to stay tight.

**8. Switch case order in `detTypeName`** (`CsI, PINDiode, CAPHRI, Invalid`) differs from the enum declaration order (`CsI, CAPHRI, PINDiode, Invalid`). Cosmetic; aligning them helps readability.

giro94:
> It looks like you didn't add a mf call, so please remove the include, thanks

that slipped, thanks

giro94:
@FNALbuild build test

giro94:
@FNALbuild build

brownd1978 (approved):

PR #1807, opened 2026-04-24
Set the EWT to be the event number in sims

michaelmackenzie:
This will at least allow for changing EWT values in simulations. The run/subrun/event values can't be packed into the 64 bit EWT, so this cannot all be encoded. The linearly increasing EWT values within a subrun will be useful for lumi stream tests, where the subrun number can be taken from the art event but the event number is lost in the buffering of objects.

### DIFF
--- a/DAQ/src/EventHeaderFromCFOFragment_module.cc
+++ b/DAQ/src/EventHeaderFromCFOFragment_module.cc
@@ -49,6 +49,7 @@ public:
     using Comment = fhicl::Comment;
     fhicl::Atom<art::InputTag> cfoTag   {Name("cfoTag"),    Comment("Input CFO fragment tag")};
     fhicl::Atom<bool>          ewm      {Name("createEWM"), Comment("Produce an event window marker object")};
+    fhicl::Atom<bool>          isSim    {Name("isSim")    , Comment("Flag to indicate simulation processing"), false};
     fhicl::Atom<int>           diagLevel{Name("diagLevel"), Comment("diagnostic level")};
   };
 
@@ -64,6 +65,7 @@ public:
 private:
   art::InputTag cfoFragmentTag_;
   bool          ewm_;
+  bool          isSim_;
   int           diagLevel_;
 };
 
@@ -76,12 +78,14 @@ art::EventHeaderFromCFOFragment::EventHeaderFromCFOFragment(
     art::EDProducer{config},
     cfoFragmentTag_(config().cfoTag()),
     ewm_(config().ewm()),
+    isSim_(config().isSim()),
     diagLevel_(config().diagLevel()) {
   produces<mu2e::EventHeader>();
   if(ewm_) {
     TLOG(TLVL_DEBUG + 2) << "Producing EventWindowMarker";
     produces<mu2e::EventWindowMarker>();
   }
+  if(isSim_) TLOG(TLVL_DEBUG + 2) << "Assuming simulation inputs!";
 }
 
 // ----------------------------------------------------------------------
@@ -93,15 +97,21 @@ void art::EventHeaderFromCFOFragment::produce(Event& event) {
   std::unique_ptr<mu2e::EventWindowMarker> ewm((ewm_) ? new mu2e::EventWindowMarker : nullptr);
   art::Handle<artdaq::Fragments>           cfoFragmentHandle;
 
-  if(!event.getByLabel(cfoFragmentTag_, cfoFragmentHandle)) {
-    // Set the EWT to be the art event number for now, useful for simulations
+  // Set values that are useful in simulations
+  if(isSim_) {
+    // Set the EWT to be the art event number for now
     evtHdr->ewt = static_cast<long unsigned int>(event.event());
+  }
+
+  // Check for the CFO fragment list
+  if(!event.getByLabel(cfoFragmentTag_, cfoFragmentHandle)) {
     event.put(std::move(evtHdr));
     if(ewm_) event.put(std::move(ewm));
     TLOG(TLVL_DEBUG + 1) << "No CFO fragments found";
     return;
   }
 
+  // Process the first fragment found
   const auto *fragments = cfoFragmentHandle.product();
   if (fragments->size()>0){
     const auto &frag = fragments->at(0);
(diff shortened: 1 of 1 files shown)


oksuzian:
# PR #1807 Review — Set the EWT to be the event number in sims

## Summary

**Author:** @michaelmackenzie
**Base:** `Mu2e:main` ← **Head:** `michaelmackenzie:EventHeaderEWTSim`
**State:** Open, draft: false, mergeable: true, mergeable_state: **unstable** (CI `build` still in queue)
**Scope:** 1 file, +2 / −0, 1 commit, 1 comment (FNALbuild CI bot)
**Reviews:** None yet
**Labels:** `build running`
**Risk:** **Low** — 2-line scoped change in a single DAQ producer module, gated by a branch that only fires when no CFO fragment handle is present (i.e., in simulation).

**Intent (from PR body):** Allow the EWT to vary in simulations so that lumi-stream tests see a linearly increasing EWT inside a subrun. The full (run/subrun/event) tuple can't be packed into the 64-bit EWT, so the event number alone is used as a best-effort proxy.

---

## Core Change

[`DAQ/src/EventHeaderFromCFOFragment_module.cc`](https://github.com/Mu2e/Offline/blob/32aa70506b264ba874167e342233a1ee1444ce7a/DAQ/src/EventHeaderFromCFOFragment_module.cc#L89-L103)

```diff name=DAQ/src/EventHeaderFromCFOFragment_module.cc url=https://github.com/Mu2e/Offline/pull/1807/files
   if(!event.getByLabel(cfoFragmentTag_, cfoFragmentHandle)) {
+    // Set the EWT to be the art event number for now, useful for simulations
+    evtHdr->ewt = static_cast<long unsigned int>(event.event());
     event.put(std::move(evtHdr));
     if(ewm_) event.put(std::move(ewm));
     TLOG(TLVL_DEBUG + 1) << "No CFO fragments found";
     return;
   }
```

Previously, when there was no CFO fragment collection in the event (the sim path), `evtHdr->ewt` was left at whatever the default-constructed `mu2e::EventHeader` initializes it to. Now it is set to `event.event()` so downstream lumi-stream / EWT-based code sees monotonically increasing values within a subrun.

---

## 🔎 Issue found — inconsistent handling of the "no CFO fragments" case

There are actually **two** places in `produce()` that represent "no usable CFO fragment" but this PR only patches **one** of them:

```c++ name=DAQ/src/EventHeaderFromCFOFragment_module.cc url=https://github.com/Mu2e/Offline/blob/32aa70506b264ba874167e342233a1ee1444ce7a/DAQ/src/EventHeaderFromCFOFragment_module.cc#L96-L119
  if(!event.getByLabel(cfoFragmentTag_, cfoFragmentHandle)) {
    // Set the EWT to be the art event number for now, useful for simulations
    evtHdr->ewt = static_cast<long unsigned int>(event.event());   // ← patched
    event.put(std::move(evtHdr));
    if(ewm_) event.put(std::move(ewm));
    TLOG(TLVL_DEBUG + 1) << "No CFO fragments found";
    return;
  }

  const auto *fragments = cfoFragmentHandle.product();
  if (fragments->size()>0){
    ...
    evtHdr->ewt = static_cast<long unsigned int>(cfo.GetEventWindowTag()...);
    ...
  } else {
    TLOG(TLVL_DEBUG + 3) << "No CFO fragments found in the event";   // ← NOT patched, ewt stays default
  }
```

If the handle is present but the fragment vector is empty (`fragments->size() == 0`), `evtHdr->ewt` is never assigned and the event is still `put()` at line 127. So the new simulation-friendly behavior only kicks in if `getByLabel` *fails outright*; any job where the CFO collection exists but is empty will still emit an EventHeader with a stale/default EWT. For the lumi-stream test use case this will look like many consecutive events with the same EWT and likely defeat the purpose of the PR.

**Suggested fix** — set the fallback EWT for both "no fragment" branches, e.g.:

```c++ name=DAQ/src/EventHeaderFromCFOFragment_module.cc
  // default EWT for the sim / no-fragment case
  evtHdr->ewt = static_cast<long unsigned int>(event.event());

  if(!event.getByLabel(cfoFragmentTag_, cfoFragmentHandle)) {
    event.put(std::move(evtHdr));
    if(ewm_) event.put(std::move(ewm));
    TLOG(TLVL_DEBUG + 1) << "No CFO fragments found";
    return;
  }

  const auto *fragments = cfoFragmentHandle.product();
  if (fragments->size() > 0) {
    ...
    evtHdr->ewt = static_cast<long unsigned int>(cfo.GetEventWindowTag()...); // overwrite with real value
    ...
  } else {
    TLOG(TLVL_DEBUG + 3) << "No CFO fragments found in the event";
    // evtHdr->ewt already seeded from event.event()
  }
```

This keeps the single-line-of-intent semantics and ensures both "simish" code paths produce a useful EWT.

---

## Other observations / possible improvements

1. **EWM not populated on the early-return path (pre-existing).** On the `!getByLabel` branch, when `ewm_` is true, the `EventWindowMarker` is `put()` without `_spillType` or `_eventLength` ever being assigned (compare to lines 121–125 on the main path). Not introduced by this PR, but worth considering a follow-up since this is the sim path you are now leaning on.
2. **Cast type.** `event.event()` returns `art::EventNumber_t` (32-bit unsigned). `static_cast<long unsigned int>` works but is platform-dependent. Since `EventHeader::ewt` is a `uint64_t`, using `static_cast<uint64_t>(event.event())` (or the exact type of `ewt`) would be clearer and match the cast used on line 112.
3. **Comment wording.** "for now" usually becomes "forever" — consider referencing a ticket or TODO so future readers know this is an intentional sim-only fallback.
4. **Run/subrun in EWT.** The PR body notes that run/subrun can't be packed in. If it ever becomes desirable, the subrun could be stashed in the high bits of the 64-bit EWT with the event number in the low bits; worth an inline comment documenting that this was considered and rejected.

---

## Merge readiness

- ✅ Change is minimal, localized, and the intent matches the diff.
- ⏳ `build` CI check is still queued — wait for that to pass.
- ⚠️ The `fragments->size() == 0` inconsistency above is the one substantive issue I'd flag before approving; everything else is polish.
- No reviewers have been requested yet.

---

## Want me to…

1. Draft an inline review comment on the PR pointing out the `fragments->size() == 0` inconsistency.
2. Propose a concrete patch (commit-ready diff) that also hoists the default-EWT assignment above the `getByLabel` check.
3. Check the `mu2e::EventHeader` default-construction to confirm what `ewt` currently defaults to when unset.
4. Look for downstream consumers of `EventHeader::ewt` in `Offline` to verify the sim-proxy value won't trip any validity checks.
5. Poll CI and report back when the `build` check finishes.

michaelmackenzie:
If there is a CFO fragment collection in the event, but no usable CFO fragment, I don't think this should be handled like simulation and set to a non-zero value. I think it's better to leave this alone and perhaps later we'll throw an error or filter out this event due to missing CFO or corrupted information.

michaelmackenzie:
@FNALbuild run build test

PR #1808, opened 2026-04-24
Stm daq develop 2026

Etho-b02:
An update version of the offline unpacking module which reads in container fragments. Container fragment represents corresponding detector data. Also includes an event and job summary of what the module reads.


DAQ/src/STMBinaryDigisFromFragments_module.cc
```diff
@@ -0,0 +1,499 @@
...
+
+void STMBinaryDigisFromFragments::produce(Event& event)
+{
+  
+  ++_totalEvents; //Increment Total Event Counter
+  
+  std::unique_ptr<mu2e::STMWaveformDigiCollection> raw_waveform_digis(new mu2e::STMWaveformDigiCollection);
+  std::unique_ptr<mu2e::STMWaveformDigiCollection> zs_waveform_digis(new mu2e::STMWaveformDigiCollection);
+  std::unique_ptr<mu2e::STMPHDigiCollection> ph_digis(new mu2e::STMPHDigiCollection);
+  std::unique_ptr<mu2e::STMWaveformDigiCollection> raw_header_waveform_digis(new mu2e::STMWaveformDigiCollection);
```
AndrewEdmonds11: I think since this module is just for writing the binary file for debugging, we don't need to create the Offline data products. You can remove these lines and others in this module

DAQ/src/STMDigisFromFragments_module.cc
```diff
@@ -0,0 +1,594 @@
...
+    bool isHPGe{false};
+
+    uint16_t ZSfromRaw{0};
+    uint16_t expectedZSRegions{0};
+    bool readRawZSinfo{false};
+
+    ContainerFragID = frag.fragmentID();
+    if (_verbosityLevel >= 3){std::cout << "\nFrag_id : " << ContainerFragID << "\n";}
+
+    if(ContainerFragID == 103) {isHPGe = true;} else if (ContainerFragID == 203){isLaBr = true ;}
```
AndrewEdmonds11: I think we don't want to hardcode the numbers 103 and 203. Is it possible to define them in the STMFragment.hh file?

DAQ/src/STMPrintFragments_module.cc
```diff
@@ -0,0 +1,132 @@
...
+
+// ======================================================================
+
+class art::STMPrintFragments : public EDAnalyzer
+{
+  public:
+  struct Config
+   {
+     fhicl::Atom<art::InputTag> stmTag {fhicl::Name("stmTag"), fhicl::Comment("stmTag for new file")};
+  //   //    fhicl::Atom<int> diagLevel{fhicl::Name("diagLevel"), fhicl::Comment("diagnostic Level")};
```
AndrewEdmonds11: You can remove commented out lines like this to clean up the code

RecoDataProducts/inc/STMWaveformDigi.hh
```diff
@@ -35,16 +35,17 @@ namespace mu2e {
...
     uint64_t _EWT;
     uint64_t _DTCtime;
     uint64_t _ADCtime;
     uint32_t _trigTimeOffset; // time offset from EWT? to first ADC value [ct]
-    double   _peak_fitTime1; // fit time of first rising edge (ns)
-    double   _peak_fitTime2; // fit time of second rising edge (ns)
-    double   _peak_sep; // separation time (ns)
+    double   _peak_fitTime1; // fit time of first rising edge (ns) // TODO: remove this parameter
+    double   _peak_fitTime2; // fit time of second rising edge (ns)  // TODO: remove this parameter
+    double   _peak_sep; // separation time (ns)  // TODO: remove this parameter
```
AndrewEdmonds11: Can we remove these parameters now, or are they being used somewhere else in the code?

RecoDataProducts/inc/STMWaveformDigi.hh
```diff
@@ -35,16 +35,17 @@ namespace mu2e {
     double                      peak_fitTime2 () const { return _peak_fitTime2; }
     double                      peak_sep      () const { return _peak_sep; }
     const std::vector<int16_t>& adcs   () const { return _adcs; }
-    
+    void set_data ( size_t n_data, int16_t const* data ) { _adcs.resize(n_data); std::copy(data, data+n_data, _adcs.begin()); } // TODO: remove this method
```
AndrewEdmonds11: I think we can remove that ```// TODO``` comment because I think we want to keep this function

STMReco/src/PlotSTMWaveformDigis_module.cc
```diff
@@ -53,18 +60,33 @@ namespace mu2e {
 
   PlotSTMWaveformDigis::PlotSTMWaveformDigis(const Parameters& config )  :
     art::EDAnalyzer{config},
+    _stmWaveformDigisTag{config().stmWaveformDigisTag()},
     _stmWaveformDigisToken(consumes<STMWaveformDigiCollection>(config().stmWaveformDigisTag())),
     _subtractPedestal(config().subtractPedestal()),
     _xAxis(config().xAxis()),
     _verbosityLevel(config().verbosityLevel()),
-    _channel(STMUtils::getChannel(config().stmWaveformDigisTag()))
+    _channel(STMChannel::findByName("HPGe")) // FIXME: don't hardcode this probably don't want to do what we had before and try to infer it from the art::InputTag like this "STMUtils::getChannel(config().stmWaveformDigisTag()))"
```
AndrewEdmonds11: Let's put this ```FIXME``` on our to-do list to fix ASAP. Maybe we can even get it in before this PR completes review...

AndrewEdmonds11 (commented):
Thanks for opening the PR, Bryan. This looks good. Just a few in-line comments we should work through, Let me know if you have any questions. Thanks

oksuzian:
Bugs / Blockers

  1. Instance-name mismatch between producer and downstream fcls — multiple files will fail at runtime. STMDigisFromFragments only ever calls produces<> with HPGe/LaBr suffixes (rawHPGe, zsHPGe, phHPGe, rawLaBr,
  zsLaBr, phLaBr). But the downstream fcls reference unsuffixed instance names:
  - STMReco/fcl/mwd.fcl — stmWaveformDigisTag : "makeSTMDigis:zs" for both phHPGe and phLaBr producers.
  - STMReco/fcl/plotSTMPHSpectrum.fcl — stmPHDigisTag : "makeSTMDigis:ph".
  - STMReco/fcl/makeSTMHits.fcl — stmPHDigisTag : "makeSTMDigis:ph" (and the other instance "phHPGe" references a module label phHPGe, which doesn't exist in this fcl either).
  - STMReco/fcl/plotSTMWaveformDigis.fcl — "makeSTMDigis:raw", "makeSTMDigis:zs", "makeSTMDigis:ph".

  These will throw "no product found" at runtime. The reason CI may be green is that each fcl has one of HPGe/LaBr commented out of trigger_paths (see #15), so end-to-end runs aren't actually exercising these
  tags.

  2. MakeSTMHits constructor hardcodes STMChannel::LaBr for both instances. STMReco/src/MakeSTMHits_module.cc:
  ,_channel(STMChannel::LaBr)
  replaces the previous STMUtils::getChannel(config().stmMWDDigisTag()). Same root cause as Andrew's FIXME on PlotSTMWaveformDigis_module.cc:68, but unmarked here. Currently masked because makeSTMHits.fcl
  comments out the HPGe path — re-enabling it will silently apply LaBr energy calibrations to HPGe data.

  3. STMPrintFragments::analyze prints the outer container block_count times instead of iterating into inner frags.
  for (size_t ii = 0; ii < contf.block_count(); ++ii){
    const auto dataBegin = frag.dataBegin();   // outer frag, ii unused
    const auto dataEnd   = frag.dataEnd();
    ...
  }
  The loop variable ii is never used; frag.dataBegin() always returns the outer container's bytes. Should be auto inner = contf.at(ii); auto dataBegin = inner->dataBegin();.

  4. _unreadInnerFrags is gated by verbosity — counters are zero at default verbosity. STMDigisFromFragments_module.cc:
  else {
    if (_verbosityLevel >= 3) {
      ++_unreadInnerFrags;     // both job and event counters live inside the if
      ++unread_InnerFrags;
    }
  }
  The endJob "Unread frags" line will always read 0 unless verbosity is ≥ 3. Move the increments outside the if.

  5. xAxis condition flipped — looks like an intended pre-existing bug fix; please call it out in the PR body. STMMovingWindowDeconvolution_module.cc:~166 changed if (_xAxis != "") to if (_xAxis == "") around the
   throw. The new condition matches the message ("No xAxis scale defined despite requesting verbosity level >= 5") and is correct, but it's a behavior change that future spelunkers will want to see noted.

  Risks / Robustness

  6. Unknown container fragment IDs are silently dropped. Beyond Andrew's hardcoding comment, the bigger issue is the silent path: when ContainerFragID is neither 103 nor 203, both isHPGe and isLaBr stay false,
  the nested if(isHPGe) ... else if(isLaBr) ... skips every branch, and no log line records that an inner frag was discarded. Add a default log at verbosity ≥ 1.

  7. ZSfromRaw != totalLen cross-check has lifetime issues. readRawZSinfo is set true after a Raw block and never reset, so it then checks every subsequent ZS block in the same container against the same
  ZSfromRaw. If only one Raw–ZS pair per container is expected, document it; if multiples are possible, reset readRawZSinfo/ZSfromRaw after each ZS check.

  8. STMBinaryDigisFromFragments::produce() allocates four unique_ptr<...Collection>s that are never event.put()'d. Andrew flagged removing them in spirit; concretely they're per-event allocations of
  STMWaveformDigiCollection × 3 + STMPHDigiCollection × 1 that immediately get destroyed.

  9. STMPrintFragments declares BTrk_difAlgebra as a build dep but never uses BTrk. DAQ/CMakeLists.txt:138. Looks copy-pasted from a CRV module — drop it.

  Code quality

  10. Five debug std::couts not gated by verbosity.
  - STMMovingWindowDeconvolution_module.cc: std::cout<<"BG"<<_xAxis<<std::endl; (author-initials marker), "event id = ", "waveformDigisHandle size = ", "waveform id =" (last three printed every waveform of every
  event).
  - PlotSTMWaveformDigis_module.cc: std::cout<<"size = "<<waveformsHandle->size()<<std::endl;

  11. STMPrintFragments header comment is wrong: "STMPrintFragments_plugin: Add CRV data products to the event". Copy-paste from a CRV module.

  12. _totalRaw is ambiguous — it includes the Empty + Zero counts. A reader of the job summary doesn't know whether to subtract. Either rename to _totalRawSeen or add a _totalGoodRaw line.

  13. Top-level non-container fragments silently no-op. The else { /* fallback non-container case */ ... } at the outer loop never logs anything — same family as #6.

  Nits

  14. Commented-out FragmentWatcher block in inspectSTMFile.fcl has three layers of # stacked (# frag: / blank / #    { / #   module_type ...). Either delete or reformat as a single comment.

  15. Test fcls ship with one detector path commented out and the choice is inconsistent across files. mwd.fcl disables LaBr; makeSTMHits.fcl disables HPGe. The PR does not say why.

  16. STMBinaryDigisFromFragments and STMDigisFromFragments include the entire ROOT graphics stack (TROOT.h, TApplication.h, TCanvas.h, TF1.h, TString.h, TLine.h, TGraph.h) but don't use any ROOT graphics. Drop.

  17. <list>, <numeric>, <random> included in both new modules, never used. Lint cleanup.

  ---

oksuzian:
@FNALbuild run build test

oksuzian:
@FNALbuild run build test

AndrewEdmonds11:
@FNALbuild run build test with Mu2e/artdaq-core-mu2e#112

rlcee:
Unfortunately the FNALbuild "with" command won't work with artdaq-core-mu2e, since this build is muse-based but artdaq-core-mu2e is cmake-only.  To do this test locally, you can use the [spack recipe](https://mu2ewiki.fnal.gov/wiki/Spack#Local_Offline_build), then we can cut a new envset with the new adcm, then the Offline can use it.

AndrewEdmonds11:
Thanks, Ray. Bryan has been building with spack locally so we'll get the artdaq-core-mu2e PR merged and then come back to sort out this PR and make sure all the checks pass

oksuzian:
@FNALbuild run build test

AndrewEdmonds11:
@FNALbuild run build tests

oksuzian:
@FNALbuild run build test

oksuzian:
# PR #1808 Review — Critical Issues

**Status:** Open, mergeable=unstable, label `build failed`, build now compiles but **ceSimReco** & **cosmicSimReco** fail (rc 90) on latest commit `854b376`. 30 files / +1793 / -647.

## 🔴 Critical (must fix before merge)

1. **Runtime "no product" failures — instance-name mismatches between producer and downstream fcls.** `STMDigisFromFragments` only `produces<>` HPGe/LaBr-suffixed instances (`rawHPGe`, `zsHPGe`, `phHPGe`, `rawLaBr`, `zsLaBr`, `phLaBr`), but downstream fcls still reference unsuffixed tags:
   - [`STMReco/fcl/mwd.fcl`](https://github.com/Mu2e/Offline/blob/stm-daq-develop-2026/STMReco/fcl/mwd.fcl): `stmWaveformDigisTag : "makeSTMDigis:LaBr"` (should be `zsLaBr`).
   - [`STMReco/fcl/plotSTMPHSpectrum.fcl`](https://github.com/Mu2e/Offline/blob/stm-daq-develop-2026/STMReco/fcl/plotSTMPHSpectrum.fcl): `stmPHDigisTag : "phHPGe"` (HPGe analyzer is missing the module label prefix — should be `"makeSTMDigis:phHPGe"`), and the LaBr analyzer is wired to `"makeSTMDigis:phHPGe"` (copy-paste, should be `phLaBr`).
   - Almost certainly the cause of the **ceSimReco / cosmicSimReco** integration failures.

2. **`STMDigisFromFragments` — `produce()` exits without `event.put()` for HPGe/LaBr collections when the respective `save*` flags are false** (only `phHPGe` / `phLaBr` are unconditional). Combined with unconditional `produces<>` at construction for `phHPGe`/`phLaBr` but **conditional** `produces<>` for raw/zs — fine, but the conditional `event.put` paths must match the conditional `produces` paths exactly; the current code does, but the comment lines `//std::unique_ptr<...> ph_waveform_digis(...)` and stale member-init order should be cleaned up.

3. **ZS-length mismatch is now a `throw cet::exception("STM_UNPACKING")`** ([`STMDigisFromFragments_module.cc` ~L347](https://github.com/Mu2e/Offline/blob/stm-daq-develop-2026/DAQ/src/STMDigisFromFragments_module.cc)). The state vars `readZSinfoFromRawHeaderHPGe`/`LaBr` are only set during a Raw block earlier in the same container. If a ZS block appears without a preceding Raw block (e.g. ZS-only running mode), the cross-check is skipped silently, but if Raw appears once and ZS twice, the second ZS will re-compare against stale `expectedZSLength*` only if reset logic fails — verify the reset block (`if (stm_frag.isHPGe()) ... = 0;`) executes whether the throw happens or not. As written, the throw aborts the event before the reset; that's fine for an error path but worth a comment.

4. **Hardcoded channel: `STMMovingWindowDeconvolution_module.cc` L148** — `channel(STMChannel::findByName("HPGe"))` with FIXME. Will silently apply HPGe calibrations to LaBr in the `phLaBr` instance. Andrew flagged this. Same pattern in `MakeSTMHits_module.cc` previously (`_channel(STMChannel::LaBr)` comment hints) — verify current state.

5. **`STMReco/fcl/makeSTMHits.fcl`** drops HPGe from `trigger_paths` (`trigger_paths: [LaBrPath ]` — note bracket spacing). Likely the cause of HPGe STMHits not being generated downstream; please justify or restore.

## 🟠 Bugs / robustness

6. **`STMDigisFromFragments` — counters semantics.** Empty/zero filtering increments `_totalEmpty*` and `_totalZero*` *and* the parent `_totalRaw` / `_totalZS` / `_totalPH`. The job summary header "RAW seen" includes filtered. Add a "Good Raw" line or rename.

7. **`STMPrintFragments`** declares `BTrk_difAlgebra` in [`DAQ/CMakeLists.txt`](https://github.com/Mu2e/Offline/blob/stm-daq-develop-2026/DAQ/CMakeLists.txt#L138) but doesn't use BTrk — drop.

8. **Header comment wrong:** `STMPrintFragments_module.cc` first line says *"Add CRV data products to the event"*.

9. **Unknown container fragment IDs** in `STMDigisFromFragments`: only `isHPGeContainer()` / `isLaBrContainer()` checked, else printed only at verbosity ≥ 1. The inner loop still runs and may double-count under `_totalContainers`. Consider `continue` on unknown.

10. **Non-container outer fragment fallback** only logs at verbosity ≥ 1 — silently dropped at default.

## 🟡 Code quality

11. **Debug `std::cout` lines not gated by verbosity** in `STMMovingWindowDeconvolution_module.cc` — `"BG"<<_xAxis` (initials marker, L171) and several event-loop prints under `verbosityLevel>=7` are fine, but the `"BG"` is unconditional inside the `if`.

12. **Unused ROOT graphics headers** (`TROOT.h`, `TApplication.h`, `TCanvas.h`, `TF1.h`, `TString.h`, `TLine.h`, `TGraph.h`) plus `<list>`, `<numeric>`, `<random>` in new `STMDigisFromFragments_module.cc` and `STMBinaryDigisFromFragments_module.cc`.

13. **xAxis logic flipped** in `STMMovingWindowDeconvolution_module.cc` L187: `if (_xAxis == "")` now throws when no axis is set at verbosity ≥ 5 — this is the *correct* sense (was `!=` previously). Call this fix out in the PR description.

14. **Whitespace + clang-tidy:** FNALbuild reports `52 errors / 134 warnings` (clang-tidy) and whitespace errors. 5 `TODO` + 2 `FIXME` introduced.

15. **`inspectSTMFile.fcl`** has stacked `#` comments and inconsistent output commands (`drop *_*_*_*` line commented vs `keep *`). Clean up.

PR #1811, opened 2026-04-25
Patch KKLine for KinKalGeom migration

brownd1978:
This addresses the exception in extracted validation following the migration of KinKalGeom to GeometryService.

### DIFF
--- a/Mu2eKinKal/inc/KKFit.hh
+++ b/Mu2eKinKal/inc/KKFit.hh
@@ -159,7 +159,6 @@ namespace mu2e {
       double addStrawMinDz_;
       int strawNBuffer_;
       bool saveHitCalib_;
-      KinKalGeom smap_;
       SurfaceIdCollection ssids_;
   };
 
@@ -207,7 +206,7 @@ namespace mu2e {
     } else {
       throw cet::exception("RECO")<<"mu2e::KKFit: unknown trajectory option "<< fitconfig.saveTraj() << endl;
     }
-    // Lookup surfaces to sample; this interface is deprecatecd and should be replaced with extrapolation TODO
+    // surfaces to sample; this interface is deprecatecd and should be replaced with extrapolation TODO
     for(auto const& sidname : fitconfig.sampleSurfaces()){
       ssids_.push_back(SurfaceId(sidname,-1)); // match all elements
     }
@@ -866,8 +865,10 @@ namespace mu2e {
   template <class KTRAJ> void KKFit<KTRAJ>::sampleFit(KKTRK& kktrk) const {
     // translate the sample and extend surface names to actual surfaces using the KinKalGeom.  This should come from the
     // extrapolation and sampling options
+    GeomHandle<mu2e::KinKalGeom> kkg_h;
+    auto const& kkg = *kkg_h;
     KinKalGeom::SurfacePairCollection tosample; // surfaces to sample the fit
-    smap_.surfaces(ssids_,tosample);
+    kkg.surfaces(ssids_,tosample);
 
     auto const& ptraj = kktrk.fitTraj();
     std::vector<TimeRange> ranges;
--- a/Mu2eKinKal/src/CentralHelixFit_module.cc
+++ b/Mu2eKinKal/src/CentralHelixFit_module.cc
@@ -121,7 +121,7 @@ namespace mu2e {
     fhicl::Atom<float> sampleTBuff { Name("SampleTimeBuffer"), Comment("Time buffer for sample intersections (nsec)") };
     fhicl::Atom<bool> useFitCharge { Name("UseFitCharge"), Comment("Set the PDG particle according to the fit charge; otherwise reject fits that don't agree with the PDG particle charge") };
     fhicl::Atom<float> minCenterRho { Name("MinCenterRho"), Comment("Minimum transverse distance from the helix axis to the Z axis to consider the fit non-degenerate (mm)") };
-};
+  };
 
   struct GlobalConfig {
     fhicl::Table<KKCHModuleConfig> modSettings { Name("ModuleSettings") };
@@ -129,7 +129,7 @@ namespace mu2e {
     fhicl::Table<KKConfig> fitSettings { Name("FitSettings") };
     fhicl::Table<KKConfig> extSettings { Name("ExtensionSettings") };
     fhicl::Table<KKMaterialConfig> matSettings { Name("MaterialSettings") };
-// helix module specific config
+    // helix module specific config
   };
 
   class CentralHelixFit : public art::EDProducer {
@@ -171,8 +171,9 @@ namespace mu2e {
       double minCenterRho_; // min center distance to z axis
       bool sampleinrange_, sampleinbounds_; // require samples to be in range or on surface
       SurfaceIdCollection ssids_;
+      KinKalGeom::SurfacePairCollection surfacess_to_sample_; // surfaces to sample the fit
       std::array<double,KinKal::NParams()> paramconstraints_;
-    };
+  };
 
   CentralHelixFit::CentralHelixFit(const Parameters& settings) : art::EDProducer{settings},
     fitflag_(TrkFitFlag::KKCentralHelix),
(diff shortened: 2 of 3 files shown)


brownd1978:
@FNALbuild run build test

oksuzian:
# PR #1811 Review — "Patch KKLine for KinKalGeom migration"

**Author:** @brownd1978 · **Base:** `Mu2e:main` · **State:** open, mergeable (clean), draft: false  
**Scope:** 3 files, +40 / −32, 2 commits, 5 conversation comments, 0 review comments, 0 reviews.  
**Reviewer requested:** @rlcee · **CI:** `build` ✅ passed at `fead67b`.

## 1. PR Summary

Following the migration of `KinKalGeom` into `GeometryService`, objects of type `KinKalGeom` cannot be safely cached as data members at module construction time (they have to come from a `GeomHandle` once the geometry is available per run/event). This PR fixes the runtime exception in extracted validation by:

- **`Mu2eKinKal/inc/KKFit.hh`** — removes the cached `KinKalGeom smap_` member and looks up `GeomHandle<KinKalGeom>` inside `sampleFit()`.
- **`Mu2eKinKal/src/CentralHelixFit_module.cc`** — drops the cached `KinKalGeom::SurfacePairCollection sample_`, stores only the `SurfaceIdCollection ssids_`, and resolves the surface pairs from the `GeomHandle` in `sampleFit()`.
- **`Mu2eKinKal/src/KinematicLineFit_module.cc`** — same idea, plus the `ExtrapolateTCRV TCRV_` member is no longer constructed at module construction; instead, the inputs (`maxdt_`, `btol_`, `minv_`, `extrapdebug_`) are saved and `ExtrapolateTCRV` is rebuilt on every call to `extrapolate()`.

**Risk:** medium. The build CI passes (3070), but there are at least two **substantive (non-refactor) changes** hiding inside what reads as a mechanical patch — see issues #1 and #2 below.

## 2. Core Changes (issues found)

### 🚨 Issue 1 — `tcrvthick_` default changed by ~1400×, with a different comment

In `KinematicLineFit_module.cc`:

```diff name=Mu2eKinKal/src/KinematicLineFit_module.cc url=https://github.com/Mu2e/Offline/pull/1811/files
-    ExtrapolateTCRV TCRV_; // extrapolation predicate based on Z values
-    double tcrvthick_ = 0.1056; // st foil thickness: should come from geometry service TODO
+    double maxdt_ = 0.0, btol_ = 0.0, minv_ = 0.0;
+    SurfaceIdCollection ssids_;
+    int extrapdebug_ = false;
+    double tcrvthick_ = 150.0; // CRV sector thickness: should come from geometry service TODO
```

`tcrvthick_` is passed straight into the `KKCRVXING` constructor and drives the material Xing thickness for the CRV. Old default was `0.1056` mm with the comment *"st foil thickness"* (clearly the wrong thing — looks like a stopping‑target foil value left over from a copy‑paste). New default is `150.0` mm with comment *"CRV sector thickness"*.

This is almost certainly the **right** fix, but it is **a physics‑affecting change** that is not mentioned in the PR description. Please:
- Call it out in the PR body so reviewers/validation know about it.
- Confirm `150.0` mm is the intended CRV sector thickness (and not, e.g., one bar — CRV sector total thickness depends on layer count).
- Ideally open / reference a TODO issue to pull this from `GeometryService` since the comment still says it should.

### 🚨 Issue 2 — `int extrapdebug_ = false;`

```cpp name=Mu2eKinKal/src/KinematicLineFit_module.cc url=https://github.com/Mu2e/Offline/pull/1811/files
int extrapdebug_ = false;
```

Type/initializer mismatch. Compiles (implicit bool→int), but it’s clearly a typo. The previous local was `int debug = settings().Extrapolation()->Debug();`. Use `int extrapdebug_ = 0;` (or make it `bool` if it’s really binary — but `Debug()` returns `int`).

### ⚠️ Issue 3 — Per‑call construction of `ExtrapolateTCRV` is a perf regression

```cpp name=Mu2eKinKal/src/KinematicLineFit_module.cc url=https://github.com/Mu2e/Offline/pull/1811/files
void KinematicLineFit::extrapolate(KKTRK& ktrk) const {
  GeomHandle<mu2e::KinKalGeom> kkg_h;
  auto const& kkg = *kkg_h;
  // extrapolate to the front of the tracker
  auto TCRV = ExtrapolateTCRV(maxdt_,btol_,intertol_,minv_,*kkg.TCRV(),extrapdebug_);
  ...
}
```

`extrapolate()` is called per fit. Building a fresh `ExtrapolateTCRV` (and re‑resolving the surface pair collection in `sampleFit`) on every call is wasteful — `KinKalGeom` does not change within a run.

The conventional Mu2e pattern is to fill these in `beginRun(art::Run&)` (which already exists for `CentralHelixFit` and is the natural place for run‑scoped geometry caches). Suggested refactor: keep `TCRV_` as a member, but assign it in `beginRun` from a `GeomHandle<KinKalGeom>`. Same for `sample_` in all three modules.

The new comment in `CentralHelixFit::sampleFit` is even slightly misleading:

> `// This must be done every call as the KKGeom object now comes from GeometryService`

It does **not** need to be done every call — it needs to be done after the geometry service is initialized, i.e. once per run.

### ⚠️ Issue 4 — Inconsistent treatment across the three files

The three callsites all have the same problem (cached objects that depend on geometry) but the patch solves it three different ways:

| File | sample list | TCRV |
|------|-------------|------|
| `KKFit.hh` | rebuilt per call | n/a |
| `CentralHelixFit_module.cc` | rebuilt per call | n/a |
| `KinematicLineFit_module.cc` | rebuilt per call | rebuilt per call |

Worth aligning on one pattern (preferably "cache in `beginRun`"), even if as a follow‑up.

### ℹ️ Issue 5 — New include may be unused

```cpp name=Mu2eKinKal/src/KinematicLineFit_module.cc url=https://github.com/Mu2e/Offline/pull/1811/files
#include "Offline/Mu2eKinKal/inc/KKShellXing.hh"
```

This include is added but I don’t see a new symbol from `KKShellXing` referenced in the diff (`KKCRVXING`/`KKCRVXINGPTR` were already in use before). Please confirm it isn’t already pulled in transitively — if so, drop it.

### ℹ️ Issue 6 — Pre‑existing typo retained, plus stray blank line

- `KKFit.hh` keeps the typo `// surfaces to sample; this interface is deprecatecd` (was already there — easy to fix while editing the line).
- `KinematicLineFit_module.cc` leaves a stray double blank line where the `ExtrapolateTCRV` block was removed.

### ℹ️ Issue 7 — clang‑tidy delta

The CI table on the latest build (`fead67b`) reports `4 errors / 71 warnings` and `TODO (19) FIXME (6) in 3 files`, up from `2 errors / 20 warnings` and `TODO (2) FIXME (1) in 1 files` on the first push. Most of this is likely because the second push touches more files, but worth a glance at the [clang‑tidy log](https://buildmaster.fnal.gov/buildmaster/job/GitHubPRTests/job/mu2e-offline-build-test//3070/artifact/clang-tidy.log) to confirm none of the new diagnostics are introduced by this PR.

## 3. Merge Readiness

| Item | Status |
|------|--------|
| Mergeable | ✅ clean |
| Build CI | ✅ passed (job 3070) |
| Reviews | ⏳ none yet (review requested from @rlcee) |
| Validation | The PR description says it fixes an exception in extracted validation — please post the validation/extracted‑run output as evidence before merge |
| Blocking issues | None compile‑level; **Issue 1** (physics constant change) and **Issue 2** (typo) should be addressed first |

## 4. Suggested Improvements

1. Update the PR description to explicitly call out the `tcrvthick_` value/semantics change.
2. Fix `int extrapdebug_ = false;` → `int extrapdebug_ = 0;`.
3. Move the `GeomHandle<KinKalGeom>` lookup, `sample_` resolution, and `ExtrapolateTCRV` construction into `beginRun()` to avoid per‑track work.
4. Drop the (apparently) unused `KKShellXing.hh` include.
5. Fix the pre‑existing `deprecatecd` typo in `KKFit.hh` while you’re there.
6. Align the three modules on one caching pattern in a follow‑up.

## 5. Want me to…

1. Post these review comments inline on the PR (you’d need to confirm the file/line targets).
2. Open a small follow‑up issue tracking "cache `KinKalGeom`‑derived objects in `beginRun`" across `Mu2eKinKal`?
3. Pull and read the clang‑tidy log to confirm whether any new diagnostics are introduced by this PR?
4. Look up the actual CRV sector thickness from `GeometryService` to verify `150.0` mm is correct?

rlcee (approved):

brownd1978:
Addressed AI issues:
1- original is correct
2- fixed
3- fixed
4-fixed
5- original is correct
6-clang-tidy errors are inside art, warnings are incorrect/irrelevant

oksuzian:
@FNALbuild run build test

PR #1812, opened 2026-04-26
more protections against data corruption

pavel1murat:
- protect code against couple of more observed data corruption cases

### DIFF
--- a/DAQ/src/StrawDigisFromArtdaqFragments_module.cc
+++ b/DAQ/src/StrawDigisFromArtdaqFragments_module.cc
@@ -74,58 +74,56 @@ public:
     fhicl::Atom<bool>            keyOnMnid        {fhicl::Name("keyOnMnid"        ), fhicl::Comment("true if need to key on MnID, default:false")};
     fhicl::Atom<bool>            allowOfflineFallbackWhenPanelMapMissing{
       fhicl::Name("allowOfflineFallbackWhenPanelMapMissing"),
-      fhicl::Comment("If TrackerPanelMap lookup fails, decode using offline StrawId(dtc,link,straw)"),
-      false};
+      fhicl::Comment("If TrackerPanelMap lookup fails, decode using offline StrawId(dtc,link,straw)")};
     fhicl::Atom<bool>            forceOfflineAddressing{
       fhicl::Name("forceOfflineAddressing"),
-      fhicl::Comment("Ignore TrackerPanelMap/mnid and decode StrawId directly as (dtc,link,straw)"),
-      false};
+      fhicl::Comment("Ignore TrackerPanelMap/mnid and decode StrawId directly as (dtc,link,straw)")};
 
-    // individual tuple specifying a minnesota label, e.g. MN123,
-    // with geographic plane/panel numbers, i.e. from DocDB-#888
-    struct GeographicTuple{
-      fhicl::Atom<std::string> minnesota{
-        fhicl::Name("minnesota"),
-        fhicl::Comment("Minnesota # label")
-      };
-      fhicl::Atom<uint16_t> plane{
-        fhicl::Name("plane"),
-        fhicl::Comment("Geographic plane [DocDB-888]")
-      };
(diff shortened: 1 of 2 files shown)


oksuzian:
@FNALbuild run build test

brownd1978 (approved):
This code needs cleaning, but the substantive changes are needed to run Pass1. We can do a thorough cleanup after debugging the hardware  + DAQ.

DAQ/src/StrawDigisFromArtdaqFragments_module.cc
```diff
@@ -438,9 +469,9 @@ void mu2e::StrawDigisFromArtdaqFragments::produce(art::Event& event) {
               uint16_t channel = static_cast<uint16_t>(hit_data->StrawIndex);
               uint16_t chid   = mu2e::StrawId(channel).straw(); // channel ID within the panel
 
-              if (chid > StrawId::_nstraws) {
-                print_(std::format("ERROR: hit with corrupted chid:{:04x} : straw:{} / dtc_id:{} link_id:{}, SKIPPING",
-                                   hit_data->StrawIndex, chid, dtc_id, link_id));
+              if (chid >= StrawId::_nstraws) {
```
edcallaghan, line 472: Should this really include equality? I don't see why a straw-number of exactly 96 would be valid.

edcallaghan (approved):
I am not sure if the `>` -> `>=` change is correct, but the rest of these changes look sane to me.

michaelmackenzie (changes requested):
This fails the trigger CI test, likely missing an update to prologs/trigger fcl files. Please add a mu2e-trig-config PR if this is needed

DAQ/src/StrawDigisFromArtdaqFragments_module.cc
brownd1978: I had the same question, but Richie pointed out this is an error test; the change flags exactly 96 as an error, while previous version let it pass as good.

brownd1978:
@FNALbuild run build test

DAQ/src/StrawDigisFromArtdaqFragments_module.cc
edcallaghan: That makes sense; no complaints from me, then.

DAQ/src/StrawDigisFromArtdaqFragments_module.cc
```diff
@@ -74,58 +74,56 @@ class mu2e::StrawDigisFromArtdaqFragments : public art::EDProducer {
...
+  //       fhicl::Comment("Geographic plane [DocDB-888]")
+  //     };
+  //     fhicl::Atom<uint16_t> panel{
+  //       fhicl::Name("panel"),
+  //       fhicl::Comment("Geographic panel [DocDB-888]")
+  //     };
+  //   };
+  //   // mandatory listing of geographic entries, to enable translation
+  //   // of panel-labelings in the data to Offline StrawIds
+  //   fhicl::Sequence< fhicl::Table<GeographicTuple> > geography{
```
brownd1978, line 100: Commenting out this block breaks the trigger configuration.

brownd1978 (changes requested):

brownd1978:
CI was failing because the forked branch was out-of-date. I was able to merge the head of main in cleanly, then you get to a real error;
 Module label: TrigMakeSD
  module_type : StrawDigisFromArtdaqFragments  
  Any parameters prefaced with '#' are optional.
  Unsupported parameters:
   + geography                      [ /exp/mu2e/app/users/brownd/KFTrack/mu2e-trig-config/core/producers/trigTrkProducers.fcl:5 ]

The 'geography' parameter is still referenced in prolog but gets commented out by this PR. If this removal is permanent,  please do these things:
- merge my PR into your fork (to get the merge with origin/main HEAD)
- delete the code block (not just comment it out). Github does a better job of following history than we can
- Put in a PR to mu2e-trig-config removing the above fcl line
- test by running:  mu2e -c Production/Validation/cosmicOffSpill.fcl

PR #1814, opened 2026-04-30
Include calorimeter in digi harvesting

edcallaghan:
This PR includes calorimeter digis in the workflow of extracting digis from an event which participate in a reconstructed track and doctoring their timing to change where in the event window they (and, as a consequence, a track reconstructed them) occur in the event window, which is necessary to include the calorimeter in the proposed salt-with-cosmics blinding scheme.

Changes to existing modules:
  - TrackDigiExtractor: Identify all calorimeter digis which ultimately contribute to a calorimeter hit associated with a given fit track, and store them in a dedicated data product, similar to how tracker digis are handled.
  - DisplaceDigiTimes: Deep-copy of all calorimeter digis in a given collection, with the starting timestamp shifted by a predetermined interval (subject to the sample spacing of the waveforms, which are the timestamp units).
 
  One note is that the sample spacing of the calorimeter is not stored condition available via the usual handes, but (where used elsewhere) as a fcl parameter; this implementation thus also supplies this value as a fcl parameter, which may be coordinated as a prolog value when production fcl for this workflow materializes.

This also sneaks in some extra declarations of data products to be serialized to disk, which are necessary when running certain ensembling workflows in the process of testing the blinding mechanics.

### DIFF
--- a/Blinding/src/DisplaceDigiTimes_module.cc
+++ b/Blinding/src/DisplaceDigiTimes_module.cc
@@ -27,8 +27,10 @@
 #include "fhiclcpp/types/Sequence.h"
 
 // mu2e
+#include "Offline/CaloMC/inc/CaloDigiWrapper.hh"
 #include "Offline/DataProducts/inc/SurfaceId.hh"
 #include "Offline/Mu2eUtilities/inc/BinnedSpectrum.hh"
+#include "Offline/RecoDataProducts/inc/CaloDigi.hh"
 #include "Offline/RecoDataProducts/inc/KalIntersection.hh"
 #include "Offline/RecoDataProducts/inc/KalSeed.hh"
 #include "Offline/RecoDataProducts/inc/StrawDigi.hh"
@@ -44,6 +46,10 @@ namespace mu2e{
       struct Config{
         fhicl::Atom<art::InputTag> tracker_digi_tag{
           fhicl::Name("StrawDigiCollection"),
+          fhicl::Comment("art::InputTag of digis to displace")
+        };
+        fhicl::Atom<art::InputTag> calo_digi_tag{
+          fhicl::Name("CaloDigiCollection"),
           fhicl::Comment("art::InputTag of digis to displace")
         };
         fhicl::Atom<art::InputTag> kalseed_tag{
@@ -67,6 +73,7 @@ namespace mu2e{
 
     protected:
       art::InputTag _tracker_digi_tag;
+      art::InputTag _calo_digi_tag;
       art::InputTag _kalseed_tag;
       SurfaceIdCollection _surface_ids;
       ProditionsHandle<StrawElectronics> _tracker_conditions_handle;
@@ -82,6 +89,7 @@ namespace mu2e{
   DisplaceDigiTimes::DisplaceDigiTimes(const Parameters& config):
       art::EDProducer(config),
       _tracker_digi_tag(config().tracker_digi_tag()),
+      _calo_digi_tag(config().calo_digi_tag()),
       _kalseed_tag(config().kalseed_tag()),
       _engine{createEngine(art::ServiceHandle<SeedService>()->getSeed())}{
     // initialize target distribution
@@ -106,6 +114,8 @@ namespace mu2e{
     this->produces<StrawDigiADCWaveformCollection>();
 
     // calorimeter...
+    this->consumes<CaloDigiCollection>(_calo_digi_tag);
+    this->produces<CaloDigiCollection>();
   }
 
   void DisplaceDigiTimes::produce(art::Event& event){
@@ -118,6 +128,9 @@ namespace mu2e{
            + " and " + std::to_string(adcs_handle->size());
       throw cet::exception("DisplaceDigiTimes") << msg << std::endl;
     }
+
+    // calorimeter
+    auto cdigi_handle = event.getValidHandle<CaloDigiCollection>(_calo_digi_tag);
 
     // first, check that there is only a single KalSeed associated with
     // these digis
@@ -185,7 +198,27 @@ namespace mu2e{
     event.put(std::move(digis));
     event.put(std::move(adcss));
 
-    // calorimeter...
+    // apply the shift to calorimeter digis
+    // TODO need adc sampling period from fucking conditions!!!
+    double sample_period = 5.0;
+    auto cdigis = std::make_unique<CaloDigiCollection>();
+    if (0 < cdigi_handle->size()){
+      for (const auto& old: *cdigi_handle){
+        auto id = old.SiPMID();
+        auto t0 = old.t0();
+        auto peakpos = old.peakpos();
+        auto& waveform = old.waveform();
+
+        // displace waveform start times
+        auto analog = t0 * sample_period;
+        auto shifted = analog + shift;
+        t0 = static_cast<CaloDigiWrapper::sample_t>(shifted / sample_period);
+
+        // new digi is identical to input digi, with displaced start index
+        cdigis->emplace_back(id, t0, waveform, peakpos);
+      }
+    }
+    event.put(std::move(cdigis));
   }
 
   double DisplaceDigiTimes::sample_target_time(){
--- a/Blinding/src/DisplaceDigiTimes_module.cc
+++ b/Blinding/src/DisplaceDigiTimes_module.cc
@@ -64,6 +64,11 @@ namespace mu2e{
           fhicl::Name("SurfaceIds"),
           fhicl::Comment("Prioritized sequence of mu2e::SurfaceIds at which tracks may be sampled for reference timing")
         };
+        // would be better for the sampling period to come from conditions
+        fhicl::Atom<double> calo_adc_sample_period{
+          fhicl::Name("CaloADCSamplePeriod"),
+          fhicl::Comment("Sample spacing of calorimeter waveforms")
+        };
       };
 
       using Parameters = art::EDProducer::Table<Config>;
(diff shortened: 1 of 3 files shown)


oksuzian:
# PR #1814 Review: Include calorimeter in digi harvesting

## 1) Summary

**Author:** @edcallaghan • **State:** open, mergeable (clean) • **Risk:** Low–Medium
**Scope:** 3 files, +80 / −4, 2 commits, 0 reviews, 2 bot comments
**Label:** `build finished` ✅ — All FNALbuild build/integration tests pass at `a4adb5e`. Note: clang-tidy reports **4 errors, 24 warnings** ([log](https://buildmaster.fnal.gov/buildmaster/job/GitHubPRTests/job/mu2e-offline-build-test//3076/artifact/clang-tidy.log)).

The PR extends the blinding/digi-harvesting workflow to include the calorimeter — previously `// calorimeter...` placeholder stubs in two modules. It also registers some `art::Sampled<...>` dictionary entries needed when running ensembling jobs that exercise the blinding chain.

## 2) Core changes

### A. `TrackDigiExtractor_module.cc` — extract CaloDigis associated with a track
Previously only tracker digis were collected. Now, for each `KalSeed` with a calo cluster, walks `caloCluster → caloHits → recoCaloDigis → caloDigiPtr` and deep-copies each `CaloDigi` into an output `CaloDigiCollection`.

```cpp name=Blinding/src/TrackDigiExtractor_module.cc url=https://github.com/Mu2e/Offline/pull/1814/files#diff-19adf578d9fa015f337c577b9ae19b880287452b
if (kalseed.hasCaloCluster()){
  const auto& cseed = kalseed.caloHit();
  const auto& cluster = cseed.caloCluster();
  const auto& hits = cluster->caloHitsPtrVector();
  for (const auto& hit: hits){
    const auto& recos = hit->recoCaloDigis();
    for (const auto& reco: recos){
      const auto& ptr = reco->caloDigiPtr();
      auto digi = CaloDigi(ptr->SiPMID(), ptr->t0(),
                           ptr->waveform(), ptr->peakpos());
      ext_calodigis->push_back(digi);
    }
  }
}
```

### B. `DisplaceDigiTimes_module.cc` — shift calo digi start times
Adds an additional fcl input `CaloDigiCollection` and a new fcl atom `CaloADCSamplePeriod`. For each input calo digi, the existing `shift` (computed for the tracker case) is added to the analog start time `t0 * sample_period`, then converted back to integer sample units.

```cpp name=Blinding/src/DisplaceDigiTimes_module.cc url=https://github.com/Mu2e/Offline/pull/1814/files#diff-59551331265e2a9f5d02b3a2307bad8469465bdf
auto analog = t0 * sample_period;
auto shifted = analog + shift;
t0 = static_cast<CaloDigiWrapper::sample_t>(shifted / sample_period);
cdigis->emplace_back(id, t0, waveform, peakpos);
```

### C. `MCDataProducts/src/classes_def.xml` — sampled product dictionaries
Adds `art::Sampled<...>` + `art::Wrapper<...>` + the underlying `std::map<art::SubRunID,...>` entries for `ProtonBunchIntensity` and `GenEventCount`, needed for sampling/ensembling jobs.

---

## 3) Issues found

### 🔴 Bug: typo `mu2e:` (single colon) in `classes_def.xml`
Six lines use `mu2e:` instead of `mu2e::`. The same lines that include the underlying `std::map<...,mu2e::ProtonBunchIntensity>` use the correct double colon, which strongly suggests this is a typo. ROOT/genreflex will likely fail to resolve these as the intended type.

```xml name=MCDataProducts/src/classes_def.xml url=https://github.com/Mu2e/Offline/pull/1814/files#diff-344de08b3713f82231bb2e93a10ef2e539265e5e
<class name="art::Sampled<mu2e:ProtonBunchIntensity>"/>                              <!-- mu2e: -->
<class name="art::Wrapper< art::Sampled<mu2e:ProtonBunchIntensity> >"/>              <!-- mu2e: -->
<class name="std::map<art::SubRunID,mu2e::ProtonBunchIntensity>"/>                   <!-- mu2e:: ✓ -->
<class name="art::Wrapper< std::map<art::SubRunID,mu2e::ProtonBunchIntensity> >"/>   <!-- mu2e:: ✓ -->
<class name="std::map< std::string,std::map<art::SubRunID,mu2e::ProtonBunchIntensity> >"/>

<class name="art::Sampled<mu2e:GenEventCount>"/>                                     <!-- mu2e: -->
<class name="art::Wrapper< art::Sampled<mu2e:GenEventCount> >"/>                     <!-- mu2e: -->
```

**Fix:** change all `mu2e:` → `mu2e::` in the new `Sampled`/`Wrapper<Sampled>` lines.

### 🟡 Truncation (not rounding) when converting shifted time back to sample index
`static_cast<sample_t>(shifted / sample_period)` truncates toward zero. For a non-integer multiple of `sample_period`, the calo digi will land on the lower sample bin while the tracker shift uses the unmodified continuous value. Two consequences:

- The actual time shift applied to calo waveforms differs from the one used for the track/straw digis by up to one sample period (~ tens of ns for the calo).
- Behavior is asymmetric for negative `shifted` values (truncation toward zero vs. floor).

Consider `static_cast<sample_t>(std::lround(shifted / sample_period))` (or `std::floor`) and document the chosen convention. If this discrepancy is intentional (because the calo is inherently sample-quantized), a comment justifying it would help.

### 🟡 No bounds/validity check on `kalseed.caloHit()`
`hasCaloCluster()` is gated, but if `caloHit().caloCluster()` is ever a null `art::Ptr`, the `cluster->caloHitsPtrVector()` deref will segfault. Worth either a defensive `if (cluster.isNonnull())` check or a comment that the upstream contract guarantees this.

### 🟡 Potential duplicate CaloDigis across KalSeeds
If two `KalSeed`s share a `CaloCluster` (or share constituent `CaloHit`s/`RecoCaloDigi`s), the same underlying `CaloDigi` will be deep-copied multiple times into `ext_calodigis`. The tracker side has the same theoretical exposure, but for calo the cluster-sharing pattern is more common. Consider deduping by `art::Ptr<CaloDigi>` (e.g., `std::set` of pointers) before emplacing — or document that downstream consumers tolerate duplicates.

### 🟢 Minor: identical fhicl `Comment` for new and existing inputs
Both `StrawDigiCollection` and the new `CaloDigiCollection` atoms in `DisplaceDigiTimes` use the comment `"art::InputTag of digis to displace"`. Distinguishing them ("…of straw digis…" / "…of calorimeter digis…") would help fcl readers.

```cpp name=Blinding/src/DisplaceDigiTimes_module.cc url=https://github.com/Mu2e/Offline/pull/1814/files#diff-59551331265e2a9f5d02b3a2307bad8469465bdf
fhicl::Atom<art::InputTag> calo_digi_tag{
  fhicl::Name("CaloDigiCollection"),
  fhicl::Comment("art::InputTag of digis to displace")   // copy of StrawDigiCollection comment
};
```

### 🟢 Minor: `peakpos` unmodified after time shift
The shifted digi keeps the original `peakpos`. Since `peakpos` is typically expressed in samples relative to `t0`, this is probably correct, but worth confirming — and worth a one-line comment in the loop, since reviewers will ask.

### 🟢 Minor: `CaloADCSamplePeriod` as fcl param
Acknowledged in the PR description. Track this as a TODO (conditions-based sampling period) and ensure the prolog value is consistent with what `CaloDigiMaker`/`CaloHitMaker` use elsewhere, otherwise the round-trip will be inconsistent.

---

## 4) Merge readiness

| Aspect | Status |
|---|---|
| Build & integration tests | ✅ all green |
| Mergeable | ✅ clean |
| Reviews | ⏳ none yet |
| clang-tidy | ⚠️ 4 errors, 24 warnings — worth a glance to confirm none are introduced by this PR |

**Blocker for merge, in my view:** the `mu2e:` vs `mu2e::` typo in `classes_def.xml`. The other items are improvements / clarifications.

---

## 5) Suggested improvements

1. Fix the `mu2e:` → `mu2e::` typos.
2. Use `std::lround` instead of truncation for the calo `t0` conversion, or document the truncation choice.
3. Differentiate fcl `Comment` strings for the two `InputTag` atoms in `DisplaceDigiTimes`.
4. Consider deduplicating `CaloDigi`s across KalSeeds in `TrackDigiExtractor`.
5. Add a brief comment near `peakpos` in the displaced digi to clarify it is intentionally not shifted.

---

## 6) Want me to…

1. Inspect the clang-tidy log to confirm whether any of the 4 errors / 24 warnings are introduced by this PR.
2. Pull `CaloDigi.hh` / `CaloDigiWrapper.hh` to confirm the `sample_t` type and recommend `lround` vs `floor` precisely.
3. Check `KalSeed::caloHit()` / `hasCaloCluster()` semantics in the codebase for the null-pointer concern.
4. Search for existing fcl files referencing `DisplaceDigiTimes` / `TrackDigiExtractor` to verify they will need to be updated for the new required parameters (these are required atoms with no defaults — existing fcl will break until updated).
5. Draft a suggested patch comment for the `classes_def.xml` typo fix.

edcallaghan:
@FNALbuild run build test

edcallaghan:
Our robot friend identified 5 suggestions, 3 of which are (potentially) functional and 2 of which are documentary. The documentary suggestions are well-intentioned but IMO below-threshold for humans with context (e.g., I don't expect people will confused about the same fcl description being shared across two detectors, which the fcl parameter has the same effect for those two detectors).

I've implemented two of the functional suggestions:
  - Fixed the scoping-related typo, which is a great catch.
  - Implemented true rounding when digitizing the shifted timestamp of calo digis. It's a judgement call, but probably better this way.

The remaining suggestion, related to deduplicating copies of hits, is another case where the model is just missing context. The use case of the relevant module is in copying select tracks from clean offspill events --- the occupancy is low enough that there won't really ever be multiple kalseeds considered, much less  with degenerate calo clusters. Given that, I'd rather not add to the complexity and runtime by introducing a deduplication step.

Of course, I'm happy to discuss any of these further.

PR #1819, opened 2026-05-06
Calo step selection in calo dts filtering

michaelmackenzie:
This adds separate min/max particle momentum thresholds for calo steps, a total calo energy deposit selection, and some additional diagnostic printouts. In Run 1B work, we found the current filtering can bias the output by filtering out events with high deposits but they shower and separate into different sim particles. This helps resolve this issue, especially when using the new total calo energy deposit selection that will add up this total shower.

### DIFF
--- a/Filters/src/DetectorStepFilter_module.cc
+++ b/Filters/src/DetectorStepFilter_module.cc
@@ -8,6 +8,7 @@
 // art includes.
 #include "fhiclcpp/ParameterSet.h"
 #include "fhiclcpp/types/Atom.h"
+#include "fhiclcpp/types/OptionalAtom.h"
 #include "fhiclcpp/types/Sequence.h"
 #include "fhiclcpp/types/OptionalTable.h"
 #include "cetlib_except/exception.h"
@@ -53,12 +54,16 @@ namespace mu2e {
 
         fhicl::Atom<double> minPartMom { Name("MinimumPartMom"), Comment("Minimum particle momentum"), 0.0 };
         fhicl::Atom<double> maxPartMom { Name("MaximumPartMom"), Comment("Maximum particle momentum"), maxE_ };
+        fhicl::Atom<double> minCaloPartMom{ Name("MinimumCaloPartMom"), Comment("Minimum particle momentum for calo steps"), 0.0 };
+        fhicl::Atom<double> maxCaloPartMom{ Name("MaximumCaloPartMom"), Comment("Maximum particle momentum for calo steps"), maxE_ };
 
         fhicl::Atom<bool> orRequirements { Name("ORRequirements"), Comment("Take the logical OR of requirements, otherwise take the AND"), true };
 
         fhicl::Atom<unsigned> minNTrkSteps { Name("MinimumTrkSteps"), Comment("Minimum number of good tracker steps"), 10};
         fhicl::Atom<double> minSumCaloStepE { Name("MinimumSumCaloStepE"), Comment("Minimum E sum of good calorimeter steps (MeV) "), 50.0 };
         fhicl::Atom<unsigned> minNCrvSteps { Name("MinimumCrvSteps"), Comment("Minimum number of good CRV steps"), 3 };
+
+        fhicl::OptionalAtom<double> minSumCaloTotalE { Name("MinimumSumCaloE"), Comment("Minimum E sum of good calorimeter steps by all good particles (MeV) ")};
 
         fhicl::Atom<unsigned> maxNTrkSteps { Name("MaximumTrkSteps"), Comment("Maximum number of good tracker steps"), maxN_ };
         fhicl::Atom<double> maxSumCaloStepE { Name("MaximumSumCaloStepE"), Comment("Maximum E sum of good calorimeter steps (MeV) "),  maxE_ };
@@ -68,6 +73,7 @@ namespace mu2e {
 
         fhicl::OptionalTable<TimeCutConfig> timeCutConfig { fhicl::Name("TimeCutConfig") };
 
+        fhicl::Atom<int> diagLevel { Name("DiagLevel"), Comment("Diagnostic output level"), 0 };
       };
 
       using Parameters = art::EDFilter::Table<Config>;
@@ -79,17 +85,19 @@ namespace mu2e {
       bool goodParticle(SimParticle const& simp) const; // select particles whose steps to count
       bool timeCut(double time) const; // particle time
       double minTrkE_, minCaloE_, minCrvE_;
-      double minPartM_, maxPartM_;
+      double minPartM_, maxPartM_, minCaloPartM_, maxCaloPartM_;
       bool or_;
       std::vector<PDGCode::type> pdgToKeep_;
       std::vector<art::InputTag> trkStepCols_, caloStepCols_, crvStepCols_;
       unsigned minNTrk_, minNCrv_;
-      double minSumCaloE_;
+      double minSumCaloE_, minSumCaloTotalE_;
+      bool useMinSumCaloTotalE_;
       unsigned maxNTrk_, maxNCrv_;
       double maxSumCaloE_;
       bool timecut_;
       double minTime_, maxTime_;
       unsigned nEvt_, nPassed_;
+      int diagLevel_;
   };
 
   //================================================================
@@ -100,10 +108,13 @@ namespace mu2e {
     , minCrvE_(conf().minCrvStepEnergy())
     , minPartM_(conf().minPartMom())
     , maxPartM_(conf().maxPartMom())
+    , minCaloPartM_(conf().minCaloPartMom())
+    , maxCaloPartM_(conf().maxCaloPartMom())
     , or_(conf().orRequirements())
     , minNTrk_(conf().minNTrkSteps())
     , minNCrv_(conf().minNCrvSteps())
     , minSumCaloE_(conf().minSumCaloStepE())
+    , useMinSumCaloTotalE_(conf().minSumCaloTotalE(minSumCaloTotalE_))
     , maxNTrk_(conf().maxNTrkSteps())
     , maxNCrv_(conf().maxNCrvSteps())
     , maxSumCaloE_(conf().maxSumCaloStepE())
@@ -111,6 +122,7 @@ namespace mu2e {
     , minTime_(0.0), maxTime_(0.0)
     , nEvt_(0)
     , nPassed_(0)
+    , diagLevel_(conf().diagLevel())
     {
       for(const auto ikeep : conf().keepPDG()) { pdgToKeep_.emplace_back(PDGCode::type(ikeep)); }
       for(const auto& trktag : conf().trkSteps()) { trkStepCols_.emplace_back(trktag); consumes<StrawGasStepCollection>(trktag); }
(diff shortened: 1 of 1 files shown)


oksuzian:
# PR #1819 Review — Calo step selection in calo dts filtering

## PR Summary

| | |
|---|---|
| **Author** | @michaelmackenzie |
| **Branch** | `michaelmackenzie:CaloStepFiltering` → `Mu2e:main` |
| **State** | Open, not draft, mergeable (`unstable` due to clang-tidy informational) |
| **Files changed** | 1 (`Filters/src/DetectorStepFilter_module.cc`), +51/−6 |
| **Commits** | 1 |
| **Reviews** | 0 (no human reviews yet) |
| **Comments** | 2 (FNALbuild bot only) |
| **CI** | ✅ build passed at `a9839ee`, all integration tests green; ⚠️ clang-tidy reports **2 errors / 11 warnings**, ✅ whitespace clean, ✅ no FIXMEs/TODOs |

**Intent:** Adds three things to the `DetectorStepFilter` module:
1. Separate min/max particle-momentum cuts that apply only to **calorimeter** steps (`MinimumCaloPartMom` / `MaximumCaloPartMom`), independent of the global `MinimumPartMom`/`MaximumPartMom`.
2. A new optional **total** calo energy-deposit cut (`MinimumSumCaloE`) summed across all good particles, in addition to the existing per-particle sum (`MinimumSumCaloStepE`). This is motivated by Run 1B observations where high-energy showers got filtered out because their energy was split across multiple SimParticles.
3. A new `DiagLevel` knob with verbose printouts at four severity tiers.

No linked issue; no acceptance criteria to track.

---

## Core Changes

### 1. Calo-specific particle momentum cut

```cpp name=Filters/src/DetectorStepFilter_module.cc url=https://github.com/Mu2e/Offline/pull/1819/files#diff-47face9bbad8cc21aa02971adfe23dd3160bb82fR56-L57
fhicl::Atom<double> minCaloPartMom{ Name("MinimumCaloPartMom"), Comment("Minimum particle momentum for calo steps"), 0.0 };
fhicl::Atom<double> maxCaloPartMom{ Name("MaximumCaloPartMom"), Comment("Maximum particle momentum for calo steps"), maxE_ };
```

In the calo loop, `minPartM_/maxPartM_` are replaced by `minCaloPartM_/maxCaloPartM_`. Defaults preserve effectively the prior behavior (0 / maxE_), so this is **backward compatible only if users were relying on `MinimumPartMom > 0` to filter calo steps** — see Issue #1 below.

### 2. Total calo energy selection

```cpp name=Filters/src/DetectorStepFilter_module.cc url=https://github.com/Mu2e/Offline/pull/1819/files#diff-47face9bbad8cc21aa02971adfe23dd3160bb82fR67
fhicl::OptionalAtom<double> minSumCaloTotalE { Name("MinimumSumCaloE"), Comment("Minimum E sum of good calorimeter steps by all good particles (MeV) ")};
```

Track `total_edep` over all steps that pass the per-step `minCaloE_`, calo momentum cuts, `goodParticle`, and time cuts. If `useMinSumCaloTotalE_` is true and `total_edep > minSumCaloTotalE_`, set `selectcalo = true`.

### 3. Diagnostic printouts gated by `DiagLevel`
- `>0`: print event id + selecttrk/calo/crv when event passes; print when total-edep cut fires.
- `>1`: print per-particle calo edep sum.
- `>2`: print per-step calo info.
- `>3`: always print event-level decision.

---

## Issues Found

### 🔴 Issue 1 — Behavior change: `minPartM_` / `maxPartM_` no longer apply to calo steps

Previously the calo loop used `minPartM_`/`maxPartM_`. After this PR it uses **only** the new `minCaloPartM_`/`maxCaloPartM_`, whose defaults are 0 / maxE_. Any existing fhicl configuration that set `MinimumPartMom` expecting it to filter calo SimParticles will silently change behavior. Two options:

- **Recommended:** keep applying `minPartM_/maxPartM_` AND the new calo-specific cuts (intersection). That way the new params are strictly tighter when set, never looser.
- Or: explicitly document the breaking change in the comment of `MinimumCaloPartMom` and grep for downstream fcl files using `MinimumPartMom`.

```cpp name=Filters/src/DetectorStepFilter_module.cc url=https://github.com/Mu2e/Offline/blob/a9839ee/Filters/src/DetectorStepFilter_module.cc#L177-L180
if(edep_step > minCaloE_ &&
    css.momentumIn() > minCaloPartM_ && css.momentumIn() < maxCaloPartM_ &&
    goodParticle(*css.simParticle()) &&
    timeCut(fmod(css.time(),mbtime))) {
```

### 🟠 Issue 2 — fhicl name `MinimumSumCaloE` collides conceptually with `MinimumSumCaloStepE`

The new parameter is named `MinimumSumCaloE` but the C++ variable is `minSumCaloTotalE_`. Users will have to distinguish:
- `MinimumSumCaloStepE` — min per-particle sum
- `MinimumSumCaloE` — min total sum across all particles

These are easy to confuse. Suggest renaming the fhicl key to **`MinimumSumCaloTotalE`** so it matches the variable and is unambiguous. (Since this is a brand-new param, renaming is cheap.)

### 🟠 Issue 3 — `>` vs `>=` inconsistency on the new total cut

The new total cut uses strict `>`, while the per-particle sum cut uses `>=`:

```cpp name=Filters/src/DetectorStepFilter_module.cc url=https://github.com/Mu2e/Offline/blob/a9839ee/Filters/src/DetectorStepFilter_module.cc#L194-L213
if(icalo->second >= minSumCaloE_ && icalo->second <= maxSumCaloE_ ){   // existing: >=
    ...
}
...
if(total_edep > minSumCaloTotalE_) {                                    // new: >
    selectcalo = true;
```

For consistency (and so a configured threshold of e.g. 50.0 means "≥ 50 MeV passes" everywhere), use `>=`.

### 🟠 Issue 4 — No upper bound on the new total cut

Existing per-particle sum has both `MinimumSumCaloStepE` and `MaximumSumCaloStepE`. The new total selection has only a min — there is no `MaximumSumCaloE`. If symmetry is desired (and to allow rejecting pathological events), add a matching `MaximumSumCaloE` (also OptionalAtom). At minimum, document why the asymmetry is intentional.

### 🟡 Issue 5 — `total_edep` declared outside the calocoltag loop (likely correct, worth confirming)

```cpp name=Filters/src/DetectorStepFilter_module.cc url=https://github.com/Mu2e/Offline/blob/a9839ee/Filters/src/DetectorStepFilter_module.cc#L173-L221
float total_edep = 0.; // for total calorimeter deposits
for(const auto& calocoltag : caloStepCols_) {
    CES caloESum;
    ...
    for(const auto& css : *csscolH ) {
       ...
       total_edep += edep_step;
       ...
    }
    ...
    if(useMinSumCaloTotalE_) {
      if(total_edep > minSumCaloTotalE_) { selectcalo = true; ... }
    }
    if(selectcalo) break;
}
```

`total_edep` accumulates **across** calo step collections, which is probably what you want ("total shower"), but the per-particle `caloESum` is reset per collection. Worth a one-line comment to make the intent explicit, since a future reader may "fix" it the wrong way.

### 🟡 Issue 6 — "Total shower" label is slightly misleading

The PR description says the new selection sums the "total shower," but `total_edep` only includes steps that already passed `minCaloE_`, calo momentum cuts, and `goodParticle()`. If a shower spawns sub-threshold daughters or non-keep-PDG daughters, their energy is excluded. This is recoverable by configuring `MinimumCaloPartMom = 0`, `MinimumCaloStepE = 0`, and a permissive `keepPDG`, but consider whether the cuts should be intentionally relaxed for the total-edep accumulation, or document the caveat in the fhicl comment.

### 🟡 Issue 7 — `total_edep` not reported in event diagnostic when selection comes from per-particle path

`diagLevel_ > 0` prints `total_edep` only when the **total** cut fires. When `selectcalo` is set via the per-particle path, the event-level printout (`diagLevel_ > 0 && retval`) does not include `total_edep`. Including it always would make Run 1B–style debugging easier.

### 🟡 Issue 8 — Member declaration order vs. init list (minor, currently safe)

Header order:
```
double minSumCaloE_, minSumCaloTotalE_;
bool useMinSumCaloTotalE_;
```
Init list initializes `useMinSumCaloTotalE_` via `conf().minSumCaloTotalE(minSumCaloTotalE_)` which writes to `minSumCaloTotalE_` by reference. It's safe today, but `minSumCaloTotalE_` is left uninitialized when the optional is absent. Consider explicitly initializing it (e.g. `, minSumCaloTotalE_(0.)`) before the line that reads the optional, to avoid future surprises and silence any uninitialized-read static analyzers.

### 🟡 Issue 9 — clang-tidy reports 2 errors / 11 warnings

Bot output: [clang-tidy.log](https://buildmaster.fnal.gov/buildmaster/job/GitHubPRTests/job/mu2e-offline-build-test//3085/artifact/clang-tidy.log). Worth checking whether any of the 2 errors / 11 warnings were introduced (or surfaced) by this PR before merging.

---

## Merge Readiness

| Criterion | State |
|---|---|
| Build & integration tests | ✅ all green |
| Code review approvals | ⚠️ none yet |
| Behavior preservation | ⚠️ Issue 1 — silent change to existing fcls that set `MinimumPartMom` |
| Naming clarity | ⚠️ Issue 2 — `MinimumSumCaloE` vs `MinimumSumCaloStepE` |
| Static analysis | ⚠️ clang-tidy errors not yet triaged |

**Recommended before merge:** address Issue 1 (decide preservation vs. breaking) and Issue 2 (rename fhicl key while it's still cheap). Issues 3–9 are nice-to-haves.

---

## Possible Improvements

1. Add a unit/integration test (or a tiny test fcl) exercising the new total-energy path so future refactors don't silently regress it.
2. Consolidate the four diagnostic blocks into a small helper or `mf::LogDebug` to keep `filter()` readable.
3. Initialize `total_edep` at the top of `filter()` per event (it currently is, since it's a local — fine, but worth a comment explaining "intentionally summed across all calo collections").
4. Bump `MaximumPartMom`'s comment in the header to mention it now applies only to tracker/CRV steps.

---

## Want me to…

1. Pull and summarize the **clang-tidy errors/warnings** from the build artifact to see if any are introduced by this PR.
2. Search downstream fcl files in `Mu2e/Offline` that set `MinimumPartMom` on `DetectorStepFilter` to quantify the blast radius of Issue 1.
3. Draft a suggested patch implementing Issues 1–4 (preservation of `minPartM_`, rename, `>=`, optional `MaximumSumCaloE`).
4. Look up other recent calo-filtering PRs / Run 1B discussions for added context on the motivation.

michaelmackenzie:

1. We should update the fcl, this is solving the issue that the current params are too strict here.
2. I can change the existing fcl parameter name if preferred, but this will require fcl updates in time with this PR merging.
3. Using `>` is necessary to not accept events with 0 steps, which is already true for the calo step check due to the for loop.
4. I don't believe a maximum is needed for this selection, but we can add one if it's necessary later.
5. This seems fine to me, it's the total deposit independent of the mixing source.
6. I think it's already described that those cuts are used when evaluating the total deposits, as is done for the per-particle deposit checks.
7. I've added an additional printout for this if `diagLevel_ > 1`.
8. The total value should only ever be used if it's enabled, so not being set when not enabled doesn't seem like an issue?
9. I don't think we enforce these.

PR #1825, opened 2026-05-12
Added diagnostic kinkal strawhitupdater

bonventre:
Simple SHU that allows you to force all hits from a single panel or plane to be inactive in the kinkal fits. If done by panel status instead no hits would be created, this allows information from that panel to still make it into EventNtuple for alignment diagnostics etc.

### DIFF
--- a/Mu2eKinKal/CMakeLists.txt
+++ b/Mu2eKinKal/CMakeLists.txt
@@ -3,6 +3,7 @@ cet_make_library(
       src/BkgANNSHU.cc
       src/CADSHU.cc
       src/Chi2SHU.cc
+      src/PanelDiagSHU.cc
       src/DriftANNSHU.cc
       src/KKBField.cc
       src/KKConstantBField.cc
--- a/Mu2eKinKal/inc/KKFitSettings.hh
+++ b/Mu2eKinKal/inc/KKFitSettings.hh
@@ -16,6 +16,7 @@
 #include "Offline/Mu2eKinKal/inc/DriftANNSHU.hh"
 #include "Offline/Mu2eKinKal/inc/BkgANNSHU.hh"
 #include "Offline/Mu2eKinKal/inc/Chi2SHU.hh"
+#include "Offline/Mu2eKinKal/inc/PanelDiagSHU.hh"
 #include "Offline/Mu2eKinKal/inc/StrawXingUpdater.hh"
 namespace mu2e {
   namespace Mu2eKinKal{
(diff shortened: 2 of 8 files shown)


brownd1978 (approved):
Looks good. you just need to add the new class to the CMakeList.

brownd1978:
@FNALbuild run build test

oksuzian:
# PR #1825 Review: Diagnostic kinkal strawhitupdater

## Summary
- **Author**: @bonventre · **State**: open · **Mergeable**: clean
- **Scope**: 8 files, +75/-3 · **Reviews**: 1 approval (@brownd1978)
- **CI**: ✅ Build passes (commit `0a543f6`)
- **Purpose**: Adds `PanelDiagSHU` — a diagnostic StrawHitUpdater that forces all hits in a chosen panel/plane to be inactive in KinKal fits. Unlike disabling via panel status, hits are still produced so they appear in `EventNtuple` for alignment diagnostics.

## Core Changes
1. New class `PanelDiagSHU` (header + impl) — uses `(StrawIdMask, StrawId, diag_level)` config. If the hit's `StrawId` matches the masked target id, sets `state_=inactive`, `algo_=PanelDiag`, `frozen_=true`.
2. Wired into the updater enum (`PanelDiag=9`) and `names_` table.
3. Hooked into `KKFitSettings.{hh,cc}` and `KKStrawHit.hh` updater dispatch.
4. Added to `Mu2eKinKal/CMakeLists.txt` (in response to @brownd1978's review).

---

## 🔴 Critical / Notable Issues

### 1. Deactivation only fires when `ca.usable()` is true
In `Mu2eKinKal/inc/KKStrawHit.hh`:

```cpp name=Mu2eKinKal/inc/KKStrawHit.hh url=https://github.com/Mu2e/Offline/pull/1825/files#diff-KKStrawHit
CA ca = unbiasedClosestApproach();
if(ca.usable()){
  auto dinfo = fillDriftInfo(ca);
  if(diagshu)whstate_ = diagshu->wireHitState(whstate_,straw_.id());
  if(cashu) ...
}
```

`PanelDiagSHU::wireHitState` only depends on the `StrawId` — not on closest approach. Burying it inside `if(ca.usable())` means panels won't be deactivated for hits where CA is unusable. For an alignment-diagnostic tool whose entire job is "force this panel's hits inactive regardless," this is a behavioral hole. Consider moving the `diagshu` invocation outside the `ca.usable()` block (or the entire branch) so deactivation is unconditional.

### 2. `npdiag` advance logic is inconsistent with sibling updaters
In `Mu2eKinKal/src/KKFitSettings.cc`:

```cpp name=Mu2eKinKal/src/KKFitSettings.cc url=https://github.com/Mu2e/Offline/pull/1825/files#diff-KKFitSettings
} else if(alg == StrawHitUpdaters::PanelDiag) {
  miconfig.addUpdater(std::any(PanelDiagSHU(paneldiagshusettings.at(npdiag))));
  if (paneldiagshusettings.size()>npdiag+1)npdiag++;
}
```

All other updaters use `at(n++)` (one config per meta-iteration). Here, the index is "sticky" at the last config, so a single config is reused for every meta-iteration. This may be intentional (you usually want the same panel forced off across all iterations), but:
- It silently diverges from the established pattern.
- There's no comment explaining why.
- If the user *does* supply N configs for N meta-iterations, the last one will still be reused for any further iterations rather than throwing.

Either add a brief comment justifying the sticky behavior, or unify with the `at(npdiag++)` pattern.

### 3. Minor: diagnostic `std::cout` in production code
`PanelDiagSHU.cc` prints to `std::cout` when `diag_>0`. The codebase typically uses MF logging; raw `cout` from a fit-time updater could clutter logs during long jobs. Low priority since gated on `diag_`.

### 4. Minor: missing default in config description
`configDescription()` returns `"StrawId mask, StrawId to set inactive, diag level"` — no documentation of valid `StrawIdMask` level names (e.g. `"plane"`, `"panel"`, `"station"`). Worth a one-liner so users of the FHiCL knob aren't guessing.

---

## Merge Readiness
- ✅ CI green, ✅ approved, ✅ CMake fix applied, ✅ no whitespace/check_cmake errors on latest commit.
- ⚠️ Issue #1 (CA-gated deactivation) is the only one I'd consider a real correctness concern — worth confirming with the author whether it's intentional before merging.

oksuzian:
@FNALbuild run build test

bonventre:
Looking at the AI review, I'd say 1 and 2 are intended behavior. If ca is unusable the hit will correctly get marked as unusable instead of just inactive, and since it is a diagnostic, it is assumed that you will want the same panel disabled in every iteration.

PR #1826, opened 2026-05-12
New tracker conditions tables for TOT and HV trips

bonventre:
Adds three new database tables. Two tables implementing the TOT=>drift time calibration, and one table to hold a list of HV trips. Since trips are expected to be within a run/subrun, they are stored as a list of panels + start and end event id

### DIFF
--- a/TrackerConditions/fcl/testTrackerAlignment.fcl
+++ b/TrackerConditions/fcl/testTrackerAlignment.fcl
@@ -3,6 +3,7 @@ process_name : TestTrkAlign
 
 source : {
   module_type : EmptyEvent
+  firstRun : 1400
 }
 services : @local::Services.Core
 
@@ -21,8 +22,10 @@ physics : {
 # turn on alignment
 services.ProditionsService.alignedTracker.useDb: true
 services.ProditionsService.alignedTracker.verbose: 2
+services.DbService.purpose : "Sim_best"
+services.DbService.version : "v1_4"
 # select nominal (perfect) alignment or a misaligned file
-services.DbService.textFile : ["Offline/TrackerConditions/data/MisalignTracker.txt"]
+services.DbService.textFile : ["align.txt"]
 # services.DbService.verbose: 2
 
 services.TFileService.fileName: "testAlign.root"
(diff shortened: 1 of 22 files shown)


TrackerConditions/fcl/testTrackerAlignment.fcl
```diff
@@ -21,8 +22,10 @@ physics : {
 # turn on alignment
 services.ProditionsService.alignedTracker.useDb: true
 services.ProditionsService.alignedTracker.verbose: 2
+services.DbService.purpose : "Sim_best"
+services.DbService.version : "v1_4"
 # select nominal (perfect) alignment or a misaligned file
-services.DbService.textFile : ["Offline/TrackerConditions/data/MisalignTracker.txt"]
+services.DbService.textFile : ["align.txt"]
```
rlcee, line 28: This looks unintentional

rlcee (changes requested):
I know we talked about this before, but I don't recall the conclusion. What will knowing trips do for reco?
If a trip can cross a subrun boundary, it might need the subrun number, since currently event number has to be reset at each subrun
If Panels trip, the row should return a PanelId
we have a guideline of one class definition per include file
I'd be concerned for the per-table overhead, and the two tables can get out of sync - you decided against the one-table designs?
The rest of the implementation looks straightforward

TrackerConditions/fcl/testTrackerAlignment.fcl
rlcee: Ah, maybe intentional in a test fcl..

bonventre: mistaken commit, will revert

bonventre:
Right now the impact in actual reco of the trip table is small (hits in those panels are assumed to be noise and excluded), but then also used to update hit status flags which we need correct in EventNtuple to do efficiency calculations etc. I'm not sure if there's a good way to do that efficiently in a post-reco analysis only step.

I've adjust the TOT table to one of the other options we discussed, which was a space separated text field for the array values instead. Performancewise you think this is the better approach here?

rlcee:
Performance-wise I think going from 3 tables to 2, is a notable improvement.  After that, it is hard to say.  Thanks

rlcee (approved):

PR #1829, opened 2026-05-14
Clean up CPR a bit

michaelmackenzie:
This includes various code cleaning changes, storing the helix fit results in fields instead of the fitter object, and other changes. I also deleted commented out code in many places as the algorithm has been fairly stable for a while now.

### DIFF
--- a/CalPatRec/inc/CalHelixFinderData.hh
+++ b/CalPatRec/inc/CalHelixFinderData.hh
@@ -132,33 +132,27 @@ namespace mu2e {
 //-----------------------------------------------------------------------------
 // circle parameters; the z center is ignored.
 //-----------------------------------------------------------------------------
-    ::LsqSums4         _sxy;
-    ::LsqSums4         _szphi;
+    ::LsqSums4         _sxy;    // circle fitter
+    ::LsqSums4         _szphi;  // dphi / dz line fitter
 
-    XYZVectorF             _center;
-    float             _radius;
+    XYZVectorF         _center; // circle fit results
+    float              _radius;
+    float              _circle_chisq_dof; // circle fit quality
 
-    //    float             _chi2;
-//-----------------------------------------------------------------------------
-// 2015-02-06 P.Murat: fit with non-equal weights - XY-only
-//-----------------------------------------------------------------------------
-    // ::LsqSums4         _sxyw;
-    // XYZVectorF             _cw;
-    // float             _rw;
-    // float             _chi2w;
 //-----------------------------------------------------------------------------
 // Z parameters; dfdz is the slope of phi vs z (=-sign(1.0,qBzdir)/(R*tandip)),
 // fz0 is the phi value of the particle where it goes through z=0
 // note that dfdz has a physical ambiguity in q*zdir.
 //-----------------------------------------------------------------------------
-    float             _dfdz;
-    float             _fz0;
+    float              _dfdz; // dphi / dz fit results
+    float              _fz0;
+    float              _dfdz_chisq_dof; // dphi / dz fit quality
 //-----------------------------------------------------------------------------
 // diagnostics, histogramming
 //-----------------------------------------------------------------------------
     Diag_t             _diag;
 //-----------------------------------------------------------------------------
-// structure used to organize thei strawHits for the pattern recognition
+// structure used to organize the strawHits for the pattern recognition
 //-----------------------------------------------------------------------------
 //    PanelZ_t                                           _oTracker[kNTotalPanels];
 //    std::array<int,kNTotalPanels*kNMaxHitsPerPanel>     _hitsUsed;
--- a/CalPatRec/inc/CalHelixFinder_module.hh
+++ b/CalPatRec/inc/CalHelixFinder_module.hh
@@ -75,7 +75,6 @@ namespace mu2e {
 // event object labels
 //-----------------------------------------------------------------------------
     std::string                           _shLabel ; // MakeStrawHit label (makeSH)
-    // std::string                           _shpLabel;
     std::string                           _timeclLabel;
 
     int                                   _minNHitsTimeCluster; //min nhits within a TimeCluster after check of Delta-ray hits
@@ -96,7 +95,6 @@ namespace mu2e {
     const ComboHitCollection*             _chcol;
     const TimeClusterCollection*          _timeclcol;
 
-    HelixTraj*                            _helTraj;
     CalHelixFinderAlg                     _hfinder;
     CalHelixFinderData                    _hfResult;
     std::vector<mu2e::Helicity>           _hels; // helicity values to fit
@@ -161,6 +159,10 @@ namespace mu2e {
     int  goodHitsTimeCluster(const TimeCluster* TimeCluster);
 
     void pickBestHelix(std::vector<HelixSeed>& HelVec, int &Index_best);
+
+    void fillDiagnosticInfo(const std::vector<HelixSeed>& helix_seed_vec,
+                            const std::vector<float>& nHitsRatio_vec,
+                            int index_best, const int nGoodTClusterHits);
   };
 }
 #endif
--- a/CalPatRec/inc/CalHelixFinder_types.hh
+++ b/CalPatRec/inc/CalHelixFinder_types.hh
@@ -39,7 +39,6 @@ namespace mu2e {
 
     struct Data_t {
       const art::Event*    event;
-      std::string          shLabel;
 
       enum  { kMaxSeeds = 100, kMaxHits = 200 };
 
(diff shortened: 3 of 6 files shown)


oksuzian:
# Code Review — Mu2e/Offline PR #1829: "Clean up CPR a bit"

This PR is mostly a cleanup of `CalPatRec`: deleting commented-out code, refactoring the diagnostic-fill block of `CalHelixFinder::produce` into `fillDiagnosticInfo`, replacing the O(N×M) helix-overlap loop in `pickBestHelix` with an `unordered_set`, guarding `_helTraj`/`shLabel` removal, gating histogram booking on `_diagLevel`, and adding new `_circle_chisq_dof` / `_dfdz_chisq_dof` cached fit-quality fields on `CalHelixFinderData`.

Focused review on **correctness, framework contracts, and Proditions/services** (default focus, since no specific area was requested).

---

## 🔴 Critical Issues

- **`CalPatRec/src/CalHelixFinderAlg.cc:1066` — wrong source for `_dfdz_chisq_dof` (copy/paste bug).**
  After a successful `doLinearFitPhiZ`, the new code assigns the φ-z (dφ/dz line) fit quality from the **circle** fitter:

  ```c++ name=CalPatRec/src/CalHelixFinderAlg.cc url=https://github.com/Mu2e/Offline/blob/649d2e6fd0df2e80e7c7f268c92bd3d000577e10/CalPatRec/src/CalHelixFinderAlg.cc#L1063-L1067
  else if (success) {                               // update helix results
    Helix._fz0  = Helix._szphi.phi0();
    Helix._dfdz = Helix._szphi.dfdz();
    Helix._dfdz_chisq_dof = Helix._sxy.chi2DofCircle();   // <-- BUG: this is the XY-circle chi2/dof
  }
  ```

  `_dfdz_chisq_dof` is documented in the header as "dphi / dz fit quality" (`CalHelixFinderData.hh` L153) and is then compared against `_chi2zphiMax` in `fitHelix` ([line 346](https://github.com/Mu2e/Offline/blob/649d2e6fd0df2e80e7c7f268c92bd3d000577e10/CalPatRec/src/CalHelixFinderAlg.cc#L346)):

  ```c++ name=CalPatRec/src/CalHelixFinderAlg.cc url=https://github.com/Mu2e/Offline/blob/649d2e6fd0df2e80e7c7f268c92bd3d000577e10/CalPatRec/src/CalHelixFinderAlg.cc#L343-L348
  else if ((Helix._nXYSh < _minNHits) || (Helix._circle_chisq_dof > _chi2xyMax)) {
    Helix._fit = TrkErrCode(TrkErrCode::fail,3); // xy reconstruction failure
  }
  else if ((Helix._nZPhiSh < _minNHits) || (Helix._dfdz_chisq_dof > _chi2zphiMax) ||
           (fabs(Helix._dfdz) < _minDfDz) || (fabs(Helix._dfdz) > _maxDfDz)) {
    Helix._fit = TrkErrCode(TrkErrCode::fail,4); // phi-z reconstruction failure
  }
  ```

  With the current PR, a helix whose circle fit happens to have `chi2/dof > _chi2zphiMax` will be rejected with `fail,4` even though its actual φ-z line fit is fine; conversely, a track with a poor φ-z line fit but a tight circle fit will sneak past the cut. This will change physics output (`_fit` success flag, downstream HelixSeed acceptance), not just diagnostics.

  **Fix:**
  ```c++
  Helix._dfdz_chisq_dof = Helix._szphi.chi2DofLine();
  ```
  Rationale: invariant — `_dfdz_chisq_dof` must always be the φ-z line χ²/dof, since that is the value the success criterion in `fitHelix` is based on.

---

## 🟡 Suggestions

- **`_dfdz_chisq_dof` semantics: 0.f means "no chi2" but the cut treats it as "very good".**
  In several paths the value is intentionally set to `0.f` after `findDfDz` (which is a histogram-based estimate, not an LSQ line fit):
  - `CalHelixFinderAlg.cc` [L1787](https://github.com/Mu2e/Offline/blob/649d2e6fd0df2e80e7c7f268c92bd3d000577e10/CalPatRec/src/CalHelixFinderAlg.cc#L1787) (after `findDfDz` in `doPatternRecognition`)
  - `CalHelixFinderAlg.cc` [L2960](https://github.com/Mu2e/Offline/blob/649d2e6fd0df2e80e7c7f268c92bd3d000577e10/CalPatRec/src/CalHelixFinderAlg.cc#L2960) and [L2987](https://github.com/Mu2e/Offline/blob/649d2e6fd0df2e80e7c7f268c92bd3d000577e10/CalPatRec/src/CalHelixFinderAlg.cc#L2987) in `findTrack`
  - `CalHelixFinderAlg.cc` [L3091](https://github.com/Mu2e/Offline/blob/649d2e6fd0df2e80e7c7f268c92bd3d000577e10/CalPatRec/src/CalHelixFinderAlg.cc#L3091) at end of `findTrack`

  Once the bug above is fixed, `0.f` will silently bypass the `_chi2zphiMax` cut at L346. If that is intentional ("we don't have a line χ² yet, accept on hit count alone"), please add a comment; otherwise consider a sentinel (`-1.f`) and adjust the cut to `(_dfdz_chisq_dof >= 0.f && _dfdz_chisq_dof > _chi2zphiMax)`. Either way the choice should be explicit, not implicit.

- **`CalHelixFinder_module.cc` — histogram booking now gated, but the corresponding `_hmanager` may not be safe to call in `produce` if it was never booked.**
  ```c++ name=CalPatRec/src/CalHelixFinder_module.cc url=https://github.com/Mu2e/Offline/blob/649d2e6fd0df2e80e7c7f268c92bd3d000577e10/CalPatRec/src/CalHelixFinder_module.cc
  void CalHelixFinder::beginJob(){
    if(_diagLevel != 0) { // only book histograms if running diagnostics
      art::ServiceHandle<art::TFileService> tfs;
      _hmanager->bookHistograms(tfs);
    }
  }
  ```
  The fill site is guarded the same way (`if (_diagLevel > 0) _hmanager->fillHistograms(&_data);`), so this is fine **as long as** every code path that calls `_hmanager->fillHistograms` is gated on `_diagLevel > 0`. The booking gate uses `!= 0` while the fill gate uses `> 0`; if `_diagLevel` can be negative, booking happens but fill doesn't — harmless, but inconsistent. Use the same predicate in both places.

- **`pickBestHelix` early return leaves `Index_best` uninitialized for the caller.**
  ```c++ name=CalPatRec/src/CalHelixFinder_module.cc url=https://github.com/Mu2e/Offline/blob/649d2e6fd0df2e80e7c7f268c92bd3d000577e10/CalPatRec/src/CalHelixFinder_module.cc
  if (HelVec.size() < 2) { return; }
  ```
  This is preexisting, but since the function is now being touched, please make it explicit — e.g. `Index_best = (HelVec.empty() ? -1 : 0); return;`. Otherwise the caller's `index_best` may be whatever default the caller wrote (0), which is correct only by accident.

- **`fillDiagnosticInfo` takes `int index_best` by value, then mutates it.**
  ```c++ name=CalPatRec/src/CalHelixFinder_module.cc url=https://github.com/Mu2e/Offline/blob/649d2e6fd0df2e80e7c7f268c92bd3d000577e10/CalPatRec/src/CalHelixFinder_module.cc
  if (index_best == 2){ index_best = 0; }
  ```
  Functionally fine (it's a copy), but consider naming the parameter differently or doing the remap in the caller — `index_best == 2` means "both helices kept", and silently aliasing it to helix 0 only in diagnostics hides a real semantic question: which seed's diagnostics should be reported when both are kept? Worth a one-line comment.

- **Helix overlap comparison uses float threshold against int count.**
  ```c++ name=CalPatRec/src/CalHelixFinder_module.cc url=https://github.com/Mu2e/Offline/blob/649d2e6fd0df2e80e7c7f268c92bd3d000577e10/CalPatRec/src/CalHelixFinder_module.cc
  constexpr float overlap_threshold(0.5f);
  ...
  if ((natc > nh1*overlap_threshold) || (natc > nh2*overlap_threshold)) {
  ```
  Matches the previous `natc > nh1/2.` semantics, but `int * 0.5f` for large `nh1` is fine and equivalent to `nh1/2` only because 0.5 is representable. For clarity and to keep behavior identical across compilers, prefer integer arithmetic: `natc * 2 > nh1` (i.e., strictly more than half).

- **Header field grouping in `CalHelixFinderData.hh`** — `_circle_chisq_dof` is placed next to `_radius`, and `_dfdz_chisq_dof` next to `_fz0`. Good. Consider initializing both to a sentinel (e.g., `-1.f`) in `clearTempVariables()` / `clearResults()` (not visible in the diff but worth checking) so a never-fit helix doesn't quietly read 0 and pass the cut.

### Summary
One real correctness bug (`_dfdz_chisq_dof = Helix._sxy.chi2DofCircle()` at L1066 must be `_szphi.chi2DofLine()`) plus a related semantics question about the `0.f` sentinel — both worth fixing before merge because they directly affect the `TrkErrCode::fail,4` decision and hence the produced `HelixSeedCollection`. Everything else is cleanup that improves the code.

michaelmackenzie:
@FNALbuild run build test

Issue #1804, opened 2026-04-21
CalPatRec: static-review audit — correctness, robustness, convention, dead-code findings


oksuzian:
### Preamble

The items below come from a static review of `CalPatRec/` on `main` — no build, no runtime test, no physics validation was performed. Each finding is either a direct read of the source or (where noted) a cross-reference against sibling modules. I've verified the top-tier bugs by re-fetching the exact lines against `main`; the rest is best-effort.

Intent is to give maintainers a single triaged punch list. Items marked **Bugs / Blockers** are high-confidence; the rest decrease in certainty as the tier name suggests.

---

## Bugs / Blockers

1. **`src/DeltaFinderAlg.cc:596`** — `float y1 = dc1->Xc();` should be `dc1->Yc();`. The subsequent `d2 = (x1-x2)² + (y1-y2)²` compares the wrong coordinate, so two-delta merging uses `(Xc, Xc)` vs `(Xc, Yc)`.
2. **`src/DeltaFinderAlg.cc:792`** — in `pruneSeeds`: `float chi2_ds2 = ds1->Chi2TotN() + ds1->Chi2Time();` uses `ds1` twice; should be `ds2`. `ds2` is currently only rejected on `ds1`'s recomputed value.
3. **`src/PhiClusterFinder_module.cc:416–420`** — right-walk loop has no `nsteps`/`nbx` guard that the left-walk has. If every bin ≥ `_threshold`, infinite loop.
4. **`src/PhiClusterFinder_module.cc:404`** — `if (bincheck > _threshold) bincheck--;` compares a bin index against a content threshold (category error). Should be `bincheck > 1`.
5. **`src/AgnosticHelixFinder_module.cc:831, 843`** — `for (size_t i = 0; i < _tcHits.size() - 2; i++)` unsigned-underflows when `_tcHits.size() < 2`. Add `if (_tcHits.size() < 3) return false;` before the triplet loops.
6. **`src/CalHelixFinderAlg.cc:1916`** — `Helix._sxy.addPoint(fCaloX, fCaloY, 1./100.);` is called unconditionally. When `setCaloCluster` took the "no cluster" path, `fCaloX/Y = -9999` still enters the weighted circle fit with weight 1/100, biasing the fit. Line 834 (`addCaloClusterToFitPhiZ`) has the same pattern with `fCaloZ`. The correct guard exists elsewhere (line 1175, `fCaloTime > 0`).
[...]

PR #1832, opened 2026-05-14
Add protection in phi peak finding

michaelmackenzie:
Address the bug reported in issue https://github.com/Mu2e/Offline/issues/1804

### DIFF
--- a/CalPatRec/src/PhiClusterFinder_module.cc
+++ b/CalPatRec/src/PhiClusterFinder_module.cc
@@ -400,27 +400,32 @@ namespace mu2e {
     // Check to the left of the peak bin
     int bincheck = max_bin;
     while(_hist1.GetBinContent(bincheck) >=_threshold) {
-      //if(_debug>3) std::cout<<"bincontent = "<<_hist1.GetBinContent(bincheck)<<" bincheck = "<<bincheck<<"   "<<_hist1.GetBinCenter(bincheck)<<std::endl;
-      if (bincheck > _threshold) bincheck--;
+      if (bincheck > 1) bincheck--;
       else bincheck = nbx;
       nsteps++;
       if (nsteps >= nbx) {
         cluphimin = 0;
         cluphimax = 2*M_PI;
-        if (_debug>2) printf("Phi min = %10.3f max : %10.3f\n",cluphimin,cluphimax);
+        if (_debug>2) printf("Hit maximum steps --> Phi min = %10.3f max : %10.3f\n",cluphimin,cluphimax);
         return;
       }
     }
     // Check to the right of the highest bin
+    nsteps = 0;
     int bincheckr=max_bin;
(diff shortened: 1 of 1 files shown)


oksuzian:
## Review: PR #1832 — Add protection in phi peak finding

This is a small, well-scoped bug fix in `PhiClusterFinder::clusterminmax()`. Touches one file (`CalPatRec/src/PhiClusterFinder_module.cc`, +10/-5). No framework / Proditions / Geant4 / fhicl surface changes.

### 🔴 Critical Issues
None. The original code was buggy in two ways, both now fixed:

- **Left-loop wrap condition fixed** (`PhiClusterFinder_module.cc:403`): the previous `if (bincheck > _threshold) bincheck--;` compared a **bin index** against a **hit-count threshold** — clearly wrong (mixed semantics, and would cause incorrect wrap when `_threshold > 1`). New `if (bincheck > 1) bincheck--;` correctly wraps at the first ROOT bin (ROOT histogram bins are `1..nbx`, with `0` = underflow). ✅
- **Right-loop infinite-loop protection added** (`PhiClusterFinder_module.cc:414-425`): the right-side scan previously had no `nsteps` guard, so if every bin in `_hist1` was ≥ `_threshold` (a hot/uniform event) the `while` would never terminate. The new `nsteps` reset + early return mirrors the left side. ✅

### 🟡 Suggestions

- **Symmetry / DRY** (`PhiClusterFinder_module.cc:402-426`): the two while-loops are mirror images. Consider factoring into a small lambda or helper `int scanDirection(int start, int step, int nbx)` to avoid drift between the two halves in future fixes. Optional, not blocking.
- **Termination invariant comment**: a one-line comment near line 402 explaining that bins wrap (`1 → nbx`, `nbx → 1`) and that the histogram has at least one bin at `< _threshold` would help future readers understand why the loop terminates in the common case.
- **`max_bin` underflow/overflow guard** (line 399): `TH1::GetMaximumBin()` can in principle return the under/overflow bin if those were filled; here `Fill(phi, …)` with `phi ∈ [0, 2π]` should keep it in-range given `_phimin/_phimax`, but if `_phimin/_phimax` ever drift this would silently misbehave. Consider `assert(max_bin >= 1 && max_bin <= nbx)` in debug builds. Minor.

### Framework / contracts checklist
- `produces<TimeClusterCollection>()` ↔ `event.put(std::move(tccol1))` — matches.
- `consumes<…>()` on both input tags — matches `getValidHandle` usage.
- No new conditions/geometry access; no Geant4 surface; no new ROOT file / TFileService misuse; no fhicl changes.

**Verdict:** LGTM with the optional refactor suggestion. The fix correctly addresses the infinite-loop / wrong-comparison bug from #1804.

oksuzian:
@FNALbuild run build test

NamithaChitrazee (approved):

PR #1833, opened 2026-05-16
KKMaterial migration

brownd1978:
This PR moves KKMaterial from Mu2eKinKal to GeometryService. Previously, each client created its own instance of KKMaterial, which required duplicate configuration. Now there's one instance, with a single configuration in standard_services. This PR makes no functional change.
There are knockon changes required in Production, EventNtuple and mu2e-trig-config due to configuration and link dependency changes.
This PR is a precursor to creating material objects for KinKal extrapolation.


michaelmackenzie:
@FNALbuild run build test with https://github.com/Mu2e/mu2e-trig-config/pull/162

brownd1978:
Thanks Michael!

brownd1978:
@FNALbuild run build test with Mu2e/mu2e-trig-config#162, Mu2e/Production#541

fcl/standardServices.fcl
```diff
@@ -42,6 +42,20 @@ Services : {
         inputFile           : "Offline/Mu2eG4/geom/geom_common.txt"
         bFieldFile          : "Offline/Mu2eG4/geom/bfgeom_v01.txt"
         simulatedDetector   : { tool_type: "Mu2e" }
+        KinKalMaterial : {
```
rlcee: this seems like a good place to use @table from a KK prolog since the content is all static and relevant only to KK, it should be located in a KK directory

KinKalGeom/inc/KKMaterial.hh
```diff
@@ -8,14 +8,19 @@
...
+#include "Offline/ConfigTools/inc/ConfigFileLookupPolicy.hh"
+#include "Offline/Mu2eInterfaces/inc/Detector.hh"
+#include "Offline/Mu2eInterfaces/inc/ProditionsEntity.hh"
 
 #include <memory>
 #include <string>
 
 namespace mu2e {
-  class KKMaterial {
+  class KKMaterial : public MatEnv::FileFinderInterface, public Detector, public ProditionsEntity {
```
rlcee: I don't think this needs to be a proditionsentity unless it will have run dependence within one geometry file config, like alignment.

fcl/standardServices.fcl
brownd1978: I debated this mentally as well. These parameters are only used in KinKal but they describe the general detector.
Currently we describe the entire detector geometry for simulation and reconstruction with Mu2eG4/geom. This breaks the design intent that GeometryService be agnostic, since we can't run reconstruction without Mu2eG4 (and its Geant4 dependency) in the release/path. A truely agnostic solution is to put base geometry and material definitions in a neutral package. That would be a natural place then to put this KinKal material config.

For now I'll follow this suggestion, and create an issue for the bigger issue.

KinKalGeom/inc/KKMaterial.hh
brownd1978: I was following an existing pattern (Tracker) without thinking. Since we've firmly committed to not allowing multiple base geometries in 1 job I'll  remove ProditionsEntity here and in KKGeom.

brownd1978:
@FNALbuild run build test with Mu2e/mu2e-trig-config#162, Mu2e/Production#541

bonventre (approved):

brownd1978:
@rlcee I believe I addressed your issues, can you approve or provide further feedback?

rlcee (approved):
Thanks

brownd1978:
@FNALbuild run build test with Mu2e/mu2e-trig-config#162, Mu2e/Production#541

brownd1978:
@FNALbuild run build test with Mu2e/mu2e-trig-config#162, Mu2e/Production#541

oksuzian:
# PR #1833 — KKMaterial migration review

**State**: open · mergeable (clean) · 2 approvals (rlcee, bonventre) · 35 files, **+1,908 / −1,909**, 14 commits · label: `build finished`.

**Intent**: Move `KKMaterial` from `Mu2eKinKal` into `KinKalGeom` and instantiate it once via `GeometryService` instead of letting every fit module construct its own. Configuration moves from per-module `MaterialSettings` to a single `KinKalMaterial` block in `standardServices`. Author claims no functional change.

## ✅ Core change — looks correct

- `KKMaterial` now lives in `KinKalGeom/`, derives from `Detector` (after dropping `ProditionsEntity` per review feedback) and implements `MatEnv::FileFinderInterface` directly (replacing the deleted `KKFileFinder`).
- `GeometryService` builds and registers `KKMaterial` once in `preBeginRun` (`GeometryService.cc`), with config plumbed through `Config::matSettings` (`KinKalMaterial`).
- Fit modules (`LoopHelixFit`, `KinematicLineFit`, `CentralHelixFit`, `RegrowLoopHelix`, `RegrowKinematicLine`) drop their `KKMaterial kkmat_` member and `MaterialSettings` table, fetching via `GeomHandle<KKMaterial>` per event. `KKExtrap` constructor likewise drops the `KKMaterial&` arg.
- New `KinKalGeom/fcl/prolog.fcl` provides `KinKalGeom.KKMaterial`; wired into `Services.Core.GeometryService.KinKalMaterial` via `@local`.

## 🔴 Critical issues

### 1. `KinKalGeom/inc/ShellMaterial.hh` is a stub — should not be committed
```cpp name=KinKalGeom/inc/ShellMaterial.hh url=https://github.com/Mu2e/Offline/blob/surfgeom/KinKalGeom/inc/ShellMaterial.hh
namespace mu2e {
  class ShellMaterial {
    using SurfacePtr = std::shared_ptr<KinKal::Surface>;
    public:
    private:
    SurfacePtr surf_; // surface for this materia
  };
}
```
Empty `public:`, no methods, unused private member, no include guard end note, missing `#endif` consistency, typo "materia". It is unused anywhere in the diff. Either flesh it out or remove it from this PR.

### 2. Removed `eloss` Config field, but the FHICL key is still set
The header dropped:
```cpp
fhicl::Atom<int> eloss { Name("ELossMode"), ... }
```
…but the new `KinKalGeom/fcl/prolog.fcl` (and the old `Mu2eKinKal.MAT`) still sets `IonizationEnergyLossMode : 1`. With `fhiclcpp` strict validation this is normally an error ("unused key"). Even if it isn't, the previous `dmconf.eloss_` setting is now gone — energy-loss model defaults to whatever `MatDBInfo` initializes. **This contradicts the "no functional change" claim.** Please confirm `dmconf.eloss_` is still being set, or remove `IonizationEnergyLossMode` from the prolog.

### 3. Large unrelated reformatting bundled into this PR
Whitespace-only / re-indentation changes account for the bulk of the diff and have nothing to do with KKMaterial:
- `Mu2eG4/src/constructTargetPS.cc` — **1682 changes (841/841)**, pure reformat.
- `ProductionTargetGeom/inc/ProductionTarget.hh` — **768 changes (384/384)**, pure indent change.
- `GeometryService/src/ProductionTargetMaker.cc` — **510 changes (255/255)**, pure reformat.
- `Mu2eG4/geom/ProductionTarget_Stickman_v1_0.txt` — **68 changes**, pure indent.
- `GeometryService/src/GeometryService.cc` — the new code adds 2 lines, but the body of `addDetector`, `addDetectorAliasToBaseClass`, and `preBeginRun` got re-indented one extra level, producing inconsistent style (function header at one indent, body deeper). Diff: 184/184.

These should be split into a separate "whitespace cleanup" PR. They make this one nearly impossible to review, will pollute `git blame`, and conflict with anything else in flight on those files. They also don't belong with a "no functional change" migration label.

### 4. Hard-coded magic numbers preserved/extended in `KinKalGeomMaker::makeDS()`
```cpp name=GeometryService/src/KinKalGeomMaker.cc url=https://github.com/Mu2e/Offline/blob/surfgeom/GeometryService/src/KinKalGeomMaker.cc#L82-L114
auto inner= std::make_shared<Cylinder>(VEC3(0,0,1),VEC3(0,0,-1482),ds->rIn1(),ds->halfLength());
auto outer= std::make_shared<Cylinder>(VEC3(0,0,1),VEC3(0,0,-1482),ds->rOut2(),ds->halfLength());
...
auto ipa= std::make_shared<Cylinder>(VEC3(0,0,1),VEC3(0,0,-2770),300.0,500.0);
```
Z origin `-1482` and the entire IPA (`-2770`, r=300, l=500) are still hard-coded despite this PR being the migration touchpoint. Also ~25 lines of commented-out `std::cout` debug blocks were left in. Either delete them or guard behind `_diag`.

### 5. `KKStrawMaterial::pathLengths`/`findXings` API change is a silent semantic change
Signature changed from taking `StrawXingUpdater const& caconfig` (which carried `nsig_`, `diag_`, and *other* flags) to taking just `double nsig` and `int diag`. Any other field of `StrawXingUpdater` that was previously read inside these methods is now dropped. The diff in `KKStrawMaterial.cc` shows only `caconfig.nsig_` and `caconfig.diag_` were used, so this looks safe — but `RecoDataProducts/inc/TrkStraw.hh` and `Mu2eKinKal/inc/KKStrawXing.hh` were updated to pass only `nsig_` (no `diag_`). Worth a quick confirmation that no path through `StrawXingUpdater` previously influenced material xing shape; otherwise reconstruction output may shift slightly.

### 6. Removed defaults on `KKStrawMaterial` ctor parameters
`const std::string& wallmat="straw-wall"` etc. defaults dropped. Anyone constructing `KKStrawMaterial` directly outside this repo (e.g., the mentioned downstream `EventNtuple` / `Production` / `mu2e-trig-config`) will break. Mention this clearly in the companion PR descriptions.

## ⚠️ Minor

- `KinKalGeom/inc/KKMaterial.hh` — `mutable` on `policy_` plus `mutable` on `matdbinfo_`/`smat_`: file-finder policy doesn't really need `mutable`; only used in `const` calls.
- `KinKalGeom/src/SConscript` — adds `KinKal_MatEnv` and `cetlib` but `CMakeLists.txt` only adds `KinKal::MatEnv`. Minor inconsistency, but ensure both builds were exercised (label says yes).
- `KKMaterial::matSettings` is plumbed via `pars().matSettings()` (struct view) while `simulatedDetector` is fetched as `pars.get_PSet().get<...>("simulatedDetector")` — fine, just inconsistent style.
- The new `KKMaterial::Config` lacks the `verbose`, `dahlLynchScatteringFraction`, `intersectionTolerance`, `strawElementOffset`, `maximumIntersectionRadiusFraction` parameters that were in the deleted `Mu2eMaterial` block in `TrackerConditions/fcl/prolog.fcl`. Confirm these were intentionally retired (none referenced in the new code) and that no downstream FHICL still tries to set them.

brownd1978:
responding to AI review critical issues:
1  true but irrelevant: this is an intermediate commit leading to a larger goal where this code will be used
2 erroneous: there were 2 similairly named fhicl keys, one redundant/irrelevant, I removed it but AI confused the 2
3 The reformatting was an unintended knockon and cannot be reverted as the original files don't pass pre-commit check (whitespace errors).
4 again this is an intermediate commit, complete remove of 'magic numbers' requires more development
5 API change was required to avoid dependency loops, correct behavior has been verified
6 intentional change (improvement) to remove default constructor argument

Some of the minor comments are legit and will be addressed in a subsequent PR, none affect behavior/performance.

PR #1836, opened 2026-05-20
Updated nominal map for laser ROC

giro94:
This was never configured as cabled


giro94:
@FNALbuild run build test

rlcee (approved):
I can't verify the content, but the format and counts appear valid

PR #1841, opened 2026-05-23
Fix CMake build issues

brownd1978:
This resolves the dependency loop problem found in cmake.

### DIFF
--- a/DAQ/CMakeLists.txt
+++ b/DAQ/CMakeLists.txt
@@ -146,7 +146,6 @@ cet_build_plugin(STMPrintFragments art::module
   Offline::ProditionsService
   Offline::DataProducts
   Offline::RecoDataProducts
-  BTrk_difAlgebra
 )
 
 cet_build_plugin(StrawDigisFromArtdaqFragments art::module
--- a/GeometryService/src/GeometryService.cc
+++ b/GeometryService/src/GeometryService.cc
@@ -356,10 +356,13 @@ namespace mu2e {
       KinKalGeomMaker kkgm;
       addDetector( std::move(kkgm.makeKKG()) );
       // directly build KKMaterial; it's constructor does everything
-      DetMap::iterator itrk = _detectors.find("Tracker");
-      Tracker const& tracker = static_cast<Tracker const&>(*itrk->second);
-      addDetector( std::make_unique<KKMaterial>(_kkMat,tracker));
-      return;
+      auto trkptr =  getElement<Tracker>();
+      if(trkptr){
+        Tracker const& tracker = *trkptr;
+        addDetector( std::make_unique<KKMaterial>(_kkMat,tracker));
+        return;
+      } else
+        throw cet::exception("GEOM") << "No Tracker found! " << "\n";
 
     } // preBeginRun()
 
--- a/GeometryService/src/GeometryService.cc
+++ b/GeometryService/src/GeometryService.cc
@@ -356,8 +356,10 @@ namespace mu2e {
       KinKalGeomMaker kkgm;
       addDetector( std::move(kkgm.makeKKG()) );
       // directly build KKMaterial; it's constructor does everything
-      addDetector( std::make_unique<KKMaterial>(_kkMat));
-      return ;
+      DetMap::iterator itrk = _detectors.find("Tracker");
+      Tracker const& tracker = static_cast<Tracker const&>(*itrk->second);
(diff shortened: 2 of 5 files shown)


oksuzian:
## PR #1841 Review — "Fix CMake build issues"

**Status:** open, mergeable=true but `mergeable_state=unstable`. CI **failing** with widespread `Return Code 139` (SIGSEGV) on nearly every sim/reco test (ceSimReco, g4test_03MT, transportOnly, POT, cosmicSimReco, ceMix, rootOverlaps, etc.). Build (prof) and check_cmake pass; the crashes are runtime.

### What the PR does
- Drops circular CMake deps (`KinKalGeom` ⇄ `GeomPrimitives`, and `DAQ` → `BTrk_difAlgebra`).
- Adds `install_fhicl` for `KinKalGeom/fcl`.
- Refactors `KKMaterial` to take a `Tracker const&` at construction time instead of lazily fetching via `GeomHandle<Tracker>` in `strawMaterial()`. Removes `mutable` state.
- Updates `GeometryService::preBeginRun()` to look up the Tracker from `_detectors` and pass it into `KKMaterial`'s constructor.

### 🚨 Critical issue — almost certainly the cause of the SIGSEGVs

In `GeometryService/src/GeometryService.cc` (lines 359–361):

```cpp name=GeometryService/src/GeometryService.cc url=https://github.com/Mu2e/Offline/blob/3a663fb5e333d61dc7b7548a4a113b6e1ef0c0c0/GeometryService/src/GeometryService.cc#L359-L361
DetMap::iterator itrk = _detectors.find("Tracker");
Tracker const& tracker = static_cast<Tracker const&>(*itrk->second);
addDetector( std::make_unique<KKMaterial>(_kkMat,tracker));
```

The `_detectors` map is keyed by **`typeid(DET).name()`** (the C++ ABI mangled name, e.g. `N3mu2e7TrackerE`) — see `addDetector()` and `getElement()` in `GeometryService.hh`/`.cc`. It is **not** keyed by the human string `"Tracker"`.

Consequence:
1. `itrk == _detectors.end()` always.
2. `*itrk->second` dereferences the end iterator → **undefined behavior / segfault**, matching the Return Code 139 across the test suite.
3. Even if found, `static_cast<Tracker const&>(*itrk->second)` casts from `Detector` to `Tracker` without RTTI safety — should be `dynamic_cast`, as in `getElement<DET>()`.

**Fix:** use the existing accessor (which already does the typeid lookup, type check, and null check):

```cpp name=GeometryService/src/GeometryService.cc
Tracker const& tracker = *getElement<Tracker>();
addDetector( std::make_unique<KKMaterial>(_kkMat, tracker));
```

…or at minimum `_detectors.find(typeid(Tracker).name())` with an end-check and `dynamic_cast`.

### ⚠️ Secondary issue — no guard on `hasTracker`

The Tracker is only added when `_config->getBool("hasTracker", false)` is true (line 279). The new code unconditionally looks up a Tracker and constructs `KKMaterial` from it. Non-tracker geometries (e.g. some `g4study`-style configs, or future detector-less jobs) will crash.

Either:
- Guard the `KKMaterial` construction with `if (_config->getBool("hasTracker", false))`, or
- Move the `KKMaterial` add inside the existing `hasTracker` block.

### Minor

- `KinKalGeom/CMakeLists.txt`: dropping `Offline::GeomPrimitives` is fine only if no header in `KinKalGeom` still includes `GeomPrimitives` types transitively as part of the public interface — worth double-checking with a clean build of downstream packages that consume `KinKalGeom` headers.
- clang-tidy now reports **3 errors / 4 warnings** (was clean previously?) — worth a glance at the log.
- `KKMaterial.cc`: trailing `}` on `findFile` is now on same line as `return policy( basename ); }` — style nit only.

### Merge readiness

**Not ready.** The critical lookup bug almost certainly explains the 139s and must be fixed before merge. Once fixed, re-run CI and verify clang-tidy errors.

### Want me to
1. Open a follow-up PR with the suggested `getElement<Tracker>()` fix and `hasTracker` guard.
2. Pull the `ceSimReco.log` or `trigger.log` from buildmaster to confirm the segfault stack trace.
3. Check whether any other code path constructed `KKMaterial` (e.g., a module) and now needs updating for the new constructor signature.
4. Audit other `_detectors.find(...)` style lookups to make sure none use raw strings.

brownd1978:
@FNALbuild run build test

PR #1842, opened 2026-05-27
Migrate KinematicLine_Fit and CentralHelix_Fit to use KKExtrap. 

brownd1978:
This PR migrates KinematicLine_Fit and CentralHelix_Fit to use KKExtrap, bringing them in line with LoopHelix_Fit. To get extrapolation to work on fits with no BField (KinematicLine) required extensions in core KinKal, so this PR requires a new KinKal tag and envset.

This PR also migrates the KinKalGeom description of the CRV to be driven from GeometryService, as part of the downstream KinKalGeom service. Each CRV 'shield' (sector) is used to define a rectangular surface at the midpoint, where KinKal track intersections are recorded. Currently the CRV module material is defined to be pure aluminum, it should be updated to be a mix of aluminum and scintillator, that's left for the next PR. Also left is modeling the DS and shielding blocks, needed to get the correct scattering covariance at the extrapolation.

PR #1843, opened 2026-05-28
p101 for kinkal 3.6.0

rlcee:
Update envset pointer to update kinkal dependence to version 3.6.0
We expect this build to fail CI unless built with #1842, kinkal updates


brownd1978:
thanks Ray. I suggest to close this PR and merge 1842, which now includes it

rlcee:
This change is now included in #1842

PR #1844, opened 2026-06-01
Overlap fixes

michaelmackenzie:
This resolves geometry overlap issues seen with Run 1B geom v40. There were two sources: 1) restoring Run 1A elements introduced overlaps due to code changes made to accommodate Run 1B and 2) the degrader mother volume wasn't quite correct and the thicker degrader also needed adjustments in its support plate. Both sources should now be resolved, which I tested using a surface check with the DS-off and DS-on v40 configuration files.

### DIFF
--- a/Mu2eG4/geom/geom_run1_b_v40.txt
+++ b/Mu2eG4/geom/geom_run1_b_v40.txt
@@ -8,6 +8,8 @@ string degrader.filter.materialName = "StoppingTarget_Al"; // stopping target ma
 double degrader.filter.halfLength   = 8.75; // 1.75 cm plate
 double degrader.filter.rIn          = 0.0;
 double degrader.filter.rOut         = 150.0; // 150 mm radius disk
+double degrader.supportArm.offsetz  = 245.0;
+double degrader.supportArm.dz       = 186.0;
 
 // Use the TSdA as an aluminum collimator, helping Run 1B
 bool hasTSdA             = true;
--- a/Mu2eG4/src/constructProtonAbsorber.cc
+++ b/Mu2eG4/src/constructProtonAbsorber.cc
@@ -1373,10 +1373,9 @@ namespace mu2e {
                                                       *CLHEP::degree);
         double yLocMoBox = pivotPos.at(1) + dptoc*sin(pabs->degraderRotation()
                                                       *CLHEP::degree);
-        CLHEP::Hep3Vector locationInMu2e (xLocMoBox, yLocMoBox,
-                                          pabs->degraderZ0()
-                                          + 2.0*filterDims.at(2)
-                                          + frameDims.at(2) );
+        const float dgr_z0 = pabs->degraderZ0(); // front face of the degrader
+        const double mother_half_width = filterDims.at(2) + frameDims.at(2) + 0.1; // contain both the frame and the filter
+        CLHEP::Hep3Vector locationInMu2e (xLocMoBox, yLocMoBox, dgr_z0 + mother_half_width);
 
         // Make mother volume for degrader
         std::string motherName("Degrader");
@@ -1387,8 +1386,7 @@ namespace mu2e {
         G4Box* motherBox = new G4Box ( "degraderOutline",
                                        lengthPDMoBox/2.0,
                                        frameDims.at(1),
-                                       2.0*filterDims.at(2) + frameDims.at(2)
-                                       + 1 );
+                                       mother_half_width);
         degraderMother.solid = motherBox;
 
         // Now put degraderMother in DS2Vacuum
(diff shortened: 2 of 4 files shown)


oksuzian:
I'll review this PR. Let me load the appropriate abilities first.## PR #1844 Review — "Overlap fixes"

```list type="pr" id=a9ae9536-c3c1-49c5-88c0-823148d16e4e
```

**Author:** michaelmackenzie · **State:** open · **Reviews:** none yet · **Risk:** ⚠️ Medium — two changes silently alter the *default* (non‑Run1B) geometry.

### What the PR does
Resolves Geant4 volume overlaps for the Run 1B configuration while keeping Run 1A behavior. Four files:

| File | Change |
|---|---|
| `Mu2eG4/src/constructProtonAbsorber.cc` | Rework pion‑degrader mother volume sizing & re‑order filter/frame/rod placement |
| `Mu2eG4/src/constructStoppingTarget.cc` | Size the ST mother volume to enclose the largest foil/support radius + add verbosity logging |
| `Mu2eG4/src/constructVirtualDetectors.cc` | Swap `isDumbbell` parent selection to `tracker.inDS2Vacuum` |
| `Mu2eG4/geom/geom_run1_b_v40.txt` | Add `degrader.supportArm.offsetz/dz` overrides |

---

### 🔴 Issue 1 — Default VD parent volume changes for *all* geometries (`constructVirtualDetectors.cc`)

The condition was flipped in 6 places:

```diff
- VolumeInfo const & parent = ( _config.getBool("isDumbbell",false) ) ?
+ VolumeInfo const & parent = ( !_config.getBool("tracker.inDS2Vacuum",false) ) ?
    _helper->locateVolInfo(theDS3) :
    _helper->locateVolInfo("DS2Vacuum");
```

- **Old default** (`isDumbbell` defaults `false`) → parent = **DS2Vacuum**
- **New default** (`tracker.inDS2Vacuum` defaults `false`) → `!false` = `true` → parent = **DS3Vacuum**

This means the nominal geometry's virtual detectors now land in **DS3Vacuum instead of DS2Vacuum**, a behavior change for every standard configuration — not just Run 1B. If the intent was "only Run1B moves these to DS2Vacuum," the logic likely needs to preserve the old default, e.g. keep selecting DS2Vacuum unless explicitly told otherwise. Please confirm this is intended and validate that standard (Run2/nominal) geometries are unaffected.

### 🔴 Issue 2 — `EMC_0_Front` virtual detector disabled in nominal geometry (`constructVirtualDetectors.cc` ~L2017)

```diff
- if ( !_config.getBool("isDumbbell",false) ){
+ if (_config.getBool("tracker.inDS2Vacuum",false) && !_config.getBool("isDumbbell",false) ){ // for now, only include in altered setup
```

Previously this VD was built whenever **not** dumbbell (i.e. the nominal case). Now it is built **only** when `tracker.inDS2Vacuum` is true. So in standard geometry the `EMC_0_Front` VD is **no longer created**. The inline comment ("for now, only include in altered setup") acknowledges this, but it is a functional regression for nominal running that could break downstream code/analyses expecting that VD. This should be gated so the default geometry keeps its previous behavior.

### 🟡 Issue 3 — `tracker.inDS2Vacuum` has no default in committed geom files
The flag is only defined in `geom_run1_b_v01.txt` (and now drives `constructVirtualDetectors.cc`). Because the C++ uses `getBool("tracker.inDS2Vacuum", false)`, every other geometry implicitly gets `false`, which triggers Issues 1 & 2. Worth double‑checking the standard geometry `.txt` files and confirming whether a default should be set centrally.

### 🟡 Issue 4 — `const float dgr_z0` precision inconsistency (`constructProtonAbsorber.cc` L1376)
```cpp
const float dgr_z0 = pabs->degraderZ0(); // front face of the degrader
```
`degraderZ0()` returns `double` and every other coordinate here is `double`. Narrowing the front‑face Z to `float` introduces unnecessary rounding in a geometry that you're tuning to ~0.1 mm overlap buffers. Recommend `const double dgr_z0`.

### 🟢 Minor observations
- **`constructStoppingTarget.cc`:** `maxMotherRadius` now correctly bounds the mother tube to enclose support structures (`length + maxFoilRadius + 0.1`). Good. But note the local `cylinderRadius` (L91) is now assigned `maxMotherRadius` and the overlap check at L112 (`rAtZClosest < cylinderRadius + 0.001`) now uses the *enlarged* radius — verify this doesn't spuriously trigger the conic‑section fallback for configs that previously passed.
- **Degrader re-ordering** (filter → frame → rod) with `mother_half_width = filterDims.at(2) + frameDims.at(2) + 0.1` looks self‑consistent; the `+0.1` buffer matches the new geom overrides (`offsetz=245`, `dz=186`).
- The added `verbosity > 1` logging in `constructStoppingTarget.cc` is helpful and low‑risk.

---

### Merge readiness
Not ready as-is. Issues 1 and 2 change the **default geometry** behavior, which I suspect is unintentional given the PR title is scoped to "overlap fixes" for Run 1B. I'd block on clarifying/guarding those two before merge. No CI checks or reviews are attached yet.

Want me to:
1. Trace where `isDumbbell` / `tracker.inDS2Vacuum` are consumed elsewhere to confirm the default-geometry impact?
2. Check which geom `.txt` files set `tracker.inDS2Vacuum` and whether nominal configs are affected?
3. Draft suggested diffs that preserve the old default while enabling the Run 1B behavior?
4. Pull the diff of `geom_run1_b_v40.txt` vs the base degrader geometry to verify the offset math?

michaelmackenzie:
1. This changes the defaults for all geometries because earlier edits broke the VD volume placement in Run 1A running.
2. I disabled it for now since I don't exactly know where it should go in the nominal geometry, it's probably best for the original author to update it to work in the nominal configuration mode.
3. The retrieval defaults to false so past geometry files don't need to be changed.
4. I changed float --> double.

It looks like the CI failures are not due to changes here, but I'm not positive why we're seeing them...

PR #1849, opened 2026-06-02
Run1B->Main

oksuzian:


Mu2eG4/geom/geom_common.txt
```diff
@@ -3,7 +3,7 @@
 // The file that is included will be time dependent.
 //
 
-#include "Offline/Mu2eG4/geom/geom_run1_a_stickman.txt"
+#include "Offline/Mu2eG4/geom/geom_run1_b_v01.txt"
```
michaelmackenzie, line 6: We shouldn't change the default geometry

Mu2eG4/geom/geom_run1_b_v01.txt
```diff
@@ -37,10 +37,16 @@ vector<double> calorimeter.diskZMotherShift   = {0.0,2000};
 // Set stopping target position AFTER calorimeter configuration
 double stoppingTarget.z0InMu2e = 4185;
 vector<double> stoppingTarget.radii  =  {600.00};
+// Set position of EMC Source Virtual Detectors
```
michaelmackenzie, line 40: Should these be in all Run 1B geometry files?

GeometryService/src/VirtualDetectorMaker.cc
```diff
@@ -185,16 +185,34 @@ namespace mu2e {
       vd->addVirtualDetector( VirtualDetectorId::ST_Out,
                                ds2centerInMu2e, 0, targetOffset+shift);
 
+      // VD EMC_Source is placed inside DS2 in front of the tracker
+      const double zVDcenterInMu2e = c.getDouble("zEMCSourceInMu2e",5300.);
```
michaelmackenzie, line 189: I don't know if we should have default z values for these

Mu2eG4/src/constructVirtualDetectors.cc
```diff
@@ -565,8 +567,9 @@ namespace mu2e {
 
           std::string theDS3("DS3Vacuum");
           if ( _config.getBool("inGaragePosition",false) ) theDS3 = "garageFakeDS3Vacuum";
-
-          VolumeInfo const & parent = _helper->locateVolInfo(theDS3);
+          VolumeInfo const & parent = ( _config.getBool("isDumbbell",false) ) ?
```
michaelmackenzie, line 570: It seems like this is a case I missed when fixing these parent volume changes? This should probably be doing the tracker in DS2 check

michaelmackenzie (commented):
By eye it's hard to really review this large of a diff and be confident that it doesn't change the nominal running. I commented on a couple of items I notice though. As virtual detectors have been added, this will likely change the simulation results such that comparisons would probably need high statistics and distribution shape comparisons to ensure they're correct. We could try doing a gdml dump on main and here, and looking to see if any volumes change between the two versions?

PR #1851, opened 2026-06-05
Added Calo crystal sim conditions

bechenard:
Implementation of conditions for calo simulation parameters (Ray should review). ALso renamed conditions for CaloDAQMap into CalDAQMap to have consistent names. Let pePerMev at 30, will move to 20 when we include the energy calibration. The fcl entry for pePerMev is not yet removed - will do when I revamp the noise, but I need to add support for the disk/crystal position conditions in priority.


rlcee (changes requested):
I've looked this over carefully, and very nice, you've followed all the db recommendations.  One request - you will need an entry in DbTables/inc/DbTableFactory.hh

bechenard:
I have largely copied what other did, so I don't have much merit! I'm working on the alignment right now, and I want to finish before coming back to this. This will take a few days

bechenard:
@FNALbuild run build test

rlcee (approved):
Looks good, thanks

oksuzian:
### 🔴 Critical / correctness

**1. Silent out-of-bounds write on invalid crystal id** — [`CalSimCrystalMaker.cc#L40-L47`](https://github.com/Mu2e/Offline/blob/643f7dac2f63b357d3e5ac9dcdbae6f349717e7b/CaloConditions/src/CalSimCrystalMaker.cc#L40-L47)

The `fromFcl()` range check lets `_invalid` through, then uses `sid` as an array index:

```c++
if (sid != CaloConst::_invalid && sid >= CaloConst::_nCrystal) { throw ... }
crystalIds[sid]  = CrystalId(sid);   // if sid == _invalid, this indexes way out of bounds
LRUs[sid]        = lru;
pePerMeVs[sid]   = std::vector<float>{npe0,npe1};
```

`_invalid` is a large sentinel (typically `0xFFFF`), so `crystalIds[sid]` on a `std::array<…, _nCrystal>` is undefined behavior. The `fromDb()` path guards with `crid.id()` bounds, but `fromFcl()` writes directly with the raw `sid`. Since `sid` is a plain array index here (not a sentinel-carrying field), the `!= _invalid` exclusion is wrong — any `sid >= _nCrystal` should throw.

**2. The `pePerMeVs.back()` validation is misplaced and weak** — [`CalSimCrystalMaker.cc#L50-L53`](https://github.com/Mu2e/Offline/blob/643f7dac2f63b357d3e5ac9dcdbae6f349717e7b/CaloConditions/src/CalSimCrystalMaker.cc#L50-L53)

```c++
if (pePerMeVs.back().size() != CaloConst::_nSiPMPerCrystal ) { throw ... }
```

- It only checks the **last** element. Every row is hard-coded to `{npe0, npe1}` (size 2), so this can never catch a malformed row — it validates nothing meaningful.
- If the file is short (fewer than `_nCrystal` rows read), `back()` is a default-constructed empty vector, so this throws `"too many nPE values"` — a misleading message for a *too few rows* condition (the real count check is below it at L55).

### 🟠 Robustness

**3. No handle on partial/short files beyond the count** — the `while(!eof)` loop with `if (eof) break;` is the classic fragile pattern; a malformed line silently stops parsing and you rely solely on the final `nRead != _nCrystal` check. Fine for the committed nominal file, but brittle for hand-edited DB dumps.

**4. `Sim_crystal.txt` naming mismatch** — the fcl (`prolog.fcl`) and `CMakeLists.txt` reference `Sim_crystal.txt`, but `CMakeLists.txt` also has a leftover `configure_file(... data/Sim_crystal.dat ...)` line. Confirm `.dat` vs `.txt` — looks like a stray `.dat` entry.

### 🟡 Minor / consistency

**5. Header guard typo** — [`CalSimCrystalConfig.hh#L1-L4`](https://github.com/Mu2e/Offline/blob/643f7dac2f63b357d3e5ac9dcdbae6f349717e7b/CaloConfig/inc/CalSimCrystalConfig.hh): guard is `CaloConditions_CaloSimCrystalConfig_hh` and the comment says `Initialize CaloMCCalib from fcl` — both are copy-paste leftovers referring to old names.

**6. `DbTable` name inconsistency** — [`CalCrystals.hh`](https://github.com/Mu2e/Offline/blob/643f7dac2f63b357d3e5ac9dcdbae6f349717e7b/DbTables/inc/CalCrystals.hh): the class is `CalCrystals` but `cxname = "CalSimCrystal"` and the factory maps `"CalSimCrystal" → new CalCrystals()`. Given the PR's stated goal of consistent naming, the class/table/cxname triad is mixed (`CalCrystals` vs `CalSimCrystal` vs table `cal.simcrystal`).

**7. `pePerMeV` still hard-coded to 30 in fcl** — [`CaloMC/fcl/prolog.fcl`](https://github.com/Mu2e/Offline/blob/643f7dac2f63b357d3e5ac9dcdbae6f349717e7b/CaloMC/fcl/prolog.fcl) sets `readoutPEPerMeV: 30`, which the author already flagged as intentional (moving to 20 with energy calibration later) — noting for completeness, not a blocker.

rlcee:
AI comment 6 is not minor, it will prevent the code from running

bechenard:
There were conflict that I resolved, but they will roll back some changes made by Paolo in the previous PR. Working on fixing these as well

rlcee:
There are still inconsistencies in the naming: CalSimCrystal vs CalSimCrystals

bechenard:
> There are still inconsistencies in the naming: CalSimCrystal vs CalSimCrystals

I need a break.... And now the PR defeats what Paolo has done for the LYSO response, so I also need to fix this (and this isn't trivial, unfortunately).

PR #1853, opened 2026-06-08
Adds ParameterHit info to KalSeed so it can be regrown

bonventre:
KKLine uses parameterhit to constrain momentum etc., plus we expect to add constraints for reflecting cosmics etc. This adds the info to the persistent object so we can regrow it

### DIFF
--- a/Mu2eKinKal/inc/KKFit.hh
+++ b/Mu2eKinKal/inc/KKFit.hh
@@ -102,6 +102,7 @@ namespace mu2e {
       bool makeStrawHits(Tracker const& tracker,StrawResponse const& strawresponse, BFieldMap const& kkbf,
           PTRAJ const& ptraj, ComboHitCollection const& chcol, StrawHitIndexCollection const& strawHitIdxs,
           KKSTRAWHITCOL& hits, KKSTRAWXINGCOL& exings) const;
+      void makeSeedParamHit(KTRAJ const& seedtraj, std::vector<double> const& paramconstraints, PARAMHITCOL& paramhits) const;
       // regrow KKTrack components from a KalSeed
       bool regrowComponents(KalSeed const& kseed, ComboHitCollection const& chcol, mu2e::IndexMap const& strawindexmap,
           Tracker const& tracker,Calorimeter const& calo, StrawResponse const& strawresponse,BFieldMap const& kkbf,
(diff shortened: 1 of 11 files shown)


Mu2eKinKal/src/KinematicLineFit_module.cc
```diff
@@ -288,12 +296,34 @@ namespace mu2e {
           }
           // set the seed range given the hit TPOCA values
           seedtraj.range() = kkfit_.range(strawhits,calohits, strawxings);
+
```
brownd1978, line 295: This code block is identical to the one in KinematicLineFit, and might be useful in LoopHelix as well. Can it be moved to KKFit or some other shared utility?

Mu2eKinKal/inc/KKFit.hh
```diff
@@ -368,7 +377,7 @@ namespace mu2e {
         << addcalohits.size() << " CaloHits and "
         << addstrawxings.size() << " Straw Xings" << std::endl;
     }
-    kktrk.extendTrack(exconfig,addstrawhits,addstrawxings,addcalohits);
+    kktrk.extendTrack(exconfig,addstrawhits,addstrawxings,addcalohits,addparamhits);
```
brownd1978: I don't understand why parameters hits are needed in the extendTrack interface. The original intent of having this call here was to allow additional schedule steps beyond simply reconstituting a regrown fit, perhaps to add new hits, or regrowing a seed fit to a drift fit, etc. I don't see how new parameter hits could be added in that context. In this call the parameter hits vector is hard-coded to empty, what is the use case for a call with a non-zero vector? I don't see such a call to it in this PR.

brownd1978 (changes requested):
Thanks for filling in this missing piece to regrowing. I'm probably missing something, but I don't understand why the extendTrack method needs parameter hits. I'd also like to see the code duplcation removed if possible.

Outside the scope of this PR; some codes will need updating to handle the reflection constraint, but I don't see anything here that breaks that. That constraint is based on ParameterConstraint but also has payload for the coupling between the tracks. I believe the persistent form can either inherit from or embed TrkParamHitSeed.

brownd1978 (approved):
Thanks for making the requested changes, all looks good now.

bonventre:
@FNALbuild run build test

PR #1857, opened 2026-06-11
Stm unpacking

Etho-b02:
Cleaned Up the if statement in line 187 in STMMovingWindowDeconvolution_module.cc. 
Fixed instance-name mismatch in modules. 
Cleared up getChannel  in plotSTMWaveformDigis_module.cc  
Also cleaned up last comment in inspectSTMFile.fcl. 
Added STMFragmentSummary to collect container and inner frag counts from unpacking module. 
Addressed fragment count issues from last PR.

### DIFF
--- a/Mu2eUtilities/src/STMUtils.cc
+++ b/Mu2eUtilities/src/STMUtils.cc
@@ -9,6 +9,7 @@ namespace mu2e {
   namespace STMUtils {
     // Function to get the STMChannel from the art::InputTag
     // (we will keep data from HPGe and LaBr in separate collections)
+    /*
     STMChannel getChannel(art::InputTag const& tag) {
       if (tag.instance() != "") {
         // If we use instance name, it will only contain the channel name
@@ -18,6 +19,22 @@ namespace mu2e {
         std::string label = tag.label();
         // Look at last four characeters of module label to decide channel
         return STMChannel(STMChannel::findByName(label.substr(label.length()-4,4)));
+      }
+      }
+    */
+    STMChannel getChannel(art::InputTag const& tag){
+      std::string name = tag.instance();
+      if (name.empty()){
+        name = tag.label();
+      }
+      if (name.find("HPGe") != std::string::npos){
+        return STMChannel::findByName("HPGe");
+      }
+      else if (name.find("LaBr") != std::string::npos){
+        return STMChannel::findByName("LaBr");
+      }
+      else {
+        return STMChannel::findByName("Unknown");
       }
     }
 
--- a/Mu2eUtilities/src/STMUtils.cc
+++ b/Mu2eUtilities/src/STMUtils.cc
@@ -3,34 +3,27 @@
 #include "Offline/Mu2eUtilities/inc/STMUtils.hh"
 
 #include "cetlib_except/exception.h"
+#include <algorithm>
 
 namespace mu2e {
 
   namespace STMUtils {
     // Function to get the STMChannel from the art::InputTag
     // (we will keep data from HPGe and LaBr in separate collections)
-    /*
-    STMChannel getChannel(art::InputTag const& tag) {
-      if (tag.instance() != "") {
-        // If we use instance name, it will only contain the channel name
-        return STMChannel(STMChannel::findByName(tag.instance()));
-      }
-      else {
-        std::string label = tag.label();
-        // Look at last four characeters of module label to decide channel
-        return STMChannel(STMChannel::findByName(label.substr(label.length()-4,4)));
-      }
-      }
-    */
     STMChannel getChannel(art::InputTag const& tag){
       std::string name = tag.instance();
       if (name.empty()){
         name = tag.label();
       }
-      if (name.find("HPGe") != std::string::npos){
+
+      std::transform (name.begin(), name.end(), name.begin(),
+                      [](unsigned char c) {
+                        return static_cast<char>( std::tolower(c));
+                      });
+      if (name.find("hpge") != std::string::npos){
         return STMChannel::findByName("HPGe");
       }
-      else if (name.find("LaBr") != std::string::npos){
+      else if (name.find("labr") != std::string::npos){
         return STMChannel::findByName("LaBr");
       }
       else {
--- a/Mu2eUtilities/src/STMUtils.cc
+++ b/Mu2eUtilities/src/STMUtils.cc
@@ -21,13 +21,13 @@ namespace mu2e {
                         return static_cast<char>( std::tolower(c));
                       });
       if (name.find("hpge") != std::string::npos){
-        return STMChannel::findByName("HPGe");
+        return STMChannel(STMChannel::HPGe);
       }
       else if (name.find("labr") != std::string::npos){
-        return STMChannel::findByName("LaBr");
+        return STMChannel(STMChannel::LaBr);
       }
       else {
-        return STMChannel::findByName("Unknown");
+        return STMChannel(STMChannel::unknown);
       }
     }
 
--- a/DAQ/src/STMDigisFromFragments_module.cc
+++ b/DAQ/src/STMDigisFromFragments_module.cc
@@ -11,6 +11,7 @@
 
 #include "Offline/ProditionsService/inc/ProditionsHandle.hh"
 #include "Offline/RecoDataProducts/inc/STMWaveformDigi.hh"
+#include "Offline/RecoDataProducts/inc/STMFragmentSummary.hh"
 #include "Offline/RecoDataProducts/inc/STMPHDigi.hh"
 #include "art/Framework/Principal/Handle.h"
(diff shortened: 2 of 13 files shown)


Mu2eUtilities/src/STMUtils.cc
```diff
@@ -19,6 +20,22 @@ namespace mu2e {
         // Look at last four characeters of module label to decide channel
         return STMChannel(STMChannel::findByName(label.substr(label.length()-4,4)));
       }
```
AndrewEdmonds11: You can just remove this commented out block. We can always go back in the history to find it again

Mu2eUtilities/src/STMUtils.cc
```diff
@@ -19,6 +20,22 @@ namespace mu2e {
...
+      }
+    */
+    STMChannel getChannel(art::InputTag const& tag){
+      std::string name = tag.instance();
+      if (name.empty()){
+        name = tag.label();
+      }
+      if (name.find("HPGe") != std::string::npos){
```
AndrewEdmonds11: Maybe we also want to make this case insensitive?

AndrewEdmonds11 (approved):
Looks good. Thanks, Bryan. A couple of small comments below

AndrewEdmonds11 (approved):

AndrewEdmonds11 (approved):

Etho-b02:
The PR is ready.

oksuzian:
I'll load the PR review ability and gather the PR details first.I'll review this PR. Let me gather the key data first.I have enough to give a concise review. Let me flag the most important issue I spotted in the diff — there's a real bug in the LaBr `art::Ptr` construction.

# PR #1857 — Stm unpacking — Review

**Assessment:** Functionally solid cleanup + new `STMFragmentSummary` product and ZS→Raw parent linking. Already approved 3× by @AndrewEdmonds11. CI label is `build pending`, `mergeable_state: unstable`. **One real bug** worth fixing before merge.

---

## 🐛 Bug — LaBr parent `Ptr` uses the wrong collection/index

In `STMDigisFromFragments_module.cc`, the LaBr raw branch builds its `art::Ptr` from the **HPGe** collection size instead of LaBr:

```cpp name=DAQ/src/STMDigisFromFragments_module.cc url=https://github.com/Mu2e/Offline/blob/stm-unpacking/DAQ/src/STMDigisFromFragments_module.cc
if(_saveRawWaveform_LaBr){
  stm_waveform.set_data(payloadWords, payloadPtr);
  raw_LaBr_waveform_digis->emplace_back(stm_waveform);
  size_t parentRawWfmIdx = raw_HPGe_waveform_digis->size()-1;   // ❌ HPGe, should be LaBr
  lastRawLaBrPtr = art::Ptr<mu2e::STMWaveformDigi>(rawLaBrProductID,parentRawWfmIdx,
                                             event.productGetter(rawLaBrProductID));
  ...
  lastRawIndexLaBr = static_cast<int>(raw_LaBr_waveform_digis->size()) -1;
}
```

Compare with the HPGe branch just above, which correctly uses `raw_HPGe_waveform_digis->size()-1`. The LaBr `parentRawWfmIdx` must come from `raw_LaBr_waveform_digis->size()-1`. As written, LaBr ZS digis will point at the wrong (or non-existent) raw waveform index — and if no HPGe raws were saved, `size()-1` underflows (`size_t` wraps to a huge value).

**Fix:** `size_t parentRawWfmIdx = raw_LaBr_waveform_digis->size()-1;`

---

## ⚠️ Minor issues

1. **Typo in verbose output** — `"/n"` should be `"\n"`:
   ```cpp name=DAQ/src/STMDigisFromFragments_module.cc url=https://github.com/Mu2e/Offline/blob/stm-unpacking/DAQ/src/STMDigisFromFragments_module.cc
   << zsDigi.trigTimeOffset()
   << "/n";          // ❌ should be "\n"
   ```

2. **Unresolved reviewer request still pending** — @AndrewEdmonds11 asked to *"just remove this commented-out block"* in `STMUtils.cc`. The old `getChannel` logic is still left as a comment block (`*/` visible in the diff hunk). The case-insensitivity request **was** addressed (`std::tolower` added) — good.

3. **`inspectSTMFile.fcl` left at debug settings** — `verbosityLevel : 6` and `saveSTMFragSummary : true`. That's fine for a test/inspection fcl, but verbosity 6 is very chatty; confirm that's intended for committed config.

4. **Dead/leftover comments** — stray `//add if statemnets to guard`, `//Record parent waveform` (empty), and commented-out `segADCS`/`stm_waveform` lines in the ZS block add noise. Minor, but this PR is largely a cleanup so worth tidying.

5. **Typo in a comment** — `"Counts all HPGe and LaBr frags, no fiklter"` (fiklter → filter).

---

## ✅ Looks good
- `STMFragmentSummary.hh` + dictionary entries (`classes.h`, `classes_def.xml`) are consistent and complete, including `art::Ptr<mu2e::STMWaveformDigi>`.
- ZS region-count mismatch check added alongside the existing length check — good defensive validation.
- The instance-name fixes in `mwd.fcl` (`makeSTMDigis:zsLaBr`), `plotSTMPHSpectrum.fcl` (HPGe/LaBr mismatch), and `makeSTMHits.fcl` (adding `HPGePath`) all correct prior copy-paste errors.

### Want me to…
1. Push a commit fixing the LaBr `Ptr` bug, the `"/n"` typo, and removing the commented-out `getChannel` block?
2. Post the LaBr `Ptr` bug as an inline review comment on the PR?
3. Check the `build pending` CI status / failing job for this PR?

oksuzian:
### 🔴 Critical — `getChannel` returns "Unknown" which doesn't exist

In `Mu2eUtilities/src/STMUtils.cc`, the fallback uses `"Unknown"`, but the enum names (`DataProducts/inc/STMChannel.hh:28-29`) are `"HPGe"`, `"LaBr"`, `"unknown"` (lowercase u).

```cpp name=Mu2eUtilities/src/STMUtils.cc url=https://github.com/Mu2e/Offline/blob/stm-unpacking/Mu2eUtilities/src/STMUtils.cc
      if (name.find("hpge") != std::string::npos){
        return STMChannel::findByName("HPGe");
      }
      else if (name.find("labr") != std::string::npos){
        return STMChannel::findByName("LaBr");
      }
      else {
        return STMChannel::findByName("Unknown");   // ⚠️ no match → still returns unknown, but by accident
      }
```

`findByName("Unknown")` won't match `"unknown"` (case-sensitive compare in `STMChannel.cc:30`). It still happens to return `STMChannel(unknown)` via the not-found fallback, so it works — but only by luck. Use `"unknown"` (or better, return `STMChannel(STMChannel::unknown)` directly) to make the intent explicit and avoid a silent trap if that fallback ever changes.

---

### 🟠 Consistency — instance names renamed but not all consumers updated

`STMDigisFromFragments` produces instances `zsHPGe`/`zsLaBr`/`phHPGe`/`phLaBr`/`rawHPGe`/`rawLaBr`. The fcl files were partially realigned:
- ✅ `mwd.fcl`: `makeSTMDigis:LaBr` → `makeSTMDigis:zsLaBr`
- ✅ `plotSTMPHSpectrum.fcl`: fixed `phHPGe`→`phHPGe`/`phLaBr` mismatch and added HPGe path

Double-check there are no remaining references to the old bare instance names (`:HPGe` / `:LaBr` / unprefixed `phHPGe`) elsewhere in fcl, since the `getChannel` logic now also depends on the instance/label containing `hpge`/`labr`. Instances like `zsHPGe` and `rawHPGe` still resolve correctly (substring match), which is good.

---

### 🟡 Minor issues

1. **Unaddressed review comment** — @AndrewEdmonds11 asked to *remove the commented-out block* in `STMUtils.cc` (the `*/` and old commented code above `getChannel`). The diff still shows leftover comment artifacts (`}` / `*/`) around the old implementation. Worth confirming it's fully cleaned up.

2. **`PlotSTMWaveformDigis_module.cc`** — `histtitle2` is still built then largely discarded: the histogram is created with `histtitle.str()` (not `histtitle2`), so the parent-raw-index / offset info added to `histtitle2` never reaches the plot title:
   ```cpp
   hWaveformOffset = tfs->make<TH1F>( histname2.str().c_str(), histtitle.str().c_str(), ... );
   //                                                          ^^^^^^^^^^ should likely be histtitle2
   ```
   Likely a copy/paste leftover — the offset/parent labeling work is effectively dead.

3. **Stray leftover comments / typos** in `STMDigisFromFragments_module.cc` event summary: `"no fiklter"`, and a commented `//#include ".../Assns.h"` left in. Cosmetic.

4. **`mergeable_state: unstable`** + label `build pending` — CI hasn't confirmed green. No check runs are reported on the head commit yet; the PR should not merge until the Mu2e build passes, especially given the new dictionary entries (`classes_def.xml` / `classes.h`) for `STMFragmentSummary` and `art::Ptr<STMWaveformDigi>`.

oksuzian:
@FNALbuild run build test

PR #1861, opened 2026-06-19
update RunTool for new run_info database schema

rlcee:
The run_info database schema ahs been developed over the last few months.  It is at a stable point and so in this PR we adapt the RunTool to the new schema.  The content is basically the same, but the tables have be reorganized.  The old schema is not supported.  Several classes are added to represent the new organization. The command line tool has the same interface, but the output is updated.  RunTool can be include in c++ code and used anywhere.  The python interface generated by pywrap has been tested.  This update is necessary to move ahead pass1 automation.  There should be no effect on validation or operations, so low risk merge.

### DIFF
--- a/DbService/inc/RunConfig.hh
+++ b/DbService/inc/RunConfig.hh
@@ -1,0 +1,41 @@
+#ifndef DbService_RunConfig_hh
+#define DbService_RunConfig_hh
+
+// holds info for one row in the config table
+
+#include <string>
+#include <vector>
+
+namespace mu2e {
+
+class RunConfig {
+ public:
+  typedef std::vector<RunConfig> RunConfigVec;
+
+  RunConfig() : _run_number(0), _version(0) {}
+
+  int runNumber() const { return _run_number; }
+  const std::string& subsystem() const { return _subsystem; }
+  const std::string& settings() const { return _settings; }
+  const std::string& createTime() const { return _create_time; }
+  int version() const { return _version; }
+
+  void setRunNumber(int run_number) { _run_number = run_number; }
+  void setSubsystem(const std::string& subsystem) { _subsystem = subsystem; }
+  void setSettings(const std::string& settings) { _settings = settings; }
+  void setCreateTime(const std::string& create_time) {
+    _create_time = create_time;
+  }
+  void setVersion(int version) { _version = version; }
+
+ private:
+  int _run_number;
+  std::string _subsystem;
+  std::string _settings;  // jsonb stored as string
+  std::string _create_time;
+  int _version;
+};
+
+}  // namespace mu2e
+
+#endif
--- a/DbService/inc/RunInfo.hh
+++ b/DbService/inc/RunInfo.hh
@@ -1,9 +1,13 @@
 #ifndef DbService_RunInfo_hh
 #define DbService_RunInfo_hh
 
-// holds info for one run for runTool
+// holds info for one run, including related records from multiple tables
 
-#include <string>
+#include "Offline/DbService/inc/RunConfig.hh"
+#include "Offline/DbService/inc/RunRecord.hh"
+#include "Offline/DbService/inc/RunSubRun.hh"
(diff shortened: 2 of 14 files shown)


oksuzian:
### 1. 🔴 `setenv("TZ", …)` is never restored — global side effect (`RunTool.cc`, printRun)
```c++ name=DbService/src/RunTool.cc url=https://github.com/Mu2e/Offline/blob/7e3c704575e655c51279a18fdf2870896b9a5201/DbService/src/RunTool.cc#L419-L426
      // Set timezone to Central US (America/Chicago)
      setenv("TZ", "America/Chicago", 1);
      tzset();
      std::strftime(start_buf, sizeof(start_buf), "%Y-%m-%d %H:%M:%S",
                    std::localtime(&start_t));
```
`RunTool` is advertised as includable "anywhere" in C++. Permanently mutating the process `TZ` env var (and not restoring it) will silently corrupt any later time handling in the host program. It's also set inside the per‑subrun loop, so it runs on every iteration. **Fix:** save/restore the prior `TZ`, or use a thread‑safe `localtime_r`/`gmtime_r` without touching the environment, and do it once outside the loop.

### 2. 🟠 `nEvents()`/`minEwt()` etc. return `long` but setters parse `std::stol`, while time fields are truncated to `int`
```c++ name=DbService/inc/RunSubRun.hh url=https://github.com/Mu2e/Offline/blob/7e3c704575e655c51279a18fdf2870896b9a5201/DbService/inc/RunSubRun.hh#L25-L33
  int startTimeUnix() const { return _start_time_unix; }
  int stopTimeUnix() const { return _stop_time_unix; }
```
`_start_time_unix` / `_stop_time_unix` are stored as `int` and parsed with `std::stoi`. These are Unix timestamps — already in the 1.7‑billion range, well within `int32` for now, but this is exactly the pattern that bites on 32‑bit `time_t`/2038. Since `printRun` assigns them to `std::time_t`, prefer storing as `long`/`std::time_t` and parsing with `std::stol` for consistency with the other count fields.

### 3. 🟠 `version` is read but never populated for configs
`RunConfig` has `version()` and `printRun` prints `config.version()`, but in `listRuns` the config parser only sets run_number, subsystem, and settings — `setVersion()` / `setCreateTime()` are never called. The custom CSV splitter stops after extracting `settings` and discards the trailing `create_time,version` fields, so **config version will always print as 0**.

```c++ name=DbService/src/RunTool.cc url=https://github.com/Mu2e/Offline/blob/7e3c704575e655c51279a18fdf2870896b9a5201/DbService/src/RunTool.cc#L262-L267
            RunConfig config;
            if (!run_num_str.empty())
              config.setRunNumber(std::stoi(run_num_str));
            config.setSubsystem(subsystem);
            config.setSettings(settings);
            runInfo.addConfig(config);
```

### 4. 🟡 Subrun field ordering looks inconsistent / fragile
In the subrun parser, `nNull` is read from index 11 while `eventModeCounts` (index 9) and `createdAt` (index 10) are read positionally without the `"None"` guard used by every numeric field. If the DB column order isn't exactly what's assumed, values land in the wrong fields silently (rows are accepted with only `size() >= 2`). Worth a comment documenting the exact expected column order, mirroring the one added for the `run` table.

### 5. 🟡 Hand-rolled CSV/JSON parsing is duplicated and brittle
The config block manually scans quotes/escapes and `runTool_main.cc` re-implements JSON pretty-printing by hand. This is a lot of bespoke string logic that `splitString` was presumably meant to handle (the comment says JSONB escapes force the manual path). Not a blocker, but a likely future maintenance/bug source — a single CSV-field helper would reduce risk.

rlcee:
AI comments addressed
1) reasonable comment, fixed
2) this reflects the schema, will work to fix, OK for now
3) lazy AI (and, I guess, operator) fixed
4) hardened all select statements
5) much better now, hacks removed

rlcee:
@FNALbuild run build test

PR #1862, opened 2026-06-19
db table hash infrastructure

rlcee:
One of the issues with transferring database tables from online to offline is that some tables (cat3 in doc 54331) might already have the same content in the offline database or might have new content added online.  So we need to ask does this content already exist offline?  This PR adds a hash to tables when their content is committed, so we can can just ask if the hash already exists.  When table content is stored and retrieved by Query Engine, the text can be tweaked a bit to deal with quoting, so that is removed in the hash process.  This PR just establishes the hash, it does not store it in the ValCalibrations table yet, that will be a second PR.  In this PR there should be no effect on validation or operations.

### DIFF
--- a/DbService/src/DbTool.cc
+++ b/DbService/src/DbTool.cc
@@ -943,6 +943,7 @@ int mu2e::DbTool::commitCalibration() {
   args["file"] = "";
   args["addIOV"] = "";
   args["addGroup"] = "";
+  args["verify"] = "";
   if ((rc = getArgs(args))) return rc;
 
   if (args["file"].empty()) {
@@ -952,6 +953,7 @@ int mu2e::DbTool::commitCalibration() {
 
   bool qai = !args["addIOV"].empty();
   bool qag = !args["addGroup"].empty();
+  bool qverify = !args["verify"].empty();
 
   if (qag && !qai) {
     std::cout << "commit-calibration: addGroup requested without addIOV "
@@ -976,6 +978,16 @@ int mu2e::DbTool::commitCalibration() {
     std::cout << "commit-calibration: no table found in file " << args["file"]
               << std::endl;
     return 2;
+  }
+
+  if (qverify) {
+    for (auto lt : coll) {
+      std::cout << std::setw(25) << lt.table().name()
+                << std::setw(30) << lt.iov().to_string(true)
+                << std::setw(18) << lt.table().hash() << std::endl;
+      if (_verbose > 5) std::cout << lt.table().csv();
+    }
+    return 0;
   }
 
   rc = commitCalibrationList(coll, qai, qag, _admin);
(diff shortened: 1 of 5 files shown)


oksuzian:
### 🔴 1. Persisted hash uses a non-portable, non-stable algorithm
`std::hash<std::string>` is **implementation-defined**. It is not guaranteed to be stable across compilers, libstdc++/libc++, or even library versions, and it can be randomized between runs/builds.

```c++ name=DbTables/src/DbUtil.cc url=https://github.com/Mu2e/Offline/blob/8f858628a34829cb37c3401cf32b98c15f2237a7/DbTables/src/DbUtil.cc#L322-L323
  std::string fullhash = std::to_string( std::hash<std::string>()(nominal) );
  return fullhash.substr(0,16);
```

Since the whole point is to compare hashes computed **online vs. offline** (potentially different builds/times) and store them long-term, this is the central risk. The two environments could produce different hashes for identical content, defeating the feature. Recommend a deterministic algorithm (e.g., a fixed CRC/MD5/SHA via a library already in the stack) instead of `std::hash`.

Also: `std::to_string()` of a 64-bit hash yields a decimal string whose length varies (1–20 digits). Taking `substr(0,16)` of a **decimal** representation keeps only the high-order digits and silently discards entropy — and for small values returns the whole string. If a fixed-width digest is wanted, hex-format it.

### 🟠 2. Quote-stripping only triggers for length > 1, ignores trimming for length ≤ 1
The quote-removal block is guarded by `temp.size()>1`, but the inner check requires both first and last chars to be `"`. A single `"` (size 1) or empty quoted field `""` (size 2, handled) behave inconsistently. Minor, but worth confirming against what Query Engine actually stores so the offline hash matches.

```c++ name=DbTables/src/DbUtil.cc url=https://github.com/Mu2e/Offline/blob/8f858628a34829cb37c3401cf32b98c15f2237a7/DbTables/src/DbUtil.cc#L310-L317
        if(temp.size()>1) {
          if(temp[0]=='"' && temp[temp.size()-1]=='"') {
            temp = temp.substr(1,temp.size()-2);
            boost::trim(temp);
          }
        }
```

### 🟡 3. No column/row delimiters in the canonical string → collision risk
`nominal` is built by appending every trimmed word with **no separator**. Rows `["ab","c"]` and `["a","bc"]` hash identically. For calibration content this is unlikely to matter, but a cheap delimiter (e.g., append `,` per field and `\n` per row) would remove the ambiguity.

### 🟡 4. Typos in user-facing help text
```c++ name=DbService/src/DbTool.cc url=https://github.com/Mu2e/Offline/blob/8f858628a34829cb37c3401cf32b98c15f2237a7/DbService/src/DbTool.cc#L2903
                 "    --verify : only read fle and summarize with hash\\n"
```
"fle" → "file". Also "canoniacal" → "canonical" in `DbUtil.hh` (line ~22) and "surounding" in `DbUtil.cc` (line 311).

rlcee:
Fixed AI comments:
1) I think I knew this back when I started, but forgot, so a good catch.  Switched to sha1
2) this case is prevented elsewhere
3) very rare case, but interesting point, fixed
4) cleaned up

oksuzian:
@FNALbuild run build test

PR #1864, opened 2026-06-24
Added muon taggers to CRV KPP setup

ehrlich-uva:
Older files without muon taggers can still be processed after these changes.
The geometry of the CRV modules did not change.
Still need to add new calibration tables to the DB, since CRV channel number increased.


oksuzian:
## Issues found

**1. `clang-tidy` reports 2 errors + 17 warnings** — the CI table flags these but they don't block the build. Worth checking the [clang-tidy log](https://buildmaster.fnal.gov/buildmaster/job/GitHubPRTests/job/mu2e-offline-build-test//3165/artifact/clang-tidy.log) to confirm none come from the new `CrvPhotonGenerator_module.cc` edits.

**2. Three large calibration files appear to be committed without a clear consumer / naming convention mismatch.** The PR adds five `*_extracted_*.txt` data files but the fcl only bumps to `v04`/`v05`. Note the inconsistent date-vs-version naming:
- `calib_extracted_20260624.txt`, `status_extracted_20260624.txt` (date-stamped)
- `calib_extracted_v04.txt`, `calib_extracted_v04_perfect.txt`, `extracted_v04.txt`, `photonyield_extracted_v04*.txt` (version-stamped)

The PR body says "Still need to add new calibration tables to the DB" — so these committed text files are likely staging inputs for the DB upload, not consumed directly by Offline. Confirm whether all five need to live in the repo or whether some are throwaway generation inputs (the `_perfect` zero-deviation file especially).

**3. `printChannelMap.fcl` switches the default geometry via commented-out line swap** rather than cleanly. This leaves a dangling commented `geom_run1.txt` line:
```diff name=CRVReco/test/printChannelMap.fcl url=https://github.com/Mu2e/Offline/blob/059242e522a36c4ddd1a3ef94c2369bbc4070e42/CRVReco/test/printChannelMap.fcl#L35
-services.GeometryService.inputFile                   : "Offline/Mu2eG4/geom/geom_run1.txt"
+#services.GeometryService.inputFile                   : "Offline/Mu2eG4/geom/geom_run1.txt"
+services.GeometryService.inputFile                   : "Offline/Mu2eG4/geom/geom_common_extracted.txt"
```
Confirm this default change is intended for committed test code (it changes what `printChannelMap` operates on for everyone), or whether it should be reverted before merge.

**4. `nSupportStructures = 0` is explicitly a placeholder.** In `crv_counters_extracted_v04.txt`:
```text name=Mu2eG4/geom/crv_counters_extracted_v04.txt url=https://github.com/Mu2e/Offline/blob/059242e522a36c4ddd1a3ef94c2369bbc4070e42/Mu2eG4/geom/crv_counters_extracted_v04.txt#L233
int crs.nSupportStructures = 0;                                   //TODO: this is only a place holder so far
```
This is the 1 TODO flagged by CI. Fine for KPP staging, but worth a tracking note so the support-structure geometry isn't forgotten.

## Possible improvements
- **`epilog_extracted_v05.fcl`**: the 8 muon-tagger lookup tables all reuse `LookupTable_2370_1` (the CRV-DS short module), with an inline comment acknowledging "no lookup table for 2200mm counters." Since M1–M8 are 2200 mm bars, consider a tracking issue to generate a proper 2200 mm table so the photon yield isn't approximated long-term.
- **DB dependency is a hard blocker for real use**: per the PR body, reconstruction of new-format files will fail until the 2304-channel calibration tables are uploaded to the conditions DB. Recommend noting the DB PR/IOV here so reviewers know the merge order.
- **Scintillation-yield comment**: the M1–M8 yields are set to 28610 ("assumed to be similar to CRV-L") — flag as an assumption to validate against measured tagger data when available.

ehrlich-uva:
Updated SiPM and Time calibrations for the muon taggers for real data.
The new geometry has more channels and requires the following database for MC:
```
services.DbService.purpose                           : Sim_best
services.DbService.version                           : v1_5
```

PR #1865, opened 2026-06-27
extract cat3 tables from blobs

rlcee:
This adds a routine to search run_info json blobs to find the keyword DBServiceTables and pull out the following values, which will be cat3 table content.  This is niche development, no change to validation or ops.

### DIFF
--- a/DbService/CMakeLists.txt
+++ b/DbService/CMakeLists.txt
@@ -12,6 +12,7 @@ cet_make_library(
       src/GrlList.cc
       src/GrlTool.cc
       src/OpenTool.cc
+      src/RunConfig.cc
       src/RunInfo.cc
       src/RunTool.cc
     LIBRARIES PUBLIC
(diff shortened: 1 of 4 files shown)


oksuzian:
Findings:

Duplicate-key collision in JSON mode — out[p.first] = p.second; silently overwrites when the same table key appears under multiple DBServiceTables blocks. Flat mode (json=false) preserves all values; JSON mode drops dupes. Inconsistent, and likely surprising since the whole point is merging across depth. 

Parameter shadowing — dbTables3(bool json) shadows using nlohmann::json;. Inside the function json is the bool, so all calls must use nlohmann::json. Compiles, but fragile/confusing. Rename param to as_json. 

-q short flag is non-obvious for "dbtables" (q for query?). Minor, but worth a clearer letter.

rlcee:
Address AI comments from Yuri:
- I was going to address this when the rules were final, but I think it is a good idea in any case, so added
- good comment, fixed
- I considered this, -q is the best I could do

PR #1866, opened 2026-06-27
enable creating db tables from text in memory

rlcee:
Add a function to create DbTables from text in memory instead of just from a text file.  This will help make the online to offline work a little simpler. niche development, no change to ops or validation

### DIFF
--- a/DbTables/inc/DbUtil.hh
+++ b/DbTables/inc/DbUtil.hh
@@ -9,6 +9,7 @@ namespace mu2e {
 class DbUtil {
  public:
   static DbTableCollection readFile(std::string const& fn, bool saveCsv = true);
+  static DbTableCollection readString(std::string const& txt, bool saveCsv = true);
   static void writeFile(std::string const& fn, DbTableCollection const& coll);
 
   // split a csv string into lines on \n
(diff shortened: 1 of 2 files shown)


oksuzian:
**1. Behavior change: missing terminal newline now throws (medium)**
The old loop used `std::getline(myfile, line)`, which happily reads a final line that has no trailing `\n`. The new path routes through `splitCsvLines()`, which **throws** `DBUTIL_NO_TERMINAL_NEWLINE` if the text doesn't end in a newline ([DbUtil.cc#L199-L202](https://github.com/Mu2e/Offline/blob/80f68938a763eab36d78f6ed39c51bf4731f4f05/DbTables/src/DbUtil.cc#L199-L202)). So a previously-valid file/string lacking a final newline will now fail. Worth confirming this is intentional — if `readString` is meant to be a convenient in-memory entry point, callers constructing text by hand are a likely source of newline-less input. Consider appending a `\n` if missing, or documenting the requirement on the new function.

**2. Performance nit: copying each line (low)**
```cpp
for (std::string line : lines)   // copies every line
```
`splitCsvLines` already allocated the vector; `boost::trim(line)` mutates in place, so iterating by value forces a needless copy per line. Use `for (std::string& line : lines)` (or `auto&`) to trim in place without copying.

rlcee:
replying to AI comments:
1) I think the new behavior is slightly better
2) OK, fixed

PR #1869, opened 2026-07-01
KinKal truth-seeded track extrapolation

RobMina:
... for MC. Helps to disentangle fit uncertainty from extrapolation bias/resolution. Turned off by default, comments in prolog.fcl show how to enable it.

This is independent of the material definition PR.


brownd1978 (commented):
Hi Rob, this is a nice addition to our functionality.
I'm uncomfortable adding MC-specific code to reconstruction modules intended nominally to run on real data, as that adds an otherwise unnecessary dependency. Eventually we will need to split MC-specific codes out of Offline, and this coupling will make that harder.
An alternative is to write a dedicated module that produces a KKTrack based purely on MC information. If you need to associate that with the reco KKTrack you could do that via the KalSeedMC. What do you think?

RobMina:
Thanks Dave! I've implemented the new modules as you suggested, splitting the MC-specific code out of the two reco fit modules.

brownd1978 (commented):
Hi Rob, thanks for splitting out the truth track, it's much cleaner this way. I hope it wasn't hard to change your analysis to work through the assns.

I see very little (any?) explicit dependence on trajectory type in this code, which begs the question whether you could unify these modules under a templated base class. That would also extend to the implementation to LoopHelix. You would need a runtime check to make sure the KalSeeds were of the expected type, and a switch to turn off domains, but otherwise it looks like the implementation is common. What do you think?

RobMina:
Hi Dave, that's a great idea! Re-implemented using a single module with templated base class, and validated for both CentralHelix and KinematicLine.

brownd1978:
@FNALbuild run build test

Mu2eKinKal/src/KKTruthSeed_module.cc
```diff
@@ -0,0 +1,355 @@
...
+    auto addTruthHit = [&](double hitTime, double covarianceScale){
+      auto cparams = truthpt.nearestPiece(hitTime).params();
+      for(size_t ipar = 0; ipar < KinKal::NParams(); ++ipar){
+        for(size_t jpar = 0; jpar < KinKal::NParams(); ++jpar) cparams.covariance()[ipar][jpar] = 0.0;
+        auto const sigma = truthSeedParamConstraints_.at(ipar);
+        cparams.covariance()[ipar][ipar] = covarianceScale*sigma*sigma;
+      }
+      truthParamHits.push_back(std::make_shared<PARAMHIT>(hitTime, truthpt, cparams, truthMask));
+    };
+    // a single piece gives a zero-length fit range (lowNDOF); split it into two hits a sliver apart.
```
brownd1978: I'm surprised this condition ever arrises: the recotraj should have  as many (or more) pieces as there are straw hits.

Mu2eKinKal/src/KKTruthSeed_module.cc
```diff
@@ -0,0 +1,355 @@
...
+      addTruthHit(truthHitTimes.front() - 0.5*truthHitDt, 2.0);
+      addTruthHit(truthHitTimes.front() + 0.5*truthHitDt, 2.0);
+    } else {
+      for(auto const hitTime : truthHitTimes) addTruthHit(hitTime, 1.0);
+    }
+
+    // rebuild the fit's field domains from the seed so the truth fit/extrapolation see the same BField
+    // model (empty for a field-free line; skipped entirely if RebuildDomains is off).
+    DOMAINCOL truthDomains;
+    if(rebuildDomains_){
```
brownd1978: Instead of a config parameter you could set rebuildDomains_ = (typeid(KTRAJ) != typeid(KinKal::KinematicLine))

Mu2eKinKal/src/KKTruthSeed_module.cc
```diff
@@ -0,0 +1,355 @@
...
+        double ke = pts[i].kineticEnergy();
+        double pmag = std::sqrt(ke*ke + 2.0*ke*mass); // |p| from kinetic energy
+        bestd2 = d2; found = true; tpos = pd; tmom = pmag*dir; ttime = pts[i].t();
+      }
+    }
+    return found;
+  }
+
+  template <class KTRAJ>
+  std::optional<KalSeed> KKTruthSeed::extrapolateFit(KinKal::ParticleTrajectory<KTRAJ> const& recotraj,
```
brownd1978: because extrap_ can be null and this function will still function and produce useful output the name is a bit misleading. maybe createTruthSeed instead?

brownd1978 (approved):
Thanks Rob, this looks great! I still have a few minor quibbles, nothing that would block approval.

brownd1978:
I have an immediate use case for truth extrapolation of LoopHelix (testing accuracy of reflection extrapolation). If you'd prefer not to add that I can easily make a follow-on PR, but since you did all the work you could also just add it. I understand that you might not be able to test it beyond compilation, that's fine.

RobMina:
Hi Dave, I added the LoopHelix and verified that it builds correctly. Also, I addressed the other earlier comments.

oksuzian:
@FNALbuild run build test

PR #1870, opened 2026-07-02
Expand regrowing options

brownd1978:
This PR adds the ability to regrow a Ptr list (selection) of LoopHelix KalSeeds, and to deep-copy the associated KalSeedMCs, to allow regrowing only selected tracks. If this feature is generally useful it could be added to regrowKinematicLine as well. Or perhaps it's time to merge these into a single module.

I also started moving some regrowing fcl into Offline prolog.

### DIFF
--- a/Mu2eKinKal/fcl/prolog.fcl
+++ b/Mu2eKinKal/fcl/prolog.fcl
@@ -485,8 +485,8 @@ Mu2eKinKal : {
       @table::Mu2eKinKal.KKFIT
       SampleSurfaces : ["ST_Outer","ST_Front","ST_Back"] # these are additional surfaces; surfaces used in extrapolation are also sampled
       # save trajectories in the Detector region
-     SaveDomains : true
-     SaveTrajectory : Detector
+      SaveDomains : true
+      SaveTrajectory : Detector
     }
     FitSettings : {
       @table::Mu2eKinKal.LHSEEDFIT
@@ -573,6 +573,26 @@ Mu2eKinKal : {
 
 Mu2eKinKal : {
   @table::Mu2eKinKal
+
+  RegrowLH : {
+    module_type : RegrowLoopHelix
+    CaloClusterCollection : "CaloClusterMaker"
+    StrawDigiIndexMap : "SelectReco:StrawDigiMap"
+    RefitSettings : {
+      @table::Mu2eKinKal.REGROW
+      BFieldCorrection : true
+    }
+    KKFitSettings: {
+      @table::Mu2eKinKal.KKFIT
+      SampleSurfaces : ["ST_Outer","ST_Front","ST_Back"]
+      AddHits : false
+      SaveDomains : true
+      SaveTrajectory : Detector
+    }
+    ExtrapolationSettings : @local::Mu2eKinKal.LHDRIFTXTRAP
+    Extend: false
+  }
+
   producers : {
     # seed fits: these don't use drift information or BField corrections
     KLSeedFit: @local::Mu2eKinKal.KLSeedFit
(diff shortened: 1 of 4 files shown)


brownd1978:
@FNALbuild run build test

oksuzian:
## 🔴 Bugs

**1. `mom` selection compares against wrong initial value → best-momentum pair never selected correctly**
In `SelectReflections_module.cc`, the `value` seed is `std::numeric_limits<double>::max()`, but the `mom` branch now looks for the *maximum* momentum (`>`). Since no real momentum exceeds `max()`, the `mom` case can never update `ibest` past 0.

```cpp name=TrkReco/src/SelectReflections_module.cc url=https://github.com/Mu2e/Offline/blob/refreco/TrkReco/src/SelectReflections_module.cc#L135-L157
double value = std::numeric_limits<double>::max();   // ok for the "min" cases
...
if(selbest_ == mom && std::get<2>(match) > value){     // never true: nothing > max()
  ibest = imatch;
  value = std::get<2>(match);
} else if(selbest_ == deltat && std::get<3>(match) < value){
...
} else if(selbest_ == nactive && std::get<5>(match) > value){  // same bug: never > max()
  ibest = imatch;
  value = std::get<5>(match);
}
```

The two "maximize" branches (`mom`, `nactive`) need to start from `lowest()`/`0`, while the two "minimize" branches (`deltat`, `deltap`) need `max()`. As written, a single shared `value` initialized to `max()` breaks both maximize cases — so `SelectBestPair : 3` (the new default in `prolog.fcl`) will always pick the first match, not the one with the most active hits.

**2. Missing `else` chains the comparisons incorrectly**
Because the four conditions are `if / else if`, only one can fire per match — that's fine — but combined with the shared `value` bug above, switching between metrics is unsafe. Recommend one `value` per metric or an early `switch` on `selbest_` before the loop.

## 🟡 Correctness / robustness

**3. Linear MC-assns lookup is O(N²)**
The old code indexed MC assns by position (`(*ksmca_H)[iseed]`). The new code scans the entire `KalSeedMCAssns` for every seed:

```cpp name=Mu2eKinKal/src/RegrowLoopHelix_module.cc url=https://github.com/Mu2e/Offline/blob/refreco/Mu2eKinKal/src/RegrowLoopHelix_module.cc#L287-L300
for(auto ksmci= ksmca.begin(); ksmci != ksmca.end(); ++ksmci){
  if(ksmci->first == kseedptr){ ksmcai = ksmci; break; }
}
```

Correct given the Ptr-based approach, but O(seeds × assns). Fine for small collections; worth a comment noting the tradeoff.

**4. `mayConsume` used for required products**
`kseedcol_T_`, `kseedptrcol_T_`, and `cccol_T_` are now built with `mayConsume` even when the collection is present. Since exactly one seed collection is always required (enforced at runtime), `consumes` would be more accurate for the branch that's actually configured and improves ART's dependency tracking.

**5. `getValidHandle` on `KalSeedPtrCollection` copies the whole collection**
`kseedptrs = *kseedptr_H;` copies the Ptr vector. Minor, but a `const auto&` in the `has_kseedptrcol_` branch (mirroring the `has_kseedcol_` branch) would avoid the copy — though the two branches build `kseedptrs` differently, so this needs a small refactor.

## 🟢 Style / nits

- **Whitespace-only churn**: several `prolog.fcl` hunks (the `SaveDomains`/`SaveTrajectory` re-indentation, removed `#` comment lines) are pure formatting and inflate the diff. Not blocking, but noise for reviewers.
- **`clang-tidy`**: the build reports **4 errors, 38 warnings** ([log](https://buildmaster.fnal.gov/buildmaster/job/GitHubPRTests/job/mu2e-offline-build-test//3179/artifact/clang-tidy.log)). Non-blocking per Mu2e convention (`:arrow_right:`), but worth confirming none are newly introduced by this PR (it was 4 errors / 32 warnings before the latest push).

brownd1978:
AI comments response:
1: fixed
2: NA
3: Unavoidable side effect of Assns not supporting a Ptr-based lookup
4: fixed
5: unavoidable

brownd1978:
@FNALbuild run build test

bonventre (approved):

PR #1871, opened 2026-07-02
Stortini/msd updates

matthewstortini:
Just adding field in reco data product for MSD channel ID, and a simple filter module.

### DIFF
--- a/DAQ/src/MSDHitsFromDTCEvents_module.cc
+++ b/DAQ/src/MSDHitsFromDTCEvents_module.cc
@@ -102,6 +102,9 @@ void art::MSDHitsFromDTCEvents::produce(Event& event) {
       for (size_t blockIndex = 0; blockIndex < decoder.block_count(); blockIndex++) {
         mu2e::MobileSyncDataDecoder::sync_data_t dataPacketsVec =
             decoder.GetMobileSyncPackets(blockIndex);
+        // get channel ID of data block
+        const int rocLinkID =
+            static_cast<int>(dtcSubEvent.GetDataBlock(blockIndex)->GetHeader()->GetLinkID());
         // get number of hits in packet
         const unsigned nHitsInPacket = decoder.GetNHitsPerPacket();
         if (_debugLevel > 4) {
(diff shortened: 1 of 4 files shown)


michaelmackenzie (approved):
Looks good, thanks!

oksuzian:
### 🔴 Blocking

**1. `_npass` is never incremented → summary always reports 0 passes**
In `filter()` you compute `passed` but never bump `_npass`, so `endRun()` always logs `passed 0 events`.

```cpp name=Trigger/src/MSDHitFilter_module.cc url=https://github.com/Mu2e/Offline/blob/stortini/msd-updates/Trigger/src/MSDHitFilter_module.cc#L74-L88
    passed &= naccepted >= _minHits;

    // Return the result
    return passed;
```
Add `if(passed) ++_npass;` before returning.

**2. `goodHit()` ignores `channelID` — the whole point of the PR**
The PR adds `_channelID` to `MSDHit`, but the filter only checks `hasTime()` and never uses the channel ID. Either filter on channel ID or clarify why it's unused.

```cpp name=Trigger/src/MSDHitFilter_module.cc url=https://github.com/Mu2e/Offline/blob/stortini/msd-updates/Trigger/src/MSDHitFilter_module.cc#L60-L64
  bool MSDHitFilter::goodHit(const MSDHit& hit) {
    if(!hit.hasTime()) return false; // time must be defined for the hit
    return true;
  }
```

### 🟡 Should fix

**3. `channelID` has no `hasChannelID()` guard, unlike other fields**
`MSDHit` defines `hasTime()`/`hasTOT()` but no `hasChannelID()`, despite using the same `-1` sentinel. Add one for consistency.

```cpp name=RecoDataProducts/inc/MSDHit.hh url=https://github.com/Mu2e/Offline/blob/stortini/msd-updates/RecoDataProducts/inc/MSDHit.hh#L28-L31
  int _channelID{-1}; // channel ID
```

**4. Author attribution mismatch** — the module header credits "Michael MacKenzie, 2026" but the PR is by matthewstortini. Confirm this is intentional.

**5. Trailing newline removed** at the end of `MSDHitsFromDTCEvents_module.cc` (last hunk drops the final `\n`) — minor, but flags as a diff nit.

### Notes / non-blocking
- `minHits >= naccepted` uses a count of *accepted* hits, but `goodHit` currently accepts nearly everything — behavior is effectively "at least N hits with a time."
- 2 conversation comments exist on the PR; no formal reviews yet.

matthewstortini:
1. Addressed
2. Not relevant
3. Not relevant -- the checks that exist are due to different ROC firmware versions having different payloads.

oksuzian:
@FNALbuild run build test

PR #1872, opened 2026-07-02
Updated CRV pulse reconstruction for secondary pulses

ehrlich-uva:


oksuzian:
## Issues found

**1. `failedFit` propagation logic looks wrong / suspicious** — `MakeCrvRecoPulses.cc` lines 195–200
```c++ name=CRVReco/src/MakeCrvRecoPulses.cc url=https://github.com/ehrlich-uva/Offline/blob/00dce128b1e979609201c5b947ac9eaed2d0c512/CRVReco/src/MakeCrvRecoPulses.cc#L195-L200
    if(_failedFits.size()>0)
    {
      if(_failedFits.back()) _failedFits.push_back(true);
      else _failedFits.push_back(failedFit);
    }
    else _failedFits.push_back(failedFit);
```
Once *any* fit fails, `_failedFits.back()` is `true`, so **every subsequent pulse is force-flagged as failed** regardless of its own fit quality. Since pulses are found in time order and later pulses are the newly-exposed secondaries, a single early failed fit poisons all later pulses. If the intent is "a bad fit contaminates the subtraction so downstream pulses are unreliable," that's defensible — but it deserves a comment, and it's a behavioral change worth confirming. If not intended, this is a bug.

**2. Unbounded loop risk if `FindNextPeak` doesn't advance `start`** — lines 143 & 203
```c++ name=CRVReco/src/MakeCrvRecoPulses.cc url=https://github.com/ehrlich-uva/Offline/blob/00dce128b1e979609201c5b947ac9eaed2d0c512/CRVReco/src/MakeCrvRecoPulses.cc#L143-L203
  while(FindNextPeak(g, start, peakStart, peakEnd, fitStart, fitEnd))
  {
    ...
    start=fitEnd;
  }
```
The loop advances via `start = fitEnd`. `fitEnd` can be as small as `peakEnd - ... ` after the `if(fitEnd-peakEnd>1) fitEnd--;` adjustment, and `FindNextPeak` scans from `start+1`. It's *probably* monotone because a real peak requires a falling edge after `start`, but there's no explicit guard that `start` strictly increases. After `SubtractPulse` flattens the region, re-scanning near the same index should not re-trigger — but worth a defensive check or assertion to guarantee termination.

**3. Removed config-validation guard, silently** — `CrvRecoPulsesFinder_module.cc`
The check `pulseAreaThreshold > minADCdifference → throw` was removed (correct, since those params are gone). ✅ No action — just confirming this removal is intentional and complete (both the module ctor and `MakeCrvRecoPulses` ctor guards are gone).

**4. `CrvRecoPulse` data-model break is not backward compatible** — `RecoDataProducts/inc/CrvRecoPulse.hh`
Dropping `_PEsNoFit`, `_pulseTimeNoFit`, `_pulseStart`, `_pulseEnd` changes the persisted data product. Any existing art/ROOT files written with the old class, and any downstream code / analysis / plugins calling `GetPEsNoFit()`, `GetPulseStart()`, etc., will fail to compile or read. Confirm a `classes_def.xml` / ClassVersion bump and that no other package references these getters.

**5. `int16_t` truncation in `SubtractPulse`** — lines 92–94
```c++ name=CRVReco/src/MakeCrvRecoPulses.cc url=https://github.com/ehrlich-uva/Offline/blob/00dce128b1e979609201c5b947ac9eaed2d0c512/CRVReco/src/MakeCrvRecoPulses.cc#L92-L94
    int16_t diff=static_cast<int16_t>(std::round(_f1.Eval(t)));
    g.SetPointY(bin,g.GetPointY(bin)-diff);
```
`_f1.Eval` returns a `double`; casting the *subtracted amount* to `int16_t` mimics ADC quantization (reasonable), but if `_f1.Eval` ever exceeds ±32767 the cast overflows (UB/wraparound). A large pulse height * `TMath::E()` could plausibly approach that. Consider clamping or using `float`/`int` for `diff`.

**6. clang-tidy: 2 errors, 32 warnings** — [clang-tidy log](https://buildmaster.fnal.gov/buildmaster/job/GitHubPRTests/job/mu2e-offline-build-test//3175/artifact/clang-tidy.log)
CI is green overall, but 2 clang-tidy **errors** were reported. Worth checking whether they're in the files this PR touched.

ehrlich-uva:
Responses:
1. No code change required, but I added a comment. The intended behavior ("a bad fit contaminates the subtraction so downstream pulses are unreliable") was correctly identified.
2. No code change required, because the start variable will always advance.
3. No action required.
4. The functions that were removed are not used anywhere else.
5. Switched to double to address the concern.

PR #1873, opened 2026-07-03
Add dt_hit / dt_ptoca linear fit information for hits on a track

michaelmackenzie:
The straw hit time vs. track expected time at the hit position has a non-unit slope when the wrong mass hypothesis is assumed. This slope information can be used in PID, as is relevant to PR https://github.com/Mu2e/MLTrain/pull/9. This adds a reco data product with this information and a producer module to create a collection of these products for input KalSeed or KalSeedPtr collections.

### DIFF
--- a/ParticleID/CMakeLists.txt
+++ b/ParticleID/CMakeLists.txt
@@ -15,7 +15,16 @@ cet_build_plugin(ParticleIDRead art::module
     LIBRARIES REG
       art_root_io::TFileService_service
       Offline::ParticleID
-      
+
+      Offline::RecoDataProducts
+)
+
+cet_build_plugin(TrackDtDt art::module
+    REG_SOURCE src/TrackDtDt_module.cc
+    LIBRARIES REG
+      art_root_io::TFileService_service
+      Offline::Mu2eUtilities
+      Offline::ParticleID
       Offline::RecoDataProducts
 )
 
--- a/ParticleID/src/SConscript
+++ b/ParticleID/src/SConscript
@@ -21,6 +21,7 @@ libs =  [
     'mu2e_ConfigTools',
     'mu2e_DbTables',
     'mu2e_GeneralUtilities',
+    'mu2e_Mu2eUtilities',
     'art_Framework_Core',
     'art_Framework_Principal',
     'art_Framework_Services_Registry',
--- a/ParticleID/src/TrackDtDt_module.cc
+++ b/ParticleID/src/TrackDtDt_module.cc
@@ -1,0 +1,95 @@
+// Fit the track dt_{hit} / dt_{track fit poca time} distribution
+//
+// Michael MacKenzie, 2026
+
+#include <iostream>
+#include <string>
+#include <vector>
+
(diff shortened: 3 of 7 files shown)


oksuzian:
## Possible Improvements

The following are concrete, diff-anchored points worth addressing before merge:

1. **Weight formula looks physically wrong / risks division-by-zero.**
   `ParticleID/src/TrackDtDt_module.cc:60-62`
   ```c++
   const double t_unc  = hit._udresid / hit._dvel;
   const double weight = 1./t_unc;
   ```
   `_udresid` is the unbiased **distance residual value** (see `TrkStrawHitSeed.hh:186`), not an uncertainty — so `residual/velocity` is not a time *uncertainty*. A time error would more plausibly come from a variance term (e.g. `sqrt(_udresidmvar)/_dvel`, or `_utocavar`). As written, `weight` can be negative (residuals are signed) and `1/t_unc` blows up when `_udresid == 0` or `_dvel == 0` (the legacy BTrk constructor sets `_dvel = 0`). Please confirm the intended uncertainty and guard against zero/negative weights.

2. **`LsqSums2` accessors are non-const, but you pass `seed` by const-ref.** This compiles only because `fitter` is a local non-const. Still, consider that `dof_ = fitter.qn() - 2` can go negative for tracks with <2 hits, and `chisq_ = chi2Dof()*dof_` will then be meaningless. Add a guard for `qn() < 3` (skip or set sentinel values).

3. **Struct members are uninitialized when set from the fit.** `KalSeedDtDt`'s default ctor zero-inits, but `fitTrackDtDt` creates a `result` and sets 5 fields — fine today, but if any `LsqSums2` accessor returns NaN (e.g. `det()==0` for degenerate input), it propagates silently. A quick `isnan` check before emplace would harden the product.

4. **Minor:** `#include <iostream>` guarded diagnostics use `std::cout`; consider `mf::LogInfo`/`MessageLogger` for consistency with framework conventions, though many Mu2e modules do use `std::cout` under `diagLevel`.

oksuzian:
@FNALbuild run build test

michaelmackenzie:
I updated the dt/dt slope definition to match Dave's suggestion and added debug histogramming for easy testing. I tested on flat electron and muon inputs with an electron track fit that the slope distribution is peaked at 1 for true electrons and ~1.5 for true muons.
image

brownd1978 (approved):
This looks good Michael.
What is your plan for adding this to the reconstruction sequence?

michaelmackenzie:
> This looks good Michael. What is your plan for adding this to the reconstruction sequence?

I'm not positive about the best location. It could either go in `reco` after track reconstruction or in the ntupling stage before the MVA consumers. I don't have a strong opinion on this, but we can put this in `reco` if we want to minimize any reconstruction-type producers within ntupling jobs. Do you have a preference or alternate suggestion?

brownd1978:
Putting it in the general track reco sequence will add it to every track, I'm not sure if we need that. Putting it in the PID sequence seems more targeted.

PR #552, opened 2021-07-31
Small modifications to CrvReco and CrvCoincidence modules

ehrlich-uva:
-small change to the no-fit pulse intervals in the CrvReco module 
-added 2/4 layer coincidence option
-added variables to CrvCoincidence and CrvCoincidenceClusters
-bug fix in CrvCoincidence module
-removed xxxxxCollection.hh files for CrvCoincidence and CrvCoincidenceCluster, which affects a few files outside of CRVResponse

PR #1875, opened 2026-07-07
Pawel plesniak/absorber shift

PawelPlesniak:
This PR addresses several issues found when completing the corrections to my normalization studies. 

Changes are summarized in the bullet point list for each file affected by core code changes. Critically note - there are several other files that are also affected, but the `pre-commit` config highlighted many changes that needed to happen before my changes could go in (after rebasing with branch `main`). These changes have been addressed.

 - `CountVirtualDetectorHits_module.cc` - this makes the vector of virtual detectors being summarized in the model unique, my mapping to and back from an unordered set
 - `PhotonGun_module.cc` - this now allows for a set of photons to travel in a straight line from the origin to  target position, defined by a direction vector as `delta_x`, `delta_y`, and `delta_z`. This vector is mapped to the unit vector, and the momentum is scaled accordingly.
 - `AbsorberFromSTHPGe.fcl` - driver code that will allow future unfolding studies to be performed, especially important for the HPGe where additional absorber is present, and the results have not yet been fully quantified.
 - `AbsorberFromSTLaBr.fcl` - as above but for LaBr.
 - `EventFilter.fcl` - debugging tool to help with ongoing simulation studies. Allows for looking at individual events in a data file. For my studies, it was used to demonstrate the behaviour of the ZS and MWD algorithms on a macro-spill worth of beam-on time.
 - `FirePhotonToDetector.fcl` - driver code that fires singular photons to a detector of choice, aimed for use with studying the STM detector waveform generators and analysis tools.
 - `STMMC/fcl/prolog.fcl` - rewritten to define the simulation parameters more clearly
 - `ROOTAnalysisDump.fcl` - driver code to convert the art files into ROOT format for easy plotting. I will not be rewriting this for the current analysis framework as I do not have the time for this.
 - `HPGeWaveformsFromStepPointMCs_module.cc` - addresses issues with the waveform generator, namely
   - MWD was failing with this implementation as the waveform amplitudes were set to zero during microspills that did not have photons interacting with the detector crystal, throwing off MWD
   - Concatenated events were discontinuous
   - Added preamplifier charge collection response function
 - `SimParticleAndVDBacktrace_module.cc` - new module generating ROOT files with data that backtraces of STM signal photons back through the `SimParticle` provenance. The code to plot these will be merged in the [STMAnalysis repository](https://github.com/Mu2e/STMAnalysis)
 - `SimParticleDump_module.cc` - dumps the full `SimParticle` provenance to ROOT files for plotting, with a more compact format than the previous module.
 - `STMResamplingProducer_module.cc` - Initial simulations had different data product names for `SimParticle`s propagated with the `BeamToVD` and `BeamToVD1809` drivers, this change standardizes them.
 - `VirtualDetectorTree_module.cc` - add more data to ROOT files generated for plotting.
 - `STMMovingWindowDeconvolution_module.cc` - some more data for visualizing the behaviour of analyzed waveforms

I would like to ask some questions
 - Why was the `STMMWDDigi` renamed to `STMPHDigi`, and what is the difference in analysis chains?


PawelPlesniak:
I leave this for review by @YongyiBWu and @AndrewEdmonds11

PawelPlesniak:
Note - I will review any failures that the CI workflow raises

PawelPlesniak:
@FNALbuild run build test

PawelPlesniak:
@FNALbuild run build test

PawelPlesniak:
@YongyiBWu and @AndrewEdmonds11 The CI workflow issues have been addressed, this is now ready for review

oksuzian:
## 🔴 Bugs (likely functional problems)

**1. `SimParticleDump_module.cc` — empty-file counter never resets, and threshold check is wrong**
```c++ name=STMMC/src/SimParticleDump_module.cc url=https://github.com/Mu2e/Offline/blob/74637245173494acdb9a2fca2fca9b54dde2a789/STMMC/src/SimParticleDump_module.cc
if (!stepHandle || stepHandle->empty()) { consecutiveEmptyFileCounter++; return; }   // returns BEFORE threshold check
...
if (consecutiveEmptyFileCounter > consecutiveEmptyFileThreshold)
    throw cet::exception("LogicError", "Too many consecutive empty files, stopping the job");
```
Because the empty branches `return` immediately, the threshold `throw` is only ever reached on a *non-empty* event — the exact opposite of intended. The counter is also never reset to 0 on success in the early-return paths. Compare with `SimParticleAndVDBacktrace` / `VirtualDetectorTree`, which check the threshold *before* returning. The author also flags this file as unvalidated: `// TODO before upload - validate that the code works`.

**2. `HPGeWaveformsFromStepPointMCs_module.cc` — `digitize()` clipping logic is inverted**
```c++ name=STMMC/src/HPGeWaveformsFromStepPointMCs_module.cc url=https://github.com/Mu2e/Offline/blob/74637245173494acdb9a2fca2fca9b54dde2a789/STMMC/src/HPGeWaveformsFromStepPointMCs_module.cc#L562-L565
ADC = _chargeDecayed[i] * chargeToADC;
_adcs[i] = (ADC > ADCMax) ? static_cast<int16_t>(std::round(ADC)) : ADCMax;
```
`ADCMax` is the most-negative value (`-2^15 + 1`), and the waveform is negative-going. This keeps the rounded value only when `ADC > ADCMax` and otherwise floors it to `ADCMax` — that clamps *from below* but nothing prevents positive overflow, and the ternary condition reads backwards relative to its intent. Worth double-checking the direction against the validation `if (j > 1000) throw ... "ADC values too high!"` on line 355.

**3. `HPGeWaveformsFromStepPointMCs_module.cc` — validation constant `defaultMicrospillBufferLengthCount < 1` is unreachable**
```c++ name=STMMC/src/HPGeWaveformsFromStepPointMCs_module.cc url=https://github.com/Mu2e/Offline/blob/74637245173494acdb9a2fca2fca9b54dde2a789/STMMC/src/HPGeWaveformsFromStepPointMCs_module.cc#L196-L197
if (defaultMicrospillBufferLengthCount < 1)
    throw cet::exception("RANGE", "defaultMicrospillBufferLengthCount has to be at least 1!\n");
```
`defaultMicrospillBufferLengthCount` is a hard-coded `const int = 2`, so this guard can never fire. The check should validate the user-supplied `microspillBufferLengthCount` instead.

**4. `PhotonGun_module.cc` — new energy check rejects valid low-energy photons**
```c++ name=EventGenerator/src/PhotonGun_module.cc url=https://github.com/Mu2e/Offline/blob/74637245173494acdb9a2fca2fca9b54dde2a789/EventGenerator/src/PhotonGun_module.cc
if (E < std::numeric_limits<double>::epsilon())
    throw cet::exception("RANGE") << "Energy must be greater than zero, exiting.";
```
`epsilon()` (~2.2e-16) is a floating-point precision tolerance, not "zero". Any legitimately small energy below it is rejected. Use `E <= 0.0`. (The good news: this diff *fixes* a real copy-paste bug on the same lines — `px = conf().py()` → `py = conf().py()`.)

**5. `HPGeWaveformGenerationAndAnalysis.fcl` — dangling data-product tag after module rename**
The producer `PHHPGe` was renamed to `MWDHPGe`, but the output still references the old name:
```fcl name=STMMC/fcl/HPGeWaveformGenerationAndAnalysis.fcl url=https://github.com/Mu2e/Offline/blob/74637245173494acdb9a2fca2fca9b54dde2a789/STMMC/fcl/HPGeWaveformGenerationAndAnalysis.fcl
"keep mu2e::STMPHDigis_PHHPGe_*_HPGeReco"
```
With no `PHHPGe` module in any path, this `keep` will drop the intended digis. Should be `MWDHPGe`.

## 🟠 Correctness / robustness concerns

**6. Uninitialized counters read in summaries** — In both `STMMovingWindowDeconvolution_module.cc` (`processedADCs`) and `HPGeWaveformsFromStepPointMCs`, some counters are declared without an initializer (`uint ... processedADCs;`) and later printed / accumulated (`processedADCs += nADCs`). Undefined initial value. Initialize to `0`.

**7. `SimParticleDump` uses `SimParticles.at(step.trackId())`** — `.at()` on the `SimParticleCollection` throws if the track id isn't present; other modules in this PR guard more defensively. Given the file is explicitly unvalidated, worth confirming the key type/lookup.

**8. Hard-coded 320 MHz coupling** — `produce()` now throws unless `fADC == 320`, and `decayCharge()` hard-codes the 320 MHz assumption for the preamplifier rise time. The author acknowledges this (`// TODO - this only keeps accurate time if the sampling is 320MHz`), but it silently makes the `fADC` fhicl parameter a lie for any other value.

## 🟡 Minor / cleanup

- **Typo:** `unqiueEnabledVDsVec` in `CountVirtualDetectorHits_module.cc`, and duplicate `;;` on the same line.
- **Debug leftovers:** `HPGeWaveformsFromStepPointMCs` carries `waveform_time` with a self-described `// TODO - REMOVE ME!`, plus `kept_events`/`dropped_events` labelled "for debugging".
- **Copy-pasted banner:** `STMMovingWindowDeconvolution::beginJob()` prints `"STM HPGe digitization parameters"` — wrong header for the MWD module.
- **Copy-paste in `.fcl`:** `AbsorberFromSTLaBr.fcl` uses `process_name: HPGeAbsorberSpectrumShift` (LaBr file, HPGe process name).
- **Author-acknowledged unfinished work:** several `@nil` "Populate me!" placeholders (`HPGeWaveformGenerationAndAnalysis.fcl`, `ROOTAnalysisDump.fcl`), a `makeADCPlot` path that always throws `"not yet implemented"`, and TODO blocks left in `prolog.fcl` ("TODOs before Pawel uploads").
- **Comment typo:** `"5. Ou1tputs"` and `"[BRAKCETS]"` in `HPGeWaveformGenerationAndAnalysis.fcl` / `ROOTAnalysisDump.fcl`.

YongyiBWu (approved):
I do not see any direct issue with the proposed changes

DAQ/src/STMBinaryDigisFromFragments_module.cc
AndrewEdmonds11, line 1: This looks like it's only whitespace changes. I think this file also got renamed at some point so might cause a merge issue...

STMMC/src/HPGeWaveformsFromStepPointMCs_module.cc
```diff
@@ -1,6 +1,11 @@
 // Simulates the electronics response of the HPGe detector. Simulates the pulse height, decay tail, and ADC digitization. Generates one STMWaveformDigi per micropulse.
 // Model based heavily on example provided in docDb 43617
 // See docDb 51487 for full documentation
+// Remaining TODOs for future development:
+//  - Include the commented out includes correctly
+//  - Implement the makeADCPlot functionality correctly
+//  - Implement different sampling frequencies correctly
```
AndrewEdmonds11: Just FYI these can probably be handled with the module ```PlotSTMWaveformDigis```

AndrewEdmonds11 (approved):
This looks good to me. There are several files which just have whitespace changes, which would be good to clean up before merging. There are some very useful modules in here that would be good for generating STM mock datasets: PhotonGun, HPGeWaveformsFromStepPoints etc.

DAQ/src/STMBinaryDigisFromFragments_module.cc
PawelPlesniak: I'll merge develop again before I merge, but I have a quick question - for the whitespace changes, what do you mean by "clean up before merging"? It took a while to be able to commit as the pre-commit action I had blocked the `git commit` before these were all changed, and I am unsure of how to disable this so I can re-commit the files with the whitespaces addressed

PawelPlesniak:
# Responses to AI review
## 🔴 Bugs (likely functional problems)
### 1. SimParticleDump_module.cc — empty-file counter never resets, and threshold check is wrong
> Because the empty branches return immediately, the threshold throw is only ever reached on a non-empty event — the exact opposite of intended. 

This has been extended to check that the threshold is exceeded before checking that the event data products are not empty.

> The counter is also never reset to 0 on success in the early-return paths. Compare with SimParticleAndVDBacktrace / VirtualDetectorTree, which check the threshold before returning. 

This has been updated to unify the logic

> The author also flags this file as unvalidated: // TODO before upload - validate that the code works.

Validation complete.

### HPGeWaveformsFromStepPointMCs_module.cc — digitize() clipping logic is inverted
This review is incomplete, and does not understand the physics of what is being simulated: 
> nothing prevents positive overflow

> Worth double-checking the direction against the validation if (j > 1000) throw ... "ADC values too high!" on line 355.

Positive overflow will not be encountered, unless the user sets the pedestal too high. This will be left for the user to address

### 3. HPGeWaveformsFromStepPointMCs_module.cc — validation constant defaultMicrospillBufferLengthCount < 1 is unreachable 

Fair point, corrected.

### 4. PhotonGun_module.cc — new energy check rejects valid low-energy photons

Corrected.

### 5. HPGeWaveformGenerationAndAnalysis.fcl — dangling data-product tag after module rename

Good catch!

## 🟠 Correctness / robustness concerns
### 6. Uninitialized counters read in summaries 
Corrected.

### 7. SimParticleDump uses SimParticles.at(step.trackId())
Corrected

### 8. Hard-coded 320 MHz coupling
The user is made aware of the difference if `fADC` is set differently.

## 🟡 Minor / cleanup
### Typo: unqiueEnabledVDsVec in CountVirtualDetectorHits_module.cc, and duplicate ;;
Corrected

### Debug leftovers
Removed

### Copy-paste in .fcl
Corrected

### Author-acknowledged unfinished work
This is actually for the user to populate. 

### Comment typo:
Addressed where relevant

STMMC/src/HPGeWaveformsFromStepPointMCs_module.cc
PawelPlesniak: Commit [ddcc9c3](https://github.com/Mu2e/Offline/pull/1875/commits/ddcc9c384946abd4b6cf88aeb331134f7da09cf9) reviews all `makeADCplot` references from this module, and [265298f](https://github.com/Mu2e/Offline/pull/1875/commits/265298fc64e7772fce1abcab3dd091baa2b0f7d1) removes it from the driver fcl

PawelPlesniak:
I will wait for the build to be completed, review any issues should they arise, and await clarification on what the whitespace changes should be

oksuzian:
@FNALbuild run build test

PawelPlesniak:
@FNALbuild run build test

oksuzian (changes requested):
### PR Review Summary

Reviewed at head `3cedaf72`

**Decision**
- 🔴 request changes

**Scope understood**
- STM simulation/analysis rework: HPGe waveform generation (preamplifier response, microspill carry-over, event-time accounting), MWD instrumentation, two new `SimParticle` provenance-dump analyzers, six new/rewritten STMMC driver fcls, and a full rewrite of `STMMC/fcl/prolog.fcl`.
- A data-product instance rename in `STMResamplingProducer` (`""` → `"virtualdetector"`) paired with an open companion PR, Mu2e/Production#552.
- Carried along: whitespace-only reformatting across `DAQ/`, `BTrkLegacy/`, `TrkReco/`, `RecoDataProducts/`, `STMReco/` (pre-commit fallout, per the PR body).

**Re-review accounting** (prior review posted 2026-07-09 at head `74637245`; delta `74637245..3cedaf72` = 8 files)

| Prior finding | Status |
| --- | --- |
| 1. `SimParticleDump` empty-file counter / threshold | 🟢 FIXED in `e500ede0`, verified |
| 2. `digitize()` one-sided clipping | 🟠 UNADDRESSED — carried forward (F4) |
| 3. `defaultMicrospillBufferLengthCount < 1` unreachable | 🟢 FIXED (right variable now checked) / ⚪ message still wrong (F10) |
| 4. `PhotonGun` `epsilon()` energy guard | 🟢 FIXED, verified |
| 5. Dangling `PHHPGe` in `keep` | 🟢 FIXED, verified |
| 6. Uninitialized counters | 🟢 FIXED, verified |
| 7. `SimParticles.at(...)` | 🟢 FIXED in `SimParticleDump`; ⚪ still present in `VirtualDetectorTree` (pre-existing) |
| 8. Hard-coded 320 MHz coupling | 🟢 no longer silent — now throws (F14 notes the cadence) |
| 9. `makeADCplot` / `waveform_time // TODO - REMOVE ME!` | 🟢 FIXED in `ddcc9c38`/`265298fc`, verified absent at head |
| 10. MWD banner, `Ou1tputs`, LaBr `process_name` | 🟢 FIXED, verified |
| 11. `unqiueEnabledVDsVec`, `[BRAKCETS]`, `@nil` "Populate me!" | 🟠 PARTIAL — carried forward (F9, F11) |

Other reviewers: @YongyiBWu approved (2026-07-19, no items). @AndrewEdmonds11 approved (2026-08-03) with one request — clean up the whitespace-only files before merging — **still open** (see PR hygiene). His `PlotSTMWaveformDigis` comment was addressed by `ddcc9c38`/`265298fc`.

---

**Findings**

1. 🔴 [S0] `Offline/STMMC/fcl/prolog.fcl` includes a `Production/` file and cannot resolve against Production `main`
   - Evidence: `STMMC/fcl/prolog.fcl:6` adds `#include "Production/JobConfig/pileup/STM/prolog.fcl"`. The new tables then dereference `@local::STMSimDataProducts.Stage1.CompressedOutput.StepPointMCs`, `...Stage2.CompressedOutput.StepPointMCs.Virtualdetector`, and `@local::STMPileup.ResamplingProducer.VirtualDetectorID`. `STMSimDataProducts` does not exist in Production `main` — it is introduced only by Mu2e/Production#552 (still OPEN, head `79419ad3`, where it appears at `JobConfig/pileup/STM/prolog.fcl:15` and `STMPileup.ResamplingProducer` at line 245). A code search over `Mu2e/Offline` for `Production/JobConfig` in `.fcl` returns `STMMC/fcl/prolog.fcl` plus only `*/test/*.fcl` driver files — no other Offline **prolog** depends on Production.
   - Impact: two problems. (a) Layering inversion: Offline is the base layer; making an Offline prolog unparseable without Production on `FHICL_FILE_PATH` breaks Offline-standalone use and inverts the dependency for every consumer of `Offline/STMMC/fcl/prolog.fcl`. (b) Merge-order landmine: Production `main`'s `JobConfig/pileup/STM/BeamToVD.fcl` still `#include`s `Offline/STMMC/fcl/prolog.fcl`, so merging this PR before #552 makes the live STM pileup drivers fail to parse. Offline CI cannot catch this — no validation job in the suite (`ceSimReco`, `ceDigi`, `ceMix`, `g4test_03MT`, …) runs any STMMC fcl, which is why `mu2e/buildtest` is green at `3cedaf72`.
   - Suggested fix: move the new `DataProducts` table into Production's STM prolog (that is where `STMSimDataProducts` already lives) and delete the `#include "Production/..."` line from the Offline prolog. If any of it must stay in Offline, inline the literals. Either way, state the required merge order in both PR bodies and land Production#552 first.

2. 🔴 [S0] `STMResamplingProducer` instance rename breaks three Production drivers the companion PR does not touch
   - Evidence: `STMMC/src/STMResamplingProducer_module.cc:47,66` change `produces<StepPointMCCollection>()` → `produces<StepPointMCCollection>("virtualdetector")` and `event.put(std::move(outputStepPointMCs), "virtualdetector")`. Production#552 updates `BeamToVD.fcl` and `BeamToVD1809.fcl` only. Still on Production `main` and absent from #552's 16-file list: `JobConfig/pileup/STM/BeamTo2VD.fcl:14`, `BeamTo2VD1809.fcl:14`, `NeutralsTo2VD.fcl:17`, each of which hardcodes `physics.producers.compressDetStepMCsSTM116.stepPointMCTags : ["extractVirtualDetectorSteps116"]` — a bare label, i.e. instance `""`. `Compression/src/CompressDetStepMCs_module.cc:618` resolves those with `event.getValidHandle<StepPointMCCollection>(i_tag)`, which throws `ProductNotFound` on a miss.
   - Impact: `BeamTo2VD`, `BeamTo2VD1809` and `NeutralsTo2VD` abort at the first event once this lands. These are the two-VD variants used for the VD116 studies.
   - Suggested fix: add `"extractVirtualDetectorSteps116:virtualdetector"` to those three fcls in Production#552 (they are the only remaining bare-instance consumers).

3. 🟠 [S1] HPGe microspill charge carry-over is snapshotted **before** the deposits, so the buffer is permanently zero
   - Evidence: `STMMC/src/HPGeWaveformsFromStepPointMCs_module.cc:288-292` now runs at the top of `produce()`, ahead of `depositCharge()` at lines 308-319:
     ```cpp
     _chargeCollected.clear();
     _chargeCollected.assign(_chargeCarryOver.begin(), _chargeCarryOver.end());
     _chargeCollected.insert(_chargeCollected.end(), nADCs, 0);   // tail := zeros
     _chargeCarryOver.clear();
     _chargeCarryOver.assign(_chargeCollected.begin() + nADCs, _chargeCollected.end()); // reads the zeros
     ```
     The pre-PR code ran this block at the **end** of `produce()` and in the opposite order — `_chargeCarryOver` was assigned from the tail first (after the deposits had landed there), and only then was `_chargeCollected` rebuilt. Moving the block without swapping the two halves reversed the data dependency. Concretely, with `nADCs_init = 543` and `microspillBufferLengthCount = 2`: `_chargeCollected` = 543 old-carry-over values + 543 fresh zeros, and `_chargeCarryOver` is then assigned exactly that zero tail. It is all zeros for every event after the first.
   - Impact: `depositCharge()` writes late deposits into `_chargeCollected[i]` for `i >= nADCs` (steps admitted by the `step.time() > microspillBufferLengthCount * micropulseTime` cut at line 374), and those are wiped at the start of the next event. The delayed component that `microspillBufferLengthCount` exists for — "charge deposits that are allocated to this event but happen after the microspill ends e.g. 844keV", per the member comment at line 150 — is silently discarded. This is a regression against `main`, and it is invisible: no exception, no counter, just missing charge.
   - Suggested fix: restore the original ordering — assign `_chargeCarryOver` from the tail of `_chargeCollected` *after* the deposit loops, then rebuild `_chargeCollected` for the next event. If the block must stay at the top of `produce()` (e.g. for the `resetEventNumber` interaction at 281-285), split it: rebuild `_chargeCollected` at the top, snapshot `_chargeCarryOver` at the bottom.

4. 🟠 [S1] `digitize()` clamps only the negative rail; positive overflow is undefined behaviour — *carried over from the 2026-07-09 review, unaddressed*
   - Evidence: `HPGeWaveformsFromStepPointMCs_module.cc:530`
     ```cpp
     _adcs[i] = (ADC > ADCMax) ? static_cast<int16_t>(std::round(ADC)) : ADCMax;
     ```
     with `ADCMax = static_cast<int16_t>((-1 * std::pow(2,15)) + 1)` = `-32767` (line 111). Values below `-32767` are floored; nothing bounds `ADC` above `+32767`. A `double` outside the `int16_t` range converted by `static_cast` is UB, so the result is whatever the target ISA's conversion does. The only downstream guard is `if (j > 1000) throw cet::exception("LogicError", "ADC values too high!")` at line 337 — it runs *after* the bad cast, and a hard-coded `1000` that aborts the job is itself a debug artefact rather than a saturation model.
   - Impact: with `NoiseSD` non-zero the baseline sits near 0 and positive excursions are reachable; the failure mode is either a wrapped ADC value written into the digi or a job abort with an unhelpful message.
   - Suggested fix: `_adcs[i] = static_cast<int16_t>(std::round(std::clamp(ADC, static_cast<double>(ADCMax), static_cast<double>(std::numeric_limits<int16_t>::max()))));` and replace the `> 1000` guard with a saturation counter reported in `endJob()`.

5. 🟠 [S1] `Efficiency.PhotonEnergy : @nil` makes all three Absorber drivers non-runnable, including one that worked before this PR
   - Evidence: the prolog rewrite changes `Efficiency.PhotonEnergy` from `1.809` to `@nil` (`STMMC/fcl/prolog.fcl`, `Efficiency` table). Three drivers consume it unconditionally: `Absorber.fcl:31`, `AbsorberFromSTHPGe.fcl:34`, `AbsorberFromSTLaBr.fcl:34`, all `E : @local::Efficiency.PhotonEnergy`. `EventGenerator/src/PhotonGun_module.cc:47` declares `fhicl::Atom<double> E` — required, no default — and line 69 throws on `E <= 0.0`.
   - Impact: `@nil` erases the key, so the reference does not deliver a value and the required atom fails validation at module construction. `Absorber.fcl` ran before this PR (`PhotonEnergy : 1.809`); after it, none of the three start. Note the same rewrite also (correctly) repairs `Absorber.fcl`'s `module_type : MakeVirtualDetectorTree`, a module that does not exist in the repo — so this file goes from broken-one-way to broken-another-way.
   - Suggested fix: keep `PhotonEnergy : 1.809` in the prolog and let a driver override it. (The `LaBrDigitization` table's `@nil` entries are fine — nothing references them.)

6. 🟡 [S2] Seven new members in `STMMovingWindowDeconvolution` are declared and never read
   - Evidence: `STMReco/src/STMMovingWindowDeconvolution_module.cc:118-125` adds `_isFirstWaveform`, `_last_ADC`, `_last_deconvolved`, `_deconvolved_history`, `_differentiated_history`, `_last_average`, and (line 153) `waveform_time_microspill_frame`. Each string occurs exactly once in the file — the declaration. The two `std::deque`s carry the comment "Rolling history buffers to bridge the gap between chunks", i.e. they were added for a chunk-boundary continuity fix that is not in the diff. `<deque>` is also never included; it compiles only via a transitive boost/stdlib include.
   - Impact: dead state that reads as an implemented feature. A future reader will assume the chunk gap is handled.
   - Suggested fix: delete them, or land the chunk-bridging logic they belong to. (Does not gate the decision.)

7. 🟡 [S2] `SimParticleAndVDBacktrace` copies the whole `StepPointMC` collection per particle and per ancestor
   - Evidence: `STMMC/src/SimParticleAndVDBacktrace_module.cc:248` `const std::vector<StepPointMC> StepPointMCs = *stepHandle;` copies the product; then `addToTree`, `addParentToTree` and `traceSteps` all take `const std::vector<StepPointMC> steps` **by value** (declarations at lines 75-77, definitions at 158, 178, 213). `addToTree` recurses up the parent chain (lines 204-210), copying the full collection at every hop, and `traceSteps` linearly scans it each time (line 160).
   - Impact: O(N_particles x chain_depth) full-vector copies plus an O(N_steps) scan each, i.e. quadratic in a VD collection that can hold thousands of steps. This is a new module, so the cost is entirely introduced here.
   - Suggested fix: `const std::vector<StepPointMC>&` on all three signatures and `auto const& StepPointMCs = *stepHandle;` in `analyze`.

8. 🟡 [S2] Per-microspill sample budget averages 543.6 ticks where the clock gives 542.4 — the concatenated timeline stretches
   - Evidence: `HPGeWaveformsFromStepPointMCs_module.cc:273-278`
     ```cpp
     nADCs = nADCs_init;                       // 543 at 320 MHz (line 197-199)
     eventTimeBuffer = eventId % 5;
     if (!(eventTimeBuffer == 0 || eventTimeBuffer == 3)) nADCs++;
     ```
     and `eventTime += nADCs` at line 361. Over any 5 consecutive microspills that emits `543*5 + 3 = 2718` ticks. The physical budget is `5 * micropulseTime / tADC = 5 * 1695 ns / 3.125 ns = 2712` ticks. The `+1` at line 197 (`(micropulseTime/tADC) + 1`, documented as "increase the number by one due to truncation") is a per-microspill ceiling, which is right for sizing one buffer but wrong as the base of a running total.
   - Impact: 6 extra ticks (18.75 ns) per 5 microspills, ~0.22 %. Over the `nMicrospillsPerSpill : 31858` concatenation configured in the prolog that is roughly 24 µs of accumulated offset by the end of a macrospill, which lands directly on the MWD peak times.
   - Suggested fix: base `nADCs_init` on the floor (542) with `+1` on 2 of every 5 microspills, or drive the pattern from an exact accumulator over `micropulseTime/tADC`. Please confirm the intended tick budget against the DAQ spec — the old code (`+= nADCs + 1` when `eventId%5==0 || eventId%3==0`) had a different, also non-exact, pattern, so this is a behaviour change either way.

9. 🟡 [S2] Committed drivers still carry `@nil` "Populate me!" placeholders and one dangling `module_type` — *carried over, partially addressed*
   - Evidence: `HPGeWaveformGenerationAndAnalysis.fcl:90` `trigger_paths : @nil` and `:99` `SelectEvents: @nil`; `ROOTAnalysisDump.fcl:103` `o1 : ["HPGeEDeps"] # Populate me!` and `:107`; `ROOTAnalysisDump.fcl:86` sets `module_type : MWDTree` and lines 93/96/99 set `module_type : @nil`. A code search over `Mu2e/Offline` for `MWDTree` returns nothing — the module does not exist (`STMMC/src/` has `PHTree_module.cc` and `HPGeTree_module.cc`). These are inert only because they sit off the active paths.
   - Impact: `HPGeWaveformGenerationAndAnalysis.fcl` cannot be run as shipped; `ROOTAnalysisDump.fcl` ships a reference to a non-existent module. The PR body states this file will not be brought up to the current analysis framework, so it will stay in this state.
   - Suggested fix: give both drivers a runnable default path (`trigger_paths : [digitization_path_ZS]`, `SelectEvents : [digitization_path_ZS]`), delete the `@nil`/`MWDTree` analyzer blocks, or relocate these to `STMMC/test/` where non-runnable scaffolding is expected.

10. ⚪ [S3] `microspillBufferLengthCount` guard message contradicts its own check — *residual of prior finding 3*
    - `HPGeWaveformsFromStepPointMCs_module.cc:192-193`: `if (microspillBufferLengthCount < 2) throw cet::exception("RANGE", "microspillBufferLengthCount has to be at least 1!\n");`. The variable swap was the substance of the fix and is correct, but the bound is 2 (line 202 sizes `_chargeCarryOver` as `nADCs * (count - 1)`), so a user who sets `1` is told `1` is allowed. Change the message to "at least 2".

11. ⚪ [S3] Carried-over typos still present
    - `unqiueEnabledVDsVec` — `Analyses/src/CountVirtualDetectorHits_module.cc:55,60` (the duplicate `;;` on that line *was* fixed).
    - `[BRAKCETS]` — `STMMC/fcl/HPGeWaveformGenerationAndAnalysis.fcl:10` and `STMMC/fcl/ROOTAnalysisDump.fcl:10`.
    - New: `mf::LogInfo log("HPGeWaevfrmsFromStepPointMCs")` — `HPGeWaveformsFromStepPointMCs_module.cc:536`, misspelled message-facility category.

12. ⚪ [S3] Dead and duplicated configuration state
    - `STMMC/src/VirtualDetectorTree_module.cc:64` adds `uint simParticleIdKey = 0;` — declared, never used (the tree fills `simParticleId`, a different member).
    - `STMMC/fcl/prolog.fcl` defines `ComponentPositions.ST` and `Efficiency.ST` with identical values `(-3904.0, 0.0, 627.0)` in the same file — one should reference the other.
    - `DigitizationTester.HPGe.delta{x,y,z}` are annotated `# MeV/c`, but `PhotonGun::Config` (lines 44-46) documents `deltax/y/z` as position differences used to build a direction; `FirePhotonToDetector.fcl` never reads them, so nothing catches the mismatch.
    - Please confirm `Efficiency.FromSTToDet.LaBr.z : 40264.5`: `ST.z + 40264.5 = 40891.5`, whereas `ComponentPositions.LaBr.z` (added in this PR) is `40404`. The HPGe pair is self-consistent (`627 + 40072.1 = 40699.1` = `crystalCentreZ` in the module), so the LaBr offset looks like it targets a different reference point.

13. ⚪ [S3] Includes and duplication in the new/modified modules
    - `SimParticleAndVDBacktrace_module.cc` uses `std::find` (105, 163), `std::fill` (172, 174) and `std::distance` (165) but does not `#include <algorithm>`; it compiles transitively.
    - `Analyses/src/CountVirtualDetectorHits_module.cc:6` and `HPGeWaveformsFromStepPointMCs_module.cc:12` keep `#include <bits/stdc++.h>` — a GCC-internal header, non-portable and the opposite of the "include only headers actually needed" rule. Pre-existing in both, so not gating, but `CountVirtualDetectorHits` is being edited here anyway and now only needs `<algorithm>`, `<unordered_set>`, `<vector>`.
    - `SimParticleAndVDBacktrace::addToTree` (178-198) and `addParentToTree` (213-235) duplicate ~15 lines verbatim. The comment at line 203 — "Require separate definitions of addParentToTree because of data accessors" — no longer applies now that both take `const art::Ptr<SimParticle>&`; they can be one function.

14. ⚪ [S3] `override` and check cadence in `HPGeWaveformsFromStepPointMCs`
    - `beginJob()` (line 78) and the newly added `endJob()` (line 83) are declared without `override`; the base signatures match today, so they are called, but a base-class change would silently orphan them.
    - The `fADC == 320` guard at lines 271-272 is a constructor-time invariant evaluated on every event. Move it next to the `tADC` computation at line 188.

15. ⚪ [S3] `depositCharge()` writes ~1000 dead values per StepPointMC
    - Lines 471-472 replaced `_charge[tIndex] = N_ehPairs;` with a loop that fills the plateau across the entire remaining `_charge` buffer. Only `[tIndexStart, tIndex)` is read by the `_chargeCollected` update at 480-481, and `std::fill(_charge.begin(), _charge.end(), 0)` at line 484 wipes the rest. Writing just `_charge[tIndex] = _charge[tIndex - 1]` preserves the (intended) continuity change at a thousandth of the cost.

**Observed, pre-existing, not gating**
- `HPGeWaveformsFromStepPointMCs::addNoise()` (lines 519-520) constructs `std::default_random_engine _randomGen;` fresh inside the per-event call with the default seed, so every microspill receives the *identical* noise realization, and it bypasses `SeedService` entirely. Not introduced by this PR, but this PR reworks the charge path that feeds it (`decayCharge` now emits `currentIntegrated`), so the noise model is worth fixing in the same campaign: hoist to a member seeded via `art::ServiceHandle<art::RandomNumberGenerator>` / `SeedService`.
- `VirtualDetectorTree_module.cc:113` still uses `SimParticles.at(step.trackId())`, which throws `std::out_of_range` rather than a `cet::exception` with a useful category. `SimParticleDump` was fixed for exactly this in `e500ede0`; the sibling module was not.

- 🟢 Build-system parity checked in **both** systems. `STMMC/CMakeLists.txt` registers both new plugins (`SimParticleDump` at 60-66, `SimParticleAndVDBacktrace` at 68-74) with `LIBRARIES REG` matching the sibling `VirtualDetectorTree` entry (`art_root_io::TFileService_service`, `Offline::GlobalConstantsService`, `Offline::MCDataProducts`) — correct for their include sets; `VirtualDetectorId` comes in transitively via `MCDataProducts`, same as the existing `ShiftVirtualDetectorStepPointMCs` and `VirtualDetectorTree` plugins. `STMMC/src/SConscript` is a generic `helper.make_plugins([...])` glob over `*_module.cc`, so it needs no per-module edit. No `configure_file` inputs are involved; `install_fhicl(SUBDIRS fcl SUBDIRNAME Offline/STMMC/fcl)` (line 143) already covers the six new fcl files. No un-mirrored plugin and no missing `LIBRARIES`.
- 🟢 `Absorber.fcl` previously specified `module_type : MakeVirtualDetectorTree` — a module that exists nowhere in the repo — plus a `VirtualDetectorId : 101` key. This PR corrects both against `VirtualDetectorTree::Config` (which declares only `StepPointMCsTag`, `SimParticlemvTag`, `consecutiveEmptyFileThreshold`).
- 🟢 `CountVirtualDetectorHits` order change is intentional and safe: the old `std::set` sorted the IDs, the new path preserves config order, and Production#552 supplies an unsorted list (`[10, 13, 15, 86, 115, 100, 101, 88, 89, 90]`) whose report order now matches the config.
- 🟢 `HPGeWaveformGenerationAndAnalysis.fcl` module labels resolve: `STMZeroSuppression` and `ConcatenateDigitizedWaveforms` are both registered plugins; `keep mu2e::STMPHDigis_MWDHPGe_*_HPGeReco` now matches the `MWDHPGe` producer label.
- 🟢 `mu2e/buildtest` is green at `3cedaf72` (2026-08-04T06:53Z) along with all 14 validation jobs — note none of them exercise STMMC, which is why findings 1, 2, 5 and 9 are invisible to CI.

**Validation check**
- Build/tests run: yes, by Mu2e CI at the reviewed head — `mu2e/buildtest` success plus `ceSimReco`, `ceDigi`, `ceMix`, `ceSteps`, `cosmicSimReco`, `cosmicOffSpill`, `g4test_03MT`, `g4study`, `transportOnly`, `POT`, `muDauSteps` all rc 0. No STM fcl is exercised by any of them, and the PR supplies no STM-specific validation evidence.
- Config contract check: **fail** — findings 1 (unresolvable `@local::` references against Production `main`), 5 (`@nil` for a required `Atom<double>`), 9 (`@nil` paths, non-existent `MWDTree`).
- Cross-repo consistency: **fail** — finding 2 (three Production drivers not covered by Production#552) and finding 1 (hard, undeclared merge-order dependency on Production#552 in both directions).

**Residual risk**
- Findings 1 and 2 are release-blocking in the ordering sense: whichever of #1875 / Production#552 merges first leaves the STM pileup chain broken until the other lands, and #552 as written does not cover `BeamTo2VD`/`BeamTo2VD1809`/`NeutralsTo2VD` at all.
- Finding 3 is a silent physics regression. Any HPGe waveform sample already produced from this branch under-counts the delayed (844 keV-class) charge, and nothing in the job output signals it.
- Finding 8 changes the microspill time base relative to `main`; MWD peak times from before and after this PR are not directly comparable.
- The whitespace-only churn (`DAQ/src/STMBinaryDigisFromFragments_module.cc` alone is +173/-173) inflates the diff and, per @AndrewEdmonds11's inline note, risks a merge conflict with a rename of that file. The physics changes are hard to isolate for review or for `git bisect`.

**Author follow-ups**
1. Decide the layering for `STMMC/fcl/prolog.fcl`: move the `DataProducts` table into Production's STM prolog and drop `#include "Production/JobConfig/pileup/STM/prolog.fcl"` from Offline. (F1)
2. Add `:virtualdetector` to the `stepPointMCTags` in Production `BeamTo2VD.fcl`, `BeamTo2VD1809.fcl` and `NeutralsTo2VD.fcl` in PR #552, and record the required Offline/Production merge order in both PR bodies. (F2)
3. Restore the carry-over ordering in `produce()` so `_chargeCarryOver` is snapshotted after `depositCharge()`. Please add a short before/after check — e.g. a single microspill with a step at `t > 1695 ns` — showing the charge appears in the next event's waveform. (F3)
4. Clamp `digitize()` on both rails and replace the `> 1000` abort with a saturation counter. (F4)
5. Restore `Efficiency.PhotonEnergy : 1.809` in the prolog so the three Absorber drivers run. (F5)
6. Either delete the seven unused MWD members or land the chunk-bridging logic they were added for. (F6)
7. Pass the `StepPointMC` collection by const reference throughout `SimParticleAndVDBacktrace`. (F7)
8. Confirm the intended ticks-per-microspill budget against the DAQ spec and reconcile the `543 + 3-in-5` pattern with `1695/3.125 = 542.4`. (F8)
9. Give `HPGeWaveformGenerationAndAnalysis.fcl` and `ROOTAnalysisDump.fcl` runnable defaults (or move them to `STMMC/test/`), and drop the `MWDTree` / `@nil` analyzer blocks. (F9)
10. Best practice reminder: please split the whitespace-only reformatting of `DAQ/`, `BTrkLegacy/`, `TrkReco/` and `RecoDataProducts/` into a separate PR so the STM physics changes stand alone — this also addresses @AndrewEdmonds11's outstanding request. Answering your question about *how*: for files where only whitespace changed, `git checkout main -- <path>` restores them; to get past the hook for that commit use `git commit --no-verify` (or `SKIP=<hook-id> git commit` for a single hook). No need to disable the config permanently.
11. Your open question — "why was `STMMWDDigi` renamed to `STMPHDigi`?" — is still unanswered, and the PR leaves the naming inconsistent: the prolog now exposes `STMMWDDigiTag` (`STMMC/fcl/prolog.fcl`, `STMMCAnalysis.PH.{HPGe,LaBr}`) while the data-product class is still `STMPHDigi` (`keep mu2e::STMPHDigis_MWDHPGe_*_HPGeReco`). Worth settling with the STMReco authors before merge so the config key and the product name agree.

oksuzian:
@FNALbuild run build test

oksuzian (changes requested):
### PR Review Summary — Mu2e/Offline#1875

Reviewed at head `dd4685bc7`. Re-review of `3cedaf72` (2026-08-04).

**Decision**
- 🔴 request changes

**Scope understood**
- Delta since my last review is three commits and seven files: two new analyzers (`GeometricShiftTree_module.cc` +173, `TSCTree_module.cc` +103), two new drivers (`GeometricShiftTree.fcl`, `GeometricShiftTreeMu31809.fcl`), a rewrite of `HPGeTree_module.cc` (+55/-88), the momentum fix in `ShiftVirtualDetectorStepPointMCs_module.cc` (+32/-13), and `CMakeLists.txt` (+22).
- The delta does not touch `STMMC/fcl/prolog.fcl`, `HPGeWaveformsFromStepPointMCs_module.cc`, `STMResamplingProducer_module.cc`, `STMMovingWindowDeconvolution_module.cc`, `SimParticleAndVDBacktrace_module.cc`, or the Absorber drivers — so every prior finding in those files stays open by definition.

**Re-review accounting** (prior review at `3cedaf72`, 15 findings)

| Prior finding | Status |
| --- | --- |
| 1. 🔴 Offline prolog `#include`s `Production/...`, unresolvable against Production `main` | 🔴 UNADDRESSED — and now broader, see finding A |
| 2. 🔴 `STMResamplingProducer` instance rename breaks three Production drivers | 🔴 UNADDRESSED — Production#552 still does not touch them |
| 3. 🟠 HPGe microspill carry-over snapshotted before the deposits | 🟠 UNADDRESSED — file untouched |
| 4. 🟠 `digitize()` clamps only the negative rail | 🟠 UNADDRESSED — file untouched |
| 5. 🟠 `Efficiency.PhotonEnergy : @nil` breaks three Absorber drivers | 🟠 UNADDRESSED — prolog untouched |
| 6. 🟡 Seven unread members in `STMMovingWindowDeconvolution` | 🟡 UNADDRESSED — `STMReco/` untouched |
| 7. 🟡 `SimParticleAndVDBacktrace` copies the collection per particle | 🟡 UNADDRESSED — file untouched |
| 8. 🟡 543-vs-542.4 tick budget per microspill | 🟡 UNADDRESSED — file untouched |
| 9. 🟡 `@nil` paths and dangling `MWDTree` in committed drivers | 🟡 UNADDRESSED — files untouched |
| 10-15. ⚪ guard message, typos, dead state, includes, `override`, `depositCharge` fill | ⚪ UNADDRESSED — files untouched |

Nothing has regressed, and the two S0s are the reason the decision does not move. @AndrewEdmonds11's request to split the whitespace-only files also remains open.

**The good news first: two real bugs were fixed here that I had never reported**

Both came out of a September package sweep of `STMMC` that was never posted to this PR, so you found and fixed them independently. Verified at this head:

- 🟢 **The shifted momentum no longer uses energy as a momentum component.** The old code set `newMomentum.setZ(preE - oldMomentum.perp() * scaleFactor)` with `preE = sqrt(|p|^2 + m^2)` — a total energy placed in a z-momentum slot and combined linearly with a scaled transverse part. It is now `newPrePz = signPz * sqrt(oldPMag2 - newPTMag2)` (`:143-145`), which conserves `|p|` exactly, preserves the original sign of `pz`, and is guarded by an explicit `newPTMag2 > oldPMag2` check (`:138-140`). The old formulation was self-consistent only for `m = 0`; for a massive particle with small `pT` it drove the module's own magnitude check over threshold.
- 🟢 **`newPrePz` and `newPostPz` are assigned before use.** They were previously declared and never written, so the absorber-shadow duplicate step was emitted with `p_z = 0` — purely transverse primaries into the downstream job. They are now set at `:144` and `:158` and read at `:184` and `:188`.
- 🟢 The magnitude checks also improved from a one-sided `a - b > epsilon` against `std::numeric_limits<double>::epsilon()` to a two-sided `std::abs(a - b) > 1e-6`. The old comparison could not fire for a shortfall and used a tolerance no floating-point sum would satisfy.
- 🟢 I checked the mirrored branch too: reusing `newPrePz` there is correct, because mirroring flips the sign of `newMomentum.x()` and `newPTMag2` depends only on its square. `oldPosition`/`oldPostPosition` are mutated inside that branch but reassigned at the top of each iteration (`:113-116`), so the mutation does not leak between steps.

**New findings**

A. 🔴 [S0] The two new drivers extend the unmerged Production dependency rather than resolving it
   - Evidence: `GeometricShiftTree.fcl:39` and `GeometricShiftTreeMu31809.fcl:37` both read `VirtualDetectorID : @local::STMPileup.ShiftVD101Steps.VirtualDetectorID`. On Production `main`, `JobConfig/pileup/STM/prolog.fcl:17` defines `STMPileup` but contains no `ShiftVD101Steps` key; that key exists only in Production#552, at `:262`. Production#552 is still OPEN, head `79419ad32`, unchanged since 2026-08-08.
   - Impact: two more committed drivers that cannot be parsed against Production `main`, on top of `STMMC/fcl/prolog.fcl` from finding 1. This is the same layering problem widening, and it is invisible to CI because no validation job runs an STMMC fcl.
   - Suggested fix: as in finding 1 — settle where the STM pileup tables live, and record the Offline/Production merge order in both PR bodies. If these two drivers are meant as analysis scaffolding rather than shipped configuration, `STMMC/test/` is the better home.

B. 🟡 [S2] The momentum fix left its scaffolding behind, including a conditions handle nothing reads
   - Evidence: the diff deletes `mass = pdt->particle(particle->pdgId()).mass();`, which was the only use of `mass` and the only use of `pdt`. At this head, seven data members occur exactly once each — their declaration on `:65`: `mass`, `preE`, `preE2`, `postE`, `postE2`, `newPrePz2`, `newPostPz2`. `GlobalConstantsHandle<ParticleDataList> pdt` (`:71`) is likewise never used, and its two includes (`:26-27`) are now unnecessary. `STMMC/CMakeLists.txt` still lists `Offline::GlobalConstantsService` in this plugin's `LIBRARIES REG`.
   - Impact: a `GlobalConstantsHandle` is still constructed per module instance for nothing, and seven dead members read as state the algorithm uses. Because they are members rather than locals, no compiler warning can flag them.
   - Suggested fix: delete the seven members, `pdt` and the two includes. The `Offline::GlobalConstantsService` entry can then go too — run a build to confirm before trimming it, since the standard asks that link-list changes be tested.

C. 🟡 [S2] Per-step values held as mutable data members is what allowed the `p_z = 0` bug to be silent
   - Evidence: `:60-72` declares roughly 25 private members that are per-step working values — `inputX/Y/Z`, `outputX/Y/Z`, `x/y/z` ("Buffer variables"), `inputRadius`, `outputRadius`, `scaleFactor`, the eight `CLHEP::Hep3Vector` old/new pairs, and `newStep` — all reassigned inside the loop in `produce()`.
   - Impact: this is the mechanism behind the bug just fixed. Had `newPrePz` been a local, reading it without assigning it would have been a `-Wuninitialized` diagnostic; as a member initialised to `0.0` it silently produced a physically wrong value for however long it shipped. The same exposure remains for every other member in that block.
   - Suggested fix: move the per-step values into the loop body as locals, keeping as members only what genuinely spans events (the token, `VirtualDetectorID`, `pdgID`, the geometry constants). The compiler then enforces what review had to catch by hand.

D. ⚪ [S3] The absorber geometry is still three unsourced literals — `:61`, `const double xHPGeAbsorber = -3944.6, HPGeAbsorberHalfWidth = 25, SSCApertureSpacing = 81.2;`. These are detector dimensions used in two position cuts (`:119`, `:169`), with no units, no origin and no way to change them without a rebuild. The coding standard puts a detector dimension in the geometry service, or at minimum wants a comment naming where the number came from. Also from the September sweep, not previously reported.

E. ⚪ [S3] Two small items, grouped
   - `:97` de-indents `void ShiftVirtualDetectorStepPointMCs::produce` by two spaces with no other change on the line — gratuitous churn in a PR already carrying a whitespace-split request.
   - Pre-existing, but the file is open here: `:99-100` throws `cet::exception("DataError", "Requested data product not found\n")` when the collection is merely *empty*. `getProduct` would already have thrown had the product been absent, so the message describes a different failure, and an event with no VD101 step is a normal physical outcome rather than a fatal error. Likewise `HPGeTree_module.cc:94-95` accepts `Detector : "LaBr"` through validation and then throws "Currently this code only works for HPGe" — the `Comment` on `:55` advertises both values. Neither is new; both are in files this PR rewrites.

**Validation check**
- Build/tests run: none by me. **CI had never run at this head** — FNALbuild's last activity on this PR is 2026-08-04 at `3cedaf72`, and the three commits of 2026-09-20 have no build. The `mu2e/buildtest` "pending" status at `dd4685bc7` is a stale leftover, not a run in progress. I have triggered one; the two new plugins and the `CMakeLists.txt` additions are exactly the kind of change a build tests, so please check the result before merging.
- Config contract check: **fail** — findings 1, 5 and 9 carried, plus new finding A.
- Cross-repo consistency: **fail** — Production#552 is unchanged since 2026-08-08 and still omits `BeamTo2VD.fcl`, `BeamTo2VD1809.fcl` and `NeutralsTo2VD.fcl`, verified against its current 16-file list.
- Not checked: the physics content of the two new analyzers beyond their configuration handling, and whether the shifted distribution reproduces the un-absorbed LaBr aperture as intended — that needs a comparison plot, not a diff.

**Residual risk**
- Unchanged from the last review: findings 1 and 2 are ordering blockers in both directions, finding 3 is a silent physics regression in the HPGe waveform path, and finding 8 moves the microspill time base.
- The momentum fix changes the shifted VD101 distribution relative to every sample produced from this branch before it. Any resampled dataset made from the old code carries both the energy-as-momentum error and the `p_z = 0` duplicates, and nothing in those files records which version produced them.

**Author follow-ups**
1. The two S0s from the last review are still the blockers: settle the Offline/Production prolog layering (finding 1, now also finding A) and cover the three 2VD drivers in Production#552 (finding 2).
2. Delete the dead members, `pdt` and the stale includes left by the momentum fix (finding B).
3. Consider making the per-step values locals (finding C) — it is what would have caught the bug you just fixed.
4. Check the CI run I triggered before merging.
5. Still open from before: the whitespace split @AndrewEdmonds11 asked for, and the `STMMWDDigi`/`STMPHDigi` naming question.

oksuzian:
Three follow-ups to the review above, at the same head `dd4685bc7`. The first is a CI blocker that has nothing to do with your code; the second is one new finding; the third closes out a static sweep of this package against this head so the same items do not come back.

### 1. CI cannot run until this branch is rebased

I triggered a build after posting the review. It came back at 04:01:29Z with `It was not possible to prepare the workspace for this test.` — not a build failure. FNALbuild's `git diff --check` reports 27 leftover conflict markers across seven files once the branch is merged onto current `main`:

`DAQ/src/STMPrintFragments_module.cc`, `DAQ/test/inspectSTMFile.fcl`, `RecoDataProducts/inc/STMWaveformDigi.hh`, `STMMC/src/ConcatenationFilter_module.cc`, `STMMC/src/HPGeTree_module.cc`, `STMMC/src/VirtualDetectorTree_module.cc`, `STMMC/src/ConcatenateDigitizedWaveforms_module.cc`.

**The branch itself is clean** — I read `DAQ/src/STMPrintFragments_module.cc` at `dd4685bc7` and there is not a single marker in it. This is merge staleness against `main`, not committed garbage. A rebase (or a merge of `main`) is needed before any CI result can exist for this PR; until then a test request keeps failing at the workspace step, so item 4 of the author follow-ups in the review above cannot be acted on yet.

### 2. 🟡 [S2] This PR adds two new consumers of a hardcoded beam-centre x, and two new copies of the same literal in fcl

- Evidence at `dd4685bc7`: `STMMC/src/HPGeTree_module.cc:77` declares `double xBeamCentre = -3904.0;`. That declaration is pre-existing, but this PR **adds two new uses of it**, at `:111` and `:113` (both `+` lines in the diff), on top of the existing `:139` and `:141`. In parallel, `STMMC/fcl/prolog.fcl` now carries `-3904.0` at `:65`, `:74`, `:94` and `:285` — two of those four are `+` lines here — plus a `# ST at (-3904, 0, 627)` comment at `:301`, also added. `STMMC/src/LaBrTree_module.cc:96` on `main` holds a third C++ copy of the same number.
- Impact: one physical fact — where the HPGe/LaBr split sits — now has four fcl homes and two compiled-in homes, and neither side can see the other. Revising the fcl value leaves both tree modules bucketing steps by the stale boundary, silently misassigning steps between the HPGe and LaBr trees with no build error and no validation error. House rule 3: the same literal in C++ and fcl is the finding, not the literal.
- Suggested fix: one `fhicl::Atom<double>` on both tree modules, defaulted from `ComponentPositions.Beam.x`.
- Two things this is **not**. It is not the `-3944.6` absorber geometry in `ShiftVirtualDetectorStepPointMCs_module.cc:61` that the review above raises — different number, different file, same house rule. And it is not finding 12 of my 2026-08-04 review, which flagged `ComponentPositions.ST` and `Efficiency.ST` carrying identical values *within* `prolog.fcl`; that one is fcl-internal, this one crosses the C++/fcl boundary.

### 3. Static sweep of `STMMC`, reconciled against this head

A sweep of this package at main `665dc45db` (2026-09-19) flagged two findings in files this PR owns. Neither is a new report:

- **`addNoise()` re-seeds a default-constructed engine every event**, so the preamplifier noise is bit-identical in every event — `HPGeWaveformsFromStepPointMCs_module.cc:519-520`, called once per event from `produce()` at `:330`. The sweep found this independently, but **I already reported it on 2026-08-04**, under "Observed, pre-existing, not gating", with the same fix. It is unchanged at `dd4685bc7`: the PR rewrites this module (+115/-55) and adds a `NoiseSD [ADC bins]` print at `:253` but does not touch the engine. Recording that it is still open, not raising it again. For the record on why it matters in practice: `STMMC/fcl/HPGeWaveformGenerationAndAnalysis.fcl:36` sets `NoiseSD` from `STMMC/fcl/prolog.fcl:254`, which is `0.32 # mV` — a genuine value, so the epsilon early-return at `:515` never fires, and `HPGeWaveformGenerationAndAnalysis.fcl:108` already sets `services.SeedService.baseSeed : 1`.
- **The `resetEventNumber` carry-over reset was dead code** at main: the `std::fill` of `_chargeCarryOver` was discarded by an unconditional `clear()`/`assign()` later in the same `produce()` call. That is no longer true here — `:284` fills and `:289` reads it into `_chargeCollected`, so the fill is live. But it buys nothing while **finding 3 of the review above is still open**: `_chargeCarryOver` is snapshotted from a tail that is still zeros at `:289-292`, so the buffer is all zeros for every event after the first regardless of the reset. Zeroing a buffer that is already zero is not a fix. Nothing to report from the sweep; the ordering fix in finding 3 subsumes it.

No other sweep finding lands in a file this PR touches. Four that land elsewhere in `STMMC` (`DetectorEfficiency`, `ConcatenateDigitizedWaveforms`, `PHTree`) are going into a separate issue against `main`.

### Decision unchanged

🔴 request changes, on the two S0s in the review above. Item 2 here adds to that list without changing it, and item 1 is a rebase rather than a code change.

PR #1877, opened 2026-07-10
Add ability to change particle type when regrowing

brownd1978:
This PR provides an option to force regrown tracks to have a different particle type (mass) than what the KalSeed stores. The default is for regrown tracks to have the same particle type as the KalSeed.


brownd1978:
@FNALbuild run build test

oksuzian:
## PR Summary

**Purpose**: Adds an option to force regrown tracks to have a different particle type (mass) than what the `KalSeed` stores. Default behavior is unchanged (same particle type as the KalSeed).

- **Scope**: 7 files changed (+84, −27), no reviews yet, label `build finished`
- **Risk**: Low–medium — touches a shared reco data product (`KalSeed`/`KalSegment`) but changes are additive/backward-compatible via default arguments
- **Mergeable**: Likely; no blocking reviews present

## Core Changes

1. **`KalSegment`** – trajectory converters (`loopHelix`, `centralHelix`, `kinematicLine`) now take an optional `double mass = -1.0`. Moved from inline header to `.cc`. When `mass >= 0`, they rebuild the `ParticleStateEstimate` with the overridden mass.
2. **`KalSeed`** – the three `*FitTrajectory()` methods now accept an optional `PDGCode::type ptype = PDGCode::unknown`; when set (and different from `_tpart`), they look up the mass via `GlobalConstantsHandle<ParticleDataList>` and pass it down.
3. **`RegrowLoopHelix_module`** – new optional fhicl parameter `FitParticle` (PDG code). Overrides the particle when building `KKTRK`, preserving the sign of the original `kseed.particle()`.
4. **Build files** – add `GlobalConstantsService` dependency to `RecoDataProducts` (CMake + SConscript).

## Issues Found

**1. Behavior change unrelated to the PR's stated purpose (potential regression)** — `RecoDataProducts/src/KalSegment.cc`

The `t0` fallback was changed from a legacy computation to now throwing:

```diff
-    //      throw cet::exception(...)
-    // for now, revert to the legacy implementation...
-    auto vel = _pstate.velocity();
-    return _pstate.time() - _pstate.position3().Z()/vel.Z();
+    throw cet::exception("RECO")<<"mu2e::KalSegment: no trajectory specified in flag" << std::endl;
```

This converts a silent-fallback path into a hard exception. Any caller relying on the legacy behavior (e.g. segments without a trajectory flag set) will now throw. This is out of scope for "change particle type when regrowing" and deserves explicit callout / confirmation that no production path hits it.

**2. Suspicious cast in the module** — `Mu2eKinKal/src/RegrowLoopHelix_module.cc`

```cpp
auto tpart = tpart_ != PDGCode::unknown ? static_cast<PDGCode>(std::copysign(...)) : kseed.particle();
```

`static_cast<PDGCode>` casts to the *class/enum name* `PDGCode`, whereas elsewhere the code uses `PDGCode::type`. If `PDGCode` is a struct wrapping an enum `type`, this may not compile or may be relying on an implicit conversion. Worth verifying it should be `PDGCode::type`.

**3. Duplicate include** — `Mu2eKinKal/src/RegrowLoopHelix_module.cc`

```diff
+#include "Offline/GlobalConstantsService/inc/GlobalConstantsHandle.hh"
+#include "Offline/GlobalConstantsService/inc/ParticleDataList.hh"
```

`ParticleDataList.hh` is already included at the top of the file (line 21 in the context shown). The second include is redundant.

**4. Mass-lookup duplication** — `RecoDataProducts/src/KalSeed.cc`

The identical 4-line mass-lookup block is copy-pasted into all three `*FitTrajectory` methods. Minor — a small private helper would reduce duplication and the risk of them diverging.

bonventre (approved):

brownd1978:
Responding to AI comments:
1: This was an intentional opportunistic change to remove deprecated code related to BTrk.
2: This is intentional. the AI doesn't realize that our inputs contain a mix of positive and negative charged particles
3: again intentional to provide new functionality
4: Duplication in general will be addressed in a future PR

brownd1978:
@FNALbuild run build test

PR #1879, opened 2026-07-10
RMC phase-space approximation

michaelmackenzie:
Add RMC spectrum logic for the phase-space approximation. Also add the Ti mu- --> e+ endpoint and optional muon capture proton energy spectrum range while making updates.

### DIFF
--- a/GlobalConstantsService/data/globalConstants_01.txt
+++ b/GlobalConstantsService/data/globalConstants_01.txt
@@ -139,13 +139,14 @@ double physicsParams.C.capture.RMCInternalRate = 0.0069; //measurement of intern
 
 //Ce+ endpoint energy
 double physicsParams.Al.ePlusEndpointEnergy = 92.32;
+double physicsParams.Ti.ePlusEndpointEnergy = 98.89;
 
 // parameters for RMC spectrum shape
 double physicsParams.Al.RMCbindingEnergyFit = 0.464;
 double physicsParams.Al.RMCrecoilEnergyFit  = 0.220;
 double physicsParams.Al.RMCdeltaMassFit     = 3.121;
-
-
+double physicsParams.Al.RMCKMaxKnockout.0   = 101.87; // 0 nucleon knockout
+double physicsParams.Al.RMCKMaxKnockout.1   =  95.45; // 1 neutron knockout
 
 // Shanker interpolation coefficients - Phys. Rev. D 25, 1847 (1982)
 // - The Shanker formula includes one D, E and F coefficient for a
--- a/GlobalConstantsService/inc/PhysicsParams.hh
+++ b/GlobalConstantsService/inc/PhysicsParams.hh
@@ -162,6 +162,13 @@ namespace mu2e
     double getRMCdeltaMassFit(targetMat material = "") const {
       return doubleOrThrow(_RMCdeltaMassFit,material,"RMCdeltaMassFit");
     }
+    double getRMCKMaxKnockout(targetMat material = "", int nKnockout = 0) const {
+      if(nKnockout == 0)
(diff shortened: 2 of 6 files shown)


oksuzian:
## Potential bugs / correctness

1. **Phase-space weight normalization is inconsistent with the RMC branch** — `getRMCPhaseSpaceSpectrum` returns a *normalized PDF* (includes `norm = (power+1)(power+2)/kMax`), whereas the existing `getRMCSpectrum` returns an unnormalized shape that `BinnedSpectrum` later re-normalizes. Since `BinnedSpectrum` normalizes bins anyway, the `norm` factor is harmless but redundant — worth confirming the two branches are treated consistently.

```cpp name=Mu2eUtilities/src/MuonCaptureSpectrum.cc url=https://github.com/Mu2e/Offline/blob/develop/Mu2eUtilities/src/MuonCaptureSpectrum.cc
double MuonCaptureSpectrum::getRMCPhaseSpaceSpectrum(double e, double kMax, int nKnockout) const {
  if ( e >= kMax ) return 0.;
  const double x = e/kMax;
  const double power = 2. + 1.5*nKnockout;   // nKnockout>1 never reaches here (throws earlier), but value is used unchecked
  ...
}
```

2. **`nKnockout` bounds not enforced at the spectrum layer.** `getRMCKMaxKnockout` throws for `nKnockout > 1`, but `getRMCPhaseSpaceSpectrum` accepts any `nKnockout` and silently computes a weight. If a caller reaches this without going through `getRMCKMaxKnockout` (e.g. future code), it produces an unvalidated result. Consider validating `nKnockout` inside the phase-space method too.

3. **Hard-coded `"Al"` in the phase-space branch.** In `BinnedSpectrum.cc` the new branch calls `getRMCKMaxKnockout("Al", nKnockout)` and the new constructor sets `_MN = getAtomicMass("Al")`. The PR adds a `Ti` endpoint constant, so hard-coding `Al` here undercuts the multi-material intent. The RMC-fit branch has the same pre-existing issue, but the new code perpetuates it.

## Config / robustness

4. **`nKnockout` read with no default** — `psphys.get<int>("nKnockout")` (no default) will throw if `phaseSpaceShape=true` but `nKnockout` is omitted. The sibling params (`kMaxUserSet`, `kMaxUser`) all supply defaults. Intentional (required) vs. inconsistent is worth a comment.

5. **New `Ti.ePlusEndpointEnergy = 98.89` constant is unused in this PR.** No code reads `getEPlusEndpointEnergy("Ti")` in the diff — confirm it's consumed elsewhere or intended for a follow-up, otherwise it's a dangling constant.

## Minor / style

6. **Dead comment block left in `getWeight`** — the old commented-out `if/else` spectrum dispatch above the new dispatch is now misleading and should be removed.
7. Trailing blank line added at the end of the new constructor body in `MuonCaptureSpectrum.cc`.
8. Magic number `power = 2. + 1.5*nKnockout` — a brief comment tying it to doc-db 57178's formula would help reviewers.

michaelmackenzie:
1. I added the normalization to make it easier for people wanting the spectrum PDF. I added a comment to note it's not really needed here, and also added the closure normalization.
2. N(knockouts) > 1 are correctly described, the enforcement is just that I haven't implemented > 1 in that location but the spectrum is fine.
3. I made the material configurable, but set the default to "Al" for compatibility with existing fcl.
4. There is no "default" value of N(knockout), it must be specified and it's an error if it's not.
5. This can be used in the conversion event generator.
6. I removed old comments
7. Files are meant to have these.
8. I added a comment, but it's a formula not a magic number, the comment is just the code, power = 2 + 1.5*N(knockout) from phase space.

Let me know if there are any other changes requested!

michaelmackenzie:
@FNALbuild run build test

PR #1885, opened 2026-07-13
extract CID tables from run_info blobs

rlcee:
This PR introduces a function to extract the CID (category 2) table information from the run_info database settings blobs.  It follows the same pattern as cat3, walking the json tree to find a specific key and extracting the values.  It prints to the command line from runTool or provides a json text structure to the python interface in the pywrap build.  In the process, I found some run_info text fields contain newlines, which was not coded, so a couple of places had to be adjusted.  The adjusted code is run in several common operations.

### DIFF
--- a/DbService/inc/RunConfig.hh
+++ b/DbService/inc/RunConfig.hh
@@ -34,6 +34,13 @@ class RunConfig {
   // If json=false -> returns one VALUE per line (flat list).
   std::string dbTables3(bool qjson = false) const;
 
+  // Walk the entire JSON tree in _settings, collect all (cid, name) pairs
+  // from every "DBServiceCIDTable" array found at any depth.
+  // If json=true  -> returns a JSON list of {"cid":..., "name":...} objects.
+  // If json=false -> returns one "cid name" pair per line.
+  std::string dbTables2(bool qjson = false) const;
+
+
  private:
   int _run_number;
   std::string _subsystem;
--- a/DbService/inc/RunConfig.hh
+++ b/DbService/inc/RunConfig.hh
@@ -1,7 +1,7 @@
 #ifndef DbService_RunConfig_hh
 #define DbService_RunConfig_hh
 
-// holds info for one row in the config table
+// holds info for one row in the run_info config table
 
 #include <string>
 #include <vector>
@@ -39,7 +39,6 @@ class RunConfig {
   // If json=true  -> returns a JSON list of {"cid":..., "name":...} objects.
   // If json=false -> returns one "cid name" pair per line.
   std::string dbTables2(bool qjson = false) const;
-
 
  private:
   int _run_number;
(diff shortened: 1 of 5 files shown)


oksuzian:
## Issues found

**1. `splitCsvLines` now throws on the previously‑normal empty terminal line** — `DbTables/src/DbUtil.cc`
The old code did `lines.pop_back()` to drop the blank line produced by the terminal `\n`. The new loop pushes a line only on an *unquoted* `\n`, so a well‑formed CSV ending in `\n` leaves `current_line` empty and correctly returns — **but** the `throw` fires whenever the final line is non-empty (no terminal newline), which was previously allowed to slip through only via the pop. More importantly, the semantics inverted: the check now guards "no terminal newline," but a file with a genuinely non-empty last line + missing newline throws instead of returning the data. Confirm every caller always supplies a terminal newline.

**2. Blank/malformed lines are silently kept** — `DbUtil.cc`
The loop no longer compresses tokens or trims, so multiple consecutive `\n` yield empty entries in `lines`. Downstream `splitCsv`/`std::stoi(sv[0])` in `RunTool.cc` will throw on an empty row. Consider skipping empty `current_line` pushes.

**3. `std::stoi(sv[0])` unguarded** — `DbService/src/RunTool.cc` (line ~134)
With the new tolerant line splitter, a malformed/blank row reaches `std::stoi` and throws `std::invalid_argument`. Worth a guard.

**4. Unvalidated `sv` indexing** — `RunTool.cc`
`sv[0]`/`sv[1]` are accessed without a size check; a short row will be UB / out-of-range.

**5. Minor: cosmetic** — several files have stray double blank lines added (`RunConfig.hh`, `RunConfig.cc`, `runTool_main.cc`). Harmless but noise.

**6. Robustness in `dbTables2`** — `RunConfig.cc`
Good: uses non-throwing `json::parse` and checks `is_discarded()`. `valueToString` is assumed to handle non-string `cid`/`name` — verify it exists and is safe (reused from `dbTables3`).

rlcee:
I committed some tweaks based on AI comments.  It looks like you are using a different model since the style is different.  The comments also seem weaker.
1) this analysis is incorrect.  The old code did not trim or remove empty lines.  This routine just splits and those issues are handled elsewhere.  The old and new code require the last char in the string to be a newline, but the way that it appears is confusingly different.
2) 3) 4) should never happen, but added an explicit check and throw, which is better
5) cleaned up some comments
6) AI affirming AI

oksuzian:
> I committed some tweaks based on AI comments. It looks like you are using a different model since the style is different. The comments also seem weaker.
> 
> 1. this analysis is incorrect.  The old code did not trim or remove empty lines.  This routine just splits and those issues are handled elsewhere.  The old and new code require the last char in the string to be a newline, but the way that it appears is confusingly different.
> 2. 3. 4. should never happen, but added an explicit check and throw, which is better
> 3. cleaned up some comments
> 4. AI affirming AI

I always use Opus for GitHub reviews

PR #1886, opened 2026-07-13
Box-cut photon filter with helix veto

michaelmackenzie:
This is the Offline PR https://github.com/Mu2e/Offline/pull/1860 but with the BTrk include removed as well as some comments and a CMake change. This is a first step to this helix-veto photon filter, where future work will separate out the histogramming/MC tools and develop the corresponding photon trigger path.

### DIFF
--- a/CaloMC/src/CaloHitTruthMatch_module.cc
+++ b/CaloMC/src/CaloHitTruthMatch_module.cc
@@ -216,10 +216,14 @@ namespace mu2e {
                                        <<" / "<<hit.time()<<" / "<<hit.energyDep()<<" "<<nbin*digiSampling_<<" "<<deltaTimePlus<<std::endl;
 
 
+          // FIXME
+          // deltaTimePlus = 200.;
+          double deltaTimeMin = 100.; // 2*digiSampling_
+
           //forward until we reach the recoHit time, then loop as long as the shower time is within the recoTime window
           bool hitIsMatched(false);
           std::vector<CaloEDepMC> edeps;
-          while (showerIt != showerItEnd && ( (*showerIt)->time() < (*hitIt)->time() - 2*digiSampling_) ) ++showerIt;
+          while (showerIt != showerItEnd && ( (*showerIt)->time() < (*hitIt)->time() - deltaTimeMin ) ) ++showerIt;
           while (showerIt != showerItEnd && ( (*showerIt)->time() < (*hitIt)->time() + deltaTimePlus) )
           {
               hitIsMatched = true;
--- a/Filters/CMakeLists.txt
+++ b/Filters/CMakeLists.txt
@@ -302,7 +302,7 @@ cet_build_plugin(PhotonFilter art::module
       Offline::RecoDataProducts
       art_root_io::TFileService_service
       ROOT::Hist
-)   
+)
 
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/fcl/prolog.fcl ${CURRENT_BINARY_DIR} fcl/prolog.fcl)
 
--- a/Filters/CMakeLists.txt
+++ b/Filters/CMakeLists.txt
@@ -290,6 +290,17 @@ cet_build_plugin(WeightSamplingFilter art::module
       art_root_io::TFileService_service
 )
 
+cet_build_plugin(PhotonFilter art::module
+    REG_SOURCE src/PhotonFilter_module.cc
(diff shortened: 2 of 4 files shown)


oksuzian:
I'll review this PR to find issues. Let me load the necessary abilities first.The user wants to find issues in this PR (a code review). Let me look at the changed files.Here are the concrete issues I found in PR #1886, focused on the new `Filters/src/PhotonFilter_module.cc`.

## 🔴 Likely bugs / crashes

**1. Unguarded `caloHitsPtrVector()[0]` — potential out-of-bounds crash**
Multiple `_hCalERatio` fills dereference the first calo hit without checking that the vector is non-empty:
```cpp
cluster.caloHitsPtrVector()[0]->energyDep() / clusterEDep
```
If a cluster ever has zero hits, this is undefined behavior. Add a guard (or rely on a documented invariant).

**2. Index/size mismatch between clusters and truth collection**
```cpp
for (size_t i = 0; i < clusters.size(); ++i) {
  if (reportResiduals(clusters.at(i), helices, caloClusterTruth[i])) { ... }
```
This assumes `caloClusterTruth.size() == clusters.size()` and that they are index-aligned. There is no size check — if the MC truth collection is smaller (or ordered differently), this reads out of bounds / mismatches truth. Add an explicit size assertion or use the proper association.

**3. Division by zero risk**
- `clusterEDep/clusterHits` — if `clusterHits == 0`.
- `e2/e9/e25 = clu.e2()/clusterEDep` and `clusterEDep/bestHelixMom` — if `clusterEDep == 0` or `bestHelixMom == 0` (the latter is only filled when `bestHelixExists`, but `bestHelixMom` could still be near zero). Consider guards.

## 🟠 Design / correctness concerns

**4. MC dependency in a filter meant to become a trigger path**
The module hard-requires MC products (`CaloClusterMC`, `PrimaryParticle`, `SimParticle`) via `getValidHandle`, which **throws** if absent. This filter cannot run on data. The PR body itself says MC/histogramming should be split out later — worth flagging that as required before this can be a trigger filter.

**5. `MinDeltaT` / `MaxCT` / `MinCT` / `MinCR` defaults differ between code and fcl**
`prolog.fcl` only sets a subset of parameters (e.g. `MinDeltaR`, `MaxNCrys`, `MinCalE`...) and omits others like `MinDeltaT`, `MaxCT`, `PrimaryParticle`. `PrimaryParticle` has **no default** in the Config and is **not** in the prolog — any job using the prolog block will fail to configure. This is likely a build/run blocker (note the `build running` label).

**6. Magic numbers**
`if (hval.momentum()*0.3 < 90) continue;` and repeated `*0.3` (B-field → momentum conversion) are unexplained hardcoded constants. Should be named constants or fcl parameters.

## 🟡 Minor / cleanup

- Large amount of **duplicated histogram-fill code** across the `index`, `indexAccept`, and `indexPdg` blocks — should be factored into a helper.
- Fixed-size arrays `hist_c cluster_hists[100]` with only ~9 slots used (indices 0,1,2,10,11,12,20,21,22) — sparse and error-prone.
- Trailing whitespace in `CMakeLists.txt` (`)   `).
- Unused/under-used config atoms (`MinCHERatio`, `MinCalE9Ratio`, etc.) are declared but not in the prolog.
- `_hZRes_CloseHelix`, `_hZcoord_calo`, `_hHelixChi2_dXY` declared in the struct but some are never booked/filled → dangling `TH1F*` pointers.

---

The most important to fix before merge: **#5 (missing `PrimaryParticle` in prolog → config failure)**, **#2 (truth index alignment)**, and **#1/#3 (out-of-bounds / div-by-zero guards)**.

michaelmackenzie:
Responding to AI comments:
1. I don't think know if we should be assuming calo clusters could have 0 hits, or throwing errors on this in every module that consumes calo clusters
2. This is the model currently used in Offline, index-aligned lists. I switched to `vector::at` access so an error would be thrown if it goes out of bounds.
3.  Same response as 1.
4. This is temporary, meant to save student work since they have graduated and a few student will take it over.
5. The code defaults are meant to be very inclusive or not cut at all, the fcl ones are a starting point for a trigger selection.
6. See 4.

This code will be updated and cleaned in the future, but it would be nice to merge it so it's not lost. It's not currently used in the Trigger so it should be safe to merge before future improvements.

matthewstortini (approved):

oksuzian:
@FNALbuild run build test

PR #1888, opened 2026-07-15
Updated CRV channel map

ehrlich-uva:
Updated CRV channel map
- ROC2/FEB5 --> ROC1/FEB2
- ROC2/FEB6 --> ROC2/FEB5
- ROC2/FEB7 --> ROC2/FEB6

Updated the SiPM calibration modules and TimeOffset calibration module.


oksuzian:
### PR Review Summary — "Updated CRV channel map"

Reviewed at head `7217a5f1c`, three-dot against merge-base `704459ce`. (Note the API's `base.sha` is main's current tip, not the branch point — main has moved 20 commits since, though none touch these files.)

**The channel map itself is clean and I could not fault it** — see "Verified correct" below. The module split is also the right call: the old `TABLE CRVTime` blocks in `CrvCalibration`/`CrvPedestalFinder` only ever *echoed* the DB back at itself (`_timeOffsets[i] = calib.timeOffset(i)` → written straight back out), so two jobs each emitted a round-trip of their own input and whichever ingested last won. Replacing that with one module that genuinely measures is a real improvement.

What blocks is the new module's output. All findings below are reproduced against the real files.

**Decision**
- Request changes.

**Scope understood**
- Three FEB remaps in `CRVConditions/data/extracted_v04.txt` (hardware truth — not something I can second-guess).
- Time-offset table writing removed from `CrvCalibration` + `CrvPedestalFinder`; `CrvTimingStudies` deleted, `CrvTimeOffsets` added as the sole `CRVTime` producer.

---

### Findings

**1. [S1] `CrvTimeOffsets` writes a sparse `TABLE CRVTime`; ingestion throws deterministically**

`CrvTimeOffsets_module.cc:394` builds `_channels` with `if(!crvChannelMap.onlineExists(offlineChannel)) continue;`, and the writer at `:362` iterates only `_channels`. That map is sparse by design (`CRVOrdinalMaker.cc:31`: *"both are sparse, so this is used to catch invalid access"*). For the geometry this PR targets, `extracted_v04.txt` maps **1920** channels out of **2304**, with the first gap at channel **1024**.

`DbTables/inc/CRVTime.hh:45` requires dense, ordered rows — *"enforce order, so channels can be looked up by index"*:

```cpp
if (channel >= CRVId::nChannels || channel != _rows.size()) {
  throw cet::exception("CRVTIME_BAD_CHANNEL") << "CRVTime::addRow bad channel, saw " ...
```

Since `_channels` is a `std::map` (`:106`) it emits ascending, so ingestion accepts rows 0..1023 and then throws:

```
saw channel 1025, expected 1024  ->  CRVTIME_BAD_CHANNEL
```

And even past that, `CRVCalibMaker.cc:60` throws `CRVCALIBMAKE_BAD_N_CHANNEL` on `1920 != 2304`. That is the same failure class as the Extracted nightly breakage in #1864 (2048→2304 channels).

This is introduced, not inherited: the removed blocks sized `_timeOffsets.resize(counters.size()*CRVId::nChanPerBar)` and looped `0..size()` — **dense to full geometry**. The deleted `CrvTimingStudies` wrote no file at all, so this writer is new code.

*Suggested fix:* iterate `0..nChan-1` and emit `0` for unmapped channels, mirroring the removed dense loop. The `,0` fallback at `:370` already exists for the found-map/no-offset case and just needs to cover the unmapped case too.

**2. [S1] An empty histogram silently becomes a 0.0 ns "measurement"**

`:132` `float mean=_h->GetMean(); //FIXME: replace by mean of gaussian fit` — that line is the *only* `GetMean`/`GetEntries` occurrence in the file. `TH1::GetMean()` returns 0 for an empty histogram, and `PreparePlots` creates a histogram for every *configured* pair whether or not data lands in it. So a dead FEB, a mistyped `firstFEB`, or a low-stats run injects an exact `0.0` that is indistinguishable from a real measurement, and the chain solver propagates it down the whole chain into the DB text file. Exit 0, no error.

The sibling module already has the right pattern — `CrvCalibration_module.cc:52` `minHistEntries{..., "minimum number of entries required for a fit", 100}` and `:223` `if(hist->GetEntries()<_minHistEntries) return false; //not enough data`.

*Suggested fix:* gate on a minimum entry count and drop the pair from `measuredTimeDiffs`, so the solver reports the channel unreachable rather than mis-solving it. (`TF1.h`/`TFitResult.h` are included but nothing is fitted — the FIXME is still open, and its output now feeds committed constants.)

**3. [S1] An empty `specialTimeOffsets` yields a full all-zeros table, silently**

`specialTimeOffsets` is required but may legally be `[]`. Then the seeding loop at `:267-270` never runs, `timeOffsets` stays empty, nothing is solved, `:342` prints a `cout` warning, and `:370` writes `<channel>,0` for **every** channel. Same outcome if no event carried the collection. The job exits 0 with a syntactically valid file that would clobber the real calibration on ingest.

*Suggested fix:* throw if `timeOffsets` is empty entering `endJob`, and refuse to write the file if any channel is unresolved rather than emitting 0.

Findings 2 and 3 share one root cause: **the module has no way to say "I don't have enough data to measure this."** Every configured pair produces a number, `0.0` is a legal-looking number, and no path fails the job or withholds the file. Because the artifact is a DB text file, the failure mode is a silent bad-calibration ingest rather than a crash.

**4. [S1, conditional — a question, only you can answer] In-place edit of a versioned map**

`crv_counters_extracted_v01/v02/v03` all set `crs.name = "extracted"` and share `extracted.txt`; v04 minted a new name *and* a new file on 2026-06-24 (`059242e52`, "added muon taggers to CRV KPP setup") precisely because the map changed. This PR changes the map again but keeps the name — same filename, different content.

That matters because selection is by geometry name only (`CRVOrdinalMaker.cc:44`: `filePath + "/" + GeomHandle<CosmicRayShield>()->getName() + ".txt"`), and `:101` says *"no database dependence (i.e. cable swaps) yet, so just return nominal"* — this PR **is** a cable swap, the exact case that comment defers. There is no run-dependent map, so old-cabling data decoded with the new map lands hits on wrong offline channels, silently.

**Was any real data taken, or MC produced, against v04 as it stood since 2026-06-24?** If no, an in-place edit is correct and cheap — just say so in the PR body so the record shows it was considered. If yes, this wants `extracted_v05` (+ `crv_counters_extracted_v05.txt`, `geom_common_extracted_v05.txt`, repointing `geom_common_extracted.txt`).

**5. [S2] `extracted.txt` encodes the identical cabling and was not remapped**

`extracted.txt` (1792 rows, geometry name `extracted`, still declared by `crv_counters_extracted_v01/v02/v03`) is a strict subset of old-v04: it agrees with **old-v04 on 1792/1792** channels and disagrees with new-v04 on **exactly the 64 ROC2/FEB5 channels**. It carries the same ROC1-missing-FEB2 hole.

So: if this PR **corrects a mis-documented map**, `extracted.txt` has the identical error for the identical physical FEB and needs the same 64-row remap. If it **records a physical re-cabling**, `extracted.txt` correctly describes the earlier state and must be left alone. Which is it?

**6. [S2] Guard dropped that the predecessor had**

`:357` `std::ofstream calibFile(_calibFileName);` is unchecked. Merge-base `CrvCalibration_module.cc:160` did `if(!outputFile.is_open()) throw cet::exception("CRVCALIB") << "Couldn't create output txt file " ...`. A bad path now means silent no-output after a full calibration run.

**7. [S2] `CrvPedestalFinder` keeps a now-pointless `CRVCalib` fetch that can throw**

`_calib_h` is read at exactly one place (`:161`), whose only purpose is filling `_timeOffsets` (`:164`) — which is now never read. The whole first-event block (`:157-166`) is dead. This matters because the pedestal finder is the *bootstrap* job, run before CRV calibration exists, yet it still pulls `CRVCalib` proditions through `CRVCalibMaker::fromDb`, which can throw `CRVCALIBMAKE_BAD_N_CHANNEL` on any table/geometry mismatch. The PR made the dependency pointless but left the crash risk. (In `CrvCalibration`, `_calib_h` stays legitimately live for `_pedestals`.)

*Suggested fix:* drop `:157-166`, the `_timeOffsets` member, `ProditionsHandle<CRVCalib> _calib_h`, and the `CRVCalib.hh`/`ProditionsHandle.hh` includes from `CrvPedestalFinder`.

**8. [S2] Outputs dropped without a replacement**

- Both removed blocks made a `crvTimeOffsets` TTree (`channel`/`timeOffset`); the new module creates **no TTree at all**. Anything reading that tree from the calibration ROOT file now finds nothing.
- `CrvTimingStudies` had `treeChannelMap` (channel/roc/feb/febChannel) and `treeCalib`, with the comment *"store channel map, pedestals and calibration constants in the file, so that it can later be used to write a full calibration set"*. Both are gone. In a PR titled "Updated CRV channel map", losing the channel-map dump is worth an explicit ack.
- Workflow: `CRVTime` used to be appended to the same `tmpDBfileName` as `TABLE CRVSiPM` (`prolog_v12.fcl:21` still says `calibration.txt`); it is now a standalone file. Splitting is *not* itself a break — `DbUtil.cc:65` starts a new table on any `TABLE` line — but any operator procedure ingesting one combined file per run now needs to handle two. Worth confirming the calibration doc is updated.

**9. [S2] Green CI does not validate this change**

`mu2e/buildtest` is genuinely `success` on the head (credit where due — all four commits predate the PR open, so the run covers `7217a5f1c`; the `buildtest/last` "triggered against dde444ee" string names the merge target, not the tested code). But the PR suite is ceSimReco / cosmicSimReco / ceDigi / ceMix / ceSteps / POT / muDauSteps / cosmicOffSpill / g4study / g4test_03MT / transportOnly — **none uses extracted geometry**. Extracted coverage lives in `Production/Validation/nightly/extracted_0{0..4}.fcl`, which is nightly, not PR-gated. So the map change and the new module compile-pass but are behaviourally untested until after merge.

Blast radius is real and not data-only: `Production` main (`062945c1`) `JobConfig/digitize/Extracted.fcl`, `reco/Extracted.fcl`, `recoMC/Extracted.fcl` and `cosmic/geom_cosmic_extracted.txt` all resolve to `geom_common_extracted.txt` → `extracted_v04`, and the sim consumers `CrvWaveformsGenerator_module.cc` / `CrvDigitizer_module.cc` read `CRVOrdinal`. (`recoMC/NoFieldRun1B.fcl` uses `geom_run1_b_v01.txt` → different `crs.name`, so the Run1B field-off chain is untouched.)

*Ask:* could you post output from an extracted job on this branch — e.g. `Production/Validation/nightly/extracted_00.fcl`, or `CRVReco/test/printChannelMap.fcl`?

**10. [S2] Histogram names contain `/`, plus a typo**

`:123` `Form("TimeDiffs_FEB/FGPA_%i/%i_%i/%i",...)`. ROOT creates the object, but `tfile->Get("TimeDiffs_FEB/FGPA_1/0_2/1")` parses the slashes as directory navigation and fails. Every sibling CRV histogram uses `_` (`crvPedestalHist_%lu`, `crvCalibrationHistPulseArea_%lu`), and the predecessor used `fpgaTimeDiff_%i_%i`. Also **`FGPA`** → `FPGA` (the title on `:124` spells it correctly).

**11. [S2] Unprotected `std::cout`/`std::cerr`, no verbosity gate**

`:220`, `:342-346`, `:347-354` (one line per FPGA, unconditional), `:369`. The Config has no `diagLevel`/`verbosity` Atom. Note `:369`'s `cerr` reports a condition that *corrupts the output file* — that should be an exception, not a stderr line. House rule is verbosity-gated or MessageFacility.

**12. [S3] Assorted**

- No fcl configures `CrvTimeOffsets` anywhere, and all 12 Atoms are required with no defaults — so running it means hand-writing the whole CRV FEB/module topology. The Config comments (`//extracted: 28,29`, `//extracted: "EX_0","T2_0"`) show you have the extracted values; they belong in a prolog entry. (Fair caveat: `CrvTimingStudies` had no fcl either — but it took 3 simple Atoms, not a detector topology, and it wasn't the sole source of a required DB table.)
- Dead `_timeOffsets` member in **both** modified modules (declared/resized/zeroed/assigned, never read). No `-Werror` risk — I checked rather than assumed; members never trigger unused-variable warnings (`-Wunused-private-field` is Clang-only).
- `CrvPedestalFinder`'s `<TTree.h>` (`:29`) is now dead — its sole `tfs->make<TTree>` was in the removed block — so `ROOT::Tree` (`CRVReco/CMakeLists.txt:46`) can go with it. Runtime impact nil (`-Wl,--as-needed` already drops libTree). `CrvCalibration` must keep both — it still makes `crvPedestals` at `:149`. Note `CrvPedestalFinder` does still use TFileService elsewhere (`:96`/`:103`), so that dep stays.
- Chain solver `:317-336` discards loop closures rather than using them as a residual check — a cheap, valuable QA number for a calibration.
- `:231` section-3 neighbor pairs build `histIndex` in config order while `analyze` only fills `fpga1->first < fpga2->first` (`:450`); a neighbor with a lower `firstFEB` never fills → (per finding 2) a silent 0. Normalize the pair or assert ordering.
- Hardcoded `*4` at `:121`,`:133`,`:269` where `CRVId::nFPGAPerFEB` is used elsewhere; `CRVId::nChanPerFPGA` already exists but is recomputed at `:365`/`:424`.
- `measuredTimeDiffs`/`timeOffsets` are `float` (`:264-265`) while `GetMean()` returns `double` and the consumer (`calib.timeOffset()`) is `double`; the solver accumulates along a chain, which is exactly where float hurts. Not limiting at O(100) ns, but free to fix.
- `:225` emplaces a PDF page unconditionally while plots are only added inside `if(!moduleConfig.muonTaggers())` → a blank page (ROOT clamps `Divide(0,1)`, so no crash).
- Stale comment `:319-321` ("That's why a negative sign is used below") — the first branch uses `+`. The algebra is correct; only the comment misleads.
- `for(auto moduleConfig : _moduleConfigs)` (`:189`,`:212`,`:242`) copies a 5-Atom fhicl struct per iteration → `const auto&`. Dead includes copied from the predecessor (`TMath.h`, `TNtuple.h`, `CrvStep.hh`, `ProtonBunchTime.hh`, `CLHEP/Units/GlobalSystemOfUnits.h` — no CLHEP unit is used). Empty `beginJob()`/`~CrvTimeOffsets()`.

---

### Verified correct — no action needed

- **The channel map is exactly what the PR body says.** 1920 rows, header unchanged, offline `Channel` column byte-identical. Exactly **192** changed rows = 3 × 64, forming precisely the three claimed remaps and nothing else, with `FEBchannel` preserved as an identity within each. **`(ROC,FEB,FEBchannel)` is unique across all 1920 rows before and after — zero aliasing**, which was the real corruption risk. No collision: the old map had no ROC1/FEB2 at all. Inventory goes `(2,7)` out / `(1,2)` in, still 30 FEBs × 64. The result makes ROC1 dense `1..24` — filling a lone hole in an otherwise-full ROC, which reads as a genuine fix rather than a cosmetic edit.
- No bound is invalidated: ROC ≤ 2 (`nROC=18`), FEB ≤ 24 (`nFEBPerROC=24`, and `OfflineMap`'s FEB dimension is `nFEBPerROC+1`, so index 24 is valid), FEBchannel ≤ 63 (`nChanPerFEB=64`). Same maxima as the old map.
- Sparse `Channel` numbering (0..2302, 383 gaps) is by design, not a defect. Offline-indexed siblings (`calib_*`, `photonyield_*`, `status_*`) stay structurally valid since the offline key set is untouched — none *need* an edit.
- **Zero dangling `CrvTimingStudies` references** anywhere in the tree (the grep is non-vacuous: the same pattern returns 12 hits at merge-base). **No Production PR is needed, and the deletion is safe precisely because its only consumers are already gone** — `Pass1/Wideband/CrvPass1_wideband2modules.fcl` and `CrvReco_wideband2modules.fcl` were the only things that ever set `module_type: CrvTimingStudies`, and `Pass1/Wideband/` no longer exists on `Production` main (only `Pass1/Calo` remains). `mu2e-trig-config` main: zero hits. CMake rename is correct, and `CRVReco/src/SConscript` globs `*_module.cc` so no SConscript edit is needed. `DEFINE_ART_MODULE(CrvTimeOffsets)` is correct: `:464` is `using mu2e::CrvTimeOffsets;`, matching the CRVReco house style (7 of 9 siblings).
- **The `LIBRARIES REG` list is adequate despite listing no `ROOT::` targets** — flagging this pre-emptively since it looks like a link error. The new module ODR-uses `TCanvas`/`TText`/`gROOT` (`:272`, `:278-281`) and `Offline::CRVReco → ROOT::Hist` does *not* back-pull Gpad/Graf, but `art_root_io::TFileService_service` (already listed, `CMakeLists.txt:92`) supplies both transitively. Confirmed against linker ground truth, not inference: `Analyses/InteractiveRoot` links only `TFileService_service` + `Offline::MCDataProducts`, uses `tfs->make<TCanvas>`, and its `.so` carries `NEEDED libGpad.so.6.32` while `libmu2e_MCDataProducts.so` pulls no ROOT at all. No new CRVConditions link need either — `CRVOrdinal::onlineExists`/`online` and `CRVCalib::timeOffset` are inline in-class, and the old module's `.so` has no `libmu2e_CRVConditions` in `DT_NEEDED`.
- The `TABLE CRVTime` byte format is otherwise identical (same header, column order, separator, precision), and the `#channel,timeOffset` vs `#channel, timeOffset` space is irrelevant — `DbUtil.cc:62` drops `#` lines before splitting. Only the row set regressed (finding 1).
- Sign convention is self-consistent (`corrected = raw + offset`); chain-solver termination is sound; no division by zero; `ProditionsHandle` members are not cached across events and `GeomHandle` is local to `analyze`; the `fhicl::Sequence<fhicl::Table<>>` pattern matches `CrvCoincidenceFinder`. No `-Wall -Werror` break in the PR.

### Residual risk

- The cabling itself is authoritative and unreviewable from outside — findings 4 and 5 hinge on facts only you have.
- Regression signal won't appear until the nightly extracted jobs run post-merge (finding 9).

### Author follow-ups

1. Fix the dense/sparse row set (finding 1) — the new module's primary artifact can't currently be ingested for `extracted_v04`.
2. Add a minimum-entries gate and refuse to emit unresolved channels (findings 2, 3), reusing `CrvCalibration`'s `minHistEntries` pattern.
3. Was any data taken or MC produced against v04 as it stood since 2026-06-24 (finding 4)? And is this a correction or a re-cabling, i.e. does `extracted.txt` need the same remap (finding 5)?
4. Could you post output from an extracted job on this branch — no PR CI test covers extracted geometry (finding 9).
5. Ship a prolog entry for `CrvTimeOffsets` so it's runnable as-is.
6. Best practice: consider splitting the module rename/split from the channel remap, and noting in the PR body that the calibration workflow now produces `CRVTime` from a separate file.

ehrlich-uva:
Responses:
1. It thinks it will throw, but it doesn't. The check if the channel exists at line 394 makes sure of that.
2. Will now cause error messages, if the mean couldn't be found correctly. Will be improved later.
3. It will cause error messages.
4. No useful/important data was taken between the first version of the channel map extracted_v04.txt and it's updated version. The calibration tables are not impacted by this change, since they rely only on the Offline channels.
5. extracted.txt needs to stay this way for older data that was recorded with the channel map written in extracted.txt
6. Fixed.
7. Fixed.
8. The removal of the trees and the separation of both tables is correct.
9. I tested the timeOffset module with the extracted geometry.
10. I changed the name to remove the slashes and fixed the typo.
11. I added a diagLevel.
12. _timeOffsets removed, TTree removed, fixed problem with limited number of FPGA pairs, hard-coded 4s fixed.

oksuzian:
@FNALbuild run build test

PR #1890, opened 2026-07-18
Add EventHeader printers

rlcee:
Add three printers which are becoming more relevant.  This should not affect operations or validation


oksuzian:
## Issues found

**1. `EventHeaderPrinter::Print` has verbose gate inconsistent with the FCL config** — `Print/fcl/printLong.fcl`
The FCL sets `EventHeadersPrinter` and `RawEventHeaderPrinter` to `verbose : 0`, but every `Print()` method early-returns when `verbose() < 1`. So the two new printers configured at `verbose:0` will **never print anything**. Either the FCL defaults should be `1`, or the intent should be documented. Worth confirming this is deliberate.

**2. Unused `int ind` parameter** — `EventHeaderPrinter.cc`, `EventHeadersPrinter.cc`, `RawEventHeaderPrinter.cc`
```cpp
void mu2e::EventHeaderPrinter::Print(const mu2e::EventHeader& obj,
                                     int ind, std::ostream& os) {
```
`ind` is declared but never used in any of the three `Print(obj, ind, os)` overloads. Harmless but will likely trigger `-Wunused-parameter`. Consider dropping the name or `(void)ind;`.

**3. Unused includes** — all three `inc/*.hh`
`<cstring>` and `canvas/Persistency/Common/Ptr.h` are included but neither `Ptr` nor `cstring` functions are used. Copy-paste carryover; minor cleanup.

**4. `os` parameter ignored in the multi-arg overloads**
In `Print(const EventHeader& obj, int ind, std::ostream& os)` the code writes to `os` correctly, but the `Print(art::Event const&, os)` fan-out calls `Print(ah)` / `Print(ih)` **without forwarding `os`**, so a non-default stream passed to the top-level `Print` is silently dropped for the per-handle prints. Same pattern in all three files. Functionally fine with the default `std::cout`, but inconsistent.

**5. Build-system duplication risk** — `Print/src/SConscript` vs `Print/CMakeLists.txt`
CMake adds `artdaq-core-mu2e::Data` only; SConscript adds `artdaq-core-mu2e_Data`, `artdaq-core-mu2e_Data_dict`, and `artdaq-core_Data`. The dependency sets differ between the two build systems — verify both actually link `RawEventHeader` (from `artdaq-core`, which CMake does **not** list).

rlcee:
AI comments:
1) by design
2) cleaned up
3) cleaned up
4) fixed
5) cleaned up, checked
ready to merge

PR #1893, opened 2026-07-20
p103 for KinKal 3.7.0

rlcee:
create p103 envset, which only adds KinKal 3.7.0.  A quick test shows this builds with the current Offline head


brownd1978 (approved):
I've tested this gives identical results for existing production (which doesn't use the new energy loss functions).

PR #1894, opened 2026-07-20
Updates for Run 1B reco studies

michaelmackenzie:
This updates reconstruction elements, focusing on needs for Run 1B work. This includes:
- Updating TZClusterFinder to only output time clusters matched to calorimeter clusters if requested
- Adding flexibility to the calo cluster-centered line time cluster finder, adding hit energy selection options, and updating the prolog defaults
- Adding optional calorimeter cluster association to CosmicTrackSeeds
- Updating SelectReco<MC> to persist information related to time cluster inputs

These updates will allow for:
- Time cluster information access at the ntuple stage
- Proton rejection (selection) in CalLineTimeClusterFinder clusters
- Studies using the cosmic track seed to easily access the associated calorimeter cluster

I will make a corresponding PR to Production, which will include adding a proton line-finding sequence for Run 1B to both use as a veto in electron cluster searches and for high KE proton tail measurement studies. With this we can also more easily test using the TZ time cluster finder in these sequences, particularly for cosmic line finding, without the high CPU cost associated with fitting all lines in the event.

### DIFF
--- a/CommonReco/src/SelectReco_module.cc
+++ b/CommonReco/src/SelectReco_module.cc
@@ -30,6 +30,7 @@
 #include "Offline/RecoDataProducts/inc/KKLoopHelix.hh"
 #include "Offline/RecoDataProducts/inc/KKCentralHelix.hh"
 #include "Offline/RecoDataProducts/inc/KKLine.hh"
+#include "Offline/RecoDataProducts/inc/TimeCluster.hh"
 // Utilities
 #include "Offline/ProditionsService/inc/ProditionsHandle.hh"
 #include "Offline/TrackerConditions/inc/StrawResponse.hh"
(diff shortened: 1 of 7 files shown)


michaelmackenzie:
@FNALbuild run build test with https://github.com/Mu2e/Production/pull/558

michaelmackenzie:
@FNALbuild run build test with https://github.com/Mu2e/Production/pull/558

oksuzian:
S1 (blocker) — TimeClusterCollections is a required FHiCL key with no default, set by no shipped config. SelectReco/SelectRecoMC declare it as a mandatory
  fhicl::Sequence (no default), but nothing in Offline or Production main sets it. This PR's own CI shows the consequence: the Offline-alone build failed
  ceSimReco/cosmicOffSpill/ceDigi/ceMix with Missing parameters: TimeClusterCollections, and only passed once re-run paired with Production#558 — so the green status here
  is a paired build, not main+this PR. Merged alone (or ahead of #558), all reco/digi/recoMC + trigger-MC jobs die at startup. Fix: give the key an empty-sequence
  default (std::vector<art::InputTag>()) — makes it opt-in, green standalone, and covers out-of-tree configs (TrkAna/EventNtuple/user decks) that #558 doesn't patch.

  S2 — SelectRecoMC uses getValidHandle where SelectReco uses getHandle+isValid(). A configured-but-absent time-cluster collection is skipped in reco but throws in the MC
  job. Match them (getHandle+isValid() in SelectRecoMC).

  S2 — LineFinder caches art::Handle/product pointer as module members (_cccol, _cccolH). Correct today (refreshed each event) but against the
  don't-cache-handles-across-events rule and blocks thread-safety — prefer per-event locals passed into addClusterToSeed.

  S2 (confirm intent) — CalTimeClusterFinder prolog flips ComboHitCollectionLabel: makeSH → makePH and StoppingTargetRadius: 150 → 200 (plus enables the 3 keV electron
  cut). Not in the nominal RecoPath, so contained to the Run 1B line finder — but nothing in the code forces these, so please confirm they're deliberate for all users of
  that block.

  Nits: SelectReco output is keyed on InputTag::label() only (collision if two tags share a label — key on the full tag); the requireCCs erase intentionally leaves iiTC
  counts ahead of the emitted collection (worth a note); add a unit annotation to StrawEDepMax.

michaelmackenzie:
Responding to the (second) AI review:

>  - The root cause / the coupling. The key is still required-with-no-default in Offline. #558 only patches the configs it knows about — any deck outside it
  (TrkAna/EventNtuple, reprocessing, user fcl using @table::CommonReco.SelectReco) still fails at construction. My recommended empty-default in the Offline Config struct
  is strictly better and complementary: it makes #1894 mergeable standalone, keeps green CI without the pairing, and protects out-of-tree configs. That belongs in #1894
  regardless of #558.

I think making defaults in the C++ is not preferred unless it is something that the user by default shouldn't need to think about (e.g. turning on diagnostic printouts), and I think selecting the reco payload should be an explicit requirement in the module configuration. I don't see where the CommonReco table is defined in prolog? Or is this a suggestion to create?

 > - Findings 2, 3, 5 are Offline C++, so #558 structurally can't touch them: the getValidHandle asymmetry in SelectRecoMC (still stands — and now live in the Run 1B path,
  since #558 configures three real collections there, any absent one throws),

I updated to skip missing collections in both cases.

>  the LineFinder handle-caching, and the label()-keyed output (not triggered here since
  #558's three tags are distinct labels, but still latent).

This style is used in other places, I don't believe it's disallowed. I don't think we should assume identical collection tags would be passed.

>  Two things #558 clarifies in the author's favor:
 > - Finding 4 is effectively moot — #558 confirms the CalTimeClusterFinder block (makeSH→makePH, radius 200) is used only for Run 1B (ProtonCalTimeClusterFinder inherits
  it and adds a proton-like 2–5 keV window vs the block's electron-like 0–3 keV). Since that block isn't in the nominal RecoPath, the default changes only ever hit Run
  1B. The e-vs-p selection design reads as intended.
>  - But finding 6 goes from latent to live: #558 sets TZClusterFinder.requireCCs : true in Run 1B, so the iiTC count really does run ahead of the emitted collection there
  — still fine if deliberate, worth the one-line note.

These choices were intentional. We use the combo hit collection in CalTimeClusterFinder and should use the same collection for proton time clusters. The overlapping straw hit edep ranges were also intentional to preserve hit efficiencies in the two cases. And it's intentional to not change the time cluster counting in TZClusterFinder that would be used for the luminosity stream, but reduce the collection output to the next reco stage if requested.

oksuzian:
@FNALbuild run build test

michaelmackenzie:
@FNALbuild run build test with https://github.com/Mu2e/Production/pull/558

CommonReco/src/SelectReco_module.cc
```diff
@@ -144,10 +157,35 @@ namespace mu2e {
...
+      if (tcch.isValid()) {
+        auto const& tcc = *tcch;
+        stcc->reserve(tcc.size());
+        for (auto const& tc : tcc) {
+          for (auto const& hitIndex : tc.hits()) {
+            auto const ich = static_cast<size_t>(hitIndex);
+            (void)chc.at(ich); // bounds check vs input ComboHitCollection
+            sdindices.insert(static_cast<StrawDigiIndex>(ich));
+          }
+          if (tc.hasCaloCluster()) {
```
brownd1978, line 180: I suggest making the CaloCluster selection optional; it won't be useful for cosmic muons for instance. A more general solution would allow choosing a TC selector, but that's outside the scope of this PR.

brownd1978 (commented):
I will add Richie as reviewer for the LineFinder changes, as those affect field-off cosmics as well.

CommonReco/src/SelectReco_module.cc
michaelmackenzie: Would you also want the calo cluster inclusion to be optional for tracks (and helices, if we add them), or is this primarily a concern for time clusters? I'll start with just making the time cluster calo cluster optional. And that's an interesting point about including a generic time cluster selector, though I agree this would probably come in a separate PR

brownd1978: I misread the code: I thought it was requiring a CaloCluster to insert the tc. I see now it's saving all the tcs, and all the clusters associated with tcs. So my suggestion isn't meaningful.

brownd1978 (approved):
This looks fine as-is.
Can you say how much adding time clusters to the output increases the payload size? Also, should time clusters be added to other production stream outputs, or just run1b? I'm thinking they might also be useful for cosmic ray running.

bonventre (approved):

PR #1895, opened 2026-07-22
Add histograms for extrapolation

brownd1978:
Validate KinKal extrapolation to IPA and ST

### DIFF
--- a/Validation/inc/ValKalSeed.hh
+++ b/Validation/inc/ValKalSeed.hh
@@ -66,6 +66,8 @@ class ValKalSeed {
   TH1D* _hHPanel;
   TH1D* _hSRadLen;
   TH1D* _hSRadLenSum;
+  TH1D* _hSTdP;
+  TH1D* _hIPAdP;
 };
 }  // namespace mu2e
 #endif
--- a/Validation/inc/ValKalSeed.hh
+++ b/Validation/inc/ValKalSeed.hh
@@ -66,7 +66,9 @@ class ValKalSeed {
   TH1D* _hHPanel;
   TH1D* _hSRadLen;
   TH1D* _hSRadLenSum;
+  TH1D* _hNST;
   TH1D* _hSTdP;
+  TH1D* _hNIPA;
   TH1D* _hIPAdP;
 };
 }  // namespace mu2e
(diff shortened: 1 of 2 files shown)


brownd1978:
@FNALbuild run build test

rlcee (approved):
One note, usually these would be coordinated
_hSTdP =   ... "hSTdE",
two places

oksuzian:
### 1. Histogram name/variable mismatch (likely a copy‑paste bug)
The C++ member variables use `dP` (momentum), but the ROOT histogram names use `dE` (energy):

```cpp name=Validation/src/ValKalSeed.cc url=https://github.com/Mu2e/Offline/blob/main/Validation/src/ValKalSeed.cc#L65-L68
_hNST = tfs.make<TH1D>( "hNST", "Number of ST Foil intersections", 20, -0.5, 19.5);
_hSTdP = tfs.make<TH1D>( "hSTdE", "Momentum Change Crossing a ST Foil;MeV", 100, -1.0, 0.0);
_hNIPA = tfs.make<TH1D>( "hNIPA", "Number of IPA intersections", 20, -0.5, 19.5);
_hIPAdP = tfs.make<TH1D>( "hIPAdE", "Momentum Change Crossing the IPA;MeV", 100, -0.5, 0.0);
```

The variable is `_hSTdP` / `_hIPAdP` but the histogram name is `"hSTdE"` / `"hIPAdE"`. The title also says "Momentum Change" and the filled value is `dMom()`, so the histogram name should be `hSTdP` / `hIPAdP` to avoid confusion in downstream validation.

### 2. Inconsistent histogram naming convention
The new histograms use a lowercase `h` prefix (`"hNST"`, `"hSTdE"`, …), while every other histogram in this file uses no prefix (`"HPanel"`, `"SRadLen"`, `"SRadLenSum"`). Recommend dropping the `h` prefix for consistency (e.g. `"NST"`, `"STdP"`, `"NIPA"`, `"IPAdP"`).

### 3. Minor style nits
- Missing braces / space on the single-line loops: `for (auto stinter : stinters)_hSTdP->Fill(...)` — a space and/or braces would match the surrounding style.
- The `SurfaceId("ST_Foils",-1)` "match all foils" magic value is fine but a brief comment on why `-1` is used (already partially there) is good; consider confirming `"IPA"` is the correct SurfaceId string.

brownd1978:
@FNALbuild run build test

PR #1896, opened 2026-07-22
put db verify response in result

rlcee:
switch from std::cout to the result string so easier to use python interface.  Do not report failure return code unless answer could not be determined.

### DIFF
--- a/DbService/src/DbTool.cc
+++ b/DbService/src/DbTool.cc
@@ -2247,7 +2247,7 @@ int mu2e::DbTool::commitPatch() {
 
     // check that this group does not include any tables to be dropped
     for (int t : droptids) {
-      if (omap.find(t) == omap.end()) {
+      if (omap.find(t) != omap.end()) {
         omap.erase(t);
         remakegroup = true;
         if (_verbose > 1) {
--- a/DbService/src/DbTool.cc
+++ b/DbService/src/DbTool.cc
@@ -2518,7 +2518,7 @@ int mu2e::DbTool::verifySet() {
     if (iovv.find(t) == iovv.end()) {
       std::stringstream ss;
         ss << "Error - TID " << t
-                << " not found in the set near the requested runs\n";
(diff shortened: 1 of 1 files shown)


oksuzian:
I'll review this PR for issues.Let me see more context around the changed function.Here are the issues I found in PR #1896:

## Bugs

**1. Bug in the missing-tables loop (lines 2517–2524) — inconsistent indentation but also a latent logic point.** The indentation of `ss <<` (line 2520) is misaligned, but functionally fine. Not a real bug, just style.

**2. Behavior change: `verifySet` now always returns 0 (line 2573).** This is the intended change per the PR description ("Do not report failure return code unless answer could not be determined"). But note the *early* error returns above still return `1` (e.g. missing purpose/version/run, or `findPidVid` failure at line 2476). So the return-code contract is now: `0` when the check completed (even if incomplete), non-zero only when it couldn't be determined. That's consistent with the description — good. However, callers/scripts that previously relied on `rc=1` to detect an *incomplete set* will silently break. The real signal now lives only in `_result` text and the `nmiss`/`nbad` counts, which are not machine-readable. Consider emitting a structured indicator (e.g. a final `MISSING n` / `OK` token) so the Python interface can detect failure without parsing prose.

**3. Pre-existing bug still present at line 2263 (in `commitPatch`, unrelated to this PR):**
```cpp
if (omap.find(t) == omap.end()) {
  omap.erase(t);   // erasing a key that was just confirmed absent — no-op
```
The condition is inverted — it erases only when the table is *not* found. Out of scope for this PR but worth flagging.

rlcee:
response to AI:
1) just style, but fixed
2) double-checked, and Richie knows this change is coming
3) I need to look at this more closely, it probably is a bug in a dusty corner of another routine.  No need to tie to this PR
Ready to merge

rlcee:
Also fixed AI comment 3, it was clearly just the reversed test

PR #1897, opened 2026-07-22
Turn off path length scaling for KKShellXing by default

RobMina:
Allow to turn it on for each intersection type separately.


oksuzian:
### 🔴 `BetheCorrectionPassive` is dead — no consumer
`KKMaterial.hh` adds `applyBetheCorrectionPassive()` and the fcl adds `BetheCorrectionPassive`, but nothing calls it. `KKExtrap.hh` only wires IPA, ST, and CRV — there is no passive-material (DS cylinders / concrete) Xing that uses it. So the flag advertised in the fcl comment ("DS cylinders, concrete/strongback planes") has no effect. Either remove it or add the passive-crossing wiring (there's a `// TODO add DS and shielding material` at line 44 of `KKExtrap.hh`).

PR #1898, opened 2026-07-24
Event generator energy fraction for proton tail

michaelmackenzie:
This adds the energy fraction calculation for the muon capture proton spectrum, as well as updates this fraction evaluation. The missing tail fraction correction is now off by default and it relies on a specific spectrum configuration to know that the full input spectrum is truncated (e.g. DIO or Mu2eX). I also found the neutron tabulated spectrum was configured incorrectly in Production (table values are bin centers, not edges) which was the source of the non-unit fraction previously fixed. I think this is a less fragile implementation, where different generators with different truncated spectra should be responsible for accounting for more complicated corrections for this.

### DIFF
--- a/EventGenerator/inc/ParticleGeneratorTool.hh
+++ b/EventGenerator/inc/ParticleGeneratorTool.hh
@@ -46,18 +46,19 @@ namespace mu2e {
     virtual ~ParticleGeneratorTool() noexcept = default;
 
     double calculateBinnedSpectrumEnergyFraction(fhicl::ParameterSet pset,
-                                                 const bool correct_full_integral = true, // correct for missing low tail to full integral
                                                  std::string var_low = "elow", // default to energy spectrum variables
                                                  std::string var_high = "ehi",
-                                                 double full_var_low = 0., // if elow == ehi, it does the full spectrum
-                                                 double full_var_high = 0.) const {
+                                                 const bool correct_full_integral = false, // correct for missing low tail to full integral
+                                                 const double full_var_low = 0.) const {
+
+      // No clear fraction for a flat spectrum
+      const bool flat = pset.get<std::string>("spectrumShape", "") == "flat";
+      if(flat) return 1.;
 
       // Initialize the spectra with and without the (possible) energy restriction
       BinnedSpectrum spectrum(pset);
-      pset.erase(var_low);
+      pset.erase(var_low); // allow the spectrum to pick default values
       pset.erase(var_high);
-      pset.put(var_low, full_var_low);
-      pset.put(var_high, full_var_high);
       BinnedSpectrum fullSpectrum(pset);
 
       // Calculate the integrals
--- a/EventGenerator/src/DIOGenerator_tool.cc
+++ b/EventGenerator/src/DIOGenerator_tool.cc
@@ -50,7 +50,7 @@ namespace mu2e {
       auto fullconfig = conf().spectrum.get<fhicl::ParameterSet>();
       _emin = fullconfig.get<double>("elow", _spectrum.getXMin());
       _emax = fullconfig.get<double>("ehi", _spectrum.getXMax());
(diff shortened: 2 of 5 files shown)


EventGenerator/inc/ParticleGeneratorTool.hh
```diff
@@ -46,18 +46,19 @@ namespace mu2e {
     virtual ~ParticleGeneratorTool() noexcept = default;
 
     double calculateBinnedSpectrumEnergyFraction(fhicl::ParameterSet pset,
-                                                 const bool correct_full_integral = true, // correct for missing low tail to full integral
                                                  std::string var_low = "elow", // default to energy spectrum variables
                                                  std::string var_high = "ehi",
-                                                 double full_var_low = 0., // if elow == ehi, it does the full spectrum
-                                                 double full_var_high = 0.) const {
+                                                 const bool correct_full_integral = false, // correct for missing low tail to full integral
```
brownd1978, line 51: If this correction is going to be spectrum-dependent anyways, is there value in having a general parameter at all? It might be better to hard-code the correction in those generators which need it, instead of a default false option.

brownd1978 (approved):
This implementation will work, my only question is whether default false parameters are desirable. It could be addressed in a follow-on PR.

oksuzian:
## Issues

**1. Flat-spectrum check reads a nested pset key that may not exist (potential logic bug)**
In `ParticleGeneratorTool.hh`, the new guard reads `spectrumShape` from the pset passed in:
```cpp
const bool flat = pset.get<std::string>("spectrumShape", "") == "flat";
if(flat) return 1.;
```
But callers pass `conf().spectrum.get<fhicl::ParameterSet>()` (the already-unwrapped spectrum pset), while some generators determine flatness from `...spectrum...get<ParameterSet>().get<std::string>("spectrumShape",...)`. Confirm the key lives at the same level in the pset the function receives — if `spectrumShape` sits one level deeper, `flat` is always `false` and the guard never fires. The generators (`MuCapNeutron`, `MuCapProton`) still keep their own `_flatSpectrum` check, so the intent is duplicated; worth verifying both agree.

**2. Default of `correct_full_integral` flipped from `true` → `false` (behavioral change)**
```diff name=EventGenerator/inc/ParticleGeneratorTool.hh
-  const bool correct_full_integral = true,   // was: default corrects for missing low tail
+  const bool correct_full_integral = false,  // now: default does NOT correct
```
This is exactly what reviewer @brownd1978 flagged ("whether default false parameters are desirable"). Any current or future caller relying on the old default now silently loses the low-tail correction. Note `Mu2eXGenerator_tool.cc` was changed to call with **no** flag, so it now gets `false` where it previously passed `true` — verify that's intended (it changes the computed energy fraction).

**3. Signature change dropped `var_high`/`full_var_high` params — check all callers**
The old signature took `full_var_low` and `full_var_high`; the new one drops `full_var_high` and reorders args. `DIOGenerator_tool.cc` was updated to the new positional order (`"elow","ehi",true,0.`), but any other caller in the codebase not in this PR's diff would break or bind arguments to the wrong parameters. Worth a repo-wide grep for `calculateBinnedSpectrumEnergyFraction`.

**4. `_energyFraction` uninitialized if a code path skips assignment**
In `MuCapProtonGenerator_tool.cc`, `_energyFraction` is a new member assigned in the constructor body. It's used later in `finishInitialization` and `add_var`. Confirm every constructor path assigns it (no early return before the assignment) to avoid using an uninitialized double.

**5. Stale `FIXME` comment left in place**
```cpp
// FIXME: calculate the spectrum fraction simulated
config->add_var(SpectrumConfig::RestrictedVar(spectrumVarName, _energyFraction, ...));
```
This PR arguably resolves that FIXME (it now passes `_energyFraction` instead of `1.`). Remove or update the comment to avoid confusion.

oksuzian:
@FNALbuild run build test

PR #1899, opened 2026-07-24
provide an explicit 'match all' index value to SurfaceId

brownd1978:
Provide an explicit 'match all' value to avoid the implicit 'magic number' value for matching all indices.

### DIFF
--- a/DataProducts/inc/SurfaceId.hh
+++ b/DataProducts/inc/SurfaceId.hh
@@ -38,12 +38,10 @@ namespace mu2e {
         CRV_StrongBack=280,                                   // CRV module Al strongback (tracker-side support plate)
         DS_HatchConcrete=300                                  // detector-area hatch concrete block approximation
       };
-
-    // Update this counter whenever you add/remove surface IDs from the enum above.
-    static constexpr std::size_t nSurfaceIds = 64;
-
-    static std::string const& typeName();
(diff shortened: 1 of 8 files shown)


RobMina (commented):
There are a few other call sites in KKFit.hh, CentralHelixFit_module.cc, and KinematicLineFit_module.cc that should be updated -1 -> allIndices_.

There is a stale comment in SurfaceId.hh "Negative value is a wild card" that should be updated.

The same pattern is also used in EventNtuple (and possibly other repos), which should be updated after this PR lands.

RobMina (approved):

brownd1978:
Thanks for your comments Rob, I updated the PR.

RobMina:
On second thought, I think changing this to `static constexpr int allIndices_ = -1` in the header file makes more sense. That matches how nSurfaceIds is declared earlier in the file. Then you can delete the `int SurfaceId::allIndices_(-1)` line from the .cc file.

brownd1978:
@FNALbuild run build test

brownd1978:
> On second thought, I think changing this to `static constexpr int allIndices_ = -1` in the header file makes more sense. That matches how nSurfaceIds is declared earlier in the file. Then you can delete the `int SurfaceId::allIndices_(-1)` line from the .cc file.

You're absolutely right, constexpr is a better implementation.

PR #1900, opened 2026-07-25
Improvements for reflection tracking and energy loss

brownd1978:
Improve and simplify extrapolation through reflection, including:
- fix a bug in reflection tracking between the tracker and the IPA
- improve the test of reflection condition off-axis, where the BField is no longer parallel with Z, when extrapolating to a Z value
- fix the time interval for extrapolation testing to be a half-open interval (], not a full open (), to fully cover.
- The empirical  brehmsstrahlung tail cutoff parameter is retuned to center the reconstructed track momentum at the tracker front, which shifted after the fix to the pathlength bug. The cutoff is now at 30%, down from 40%.


Mu2eKinKal/inc/KKExtrap.hh
```diff
@@ -46,7 +46,7 @@ namespace mu2e {
     private:
       int debug_;
       double btol_, intertol_, maxdt_, maxdtstep_, minv_;
-      bool backToTracker_, extrapolateOPA_, toTrackerEnds_, upstream_, toCRV_;
+      bool backToTracker_, extrapolateOPA_, toTrackerEnds_, upstream_, toCRV_, stopAtTSDA_;
```
RobMina: stopAtTSDA_ is declared, but as far as I can tell is not initialized or read anywhere.

RobMina (approved):
This is a solid, coherent update. Just one small comment.

brownd1978:
@FNALbuild run build test

Mu2eKinKal/inc/KKExtrap.hh
brownd1978: > stopAtTSDA_ is declared, but as far as I can tell is not initialized or read anywhere.
Good catch. This was leftover from an aborted attempt to extend the reflection range into TSDA.

brownd1978:
I investigated the failure of the 'trigger' validation test. That shows a very small change in trigger rate, consistent with the small change in estimated straw energy loss. It shows a 15% increase in processing time. Comparing the log file with/without these changes, I don't see any substantial change in the LoopHelix fit time: for instance
tpr_TrkDe_50_D0200:TrigKSFDe:LoopHelixFit                                          0.000158104    0.0015691     0.0110027    0.00136202    0.000976374     1683 (this PR)
tpr_TrkDe_50_D0200:TrigKSFDe:LoopHelixFit                                          0.000157945   0.00156691     0.0109875    0.00136184    0.000975036     1683 (previously)
@michaelmackenzie please let me know if there's something more to check

michaelmackenzie:
@brownd1978 I agree this looks like normal variations after small changes in the tracking. The timing also looks fine at this level, so I wouldn't worry about the minor code in the trigger validation. Please let me know when this is merged and I'll update the reference log. Thank you for checking about this!

oksuzian:
- S2 (clean fix) — SurfaceId::allIndices_ should be static constexpr; being a mutable non-constexpr static is what forced the new KinKalGeom → Offline::DataProducts link dep.
  - S2 (confirm intent) — RegrowLoopHelix EDProducer→EDFilter with default MustRegrow:"Any" will drop events; inert in Offline but downstream paths need awareness.
  - S3 nits — unqualified min/max in boundedFoils; ST orders foils by vel.Z() while the rest uses vb; redundant if((!hitTSDA)); half-applied allIndices_ swap.

brownd1978:
> * S2 (clean fix) — SurfaceId::allIndices_ should be static constexpr; being a mutable non-constexpr static is what forced the new KinKalGeom → Offline::DataProducts link dep.
I removed the unnecessary dependency
> * S2 (confirm intent) — RegrowLoopHelix EDProducer→EDFilter with default MustRegrow:"Any" will drop events; inert in Offline but downstream paths need awareness.
I restored the old default ("None")
> * S3 nits — unqualified min/max in boundedFoils; ST orders foils by vel.Z() while the rest uses vb; redundant if((!hitTSDA)); half-applied allIndices_ swap.
The code has the correct implementation, Vb isn't applicable here.

brownd1978:
@FNALbuild run build test

PR #1903, opened 2026-07-26
Extend functionality for finding overlapping tracks and printing KalSeeds

brownd1978:
Add the ability to search for overlapping (ie duplicate) tracks in the same list to SelectSameTracks. 
Add the ability to print KalSeedPtr collections to KalSeedPrinter. 
This PR sits on top of PRs 1900 and 1899.


brownd1978:
Hi @michaelmackenzie now that PR 1900 is merged this PR should be reviewable.

michaelmackenzie (approved):
This looks good to me

michaelmackenzie:
@FNALbuild run build test

rlcee (approved):

oksuzian:
**1. [S2 — confirm intent] Same-list duplicate search double-counts every pair.**
  In `samelist_` mode the loop skips only the diagonal (`if(samelist_ && isec == ipri) continue;`), not the lower triangle. So a duplicate pair (A,B) is found twice — once as `ipri=A → sec=B` (emits `A,B`) and once as `ipri=B → sec=A` (emits `B,A`) — so the output
  `KalSeedPtrCollection` contains A and B each twice, and `nref_` counts the pair twice. For a duplicate-finder that's probably not what you want (the "found N overlapping tracks" summary is then 2× the number of pairs). If each unordered pair should be reported once, skip
  `isec <= ipri` instead of `isec == ipri`. If double-listing both members is intentional, a one-line comment (and a note on the doubled count) would clarify it.

  **2. [S3] `KalSeedPrinter::ptrcol_` is uninitialized in the default constructor.**
  `KalSeedPrinter() {}` leaves `bool ptrcol_;` indeterminate; a default-constructed instance whose `Print(event)` runs would branch on garbage (UB). Only the `KSConfig` ctor is used today (PrintModule is the sole construction site), so it's latent — but give it an in-class
  initializer: `bool ptrcol_ = false;`. Also `ptrcol_` is a public data member; per convention it should be private (with an accessor if needed).

  **3. [S3] `nmultref_` is declared and printed but never incremented.**
  The `endJob` line reports "… with `nmultref_` multiple matchess", but nothing does `++nmultref_` — so that figure is always 0. Presumably it should increment inside `if(matches.size() > 1)`.

  **4. [S3] `endJob` output: typos + unguarded print.**
  "overlaping" → "overlapping", "matchess" → "matches". The summary goes to `std::cout` unconditionally; a single per-job line is common, but strictly, production prints should be gated by a verbosity flag or use the message facility.

  **5. [S3 — confirm] The `ComparisonDirection: "Unknown"` ("test all") path may not do what the comment says.**
  The direction test is `priinter.momentum3().Z() * compdir_.dzdt() > 0.0`. If `dzdt()` is 0 for the unknown/mixed direction, that rejects every intersection, so "Unknown means test all" would actually match nothing. The shipped prolog uses `"Downstream"`, so it's latent —
  please confirm the Unknown path behaves as documented, and that `"Downstream"`/`"TT_Front"` are the exact accepted spellings (a mismatch would only surface at runtime, which CI doesn't cover here).

oksuzian:
@FNALbuild run build test

oksuzian:
@FNALbuild run build test

brownd1978:
> **1. [S2 — confirm intent] Same-list duplicate search double-counts every pair.**
fixed
> **2. [S3] `KalSeedPrinter::ptrcol_` is uninitialized in the default constructor.** `
fixed
> **3. [S3] `nmultref_` is declared and printed but never incremented.**
fixed
> **4. [S3] `endJob` output: typos + unguarded print.** 
fixed
> **5. [S3 — confirm] The `ComparisonDirection: "Unknown"` ("test all") path may not do what the comment says.**
fixed

brownd1978:
@FNALbuild run build test

PR #1907, opened 2026-07-28
simplify run_info conditions handling

rlcee:
for operational reasons outside this scope, provide json at the CLI.  Provide methods on the RunIno class to do standard collection procedures for the user.

### DIFF
--- a/DbService/inc/RunInfo.hh
+++ b/DbService/inc/RunInfo.hh
@@ -54,6 +54,22 @@ class RunInfo {
   // Convenience accessor for run number
   int runNumber() const { return _run_record.runNumber(); }
 
+  // Collect the cat-2 DbService CID table entries across the configs for this
+  // run.  Only the "TRG" config holds the CID tables, so this returns the
+  // cleaned content from that config (empty if there is no TRG config or it
+  // has no entries).
+  // If qjson=true  -> returns a JSON array of {"cid":..., "name":...} objects.
+  // If qjson=false -> returns one "cid name" pair per line.
+  std::string dbTables2(bool qjson = false) const;
+
+  // Collect the cat-3 DbService table key-value pairs across all configs for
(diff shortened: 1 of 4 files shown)


oksuzian:
1. [major — confirm] The "TRG" subsystem match is an unverified magic string. RunInfo::dbTables2 skips every config where subsystem() != "TRG". But "TRG" is a raw match against a DB value, it appears nowhere else in DbService, and CI never runs runTool against a real DB — so a case/spelling mismatch would silently make --cidtables return empty (for both JSON and text). Since the old code looped all configs with no TRG filter, a wrong literal is a silent regression, not just a JSON-path issue. Could you confirm the exact stored string (with a runTool -e -j sample) and maybe promote it to a named constant?

  2. [major] The cross-config JSON merge silently overwrites duplicate keys. RunConfig::dbTables3 throws on an intra-config duplicate key ("would silently overwrite… refuse to continue"), but RunInfo::dbTables3 does merged[key] = value, keeping the last config's value silently. Same data loss, unguarded across configs — and the text path then shows both values while JSON shows one. Apply the same check (throw/warn) or document why cross-config collisions can't happen.

  3. [major] Multi-run -q -j / -e -j produces concatenated JSON documents. The new methods avoid concatenating invalid JSON across configs, but runTool_main still prints one pretty-printed JSON per run, so a multi-run selection emits multiple concatenated docs (invalid JSON, not clean NDJSON). Fine for single-run use, but consider wrapping runs in a top-level array, emitting compact NDJSON, or documenting -j as single-run.

  4. [nit] -j is silently ignored without -q/-e; and a sample of the intended JSON output in the description would help a lot since CI can't exercise runTool.

rlcee:
AI comment responses:
1) true but OK, given the intended use and out of scope facts
2) reasonable comment, added the check
3,4) too nit-picky for me.  I think I'm the only person who will ever run this

PR #1908, opened 2026-07-29
Refactoring of the caloriemter base code and bug fixes

bechenard:
Major (and hopefully final) refactoring of the calorimeter code, including a few bug fixes, to facilitate condition integration and improve readability
 o Changed interface to navigation classes. There is still a method to call mu2e<->tracker frame transformations (a big wart) that needs to be moved to the tracker and removed
 o The SimParticle selection for the CaloShowerStep reconstruction was slightly tightened -> doesn't change the physics, but it changes the RNG sequence and the validation plots
 o Kept the calorimeter interface to avoid exposing some of the DiskCalorimeter methods -> fixed some call sites

Will fix the CMakeList files when the tests complain...


bechenard:
@FNALbuild run build test

bechenard:
Trigger test invalidated because the RNG sequence is altered by changes. The rest should be ok

oksuzian:
1. [S0] Crystal neighbor cache is now disk-LOCAL but consumed as GLOBAL → silent mis-clustering on disk 1. In GeometryService/src/DiskCalorimeterMaker.cc:259-261:
  //precompute the neighbors in the global frame        // stale comment
  thisCrystal.setID(icry+crystalOffset);                // id GLOBAL
  thisCrystal.setNeighbors(thisDisk.neighbors(icry,1)); // neighbors LOCAL (no offset)
  Disk::neighbors returns local ids; base used neighborsByLevel(icry+crystalOffset) (global), and DiskCalorimeter::neighbors(globalId,level) still adds the offset. So the cache now holds local ids where global are expected. For a disk-1 seed (globals 674–1347, 674/disk × 2 disks) crystal(id).neighbors() returns 0–673 → resolves to disk-0 crystals. It's silent (ids stay in range) and disk-0 works, so single-disk smokes pass. Affected: ClusterFinder (→CaloClusterMaker), CaloClusterFast, CaloTrigger (this PR changed it from the correct cal->neighbors(id) to the cache), ClusterUtils. Fix: offset in the maker before caching (v += crystalOffset), or switch consumers to cal->neighbors(globalId,level). Please add a disk-1 clustering check — this bug class is invisible to disk-0-only tests.

  2. [S1] crystalPos.txt is loaded at runtime but not installed by CMake. Disk::fillCrystals() opens it via ConfigFileLookupPolicy (geom key diskCrystalFile), but CalorimeterGeom/CMakeLists.txt has no install(DIRECTORY data …) (sibling packages
  Mu2eG4/TrackerConditions/ParticleID all do). A CMake/spack install or Phase-2 tarball ships without it → DISK_OPEN_FAILED. The scons buildtest reads from the source tree so CI is green regardless. One-line fix: install(DIRECTORY data DESTINATION
  ${CMAKE_INSTALL_DATAROOTDIR}/Offline/CalorimeterGeom).

  3. [S2] DiskCalorimeter copy/move ctors drop trackerCenter_ (DiskCalorimeter.cc:22-36) → zeroed on any copy → mu2eToTracker/trackerToMu2e become the identity (multi-mm error in track-calo matching). Latent (production path is a unique_ptr move) but a real regression; add trackerCenter_(rhs.trackerCenter_) to both.

  4. [S2] The usePhysVolInfo FHiCL knob is now silently dead (CaloShowerStepMaker): the phys-vol branch was dropped but usePhysVol_/mapPhysVol_ remain (rebuilt every subrun, no reader). Setting it true is silently ignored. No production impact (default false); remove or re-wire.

  5. [S3] isContainedSection→isInsideSameDisk is a real logic change (this is the RNG/plot diff) — energy-conserving, drops nothing at the production eDepThreshold: 0; add a comment. caloMotherXorig is stored sign-flipped vs its comment (dead key). Minor CMake/SConscript lib drift; dead Disk::radiusInCrystal_; stale "global frame" comment.

oksuzian:
@FNALbuild run build test

giro94 (approved):
Ok for me. This might have some consequence in EventNtuple code, e.g. https://github.com/Mu2e/EventNtuple/blob/main/src/InfoStructHelper.cc#L532C1-L532C79

oksuzian:
### Findings

  **1. [S1] `crystalPos.txt` is still not installed by the CMake build — and the new `configure_file` line doesn't fix it.**

  The file is loaded at runtime via `ConfigFileLookupPolicy` (geom key `calorimeter.diskCrystalFile`, throws `DISK_OPEN_FAILED` if absent). The new line in `CalorimeterGeom/CMakeLists.txt`:
  ```cmake
  configure_file(${CMAKE_CURRENT_SOURCE_DIR}/data/crystalPos.txt ${CURRENT_BINARY_DIR} data/crystalPos.txt)
  has two problems:
  - ${CURRENT_BINARY_DIR} is not a defined CMake variable (the standard one is CMAKE_CURRENT_BINARY_DIR); it expands to empty, which silently collapses the call to a 2-arg configure_file that copies the file into the build tree.
  - Even working as intended, configure_file is a configure-time copy, not an install rule — a CMake/spack install or a Phase-2 grid tarball still ships without the file, and the first calo-geometry job dies. The scons Jenkins CI can't catch this
  (it reads the source tree).

  Fix (one line, matches every sibling package that ships runtime data — Mu2eG4/CMakeLists.txt:306-310, TrackerConditions, ParticleID):
  install(DIRECTORY data DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/Offline/CalorimeterGeom)
  and drop the configure_file line.

  2. [S3] mapPhysVol_ is now a write-only set rebuilt every subrun.

  With the usePhysVol_ branch gone, CaloShowerStepMaker_module.cc still clears and repopulates mapPhysVol_ in beginSubRun (lines ~195-207), but nothing reads it anymore. Remove the member and the population loop (physVolInfoInput_/vols_ are still
  genuinely needed by PhysicalVolumeMultiHelper).

  3. [S3] Minor / confirm (carried over from the first pass, unchanged)
  - isContainedSection → isInsideSameDisk (CaloShowerStepMaker): a genuine logic change (old = Δz-only; new = both endpoints inside the disk envelope), not a rename — this is the RNG/validation-plot change the author flagged. It conserves
  per-crystal deposited energy at the production eDepThreshold: 0. A one-line comment that the criterion changed deliberately would help future archaeology.
  - The shim-step loop assumes every crystal row straddles x=0 (sentinels xLowMin=1e6…); fine for the shipped crystalPos.txt (all 37 rows have both sides), fragile for future asymmetric files.
  - CMake vs SConscript link-lib drift (cetlib on scons only); cosmetic. Dead Disk::radiusInCrystal_ member. DiskCal00_module.cc dropped an Origin: debug print rather than porting it. SquareShiftMapper::colFromIndex changed the negative-column
  convention (now self-consistent, internally unused — confirm no external caller relied on the old numbering).

PR #1913, opened 2026-08-01
Add optional protonabsorber.zStartInMu2e: absolute IPA position

oksuzian:
## Motivation

The MECO-style inner proton absorber is placed relative to the stopping
target: `MECOStyleProtonAbsorberMaker` computes `targetEnd` from the live
`StoppingTarget` and derives everything from it — body placement, DS2/DS3
segmentation, the cone-radius interpolation anchor, and the IPA
support-wire ring positions. But the IPA is fixed hardware;
`protonAbsorber_cylindrical_v04.txt` itself documents
`distFromTargetEnd = 625.; //mu2e z positions are 6901-7901 mm`.

Anyone varying stopping-target geometry gets a silently displaced
absorber. In our stopping-target optimization study a longer foil stack
dragged the IPA 133.33 mm downstream as a rigid body (measured with
`protonabsorber.verbosityLevel = 1`: 7034.35–8034.35 vs the design
6901.02–7901.02) through 414 grid evaluations before it was noticed. For
a cylindrical IPA this is exactly compensable by adjusting
`distFromTargetEnd`; for the conical variant it is not (the taper anchor
moves with the target) — hence a first-class option.

## Change

New optional key `protonabsorber.zStartInMu2e`. When present, the maker
derives its reference plane from the given absolute z
(`targetEnd = zStartInMu2e - distFromTargetEnd`), so the absorber's
upstream end lands exactly at the given z and its full geometry (length,
DS2/DS3 split, cone radii, support wires) is independent of the stopping
target. When absent — the default everywhere — the new branch never
executes and output is bit-identical to today. One commented example
added to `protonAbsorber_cylindrical_v04.txt`.

## Validation (Offline v13_32_10 + SimJob/Run1Bap backing)

Six single-event G4 runs with full surface check
(`protonabsorber.verbosityLevel = 1`), stopping-target stack extents
400 / 800 / 1100 mm, each extent probed twice — config-compensated
`distFromTargetEnd` (no new key) vs stock 625 + `zStartInMu2e = 6901.02`:

| probe | mechanism | absorber Z extent printed | overlaps |
|---|---|---|---|
| extent 400, expression | distFromTargetEnd = 825.0 | constructProtonAbsorber protonabs1 Z extent in Mu2e    : 6901.02, 7901.02 | 0 |
| extent 800, expression | distFromTargetEnd = 625.0 | constructProtonAbsorber protonabs1 Z extent in Mu2e    : 6901.02, 7901.02 | 0 |
| extent 1100, expression | distFromTargetEnd = 475.0 | constructProtonAbsorber protonabs1 Z extent in Mu2e    : 6901.02, 7901.02 | 0 |
| extent 400, option only | zStartInMu2e = 6901.02, dist = stock 625 | constructProtonAbsorber protonabs1 Z extent in Mu2e    : 6901.02, 7901.02 | 0 |
| extent 800, option only | zStartInMu2e = 6901.02, dist = stock 625 | constructProtonAbsorber protonabs1 Z extent in Mu2e    : 6901.02, 7901.02 | 0 |
| extent 1100, option only | zStartInMu2e = 6901.02, dist = stock 625 | constructProtonAbsorber protonabs1 Z extent in Mu2e    : 6901.02, 7901.02 | 0 |

All six print the identical absorber extent 6901.02–7901.02 with zero
overlaps. With the key unset, a golden-parity harness (real G4 preflight
replay) reproduces baseline output unchanged.

### DIFF
--- a/GeometryService/src/MECOStyleProtonAbsorberMaker.cc
+++ b/GeometryService/src/MECOStyleProtonAbsorberMaker.cc
@@ -125,6 +125,14 @@ namespace mu2e {
     // we add space for the virtual detector here
     double targetEnd = _target->centerInMu2e().z() + 0.5*_target->cylinderLength() + 2.*vdHL;;
 
+    // Optional absolute pinning: when protonabsorber.zStartInMu2e is set,
+    // the IPA's upstream end sits at that absolute z regardless of the
+    // stopping target; distFromTargetEnd then only anchors the cone-radius
+    // interpolation. Unset (default): stock target-relative behavior.
+    if ( _config.hasName("protonabsorber.zStartInMu2e") ) {
+      targetEnd = _config.getDouble("protonabsorber.zStartInMu2e") - distFromTargetEnd;
+    }
+
     // distance from target end to ds2-ds3 boundary
     double targetEndToDS2End = _ds->vac_zLocDs23Split() - targetEnd;
 
--- a/Mu2eG4/geom/protonAbsorber_cylindrical_v04.txt
+++ b/Mu2eG4/geom/protonAbsorber_cylindrical_v04.txt
@@ -29,6 +29,9 @@ bool   protonabsorber.solid        = false;
 // To make inner part shorter than MECO design
 bool   protonabsorber.isShorterCone = true;
 double protonabsorber.distFromTargetEnd = 625.; //mu2e z positions are 6901-7901 mm
+// Optional: pin the IPA at an absolute z, independent of the stopping
+// target (see MECOStyleProtonAbsorberMaker):
+// double protonabsorber.zStartInMu2e = 6901.;
 double protonabsorber.halfLength   = 500.0;
 
 bool   protonabsorber.ipa.buildSupports = true;


oksuzian:
@FNALbuild run build test

GeometryService/src/MECOStyleProtonAbsorberMaker.cc
```diff
@@ -125,6 +125,14 @@ namespace mu2e {
     // we add space for the virtual detector here
     double targetEnd = _target->centerInMu2e().z() + 0.5*_target->cylinderLength() + 2.*vdHL;;
 
+    // Optional absolute pinning: when protonabsorber.zStartInMu2e is set,
+    // the IPA's upstream end sits at that absolute z regardless of the
+    // stopping target; distFromTargetEnd then only anchors the cone-radius
+    // interpolation. Unset (default): stock target-relative behavior.
+    if ( _config.hasName("protonabsorber.zStartInMu2e") ) {
+      targetEnd = _config.getDouble("protonabsorber.zStartInMu2e") - distFromTargetEnd;
```
AndrewEdmonds11, line 133: Does the  ```distFromTargetEnd``` parameter still need to be set correctly? By default it is zero and so here the ```targetEnd``` would be the starting position of the IPA...

AndrewEdmonds11 (approved):
This is OK with me as a patch. I think ideally we would number a new version and then have new code that would not refer to the target position at all but I appreciate that that would be a lot of effort

oksuzian (commented):
### PR Review Summary — Mu2e/Offline#1913 `Add optional protonabsorber.zStartInMu2e: absolute IPA position`

Reviewed at head `a894d5da143f3392fcf3600d6f2293169eb632f9`

*Disclosure: this PR was authored with Claude Code assistance, and this review was
produced the same way. It is offered as a self-review — weigh it accordingly. Its
main purpose is to answer @AndrewEdmonds11's open inline question with numbers.*

**Decision**
- 🟡 comment — the mechanism is sound and the default path is untouched; three
  non-blocking items below, the first of which answers the open review question.

**Scope understood**
- Adds an opt-in `protonabsorber.zStartInMu2e` key that re-anchors `targetEnd` in
  `MECOStyleProtonAbsorberMaker::Build`, so the IPA's upstream face sits at an
  absolute z instead of tracking the stopping target.
- 8 code lines + 3 comment lines in `protonAbsorber_cylindrical_v04.txt`. No new
  files, no header change, no build-system change in either scons or CMake.

---

**Answer to @AndrewEdmonds11's question** (`MECOStyleProtonAbsorberMaker.cc:133`)

> Does the `distFromTargetEnd` parameter still need to be set correctly? By default
> it is zero and so here the `targetEnd` would be the starting position of the IPA...

Yes on both counts, and the consequence splits cleanly in two:

*Position: exactly independent of `distFromTargetEnd`.* Every downstream placement
expression cancels the `- distFromTargetEnd` that line 133 introduces:

| expression | head line | with key set, reduces to |
|---|---|---|
| `pabs1` off if `d > targetEndToDS2End` | 149 | `zStart > ds23split` |
| `pabs2` off if `d + 2H < targetEndToDS2End` | 154 | `zStart + 2H < ds23split` |
| `pabs1halflen` | 172 | `(ds23split - zStart)/2`, capped at `H` |
| `pabs1ZOffset` | 193 | `zStart + pabs1halflen` |
| `pabs2ZOffset` | 197/200 | `ds23split + pabs2halflen` / `zStart + pabs2halflen` |
| `ipazstart` (support wires) | 380 | `zStart` |

So when `d` defaults to 0 the IPA start is still exactly `zStartInMu2e` — line 380
adds `d` straight back onto `targetEnd`. I swept `d ∈ {0, 5, 475, 625, 825, 3000}`
× three stopping-target configurations × three `ds23split` values through a
transcription of the head math: the pabs1 extent is `(6901.0200, 7901.0200)` in all
54 combinations, reproducing the extent in your validation table.

*Radii: genuinely dependent on `distFromTargetEnd`.* Lines 220–227 anchor the cone
interpolation at `targetEnd`, and `pabs1ZOffset - pabs1halflen - targetEnd` reduces
to `d`, so `rOut0 = r1out0 + slope·d`. The size of that effect is geometry-dependent:

- **`protonAbsorber_cylindrical_v04.txt`** (this PR's file) sets
  `OutRadius0 == OutRadius1 == 300.5`, so the only slope is the `r2out1 -= 2·vdHL`
  at line 116 — `-8.0e-6` mm/mm. Over the whole legal `d` range the upstream radius
  moves **~6.6 µm**. Irrelevant.
- **`protonAbsorber_conical_v2.txt`** has `335.2 → 380.0`, slope `0.017912` mm/mm.
  There `d` matters: a 133 mm error moves `rOut0` by 2.39 mm.

So for every geometry chain that includes v04 — `geom_run1.txt`, `geom_run1_a.txt`,
`geom_2021_PhaseI{,_v02,_v03}.txt`, `geom_reduced_DSTS_shielding.txt` — leaving
`distFromTargetEnd` at 0 would be harmless. It is not harmless for a conical IPA,
and the code comment's "only anchors the cone-radius interpolation" is the accurate
statement of why.

Worth noting the clamp at lines 75–77 (`d > 2500 || d < 0 → d = 0`) fires silently.
With the new key set, tripping it leaves the position perfectly correct and shifts
only the radii — so the obvious sanity check (does the IPA start where I asked?)
cannot detect it. Cylindrical: 5 µm, ignorable. Conical: up to 44.8 mm.

---

**Findings**

1. 🟡 [S2] `PrintConfig` reports `Dist from target end` that is no longer the
   distance from the target end
   - Evidence: `MECOStyleProtonAbsorberMaker.cc:327` stores the raw config `d` into
     `_distfromtargetend`; `:572` prints it as `Dist from target end`. With
     `zStartInMu2e = 6901.02`, stock `d = 625`, and a target ending at 5871.02, the
     true gap is 1030 mm but the printout says 625 — under-reporting by 405 mm.
   - Impact: diagnostic only — I grepped the accessor
     (`MECOStyleProtonAbsorber.hh:85` `distanceFromTargetEnd()`) and its sole
     consumer in Offline is that `std::cout`, so no physics reads it. But this is
     the exact `protonabsorber.verbosityLevel = 1` output the PR description cites
     as how the original 133.33 mm displacement was caught, so the new mode blunts
     the instrument that motivated it.
   - Suggested fix: in the new branch, either recompute and store the real gap
     (`zStartInMu2e - (targetCenter + 0.5·cylinderLength + 2·vdHL)`), or add a line
     to `PrintConfig` such as `IPA z pinned at : <zStartInMu2e> (target-relative
     mode off)` so the printout is unambiguous about which mode produced it.

2. 🟡 [S2] The geometry-file comment omits the one thing a user must still get right
   - Evidence: `protonAbsorber_cylindrical_v04.txt:32-34` says only "pin the IPA at
     an absolute z, independent of the stopping target". A reader reasonably
     concludes `distFromTargetEnd` is now inert — which is what prompted the review
     question above, from someone who had already read the code.
   - Impact: for v04 the misreading is harmless, but the key is read by the maker
     for *all* IPA geometries, and the comment is the only user-facing documentation.
   - Suggested fix: extend it, e.g. `// distFromTargetEnd still sets the cone-radius
     anchor (negligible for this cylindrical file, significant for conical).`

3. ⚪ [S3] `zStartInMu2e` is accepted without a sanity check
   - Evidence: line 133 consumes the value directly; nothing verifies it is
     downstream of the real target end or upstream of `ds23split`. A value past the
     split silently drops pabs1 (line 149); a value inside the stopping target
     produces an overlap that only `g4surfaceCheck` would catch.
   - Impact: low — this is an expert-only study knob, and the failure is loud under
     surface check. But the feature invites hand-entered absolute coordinates, which
     is precisely where a typo lands.
   - Suggested fix: a `mf::LogWarning` when `zStartInMu2e` is upstream of the
     computed target end. A hard `cet::exception` is probably too strong, since an
     exotic study might legitimately want it.

- 🟢 Default behavior is unchanged. The key is commented out in v04, `hasName`
  returns false, and the branch never executes; `targetEnd` keeps its line-126
  value. All six geometry chains that include v04 are unaffected.
- 🟢 The PR's motivating claim — that config-compensation works for a cylindrical
  IPA but not a conical one — is quantitatively correct. Holding the conical IPA
  start at 6901.02 by adjusting `d` across a 400/800/1100 mm target range needs
  `d = 1230/1030/880`, which swings `rOut0` over **335.2 → 357.2 mm (6.27 mm)**.
  With `zStartInMu2e` set, the same sweep gives `rOut0 = 374.6060` unchanged in all
  three. The feature does what the description says.
- 🟢 CI green at this head: build 3257 passed all of build (prof), ceSimReco,
  g4test_03MT, transportOnly, POT, g4study, cosmicSimReco, cosmicOffSpill, ceSteps,
  ceDigi. (The earlier red at 3255 was `CosmicReco/src/LineFinder_module.os`, on a
  stale `main` — unrelated to this diff, and cleared by the re-run.)
- 🟢 Nothing else in Offline consumes `distanceFromTargetEnd()`, so re-anchoring
  `targetEnd` has no reach beyond this maker.
- 🟢 No build-file change required — `.cc`-only edit to an existing translation unit,
  so the scons/CMake dual-build blind spot does not apply here.

**Validation check**
- Build/tests run: yes, by CI — build 3257 green at `a894d5da`.
- Config contract check: pass. `hasName` gates cleanly; no FHiCL schema involved
  (SimpleConfig geometry, not validated-config).
- Cross-repo consistency: pass. Opt-in and default-off, so no `Production` or
  `mu2e-trig-config` change is implied.

**Residual risk**
- The radius/position asymmetry is the one non-obvious property of this design: a
  user who verifies only the z extent will not notice a wrong `distFromTargetEnd`.
  Negligible for the cylindrical geometries actually in use; real for conical.
- The silent `d` clamp at lines 75–77 is pre-existing, but the new mode makes its
  effect harder to spot for the reason above.

**Author follow-ups**
1. Reply to @AndrewEdmonds11 on the inline thread — the short answer is "yes, for
   conical geometries; ~µm for this cylindrical file", with the cancellation table
   above if useful.
2. Consider finding 1 (the misleading `Dist from target end` printout) the one worth
   acting on before merge; findings 2 and 3 are fine as follow-ups.

PR #1914, opened 2026-08-01
Install CalorimeterGeom/data in the CMake build (fixes spack ceSimReco)

oksuzian:
Fixes the spack/CMake runtime failure reported this morning:

```
Can't find file "Offline/CalorimeterGeom/data/crystalPos.txt"
```

`crystalPos.txt` (added in #1908) is opened at runtime via `ConfigFileLookupPolicy` from geom key `calorimeter.diskCrystalFile`, so any calo-geometry job (e.g. ceSimReco) needs it on `MU2E_SEARCH_PATH`. `CalorimeterGeom/CMakeLists.txt` has no install rule for `data/` — the `configure_file` line only stages the file into the *build* tree — so a spack/CMake install ships without it. The scons build reads the source tree, which is why CI stayed green.

One line, matching every other package that ships runtime data (`Mu2eG4`, `TrackerConditions`, `CaloConditions`, `CalPatRec`, `CRVConditions`, `CRVResponse`, `CaloFilters`, …):

```cmake
install(DIRECTORY data DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/Offline/CalorimeterGeom)
```

No behavior change for scons or build-tree runs.

### DIFF
--- a/CalorimeterGeom/CMakeLists.txt
+++ b/CalorimeterGeom/CMakeLists.txt
@@ -18,3 +18,4 @@ configure_file(${CMAKE_CURRENT_SOURCE_DIR}/data/crystalPos.txt ${CURRENT_BINARY_
 
 install_source(SUBDIRS src)
 install_headers(USE_PROJECT_NAME SUBDIRS inc)
+install(DIRECTORY data DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/Offline/CalorimeterGeom)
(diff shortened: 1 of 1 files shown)


rlcee (changes requested):
You want to remove the configure_file a few line above

oksuzian:
Removed the configure_file line in ceb9a3a — the install rule alone matches how Mu2eG4 and the other data-shipping packages handle it.

rlcee (approved):

oksuzian:
@FNALbuild run build test

PR #1916, opened 2026-08-04
Calorimeter alignment condition support

bechenard:
Support for calorimeter alignment via the condition service. Only disk movement and rotations are supported at this point. Code to move the crystals, should this be necessary (I doubt it because we already use their measured placement in the calorimeter and these crystals won't budge without explosives), is readily available.  The resulting object is called alignedCalorimeter, following the tracker. 

Deployment on consumer modules will be done in a later stage by the corresponding experts (LineFinder calo matching via diskToMu2e/mu2eToTracker, CalPatRec, TrackCaloMatching, the truth-match cog checks)


oksuzian (changes requested):
### PR Review Summary — "Calorimeter alignment condition support" (#1916)

Reviewed at head `2b0f3eda` (2026-08-04). 17 files, +464/−14: new `CalAlignDisk`/`CalAlignCrystal` DB tables, an `alignedCalorimeter` proditions entity (deep-copied `DiskCalorimeter` with disk shifts/rotations applied), and `CaloClusterMaker` switched from `GeomHandle` to `ProditionsHandle<Calorimeter>`. The core design is sound and the dangerous part — mutating a copy of the geometry through `const_cast` — is verified safe below. The blockers are all in the build/wiring layer, and two of the three are invisible to the scons CI by construction.

**Decision**
- 🔴 Request changes. One S0 (CMake configure breaks on merge) and three S1s (fromDb ignores its own params, missing factory registration, CMake link deps not updated). All are small fixes.

**Scope understood**
- Rigid-body disk alignment (dx,dy,dz,rx,ry,rz) from fcl file or DB, applied to a deep copy of the GeomHandle `DiskCalorimeter`, served as proditions entity `Calorimeter` ("alignedCalorimeter" in config).
- Crystal-level alignment plumbed but deliberately zeroed ("won't budge without explosives"); only `CaloClusterMaker` consumes the aligned geometry so far.

---

### Findings

**1. 🔴 [S0] CMake configure fails: both new `configure_file` inputs do not exist — the spack/CMake build of Offline breaks on merge, and scons CI cannot see it.**
- Evidence: `CaloConditions/CMakeLists.txt` adds `configure_file(...data/crystal_align.dat ...)` and `configure_file(...data/disk_nominal.dat → data/disk_align.dat)`. `CaloConditions/data/` at this head contains `Sim_crystal.txt, caloDMAP_nominal.dat, disk_align.txt, nominal.txt` — neither `crystal_align.dat` nor `disk_nominal.dat` exists. `configure_file` with a missing input is a hard configure-time error. The Jenkins buildtest is scons-only and green.
- Impact: identical failure mode to the `crystalPos.txt` incident (#1908 → spack break → #1914), caught pre-merge this time.
- Suggested fix: delete both lines. `install(DIRECTORY data ...)` is already present in this CMakeLists and covers the install; the `configure_file(${CURRENT_BINARY_DIR})` staging idiom is exactly what #1914 removed from CalorimeterGeom — don't extend it. (The mismatched output name `disk_nominal.dat → disk_align.dat` in the second line is a symptom of the same copy-paste.) The runtime path is safe without them: the fcl name `Offline/CaloConditions/data/disk_align.txt` resolves via `ConfigFileLookupPolicy` from the source tree or the installed `share/`.

**2. 🟠 [S1] `fromDb` computes validated parameter vectors and then ignores them — raw DB crystal rows get applied, contradicting the PR's own scope.**
- Evidence: `AlignedCalorimeterMaker.cc::fromDb` builds `disk_align_params` via `readDb(...)` (range+count validation) and a zeroed `crystal_align_params`, then calls `alignCalorimeter(ptr, cad_p->rows(), cac_p->rows())` — both locals discarded.
- Impact: for disks it's dead work (table `addRow` already enforces sequential indices, so contents are identical). For crystals it's a behavior contradiction: `fromFcl` forces zeros per the stated "crystals won't budge" scope, but `fromDb` applies whatever `cal.aligncrystal` contains. The zeroed local shows the intended call.
- Suggested fix: `alignCalorimeter(ptr, disk_align_params, crystal_align_params);` — one line.

**3. 🟠 [S1] `CalAlignCrystal` is not registered in `DbTableFactory` — the `useDb:true` path throws on first use.**
- Evidence: `DbTableFactory.cc` adds only `CalAlignDisk`; the unknown-name fallthrough is `throw cet::exception("DBFILE_BAD_TABLE_NAME")`. `AlignedCalorimeterCache::initialize()` creates `DbHandle<CalAlignCrystal>` whenever `useDb` is true.
- Impact: the DB-backed mode is dead on arrival. Latent today (prolog default `useDb:false`), first-turn-on failure later.
- Suggested fix: add the `CalAlignCrystal` case. Also worth stating in the PR body that `cal.aligndisk`/`cal.aligncrystal` need the DBA-side `val.tables` registration before `useDb` can ever be exercised.

**4. 🟠 [S1] CMake plugin dependencies not updated to mirror the SConscript — CMake is the only link check that matters (`--no-undefined`).**
- Evidence: `CaloCluster/src/SConscript` gains `mu2e_CaloConditions` (mainlib+plugins) and `mu2e_DbTables` (plugins), but `CaloCluster/CMakeLists.txt` is untouched: the `CaloClusterMaker` plugin lists neither `Offline::ProditionsService` nor `Offline::CaloConditions`. The sibling precedent is explicit: `StrawHitReco` (the canonical `ProditionsHandle` consumer) links `Offline::ProditionsService` + its conditions library in its plugin block.
- Impact: scons CI green proves nothing here; the CMake link of the module is at risk exactly per the repo's known scons↔CMake drift failure mode.
- Suggested fix: add `Offline::ProditionsService` (and `Offline::CaloConditions` if the entity headers are reached) to the `CaloClusterMaker` plugin LIBRARIES, mirroring StrawHitReco.

**5. 🟡 [S2] All three new headers have wrong include guards.**
- `AlignedCalorimeterCache.hh`: `TrackerConditions_AlignedCalorimeterCache_hh` (copy-paste from TrackerConditions).
- `AlignedCalorimeterMaker.hh`: `CalConditions_...` (typo for CaloConditions).
- `AlignedCalConfig.hh`: `CaloConditions_...` but the file lives in `CaloConfig/`.
- Suggested fix: `CaloConditions_AlignedCalorimeterCache_hh`, `CaloConditions_AlignedCalorimeterMaker_hh`, `CaloConfig_AlignedCalConfig_hh`.

**6. 🟡 [S2] `readFile`'s EOF loop hangs on a malformed token.**
- Evidence: `while (!ordFile.eof()) { ordFile >> ...; if (eof) break; ... }` — a non-numeric token sets `failbit`, never `eofbit`; every subsequent `>>` is a no-op and the loop appends the stale values forever (unbounded memory, no diagnostic). Potential risk (needs a bad file), but alignment files are exactly the kind of hand-edited input that gets typos.
- Suggested fix: `while (ordFile >> index >> dx >> dy >> dz >> rx >> ry >> rz) { ... }` and keep the count check after — fails loudly on any malformed line.

**7. 🟡 [S2] The aligned/unaligned consumer split is undocumented.**
- Evidence: only `CaloClusterMaker` migrates to the aligned geometry. Everything downstream that interprets cluster positions — `LineFinder`'s calo matching (`diskToMu2e`/`mu2eToTracker`), CalPatRec, TrackCaloMatching, the truth-match cog checks — still uses `GeomHandle` (unaligned).
- Impact: zero today (`disk_align.txt` is identity), but the first non-zero DB commit makes cluster cog live in the aligned frame while consumers transform it with unaligned geometry. That rollout plan should be written down.
- Suggested fix: a sentence in the PR body (or the maker header) naming which consumers migrate next and the constraint that sim/digitization must never see the aligned entity.

**8. ⚪ [S3] Small items.**
- `ProditionsService.cc` now includes `DAQConditions/inc/EventTimingCache.hh` twice.
- Commented-out `filenameCrystal` reads in both `fromFcl`/`fromDb` plus a config that has no `filenameCrystal` atom — either wire it or drop the comments (code-history comments).
- `ProditionsHandle<Calorimeter>` constructed per `produce()` call; sibling modules hold it as a member.
- Rotation convention undocumented: units of `rx,ry,rz` (radians?), composition order `rotateX·rotateY·rotateZ`, pivot = disk origin via `setPose(origin+shift, rotation*rot)`. One comment in `CalAlignParams.hh` fixes it.
- Title typo: "contition" → "condition".

---

### 🟢 Verified correct — no action

- 🟢 **The `const_cast` mutation cannot corrupt the GeomHandle original**: `DiskCalorimeter`'s copy ctor copies `disks_` by value (`std::vector<Disk>`), `G4Info_`, and `trackerCenter_` (the #1908 omission — fixed at main), then `rebuildCrystalPtrs()` repoints the crystal-pointer cache into the copy. The maker mutates only its private deep copy. This was the highest-risk item in the PR and it is clean.
- 🟢 **`moveDisk`/`moveCrystal` keep the object self-consistent**: pose updated via `setPose`, every crystal's global position recomputed through the new pose, and all frame transforms (`mu2eToDisk`, `diskToMu2e`, FF variants) flow through the same pose. Neighbor caches are local-id-based post-#1908 and immune to pose moves.
- 🟢 **Proditions contract**: cache registered under `Calorimeter::cxname` ("Calorimeter"), matching `ProditionsHandle<Calorimeter>`; `DiskCalorimeter → Calorimeter` entity resolution is fine; `makeSet`/`makeIov`/`makeEntity` follow the standard cache pattern with correct `useDb` guards.
- 🟢 **fcl contract**: prolog `alignedCalorimeter` table ↔ `AlignedCalConfig` keys match; registered in both the Proditions prolog and the service config; `disk_align.txt` = 2 rows × 7 zeros = identity alignment for both disks.
- 🟢 **`rowToCsv` matches the `TrkAlignElement` sibling convention exactly** (including the trailing comma — checked, it's the established pattern, not a defect).
- 🟢 **`addRow` enforces strict sequential indices** — good defensive parsing.
- 🟢 **Link transitivity for the maker library**: `GeomHandle` use in `CaloConditions` resolves via `Offline::ProditionsService`'s PUBLIC link of `Offline::GeometryService`.

### Validation check
- Build/tests: 🟡 scons CI green at `2b0f3eda` — but findings 1 and 4 live exclusively in the CMake build, which CI does not exercise. A local `cmake` configure (or spack dev-build) would catch both instantly.
- Config contract: 🟢 pass.
- Cross-repo: 🟢 none required in Production/trig-config while the entity has no non-identity data; DB-side `val.tables` registration needed before `useDb` (finding 3).

### Residual risk
- If merged as-is, the next spack build of Offline fails at CMake configure (finding 1) — the same downstream break as the crystalPos.txt incident.
- No test exercises a non-zero alignment; a one-off job with a shifted disk demonstrating the cog moves accordingly would validate the sign/frame conventions end-to-end.

### Author follow-ups
1. Delete the two `configure_file` lines (finding 1).
2. Pass the validated locals in `fromDb` (finding 2).
3. Register `CalAlignCrystal` in the factory; note the `val.tables` prerequisite (finding 3).
4. Mirror the SConscript deps into `CaloCluster/CMakeLists.txt` (finding 4).
5. Fix the three include guards (finding 5).
6. Harden `readFile` (finding 6) and document the rotation convention (finding 8).
7. State the consumer-migration plan for the aligned entity (finding 7).

rlcee:
What is the plan for calaligncrystal?  I wasn't expecting it to be a dbtable since it is a constant.  Would this table be in the db or read from a text file?

bechenard:
Let me scrub the crystal stuff (I left it for just in case but I really doubt this will be needed), this will make the code cleaner

oksuzian (commented):
### PR Review Summary — "Calorimeter alignment condition support" (#1916)

Reviewed at head `b7f10044` (re-review; previous review was at `2b0f3eda`).
18 files, +439/−18. Delta under review: `2b0f3eda..b7f10044` ("Fixed AI findings").

**Decision**
- 🟡 Comment only. All four blocking items from the `2b0f3eda` review (one S0, three S1) are fixed and verified at this head — plus @rlcee's request to scrub the crystal table, which the author implemented by deleting `CalAlignCrystal` outright. What remains is two 🟡 S2s and a handful of ⚪ S3s, none of which gate. Flip to 🟢 approve once `mu2e/buildtest` is green at `b7f10044` (currently **pending** — this head has never been built).

**Scope understood**
- Rigid-body **disk** alignment (dx,dy,dz,rx,ry,rz) from an fcl-named text file or from `cal.aligndisk`, applied to a deep copy of the `GeomHandle` `DiskCalorimeter`, served as proditions entity `Calorimeter` (config key `alignedCalorimeter`).
- Crystal-level alignment **removed entirely** in this revision (`CalAlignCrystal`, the crystal loop in `alignCalorimeter`, the crystal `DbHandle`, the commented-out `filenameCrystal` reads). Repo-wide code search for `CalAlignCrystal` at this head returns 0 hits — no dangling references.
- Only `CaloClusterMaker` consumes the aligned geometry so far; everything downstream still uses `GeomHandle`.

---

### Carry-forward from the `2b0f3eda` review

| # | Prior finding | Status at `b7f10044` |
|---|---|---|
| 1 | 🔴 S0 — `configure_file` inputs `data/crystal_align.dat` / `data/disk_nominal.dat` do not exist → CMake configure error | 🟢 **FIXED, verified** |
| 2 | 🟠 S1 — `fromDb` computes validated params then passes `cad_p->rows()` / `cac_p->rows()` | 🟡 **PARTIAL** — crystal half gone, disk half unchanged; downgraded to S2 (finding A) |
| 3 | 🟠 S1 — `CalAlignCrystal` missing from `DbTableFactory` | 🟢 **FIXED, verified** (resolved by deleting the class) |
| 4 | 🟠 S1 — `CaloCluster/CMakeLists.txt` not mirrored to the SConscript | 🟢 **FIXED, verified** |
| 5 | 🟡 S2 — wrong include guards in all three new headers | 🟢 **FIXED, verified** (all three) |
| 6 | 🟡 S2 — `readFile` `while(!eof)` infinite loop on a malformed token | 🟢 **FIXED, verified** (residual nits in finding C) |
| 7 | 🟡 S2 — aligned/unaligned consumer split undocumented | 🟠 **UNADDRESSED** — carried over (finding B) |
| 8 | ⚪ S3 — five small items | 🟡 **PARTIAL** — 3 of 5 fixed; see finding D |

Verification detail for the four cleared blockers:

- **1 (S0)** — `CaloConditions/CMakeLists.txt` at head has **no** `configure_file` lines at all; `install(DIRECTORY data DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/Offline/CaloConditions)` remains. The author went one step further than prescribed and also removed the three pre-existing lines (`nominal.txt`, `caloDMAP_nominal.dat`, `Sim_crystal.txt`). 🟢 That is correct and matches the #1914 precedent exactly: `CalorimeterGeom/CMakeLists.txt` at this same head has `install(DIRECTORY data ...)` and zero `configure_file`. (Mechanically confirmed why the original lines were even parseable: `CURRENT_BINARY_DIR` is undefined anywhere in the tree — only `CMAKE_CURRENT_BINARY_DIR` exists — so the unquoted `${CURRENT_BINARY_DIR}` expanded to *zero* arguments and the third argument became the output path. The missing-input error was therefore real, and the fix is real.)
- **3 (S1)** — `DbTables/src/DbTableFactory.cc` at head registers `CalAlignDisk` only (line 133), and `CalAlignCrystal` no longer exists (`DbTables/inc/CalAlignElement.hh` now defines `CalAlignElement` + `CalAlignDisk`). Nothing references the removed class. This also answers @rlcee's question directly.
- **4 (S1)** — `CaloCluster/CMakeLists.txt` `CaloClusterMaker` plugin now lists `Offline::CaloConditions` and `Offline::ProditionsService`. Matches the `StrawHitReco` precedent (`TrkHitReco/CMakeLists.txt`: `Offline::ProditionsService` + `Offline::TrackerConditions` in the plugin `LIBRARIES REG`).
- **5 (S2)** — `CaloConditions_AlignedCalorimeterCache_hh`, `CaloConditions_AlignedCalorimeterMaker_hh`, `CaloConfig_AlignedCalConfig_hh`. All three now match the repo convention (path words + `_hh`, no repo prefix in `Offline`).
- **6 (S2)** — `readFile` is now `while (std::getline(...))` + per-line `std::istringstream`, throwing `cet::exception` on a failed extraction and on a trailing extra token. The unbounded-append failure mode is gone.

---

### Findings

**A. 🟡 [S2] (carried over from finding 2, downgraded) `fromDb` still computes `disk_align_params` and then discards it, passing the raw `cad_p->rows()` instead.**
- Evidence: `CaloConditions/src/AlignedCalorimeterMaker.cc:151-152`
  ```cpp
  auto disk_align_params = readDb(cad_p, CaloConst::_nDisk);
  alignCalorimeter(ptr, cad_p->rows());
  ```
  `disk_align_params` is never read. The crystal half of the original finding is resolved by deletion, so this is no longer a behavior contradiction: `readDb` is still *called*, so its range and count checks still fire, and given `CalAlignElement::addRow`'s strict `index == _rows.size()` invariant the vector it builds is element-wise identical to `cad_p->rows()`. Hence S2, not S1.
- Impact: dead work plus a misleading read — a maintainer sees a validated vector built and reasonably assumes it is what gets applied. It will not trip `-Wunused-variable` (non-trivial destructor), so no compiler will ever point at it.
- Suggested fix: `alignCalorimeter(ptr, disk_align_params);` — the same one-line change the original review asked for, and it makes `fromDb` structurally identical to `fromFcl` (which does it correctly at line 137).

**B. 🟠 [S2, carried over] The aligned/unaligned consumer split is still undocumented.**
- Evidence: PR body is unchanged from `2b0f3eda` ("Deployment on consumer modules will be done in a later stage by the corresponding experts"). `CaloClusterMaker_module.cc:86-87` takes `ProditionsHandle<Calorimeter>`; every other calorimeter consumer (`LineFinder` calo matching via `diskToMu2e`/`mu2eToTracker`, CalPatRec, TrackCaloMatching, the truth-match cog checks) still uses `GeomHandle<DiskCalorimeter>`.
- Impact: zero today — `disk_align.txt` is 2 rows × 7 zeros, i.e. identity. The first non-zero DB commit makes cluster cog live in the aligned frame while its consumers transform it with unaligned geometry, with no compile-time or runtime signal. Also unstated: sim/digitization must never see the aligned entity.
- Suggested fix: a short paragraph in the PR body (or a comment block in `AlignedCalorimeterMaker.hh`) naming the consumers that migrate next and the sim/digi exclusion. Cheap now, expensive to reconstruct later.

**C. 🟡 [S2] `CaloConditions/src/SConscript` gained `'CLHEP'` but `CaloConditions/CMakeLists.txt` did not gain `CLHEP::CLHEP` — the same scons↔CMake asymmetry class as the fixed finding 4.**
- Evidence: the SConscript diff un-comments `'CLHEP'` (it was `#'CLHEP'`) because `AlignedCalorimeterMaker.cc` now uses `CLHEP::Hep3Vector` and `CLHEP::HepRotation` (lines 28-32). `CaloConditions/CMakeLists.txt` `LIBRARIES PUBLIC` at head lists no CLHEP.
- Impact: **it links today** — `Offline::CalorimeterGeom` carries `CLHEP::CLHEP` as a `PUBLIC` dep and CaloConditions links CalorimeterGeom, so the symbols arrive transitively. So this is not a build break; it is a first-order-dependency-not-declared issue (Mu2e coding standard: "require explicit first-order library dependencies in build files"), and it silently breaks the day CalorimeterGeom's CLHEP link is narrowed. The author's own SConscript edit is the proof that CLHEP is considered first-order here.
- Suggested fix: add `CLHEP::CLHEP` (and, for the same reason, `cetlib_except::cetlib_except`, already used by every `throw` in this library) to `CaloConditions/CMakeLists.txt` `LIBRARIES PUBLIC`.

**D. ⚪ [S3] Residual items from prior finding 8, and small nits in the new `readFile`.**
- 🟢 *Fixed:* duplicate `#include ".../EventTimingCache.hh"` in `ProditionsService.cc` removed; the commented-out `filenameCrystal` reads removed from both `fromFcl` and `fromDb`; PR title typo corrected.
- 🟡 *Partial:* `CalAlignParams.hh` now documents units ("dx,dy,dz = displacement of the disk in mm", "rx,ry,rz are rotations along the x,y,z axes in radians"). Still undocumented: composition order and pivot. As written (`AlignedCalorimeterMaker.cc:29-32`), `HepRotation::rotateX/Y/Z` **pre-multiply**, so the applied rotation is `Rz·Ry·Rx`, and `Disk::moveDisk` pivots about the disk origin. One more comment line closes it.
- ⚪ *Unaddressed:* `ProditionsHandle<Calorimeter>` is still constructed inside `produce()` (`CaloClusterMaker_module.cc:86`); sibling modules (e.g. `StrawHitReco`) hold it as a member. Per-event cost is negligible, so this is purely idiom-matching. (Note this is not the "don't cache handles across events" rule — `ProditionsHandle` is explicitly designed to be held as a member and queried with `get(event.id())`.)
- ⚪ *New:* `AlignedCalorimeterMaker.cc` uses `std::istringstream` (line 63) and `std::abs` on floats (lines 30-32) but includes only `<iostream>` and `<fstream>`. It compiles — `<sstream>` arrives via `DbTables/inc/CalAlignElement.hh:10` (and via `cetlib_except/exception.h`), `<cmath>` via `CLHEP/Vector/Rotation.h` — but both are speculative-transitivity. Add `<sstream>` and `<cmath>` directly. The `<cmath>` one has teeth: if only the `<cstdlib>` `std::abs(int)` overload were visible, the guards would truncate every rotation below 1 rad to zero.
- ⚪ *New:* `readFile` now throws on **any** line that does not parse, including a blank trailing line or a `#` comment. `disk_align.txt` at head is exactly two clean lines so nothing breaks, but a hand-edited alignment file that picks up a trailing newline-only line becomes a hard job abort. Consider skipping empty/`#` lines before the extraction. Also, the "nothing left on the line" guard is `float extra; if (iss >> extra)` — it catches a stray *number* but silently ignores trailing non-numeric junk; `iss >> std::ws; if (!iss.eof()) throw;` covers both. Finally, both format throws use category `ALIGNEDCAL_RANGE`, which is the range-error category — `ALIGNEDCAL_FORMAT` would read better in a log.
- ⚪ *New (simplify lens):* the `if (std::abs(disk_align.r?()) > 1e-6)` guards buy nothing — `rotateX(0.0)` is already cheap and exact — while introducing a silent 1e-6 rad dead band. Dropping the three guards is simpler and strictly more faithful.
- ⚪ *New (simplify lens):* `CaloClusterMaker_module.cc` includes no `CaloConditions` or `DbTables` header — its only new first-order dep is `ProditionsService`. So `Offline::CaloConditions` (CMake), `mu2e_CaloConditions` (scons mainlib **and** plugins) and `mu2e_DbTables` (scons plugins) are over-declarations; the scons mainlib sources (`ClusterAssociator.cc`, `ClusterUtils.cc`, `ClusterFinder.cc`) touch none of them either. Harmless, and my previous review invited the CaloConditions entry conditionally — noting it only so the dependency list stays honest. **Does not gate.**
- ⚪ *New:* with the crystal loop deleted, `Disk::moveCrystal` (`CalorimeterGeom/inc/Disk.hh`, `src/Disk.cc`) has zero callers repo-wide. It pre-dates this PR, so removing it is out of scope, but a one-line comment saying it is retained for a possible future crystal-alignment pass would stop the next reader from treating it as live.

---

### 🟢 Verified correct — no action

- 🟢 **`const_cast` mutation cannot corrupt the `GeomHandle` original** (re-verified: `alignCalorimeter` is unchanged apart from the crystal-loop deletion). `DiskCalorimeter`'s copy ctor copies `disks_` by value, plus `G4Info_` and `trackerCenter_`, then `rebuildCrystalPtrs()` repoints the crystal cache into the copy. The maker mutates only its private deep copy.
- 🟢 **`moveDisk` keeps the object self-consistent** — pose via `setPose`, crystal global positions recomputed through the new pose, all frame transforms routed through the same pose.
- 🟢 **Crystal removal is complete and clean.** `CalAlignCrystal`, the `_tacrys_p` handle, its `makeSet`/`makeIov` contributions, the `fromDb`/`alignCalorimeter` crystal parameters, and the zeroed placeholder vectors are all gone together. No half-removal, no unreachable branch, no dangling reference (verified by repo-wide search). `CalAlignElement.hh` still includes `CaloConst.hh`, correctly — `CalAlignDisk` uses `CaloConst::_nDisk`.
- 🟢 **`CalAlignElement` DB contract.** `orderBy()` is declared `const std::string orderBy() const` and the base is `virtual const std::string orderBy() const` — signatures match, so it genuinely overrides (identical to the `TrkAlignElement` sibling, which likewise omits `override`). `addRow` enforces strict sequential indices; `rowToCsv` matches the `TrkAlignElement` convention including the trailing comma.
- 🟢 **Proditions wiring.** Cache registered under `Calorimeter::cxname`, matching `ProditionsHandle<Calorimeter>`; `makeSet`/`makeIov`/`makeEntity` follow the standard pattern with correct `useDb` guards after the crystal removal; `_caches[acc->name()] = acc;` added alongside the other calo caches in `ProditionsService.cc`.
- 🟢 **fcl contract.** `CaloConditions/fcl/prolog.fcl` `alignedCalorimeter { filenameDisk, verbose, useDb }` exactly matches the three atoms of `AlignedCalConfig` at head (`filenameCrystal` correctly absent); registered in `ProditionsService/fcl/prolog.fcl` under `Proditions` and as `fhicl::Table<AlignedCalConfig>` in `ProditionsService.hh`. `disk_align.txt` is 2 rows × 7 zeros = identity for both disks, and 2 == `CaloConst::_nDisk` so the count check passes.
- 🟢 **Cross-repo config compatibility of the new non-defaulted fhicl table.** Adding a `fhicl::Table` with no defaults to the service `Config` would break any job that builds `services.ProditionsService` from scratch. Checked every consumer: `Offline/fcl/standardServices.fcl` and `mu2e-trig-config/core/trigServices.fcl` both use `ProditionsService: @local::Proditions`; all 8 `Production` hits (`JobConfig/reco|mixing|digitize/epilog.fcl`, `Processing/*.fcl`, `Validation/database-v1_1-compat.fcl`) and the Offline test/EventDisplay fcls are dotted overrides on top of the prolog. **No cross-repo change required.**
- 🟢 **No linkage loop introduced.** `Offline::ProditionsService` is an `INTERFACE` library; the service itself is a separate `art::service` plugin that links `Offline::CaloConditions`. `CaloConditions → Offline::ProditionsService (INTERFACE)` therefore does not close a cycle.
- 🟢 **Merge state.** `mergeable: true`; `main` moved `f4084747 → 1ce31dbc` but the delta is `Mu2eG4` proton-absorber geometry/materials only — no overlap with this PR.

### Validation check
- Build/tests run: 🟡 **`mu2e/buildtest` is PENDING at `b7f10044`** — the previous green run was at `2b0f3eda` and FNALbuild has already marked it out of date (main moved). The `2b0f3eda..b7f10044` delta is not cosmetic: it deletes a DB table class, changes two function signatures, and rewrites the file parser. Nothing here should fail, but nothing here has been compiled by CI either. Re-run and confirm green before merge.
- CMake build: 🟡 still not exercised by CI. Findings 1 and 4 were CMake-only and are now fixed; finding C is CMake-only and non-breaking. A local `cmake` configure would close this out in a minute.
- Config contract: 🟢 pass (fcl ↔ `Config` keys, prolog registration, data-file row count).
- Cross-repo consistency: 🟢 pass — no `Production` / `mu2e-trig-config` change required. Still outstanding on the DB side: `cal.aligndisk` needs DBA `val.tables` registration before `useDb: true` can ever be exercised (unchanged from the prior review; worth stating in the PR body so @rlcee can schedule it).

### Residual risk
- Low. With alignment identity everywhere and `useDb: false` by default, the runtime behavior change is a per-IOV deep copy of `DiskCalorimeter` plus one extra proditions cache in every job that runs `CaloClusterMaker`.
- The real risk is deferred, not removed: finding B. The first non-zero `cal.aligndisk` commit produces cluster positions in a frame that no downstream consumer knows about.
- No test exercises a non-zero alignment. A one-off job with a shifted disk showing the cog moves by the expected amount would pin the sign and frame conventions end-to-end — worth doing before the first real DB payload, not necessarily before this merge.

### Author follow-ups
1. Pass `disk_align_params` (not `cad_p->rows()`) in `fromDb` — finding A, one line.
2. Add `CLHEP::CLHEP` (and `cetlib_except::cetlib_except`) to `CaloConditions/CMakeLists.txt` to mirror the SConscript — finding C.
3. Add a paragraph on which consumers migrate to the aligned entity next, and that sim/digi must not — finding B.
4. Optional S3 polish: direct `<sstream>` / `<cmath>` includes; tolerate blank/comment lines in `readFile`; drop the `1e-6` rotation guards; document the `Rz·Ry·Rx` order and disk-origin pivot; note that `Disk::moveCrystal` is retained for a possible future pass.
5. Re-run the build test at `b7f10044` and confirm green.
6. Confirm with @rlcee that `cal.aligndisk` will be registered in `val.tables` before anyone sets `useDb: true`.

bechenard:
@FNALbuild run build test

oksuzian (approved):
### PR Review Summary — re-review of "Calorimeter alignment condition support" (#1916)

Reviewed at head `8fb8aab9` (2026-08-04). Third review pass: 🔴 at `2b0f3eda` (all four blockers since fixed and verified), 🟡 at `b7f10044` (held from approve only by the pending build test), now the delta `b7f10044..8fb8aab9` ("More AI fixes") plus a green CI run. Every remaining finding is accounted for below; fixes verified in the diff.

**Decision**
- 🟢 Approve. `mu2e/buildtest` is green at `8fb8aab9` (merged against main `1ce31dbc`) — the exact condition my previous review named — and both remaining S2s are fixed and verified. What's left is S3 polish and one DB-side coordination reminder, none of which gates. This approval supersedes my earlier request-changes.

---

### Carry-forward accounting (vs the comment review at `b7f10044`)

**1. 🟢 [was S2] `fromDb` discarding `disk_align_params` — FIXED in `8fb8aab9`, verified.**
`alignCalorimeter(ptr, disk_align_params)` — the validated vector is now what gets applied, making `fromDb` structurally identical to `fromFcl`. Compiles (CI green).

**2. 🟢 [was S2] Consumer-split documentation — FIXED.**
The PR body now names the consumers that migrate in a later stage (LineFinder calo matching, CalPatRec, TrackCaloMatching, truth-match cog checks) and the crystal-alignment rationale. Residual one-liner (⚪, optional): it still doesn't state explicitly that sim/digitization must never consume the aligned entity.

**3. 🟢 [was S2] CLHEP scons↔CMake asymmetry — FIXED, verified.**
`CaloConditions/CMakeLists.txt` gains both `CLHEP::CLHEP` and `cetlib_except::cetlib_except` — the full prescription, first-order deps now explicit.

**4. S3 batch from the prior review — mostly done:**
- 🟢 Direct `<sstream>` / `<cmath>` includes — FIXED (closes the `std::abs(int)`-truncation trap for real).
- 🟢 The `1e-6` rotation guards — FIXED (dropped; unconditional rotations, simpler and exact).
- 🟢 Trailing-junk check — FIXED with the suggested `iss >> std::ws; if (!iss.eof()) throw`, and the format throws now use their own `ALIGNEDCAL_FORMAT` category.
- 🟢 `ProditionsHandle<Calorimeter>` — FIXED (now a member `alignedCal_h`, matching the `StrawHitReco` idiom).
- ⚪ readFile still hard-aborts on a blank trailing line or `#` comment in a hand-edited alignment file — unchanged, take or leave.
- ⚪ Dependency-list honesty — PARTIAL: `Offline::CaloConditions` removed from the `CaloClusterMaker` plugin in **both** build systems (correct: the module includes no CaloConditions header, and `ProditionsHandle<Calorimeter>` needs only `Calorimeter::cxname` from CalorimeterGeom — CI green confirms), but the scons mainlib still lists `mu2e_CaloConditions` and the plugin list keeps `mu2e_DbTables`. Cosmetic.
- ⚪ `Disk::moveCrystal` zero-callers — the PR body's "code to move the crystals ... is readily available" covers the intent; an in-code one-liner would still help the next reader, optional.

**All four `2b0f3eda` blockers remain fixed** — the delta touches none of their evidence except `AlignedCalorimeterMaker.cc`, where it only improves.

---

### New in the delta (`b7f10044..8fb8aab9`)

**1. ⚪ [S3] The new rotation comment in `CalAlignParams.hh` overstates commutativity.**
- Evidence: "Order of operation is irrelevant since rotation is perfomed around disk center". Rotation composition about different axes never commutes exactly, and the pivot point has nothing to do with it — what actually makes order immaterial here is that alignment rotations are mrad-scale, so the commutator is second-order (~rx·ry ≈ 10⁻⁶ rad). Since this comment is the contract DBAs read when filling `cal.aligndisk`, precision is worth one more line.
- Suggested fix: "applied as Rz·Ry·Rx about the disk center; for small alignment rotations the order is immaterial to first order" — and "perfomed" → "performed".

### 🟢 Verified — no action
- 🟢 Build test green at `8fb8aab9` merged against main `1ce31dbc` — the head with all fixes has now been compiled and tested by CI (the gap that held back the previous review).
- 🟢 The `CaloClusterMaker` handle-as-member change is the designed `ProditionsHandle` usage (queried per event with `get(event.id())`) — not a cached-handle violation.
- 🟢 Everything verified in the two prior reviews stands: deep-copy safety of the `GeomHandle` original, `moveDisk` self-consistency, complete crystal removal, fcl↔Config contract, cross-repo config compatibility (all `Production`/`mu2e-trig-config` consumers build `ProditionsService` from `@local::Proditions`), no linkage loop.

### Validation check
- Build/tests run: 🟢 FNALbuild green at head (scons, prof). CMake still has no CI, but every CMake-only finding is now resolved and the delta's CMake edits are verified-safe dependency changes.
- Config contract: 🟢 pass (unchanged by the delta).
- Cross-repo consistency: 🟢 pass — no `Production`/`mu2e-trig-config` change required. Still outstanding on the DB side (non-gating): `cal.aligndisk` needs `val.tables` registration before `useDb: true` can be exercised — worth confirming with the DB maintainers before the first DB payload.

### Residual risk
- Low and deferred by design: until consumers migrate (per the PR body's plan), a non-zero `cal.aligndisk` payload would put cluster cog in the aligned frame while downstream uses unaligned geometry. That is the documented later stage, not this PR's defect.
- No test exercises a non-zero alignment yet; a one-off shifted-disk job pinning sign/frame conventions remains a good idea before the first real DB payload.

### Author follow-ups (non-blocking)
1. Tighten the `CalAlignParams.hh` rotation comment (+ typo).
2. Optional: tolerate blank/`#` lines in `readFile`; state the sim/digi exclusion in the PR body; drop the leftover scons mainlib `mu2e_CaloConditions` / plugin `mu2e_DbTables` entries.
3. Confirm `cal.aligndisk` gets registered in `val.tables` before anyone sets `useDb: true`.

DbTables/inc/CalAlignParams.hh
```diff
@@ -0,0 +1,36 @@
...
+//
+//  Rigid body alignment parameters for calorimeter
+//  dx,dy,dz = displacement of the disk in mm
+//  rx,ry,rz are rotations along the x,y,z axes w.r.t disk center in radians
+//  Order of operation is irrelevant since rotation is perfomed around disk center
+//
+namespace mu2e {
+  class CalAlignParams {
+    public:
+      CalAlignParams(int index, float dx, float dy, float dz,
```
rlcee: unsigned index

DbTables/inc/CalAlignElement.hh
```diff
@@ -0,0 +1,85 @@
...
+
+      void rowToCsv(std::ostringstream& sstream, std::size_t irow) const override {
+        CalAlignParams const& r = _rows.at(irow);
+        sstream << r.index() << ",";
+        sstream << std::fixed
+                << std::setprecision(3);
+        sstream << r.dx() << ",";
+        sstream << r.dy() << ",";
+        sstream << r.dz() << ",";
+        sstream << std::fixed << std::setprecision(3);
```
rlcee: prob should be 6

DbTables/inc/CalAlignElement.hh
```diff
@@ -0,0 +1,85 @@
...
+#include <string>
+
+namespace mu2e {
+
+  class CalAlignElement : public DbTable {
+    public:
+      typedef std::shared_ptr<CalAlignElement>       ptr_t;
+      typedef std::shared_ptr<const CalAlignElement> cptr_t;
+
+      CalAlignElement(const char* Name, const char* DbName, size_t nrows) :
```
rlcee: variables should not be capitalized

CaloConditions/inc/AlignedCalorimeterCache.hh
```diff
@@ -0,0 +1,61 @@
...
+          cids.insert(_tadisk_p->cid());
+        }
+        return cids;
+      }
+
+      DbIoV makeIov(art::EventID const& eid) {
+        DbIoV iov;
+        iov.setMax(); // start with full IOV range
+        if (_useDb) {
+          iov.overlap(_tadisk_p->iov());
```
rlcee, line 42: missing the get

CaloConditions/src/AlignedCalorimeterMaker.cc
```diff
@@ -0,0 +1,158 @@
...
+  //-----------------------------------------------------------------------------------------
+  void AlignedCalorimeterMaker::alignCalorimeter(AlignedCalorimeterMaker::ptr_t ptr,
+                                                 const std::vector<CalAlignParams>& disk_align_params)
+  {
+    DiskCalorimeter& cal = *ptr;
+
+    for (auto& disk : cal.disks()) {
+      // Use the const_cast bazooka here - this is the only place where
+      // Disk parameters should be modified by an external agent
+      auto& diskcc = const_cast<Disk&>(disk);
```
rlcee: I think there a couple ways to avoid this.  I guess I don't feel strongly, so it is up to you

CaloConditions/src/AlignedCalorimeterMaker.cc
```diff
@@ -0,0 +1,158 @@
...
+  AlignedCalorimeterMaker::ptr_t AlignedCalorimeterMaker::fromFcl()
+  {
+    //Deep copy of the calorimeter to keep the original untouched
+    GeomHandle<DiskCalorimeter> cal_h;
+    auto ptr = std::make_shared<DiskCalorimeter>(*cal_h);
+
+    if ( _config.verbose() > 0 )
+      std::cout << "AlignedCalorimeterMaker::fromFcl now zero aligning Calorimeter\n";
+
+    auto disk_align_params = readFile(_config.filenameDisk(), CaloConst::_nDisk);
```
rlcee, line 135: This is still ding the same as what fromDb with a text would do.  Typically fromFcl would insert an alignment from fcl parameters, or all zeros.  Since this is a style item, it is up to you.

CaloConditions/inc/AlignedCalorimeterMaker.hh
```diff
@@ -0,0 +1,33 @@
...
+    public:
+      AlignedCalorimeterMaker(const AlignedCalConfig& config):_config(config) {}
+      void alignCalorimeter(ptr_t ptr,
+                            const std::vector<CalAlignParams>& disk_align_params);
+      ptr_t fromFcl();
+      ptr_t fromDb(CalAlignDisk::cptr_t cad_p);
+
+    private:
+      std::vector<CalAlignParams> readFile(const std::string& fileName, uint16_t nRowMax);
+      std::vector<CalAlignParams> readDb(CalAlignElement::cptr_t cptr, uint16_t nRowMax);
```
rlcee, line 25: I think this should be CalAlignDisk.  Tracing the trk example, I see the cptr_t is generically defined, so we are using some implicit conversions, but since tracker works, I guess that's the bottom line.  Still change this here, I think, for clarity at least.

rlcee (changes requested):
The missing get() is important, I think the rest are clarity or style

CaloConditions/src/AlignedCalorimeterMaker.cc
bechenard: Make me a suggestion, but I would like to keep the disk interface minimal and make sure clients can't accidentally change anything (so avoiding mutable accessors).

CaloConditions/src/AlignedCalorimeterMaker.cc
bechenard: I need to fix the error message, but the goal is to have the alignment from a text file controlled by fcl. Ideally doing what the db would do so we can test it. Am I missing anything?

CaloConditions/src/AlignedCalorimeterMaker.cc
rlcee: I think the most standard way would be put a second, non-const accessor disks() on DiskCalorimeter.  The standard geometry delivers a const DiskCalorimeter so this is inaccessible to any normal user and only accessible on a new non-const DiskCalorimeter, like here, which is appropriate.  I think the sentiment would be that seeing this accessor on the DiskCalorimeter would indicate to the observer that this use case is out there.  casting hides this case. If you reject this pattern, I won't complain

CaloConditions/src/AlignedCalorimeterMaker.cc
rlcee: fromFcl() reads content from a local text file
fromDb() with local text file override reads content from a local text file
These are the effectively the same.  If you don't see the point, then please proceed, I won't block you

CaloConditions/inc/AlignedCalorimeterMaker.hh
bechenard: CalAlignParams has the alignment parameters, which happen to be the same for the crystals and disk (shifts + rotations), so this is correct. The CalAlignElement could become a CalAlignDisk, but the current solution is more general and doesn't cost anything so I prefer to keep it.

rlcee (approved):

bechenard:
@FNALbuild run build test

PR #1923, opened 2026-08-09
feat(geom): add Run1B production geometry and virtual detector ids

oksuzian:
Ports the Run1B production geometry to main.

`geom_common.txt` is deliberately **not** changed — main keeps
`geom_run1_a_stickman.txt` as its default. That was the review comment that
blocked #1849, and it is addressed by construction here.

## Why

Production main's `Tests/Run1BReco.fcl` and the prodtools Run1B digi/reco
templates already reference `Offline/Mu2eG4/geom/geom_run1_b_v40.txt`, which
today exists only on the `Run1B` branch. This repairs that dangling cross-repo
reference.

## Only v40 is ported

Run1B has two stopping targets: the thin 37-foil target shared with Run1A, and
a disk at the face of TS5. `geom_run1_b_v40.txt` models this correctly — it
inherits `geom_run1_a.txt` untouched, so the 37 foils and their support
structure survive, and adds a two-piece aluminium assembly at the TS5 face (the
TSdA as a 600 mm plate with a 135 mm aperture at z = 4195, and the pion degrader
repurposed as a 150 mm mobile target at z = 4235). `ds_on_v40` is v40 with the
mobile target rotated out of the beam.

The earlier v01–v06 line instead overrode `stoppingTarget.radii` to a single
600 mm disk, *replacing* the foil target rather than adding to it. v40
superseded that line and is what the Run1Ban and Run1Bap campaigns run, so only
v40 is ported here.

Everything v40 needs already exists on main — no `constructTSdA.cc`,
`constructStoppingTarget.cc` or `Mu2eWorld.cc` changes are required. Confirmed
from the dump: v40 yields `TSdA4` at rmax 600 / rmin 135 / z 17.5 in
`StoppingTarget_Al` (nominal gives 525 / 235 / 50.8), and its stopping-target
foil count is identical to `geom_run1_a.txt`.

## Tracker FEB virtual detectors, off by default

19 identifiers (117–135) and the code that builds them: annular virtual
detectors on the upstream face of every second tracker plane's front-end board,
plus one behind the last plane.

These are gated on a new key, **`hasTrackerFEBVirtualDetectors`, defaulting
false**, which no geometry file sets. This deliberately differs from the `Run1B`
branch, which gates them on `TrackerHasBrassRings` — a key set `true` in
`tracker_v4/v5/v6/v7.txt`, i.e. every current geometry, so on the branch these
detectors are built in nominal running. Re-gating keeps the capability available
without changing anyone's simulation.

The registration block is the only place the key is read; placement follows via
the existing `vdg->exist(vdId)` guard and is additionally fenced behind
`vdg->exist(Tracker_FEB_0_SurfIn)`, so the disabled path constructs nothing and
performs no geometry lookups. Diagnostic output is behind `verbosityLevel > 0`,
matching the surrounding code.

## Enum safety

The identifiers are appended before `lastEnum`, so no existing enumerator
changes position or value and existing art files stay readable.
`VIRTUALDETECTORID_NAMES` gains 19 matching strings in the same order; the
`BOOST_STATIC_ASSERT` at `VirtualDetectorId.cc:65` is satisfied at 136 == 136,
verified pairwise across all 136 entries. A survey of `lastEnum` consumers found
no fixed-size array dimensioned on a literal, no persisted format with a baked-in
width, and no switch assuming a maximum id.

Per review, the branch's `EMC_Source`, `EMC_Source2` and `EMC_0_Front`
identifiers are **not** included — the first two served a study calorimeter
between Coll5 and the tracker rather than production hardware, and `EMC_0_Front`
duplicates the existing `EMC_0_FrontIn`/`EMC_0_FrontOut` that already back
`isDisk0()`.

## Verification

Normalized gdml dumps (Geant4 pointer suffixes stripped) of both nominal
geometries are **byte-identical** to main:

| geometry | volumes | md5, unchanged |
|---|---|---|
| `geom_common.txt` | 13767 | `85b9f421823ab803c4250e95214ff8f1` |
| `geom_run1_a.txt` | 14355 | `1d0cbf23ebbae35a2c14660cc535f90b` |

The harness was validated before being trusted: two dumps of the same unchanged
geometry are byte-identical after normalization, while the raw pre-normalization
dumps differ by 125,118 lines from pointer addresses alone.

The FEB detectors were also exercised in the enabled state — a scratch geometry
setting `hasTrackerFEBVirtualDetectors = true` on top of v40 produces exactly 19
volumes at copy numbers 117–135 with no Geant4 overlaps, while a plain v40 dump
produces none and is byte-identical to v40 before this change.

### DIFF
--- a/DataProducts/inc/VirtualDetectorId.hh
+++ b/DataProducts/inc/VirtualDetectorId.hh
@@ -88,6 +88,25 @@ namespace mu2e {
       STM_Final, // 114
       STM_UpStrHole, //115
       STM_UpStrLarge, //116
+      Tracker_FEB_0_SurfIn, // 117 before tracker station 0
+      Tracker_FEB_1_SurfIn, // 118 before tracker station 1
+      Tracker_FEB_2_SurfIn, // 119 before tracker station 2
+      Tracker_FEB_3_SurfIn, // 120 before tracker station 3
+      Tracker_FEB_4_SurfIn, // 121 before tracker station 4
+      Tracker_FEB_5_SurfIn, // 122 before tracker station 5
+      Tracker_FEB_6_SurfIn, // 123 before tracker station 6
+      Tracker_FEB_7_SurfIn, // 124 before tracker station 7
+      Tracker_FEB_8_SurfIn, // 125 before tracker station 8
+      Tracker_FEB_9_SurfIn, // 126 before tracker station 9
+      Tracker_FEB_10_SurfIn, // 127 before tracker station 10
+      Tracker_FEB_11_SurfIn, // 128 before tracker station 11
+      Tracker_FEB_12_SurfIn, // 129 before tracker station 12
+      Tracker_FEB_13_SurfIn, // 130 before tracker station 13
+      Tracker_FEB_14_SurfIn, // 131 before tracker station 14
+      Tracker_FEB_15_SurfIn, // 132 before tracker station 15
+      Tracker_FEB_16_SurfIn, // 133 before tracker station 16
+      Tracker_FEB_17_SurfIn, // 134 before tracker station 17
+      Tracker_FEB_18_SurfIn, // 135 after tracker station 17
       lastEnum
     };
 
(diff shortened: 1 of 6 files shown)


rlcee (approved):

sdifalco:
immagine
Virtual detectors EMCSource and EMCSource2 were introduced to study rates and doses in a small calorimeter placed between Coll5 and the tracker. I think they should not be inserted.

sdifalco:
immagine
Also VirtualDetector_EMC_0_Front is obsolete. 
immagine
The existing virtual detectors surrounding Disk 0 should be enough for radiation studies

DataProducts/inc/VirtualDetectorId.hh
sdifalco, line 1: I suggest to delete EMC_Source, EMC_Source2 and EMC_0_Front

sdifalco (changes requested):

michaelmackenzie (approved):
This looks good to me

oksuzian:
@FNALbuild run build test

oksuzian:
@FNALbuild run build test

oksuzian:
@sdifalco Thanks — agreed on both counts, and all three identifiers are removed as of 615cf5cc7.

Checking before cutting them, your case was if anything stronger than stated: `EMC_0_Front` would have sat alongside `EMC_0_FrontIn` and `EMC_0_FrontOut` (49, 50), which already exist and already back `isDisk0()`. And none of the three was referenced by any code, fcl or geometry file in this PR — `geom_run1_b_v40.txt` carries no `zEMCSource*` keys, so they would have been enumerators that nothing could ever fill.

The PR now adds 19 identifiers rather than 22, all `Tracker_FEB_*` (117–135), with `lastEnum` at 136. `isFEBTracker()` uses a symbolic range so it was unaffected.

This also removes the only reason for a follow-up PR that would have registered and placed those virtual detectors, so that work is dropped rather than deferred.

One addition since your review, in `VirtualDetectorMaker.cc` and
`constructVirtualDetectors.cc`: the code that actually builds the 19 FEB
detectors, which previously had identifiers here but no implementation. It is
gated on a new key, `hasTrackerFEBVirtualDetectors`, defaulting **false**, and
no geometry file sets it. Worth flagging that this deliberately differs from the
`Run1B` branch, which gates the same code on `TrackerHasBrassRings` — a key set
`true` in `tracker_v4` through `tracker_v7`, i.e. every current geometry, so on
the branch these detectors are built in nominal running. Off-by-default keeps
the capability available without changing anyone's simulation: a normal
`geom_run1_b_v40.txt` dump is byte-identical to the same dump taken before this
code existed.

Since that is 75 lines of C++ added after the green build test at 615cf5cc7, I
have re-triggered the build on the current head.

sdifalco (approved):
I agree to having the tracker FEB virtual detector not included as default

PR #1926, opened 2026-08-09
Fix geometry overlap: nudge dirtCentralStairwellLower3 face by 1 um

### LINKED ISSUE #1925: Geometry overlap after #1918: dirtCentralStairwellLower3 vs CentralStairwellLowerStairs (ovlp=3.1e-10)
`Offline/bin/overlapCheck.C` reports an overlap in the default geometry after #1918:

```
Processing Offline/bin/overlapCheck.C("build/al9-prof-e29-p103/Offline/gen/gdml/mu2e_common.gdml")...
 = Overlap ov00000: HallAir/dirtCentralStairwellLower30x545f8d0
                    overlapping HallAir/CentralStairwellLowerStairs0x55f22b0  ovlp=3.11047e-10
```

Verified on `main` @ `1891b3289`. Originally seen in a nightly build.

### Cause

`dirtCentralStairwellLower3` was **added** by #1918 (merged 2026-08-08). `CentralStairwellLowerStairs` is pre-existing and untouched by that PR, so this is new-volume-against-old-volume.

The two share a coincident face at x = 9956.8 mm:

- `bldg/dirtCentralStairwellLower3.txt` — `xPositions` = {9956.8, 11617.2, 11617.2, 9956.8}
- `bldg/centralStairwellLowerStairs.txt` — `xPositions` span 1.0 … **9956.8**, and the volume is rotated by `angles = {0., 0., -1.5707963268}`

`-1.5707963268` is π/2 truncated at the 10th decimal, so the rotation is off true right-angle by ~1e-11 rad. Against a ~1.2 m lever arm that puts the rotated face a few times 1e-10 mm past the neighbour — which is exactly the reported 3.11e-10.

### Severity: cosmetic, but it does break the check

The magnitude is **3.11e-10 mm = 3e-13 m**, about a thousandth of an atomic diameter. There is no physics consequence; no particle transport is affected.

It matters only because `bin/overlapCheck.C` calls `CheckOverlaps(1e-12)`, and #1523 deliberately tightened that tolerance to catch small extrusions early. So the check is doing its job — this is a genuine coincident-face artifact that the stringent tolerance is designed to surface. I would **not** loosen the tolerance to silence it.

### Suggested fix

The codebase already has a precedent for exactly this, on the *other* volume of this same pair — `bldg/centralStairwellLowerStairs.txt:10`:

```
double building.central.stairwell.lower.stairs.yHalfThickness = 609.599;
  // remove 1 um to prevent conflict as the rotation angle is not exactly pi/2
```

Same root cause, same remedy. Shave ~1 µm off the new volume so the faces no longer touch, e.g. in `bldg/dirtCentralStairwellLower3.txt`:

```diff
-double dirt.central.stairwell.lower3.yHalfThickness = 1260.45;
+double dirt.central.stairwell.lower3.yHalfThickness = 1260.449;  // remove 1 um to prevent conflict, cf. centralStairwellLowerStairs
```

Whether the right axis is `yHalfThickness` or a 1 µm nudge on the `xPositions` face at 9956.8 depends on which the author intends as the free dimension — @sdifalco, you'll know which is dimensionally meaningful here. Either way it's a micron on a dirt volume.

### Scope

`dirtCentralStairwellLower3` enters via `mu2eHall_v03.txt:663`, which is reached by ~43 top-level geometries including **`geom_common.txt`** (the current default, via `geom_run1_a_stickman.txt`). So this is not confined to a frozen campaign geometry — anyone running the overlap check on default geometry sees it.

Happy to open the one-line PR if you'd rather not bother, just say which dimension to trim.

oksuzian:
Fixes #1925.

`dirtCentralStairwellLower3`, added in #1918, abuts `CentralStairwellLowerStairs` at Mu2e z = 6375.4. That neighbour is rotated by `-1.5707963268` — π/2 truncated at the 10th decimal — so its face is tilted ~5e-12 rad off square and pokes into the new volume.

`bin/overlapCheck.C` calls `CheckOverlaps(1e-12)`, so the artifact is reported on the **default** geometry (`geom_common.txt`):

```
Overlap ov00000: HallAir/dirtCentralStairwellLower30x545f8d0
  overlapping HallAir/CentralStairwellLowerStairs0x55f22b0  ovlp=3.11047e-10
```

This moves the abutting face out by 1 µm so the volumes no longer touch — the same remedy already applied to the neighbour itself, whose `yHalfThickness` carries *"remove 1 um to prevent conflict as the rotation angle is not exactly pi/2"*.

### Validation

Regenerated `mu2e_common.gdml` from the patched geometry via `gdmldump.fcl` (art exit 0) and reran `overlapCheck.C`:

| | illegal overlaps/extrusions |
|---|---|
| baseline (`main` @ `1891b328`) | **1** |
| this change | **0** |

Diffing the regenerated GDML against the baseline after stripping Geant4 pointer suffixes shows exactly the two `twoDimVertex` entries for this face:

```diff
-<twoDimVertex x="9956.8" y="-12369.8"/>
+<twoDimVertex x="9956.801" y="-12369.8"/>
-<twoDimVertex x="9956.8" y="-13589"/>
+<twoDimVertex x="9956.801" y="-13589"/>
```

Volume counts identical (9987 `physvol`), and the neighbour's rotation is untouched. No physics consequence — 1 µm on a dirt volume, against an original discrepancy of 3e-13 m.

### Why not fix the rotation angle instead

I first tried replacing the truncated literal with a full-precision π/2, which also gives 0 overlaps. Two reasons I abandoned it:

1. **`SimpleConfig` has no π constant and no expression evaluation** — values go through `strtod`, so `M_PI/2` isn't available. It would have to be the bare literal `-1.5707963267948966`, which is unpleasant to read and invites someone to "tidy" it back to something shorter later.
2. **Shortening it doesn't work.** I tested the intermediate `-1.5707963267` (10 decimals, erring *short* of π/2 on the theory it would open a gap) and it still overlaps, at `ovlp=5.78e-09` — worse than the original. Any inexact right angle tilts a corner in somewhere; only the full 16-digit value clears the tolerance. So the angle route is all-or-nothing on that literal.

The 1 µm nudge sidesteps the precision question entirely and matches the idiom already in these files.

### Note for a follow-up

Two sibling volumes carry the same truncated literal:

- `bldg/centralStairwellUpperStairs1.txt:11`
- `bldg/centralStairwellUpperStairs2.txt:11`

Neither currently overlaps anything, so they're out of scope here. Worth knowing they're the same latent pattern if a new volume is ever placed against them. @sdifalco — these are your files, happy to defer if you'd rather handle the family differently.

### DIFF
--- a/Mu2eG4/geom/bldg/dirtCentralStairwellLower3.txt
+++ b/Mu2eG4/geom/bldg/dirtCentralStairwellLower3.txt
@@ -12,11 +12,13 @@ double dirt.central.stairwell.lower3.offsetFromMu2eOrigin.z   = -3581.4;
 
 double dirt.central.stairwell.lower3.yHalfThickness           = 1260.45;
 
+// the face at 9956.8 abuts CentralStairwellLowerStairs; add 1 um to prevent
+// conflict as that volume's rotation angle is not exactly pi/2
 vector<double> dirt.central.stairwell.lower3.xPositions = {
-  9956.8,   // g1
+  9956.801,   // g1
   11617.2,   // g2
   11617.2,    // g3
-  9956.8   // g4
+  9956.801   // g4
 };
 
 vector<double> dirt.central.stairwell.lower3.yPositions = {


oksuzian:
Replaced the original approach (full-precision pi/2 on the rotation literal) with a 1 um face nudge — same result, 0 overlaps, but no 16-digit constant to maintain. Force-pushed; the PR body above describes the current change. Details on why shortening the angle literal doesn't work are in the "Why not fix the rotation angle instead" section.

sdifalco (approved):

oksuzian:
@FNALbuild run build test

Issue #1924, opened 2026-08-09
MDC2020 and other calorimeter_CsI_v1 geometries throw: No such parameter calorimeter.diskCrystalFile


oksuzian:
`geom_common_MDC2020.txt` and every other geometry on `calorimeter_CsI_v1.txt` now throw at geometry construction:

```
SimpleConfig: No such parameter calorimeter.diskCrystalFile in file
  Offline/Mu2eG4/geom/geom_common_MDC2020.txt
```

Reproduced on `main` @ `1891b3289` via `Mu2eG4/fcl/gdmldump_MDC2020.fcl`; originally reported from a nightly build.

### Cause

#1908 (`521583d3`) added an unconditional read in `GeometryService/src/DiskCalorimeterMaker.cc:76`:

```cpp
calo_->G4Info_.set("diskCrystalFile", config.getString("calorimeter.diskCrystalFile"));
```

but added the key to `calorimeter_CsI_v2.txt` only. The two calorimeter geometry files on `main` now differ by exactly this:

```
$ diff Mu2eG4/geom/calorimeter_CsI_v1.txt Mu2eG4/geom/calorimeter_CsI_v2.txt
22a23
> string calorimeter.diskCrystalFile = "Offline/CalorimeterGeom/data/crystalPos.txt";
86c87
< double calorimeter.FEBToDiskZOffset = 1.98;
---
> double calorimeter.FEBToDiskZOffset = 35;
```

`SimpleConfig::getString` with no default throws when the key is absent, so any geometry reaching v1 dies at construction — before any event is processed.

### Blast radius

Resolving the include graph over all 53 top-level `geom*.txt` on `main`, three reach v1 and therefore throw:

| Top-level geometry | Reaches |
|---|---|
| `geom_common_MDC2020.txt` | `calorimeter_CsI_v1.txt` |
| `geom_2021_PhaseI_v03.txt` | `calorimeter_CsI_v1.txt` |
| `geom_reduced_DSTS_shielding.txt` | `calorimeter_CsI_v1.txt` |

The first two are the same chain: `geom_common_MDC2020.txt` → `geom_2021_PhaseI_v03.txt` → `calorimeter_CsI_v1.txt`.

Known consumers:

- **Offline**: `Mu2eG4/fcl/gdmldump_MDC2020.fcl`, `EventDisplay/fcl/EventDisplayMDC2020.fcl`
[...]

PR #1929, opened 2026-08-10
Fixed CalorimeterCsI_v1 issue

bechenard:
This is to address issue 1924. I added a file holding the nominal position for all crystals used in v1, compatible up to double rounding errors (<1 nm). I prefer this solution to avoid a silent use of the default geometry if there is no file issued (warning messages are easily ignored). Also note that the nominal geometry couldn't be constructed from first principled since someone decided to remove crystals "too close to the outer ring" without any algorithm. 

I can reinstate the default algorithm if there is a strong feeling for it, but I would prefer this solution, I find it more robust.


oksuzian (approved):
### PR Review Summary — Mu2e/Offline#1929, "Fixed CalorimeterCsI_v1 issue"

Reviewed at head `e29bdd76` (2026-08-10). First pass. This closes [#1924](https://github.com/Mu2e/Offline/issues/1924), which I opened, so the one thing I owed this review is an independent check that MDC2020 crystal positions are actually preserved — not a reading of the PR body. I did that by reimplementing the pre-#1908 `fillCrystalsIdeal()` and comparing it to the shipped file, and it holds.

**Decision**
- 🟢 Approve. The fix is complete, MDC2020 geometry is reproduced to 1.2 × 10⁻¹³ mm, and I think you picked the better of the two options on the issue thread.

**Scope understood**
- `Mu2eG4/geom/calorimeter_CsI_v1.txt` gains the `calorimeter.diskCrystalFile` key that #1908 made mandatory in `DiskCalorimeterMaker.cc:76`, pointing at a new file rather than at v2's `crystalPos.txt`.
- `CalorimeterGeom/data/crystalPosNominal.txt` (new, 1348 lines) holds the nominal positions for the 674 crystals per disk that v1 used before #1908.

---

**On the design choice** — you took option (a) over @sdifalco's preference for option 1, and I agree with you.

My issue text argued for restoring an idealized fallback; your objection is the right one and it contradicts the argument I made three paragraphs earlier in the same issue ("warning messages are easily ignored"). A missing `diskCrystalFile` under (a) throws `DISK_OPEN_FAILED` at construction (`Disk.cc:53-54`) — which is also what @sdifalco asked for in his comment ("eventually raising a warning or an exception if a geometry file is not found"), so his ask is satisfied by this implementation rather than overridden by it. Option 1 would additionally have meant reviving `fillCrystalsIdeal()` as a permanent second code path in a class #1908 had just reduced to one. Fail-loud, one path — this is better.

One correction to the body, and it works in your favour: the nominal map **is** reconstructible from first principles. The four hand-removed crystals are hardcoded coordinate tests in the old `fillCrystalsIdeal` (`Disk.cc:60-63` at `521583d3~1`), not an undocumented decision, which is exactly why I could regenerate the file and audit it. That auditability is worth keeping — see follow-up 1.

---

**🟢 Verified — no action needed**

- 🟢 **The shipped file is bit-for-bit the pre-#1908 ideal map.** I reimplemented `Disk::fillCrystalsIdeal()` and `SquareShiftMapper` from `521583d3~1` and ran it with v1's parameters — cell size 34.3 mm (`2×(crystalXYLength/2 + 2×wrapperThickness/2)` = `2×(17.0+0.15)`), `diskCrystalRIn = 374`, `diskCrystalROut = 660`, `nRingsMax = int(1.5×660/34.3) = 28`, `nCrystalMax(28) = 2437` map slots — including `isInsideDisk` over the five apex points and the four hand-removals at `(±257.25, ±583.1)`. Result against `crystalPosNominal.txt`:
  ```
  ideal crystals/disk  = 674
  disk 0: 674 crystals, crystal-id order mismatches = 0, max |file - ideal| = 1.17e-13 mm
  disk 1: 674 crystals, crystal-id order mismatches = 0, max |file - ideal| = 1.17e-13 mm
  ```
  Count, ordering and coordinates all match, and 674/disk is `CaloConst::_nCrystalPerDisk` on the nose. The ordering check matters as much as the coordinates: `fillCrystals` assigns crystal IDs by increasing `mapIdx` (`Disk.cc:85-88`), exactly as `fillCrystalsIdeal` did, so a file in a different order would have silently renumbered every MDC2020 crystal. It does not. Your "<1 nm" is conservative by four orders of magnitude at the file level.

- 🟢 **Nothing else changed underneath that comparison.** `SquareShiftMapper`'s `lk`/`index`/`ring`/`xyFromIndex`/`nCrystalMax` are algorithmically identical pre- and post-#1908 (the diff is `int`→`unsigned`, formatting, and `float`→`double` in one `sqrt` argument). `DiskCalorimeterMaker` computes `crystalCellRadius`, `innerCrysRadius` and `outerCrysRadius` from the same keys with the same expressions at `521583d3~1:170-177` and `main:170-177`, and those four keys are unchanged in `calorimeter_CsI_v1.txt`. So the ideal map I generated is the one MDC2020 actually used.

- 🟢 **Byte-identical to `crystalPos.txt` today, and that is the point rather than an accident** — `md5 f3041c2e8e466bb72cd7454b7c7dfdc6` for both. v1 and v2 differ only in `FEBToDiskZOffset` (1.98 vs 35), which does not enter crystal x/y, so the ideal map is common to them. Duplicating 1348 lines looks redundant right now; it is what keeps MDC2020 frozen on the day `crystalPos.txt` is replaced by real survey positions, which is the whole reason that file exists. Worth one line in the file header so the next person does not "helpfully" de-duplicate them (follow-up 1).

- 🟢 **The fix covers every affected geometry.** `Mu2eG4/geom/` contains exactly two calorimeter files (`calorimeter_CsI_v1.txt`, `calorimeter_CsI_v2.txt`) and both now carry the key. I re-checked the three top-level geometries I listed in #1924 — `geom_common_MDC2020.txt`, `geom_2021_PhaseI_v03.txt`, `geom_reduced_DSTS_shielding.txt` — and **none of them overrides any `calorimeter.*` parameter**, so all three inherit v1 unmodified and the shipped map is correct for all three. That mattered: `fillCrystals` silently `continue`s on an unmatched map slot, so a geometry that shifted a radius would have quietly produced a disk with fewer crystals and no diagnostic.

- 🟢 **The new file is installed.** `CalorimeterGeom/CMakeLists.txt:20` is `install(DIRECTORY data DESTINATION ...)` — the whole directory, so `crystalPosNominal.txt` ships with no build-file change. This is the rule added by #1914, and it is the third piece of #1908 data-side fallout, so I checked it explicitly rather than assuming.

- 🟢 **`checkPosition` will not throw on this map.** The new pairwise overlap guard (`Disk.cc:94-107`) rejects a crystal when `dx < 0.499×(34.3+34.3)` **and** `dy < ...`, i.e. below 34.2314 mm in both axes. On the ideal lattice the nearest same-row neighbour is 34.3 mm in x and the nearest adjacent-row neighbour is 34.3 mm in y — a 0.0686 mm margin, three orders of magnitude above the float precision the check runs at. No CI job constructs v1, so this one is worth stating rather than inferring from a green suite.

- 🟢 **CI green at this head** (build 3288, `e29bdd76` merged into base `5c9ca42a`): `build (prof)`, all 12 art jobs, `rootOverlaps`, `g4surfaceCheck`, `trigger`, `check_cmake`, clang-tidy 0/0, whitespace clean. It exercises v2, not v1 — see residual risk — but the disk-filling code path is identical and only the filename string differs. I did not re-trigger.

---

**Findings** (none gating)

1. ⚪ [S3] The runtime deviation is ~27 nm, not <1 nm — the reader truncates to `float`.
   - Evidence: `Disk::fillCrystals` declares `float x,y;` (`Disk.cc:58`) and stores `realPos.emplace_back(x,y,0)` into a `Hep3Vector` of doubles. Round-tripping every value in the file through `float` gives a maximum displacement of **2.73 × 10⁻⁵ mm (27.3 nm)**, worst at disk 0 crystal 641 (`531.65, 308.7` → `531.6500244, 308.7000122`). That is 10⁸ times larger than the file's own 1.2 × 10⁻¹³ mm, so `float` — not the file — is what sets the agreement with the pre-#1908 geometry.
   - Impact: none physically. 27 nm against a 34 mm crystal and mm-scale cluster resolution is not a measurable quantity, and it applies equally to v2. Raising it only so the number in the record is the right one.
   - Suggested fix: `double x,y;` in the reader, whenever this file is next touched. One word, and then "reproduces the ideal map exactly" is true end to end.

2. ⚪ [S3] The data file carries no provenance, and the reader makes it awkward to add.
   - Evidence: `crystalPosNominal.txt` starts at `0 0 -188.65 377.3` with no header. The obvious remedy is a `#` comment line — but the reader is `while (crysFile >> did >> cidx >> x >> y)`, which stops at the first token that will not parse as a number. A comment line would silently truncate the crystal list rather than being skipped, and a short list is accepted without complaint (unmatched map slots just `continue`).
   - Impact: the one file that pins a frozen campaign's geometry has nowhere to say so. That is how it gets "cleaned up" as a duplicate of `crystalPos.txt` in a year.
   - Suggested fix: teach the reader to skip lines beginning with `#` (two lines), then head the file with what it is, that it is frozen for MDC2020, and how it was generated. Alternatively record it in `CalorimeterGeom/data/README` or next to the `diskCrystalFile` line in `calorimeter_CsI_v1.txt`, which costs nothing today.

3. ⚪ [S3] A short or partially-mismatched crystal file is still accepted silently.
   - Evidence: `fillCrystals` throws on out-of-order indices (`realPos.size() != cidx`) and on a file it cannot open, but a file that simply ends early, or whose positions do not land within half a cell of any map slot, yields a disk with fewer crystals and no message.
   - Impact: pre-existing in #1908's reader, not introduced here — but this PR doubles the number of position files that have to stay consistent, and `CaloConst::_nCrystal` already states the expected total. Comparing the two after the fill would turn a silent geometry change into a startup error.
   - Suggested fix: after the loop, `if (nCrystal != expected) throw`. Not for this PR.

4. ⚪ [S3] The body says "This is to address issue 1924" without a closing keyword — `Fixes #1924` would close it on merge.

---

**Validation check**
- Build/tests run by CI: **yes, fully green at `e29bdd76`** (build 3288).
- Build/tests run by reviewer: an independent Python reimplementation of `fillCrystalsIdeal` + `SquareShiftMapper` at `521583d3~1`, run with v1's parameters and diffed against the shipped file (results above). No Offline build.
- Config contract check: **pass** — both calorimeter geometry files now supply `calorimeter.diskCrystalFile`; no top-level geometry overrides the parameters the map depends on.
- Cross-repo consistency: **pass** — no code change, so nothing to update in `Production` or `mu2e-trig-config`. `Production@MDC2020`'s `JobConfig/common/epilog.fcl` and its validation fcl are unblocked by this without an edit.
- Build-file check: **pass** — covered by the existing `install(DIRECTORY data ...)`.

**Residual risk**
- **Nothing in Offline CI constructs a frozen-campaign geometry**, so this fix is verified by construction and by CI-on-v2, not by a job that loads v1. That is the same gap that let #1924 through, and the `SurfaceIdEnum invalid enum name : CRV_C3` breakage before it. A `gdmldump` over the frozen top-level geoms is cheap and would have caught both at PR time; I raised it in #1924 and it is still the systemic fix, separate from this PR.
- The 27 nm `float` truncation applies to v1 and v2 alike, so MDC2020 reconstruction is reproducible to that scale rather than exactly. Irrelevant physically; stated so the record is accurate.

**Author follow-ups** (none blocking)
1. Say somewhere durable what `crystalPosNominal.txt` is, that it is frozen for MDC2020, and that being identical to `crystalPos.txt` today is deliberate — ideally in the file, which needs the two-line comment-skip in the reader first (finding 2).
2. `float` → `double` in the `fillCrystals` reader next time you are in there (finding 1).
3. `Fixes #1924` in the body (finding 4).
4. Separately, and not yours alone: a geometry-construction smoke test over the frozen top-level geoms.

PR #1930, opened 2026-08-10
add expected Proditions get calls

rlcee:
These calls should be here to be consistent with the standard pattern.  In briefly reviewing the current calling code, I think they may be redundant, with negligible cost, but until all the Proditions can be rewritten, possibly, with a more compact approach, I'd prefer the patterns are consistent across the code base.

### DIFF
--- a/TrackerConditions/inc/AlignedTrackerCache.hh
+++ b/TrackerConditions/inc/AlignedTrackerCache.hh
@@ -54,6 +54,10 @@ namespace mu2e {
         DbIoV iov;
         iov.setMax(); // start with full IOV range
         if(_useDb) {
+          _tatr_p->get(eid);
+          _tapl_p->get(eid);
+          _tapa_p->get(eid);
+          _tast_p->get(eid);
           iov.overlap(_tatr_p->iov());
           iov.overlap(_tapl_p->iov());
           iov.overlap(_tapa_p->iov());
--- a/TrackerConditions/inc/TrackerStatusCache.hh
+++ b/TrackerConditions/inc/TrackerStatusCache.hh
@@ -44,6 +44,10 @@ namespace mu2e {
         DbIoV iov;
         iov.setMax(); // start with full IOV range
         if(_useDb) {
+          _tpls_p->get(eid);
+          _tpas_p->get(eid);
+          _tssl_p->get(eid);
+          _tsss_p->get(eid);
           iov.overlap(_tpls_p->iov());
           iov.overlap(_tpas_p->iov());
           iov.overlap(_tssl_p->iov());


oksuzian (approved):
### PR Review Summary — Mu2e/Offline#1930, "add expected Proditions get calls"

Reviewed at head `dd0a532f` (2026-08-10). First pass. Eight lines, so the useful work here was checking the two claims in the body — that the calls are redundant, and that this makes the pattern consistent across the code base — rather than re-reading the diff.

**Decision**
- 🟢 Approve. Both claims check out, and after this merges the pattern is uniform with **no** remaining exceptions in Offline.

**Scope understood**
- `AlignedTrackerCache::makeIov` and `TrackerStatusCache::makeIov` each gain the four `get(eid)` calls on their `DbHandle`s ahead of the `iov()` reads, inside the existing `if(_useDb)`.

---

**🟢 Verified — no action needed**

- 🟢 **"They may be redundant" — confirmed, not assumed.** `ProditionsCache::update` calls `makeEntity(eid)` → `makeSet(eid)` → `makeIov(eid)` in that order (`Mu2eInterfaces/inc/ProditionsCache.hh:112-115`). In both of these caches `makeEntity` calls `getPtr(eid)` on all four handles and `makeSet` calls `get(eid)` on all four, and `getPtr` is `get(eid); return _table;`. So every handle is already current twice over by the time `makeIov` runs. I also searched for other callers of `makeIov`: every hit in the repo is a definition, so `ProditionsCache::update` is the only caller and the redundancy holds universally, not just on the common path.

- 🟢 **"Negligible cost" — confirmed.** `DbHandle<T>::get` returns immediately when `current(eid)` is true, and `current` is `_liveTable.iov().inInterval(run, subrun)` — a couple of integer comparisons, no engine call. And `makeIov` only runs on a cache miss, i.e. at IOV boundaries, not per event. Eight extra `inInterval` calls per boundary crossing.

- 🟢 **The consistency claim is now exact — I swept every cache rather than spot-checking.** Across all `ProditionsCache` subclasses in Offline, every one that reads a `DbHandle::iov()` calls `get(eid)` on that handle first: `MVACatalogCache` (and `TrkQualCatalogCache`, which is just a typedef of it), `CalCalibCache`, `CalDAQMapCache`, `CalSimParamsCache`, `AlignedCalorimeterCache`, `CRVCalibCache`, `CRVPhotonYieldCache`, `CRVStatusCache`, `STMEnergyCalibCache`, `SimBookkeeperCache`, `StrawElectronicsCache`, `StrawPhysicsCache`, `StrawResponseCache`, `TrackerPanelMapCache`, plus the two in this PR. `CRVOrdinalCache`, `EventTimingCache`, `FullReadoutStrawCache` and `StrawDriftCache` return `setMax()` with no handle read, so they have nothing to pair. `AlignedTrackerCache` and `TrackerStatusCache` were the only two exceptions in the repo, and this PR is both of them. (`GlobalConstantsService/inc/MassCache.hh` matches the filename pattern but is a PDG mass memo, not a Proditions cache.)

- 🟢 **The `AlignedCalorimeterCache` case is already closed.** #1916 merged into `main` as `bb4b00e6` shortly before this review, and its `makeIov` carries `_tadisk_p->get(eid)` — the omission you flagged there was fixed before merge, so this PR does not leave a third instance behind. Worth stating explicitly because `AlignedTrackerCache` was the donor those copies came from: fixing it here removes the source, not just the symptoms.

- 🟢 **`AlignedTrackerCache` is a template, so both `AlignedTrackerCacheReco` and `AlignedTrackerCacheSim` pick this up.** No second edit needed for the Sim tables.

- 🟢 **CI fully green at this head** (build 3290, `dd0a532f` merged into base `0e06c26f`): `build (prof)`, all 12 art jobs, `rootOverlaps`, `g4surfaceCheck`, `trigger`, `check_cmake`, clang-tidy 0/0, whitespace clean. This PR's base predates the #1916 merge; the files do not overlap and GitHub reports `MERGEABLE` / `CLEAN`.

---

**Findings**

1. ⚪ [S3] Worth one comment saying why these calls stay, given they are provably dead today.
   - The thing making them non-redundant is not the current code — it is that `makeIov` should not depend on `ProditionsCache::update`'s call order, which lives in a different package and is invisible from these headers. That coupling is real: `findByRun` keys the cache on the IOV, so computing the IOV *before* building the entity is arguably the more natural order, and under the pre-PR code that reordering would have produced an IOV built from whichever run was loaded last — a silently wrong cache interval, not a crash.
   - Suggested fix: a short comment on `ProditionsCache::makeIov`'s declaration (`ProditionsCache.hh:56`) — "implementations must refresh their own handles; do not rely on update() having called makeEntity/makeSet first" — would keep a future reader from removing these as no-ops. Cheaper than the compact rewrite you mention, and it survives that rewrite.

**Validation check**
- Build/tests run by CI: **yes, green at `dd0a532f`.**
- Build/tests run by reviewer: none. Traced from the two changed headers, `ProditionsCache::update`, `DbHandle::get`/`current`/`iov`, and a sweep of every `*Cache.hh` in the repo plus `SimBookkeeperCache.cc`.
- Config contract check: **n/a** — no FHiCL touched.
- Cross-repo consistency: **pass** — header-only change in `TrackerConditions`, no interface or behaviour change to propagate.

**Residual risk**
- None from this diff. The behaviour is identical today by construction; the change buys insurance against a future reordering of `ProditionsCache::update`.

**Author follow-ups**
1. Optional: the comment in finding 1, so the next reader does not "simplify" these calls back out.

PR #1932, opened 2026-08-14
delete obsolete STM shielding Poly

sdifalco:
Following request from Yongyi.
New STM geometry configuration file (v10) created
All geometry configuration files changed.


Mu2eG4/geom/geom_run2.txt
YongyiBWu, line 1: I think run2 geometry will still have the poly. The only reason it is removed from run1 geometry is because the external shielding downstream blocks are removed in run1. The poly is embedded in the blocks so it cannot float in the air. For run2 geometry the shielding blocks are still there so there is no need to change this file.

YongyiBWu (commented):
I believe the change does not apply to Mu2eG4/geom/geom_run2.txt. All the rest changes are good.

PR #1933, opened 2026-08-14
Haunchfix

sdifalco:
The length of the haunch above the Remote Handling Room Door is changed from 298'' to 274'' following George's measurements.
immagine
immagine

### DIFF
--- a/Mu2eG4/geom/STM_v10.txt
+++ b/Mu2eG4/geom/STM_v10.txt
@@ -1,0 +1,11 @@
+// (Muon) Stoping Target Monitor (STM)
+//
+// Author: Yongyi Wu
+
+#include "Offline/Mu2eG4/geom/STM_v09.txt"
+bool stm.shieldingHolePoly.build = false;
+
+// This tells emacs to view this file in c++ mode
+// Local Variables:
+// mode:c++
+// End:
--- a/Mu2eG4/geom/bldg/remoteHandlingCeilingStrip.txt
+++ b/Mu2eG4/geom/bldg/remoteHandlingCeilingStrip.txt
@@ -21,8 +21,8 @@ vector<double> building.remote.handling.ceiling.strip.xPositions = {
   3708.4,   // RHCS08
   3708.4,   // RHCS09
   1544.32,   // RHCS10
-  -4140.2,   // RHCS11
-  -4140.2,   // RHCS12
+  -4749.8,   // RHCS11
+  -4749.8,   // RHCS12
   -11709.4   // RHCS13
 };
 
--- a/Mu2eG4/geom/geom_reduced_DSTS_shielding.txt
+++ b/Mu2eG4/geom/geom_reduced_DSTS_shielding.txt
@@ -60,7 +60,7 @@ double mu2e.detectorSystemZ0 = 10171.;   // mm  G4BL: (17730-7292=9801 mm)
 #include "Offline/Mu2eG4/geom/TSdA_v02.txt"
 #include "Offline/Mu2eG4/geom/muonBeamStop_v08.txt"
 
-#include "Offline/Mu2eG4/geom/STM_v09.txt" // (muon) stopping target monitor
+#include "Offline/Mu2eG4/geom/STM_v10.txt" // (muon) stopping target monitor
 
 // Proton Absorber
 #include "Offline/Mu2eG4/geom/protonAbsorber_cylindrical_v04.txt"
--- a/Mu2eG4/geom/geom_run1.txt
+++ b/Mu2eG4/geom/geom_run1.txt
@@ -60,7 +60,7 @@ double mu2e.detectorSystemZ0 = 10171.;   // mm  G4BL: (17730-7292=9801 mm)
 #include "Offline/Mu2eG4/geom/TSdA_v02.txt"
 #include "Offline/Mu2eG4/geom/muonBeamStop_v08.txt"
 
-#include "Offline/Mu2eG4/geom/STM_v09.txt" // (muon) stopping target monitor
+#include "Offline/Mu2eG4/geom/STM_v10.txt" // (muon) stopping target monitor
 
 // Proton Absorber
 #include "Offline/Mu2eG4/geom/protonAbsorber_cylindrical_v05.txt"
--- a/Mu2eG4/geom/geom_run1_a.txt
+++ b/Mu2eG4/geom/geom_run1_a.txt
@@ -60,7 +60,7 @@ double mu2e.detectorSystemZ0 = 10171.;   // mm  G4BL: (17730-7292=9801 mm)
 #include "Offline/Mu2eG4/geom/TSdA_v02.txt"
 #include "Offline/Mu2eG4/geom/muonBeamStop_v08.txt"
 
-#include "Offline/Mu2eG4/geom/STM_v09.txt" // (muon) stopping target monitor
+#include "Offline/Mu2eG4/geom/STM_v10.txt" // (muon) stopping target monitor
 
 // Proton Absorber
 #include "Offline/Mu2eG4/geom/protonAbsorber_cylindrical_v04.txt"
--- a/Mu2eG4/geom/geom_run1_a_stickman.txt
+++ b/Mu2eG4/geom/geom_run1_a_stickman.txt
@@ -60,7 +60,7 @@ double mu2e.detectorSystemZ0 = 10171.;   // mm  G4BL: (17730-7292=9801 mm)
 #include "Offline/Mu2eG4/geom/TSdA_v02.txt"
 #include "Offline/Mu2eG4/geom/muonBeamStop_v08.txt"
 
-#include "Offline/Mu2eG4/geom/STM_v09.txt" // (muon) stopping target monitor
+#include "Offline/Mu2eG4/geom/STM_v10.txt" // (muon) stopping target monitor
 
 // Proton Absorber
 #include "Offline/Mu2eG4/geom/protonAbsorber_cylindrical_v05.txt"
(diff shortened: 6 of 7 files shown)


oksuzian (approved):
### PR Review Summary — Mu2e/Offline#1933

Reviewed at head `f2d98e8457998de19dce3af656223644d47d5907`. First pass.

**Decision**
- 🟢 approve

The haunch change itself is arithmetically exact and geometrically consistent
with the surrounding building volumes — details under "Verified" below. The two
🟡s are description/maintenance issues, not defects in the geometry.

**Scope understood**
- `bldg/remoteHandlingCeilingStrip.txt`: vertices RHCS11/RHCS12 move in x from
  `-4140.2` to `-4749.8`, shortening the haunch above the Remote Handling Room
  door per George's measurements.
- The other six files are **not** part of the haunch fix: they are commit
  `d53f259f`, which is the entirety of open PR #1932 ("delete obsolete STM
  shielding Poly"). This branch is stacked on that one.

**Findings**

1. 🟡 [S2] The PR silently contains all of #1932, and the description does not say so
   - Evidence: `gh pr view 1933 --json commits` lists `d53f259f` ("delete
     obsolete STM shielding Poly") followed by `f2d98e84` ("change haunch
     longitudinal length"). `d53f259f` is the *same SHA* as the head of #1932.
     The `STM_v10.txt` blobs in the two PRs are identical. The body of #1933
     describes only the haunch, so six of the seven changed files arrive with no
     explanation to a reader of this PR.
   - Impact: a reviewer who reads the description and approves is approving an
     STM geometry change (`stm.shieldingHolePoly.build = false`, applied to five
     geometry configurations) that they were never told was here. If #1933
     merges first, that change lands under the title "Haunchfix" and #1932
     becomes an empty PR. This is a description and merge-order problem, not a
     content problem — the split into two PRs is itself correct.
   - Suggested fix: add one line to the body — "stacked on #1932; merge that
     first and this reduces to the two-vertex haunch change" — and merge in that
     order. No rebase needed; git will handle the shared commit cleanly.

2. 🟡 [S2] Hand-edit to a file marked auto-generated, with no annotation
   - Evidence: `remoteHandlingCeilingStrip.txt` opens with "SimpleConfig
     geometry file automatically produced for original file: geom/
     geom_remote_ceilingStrip.ccl". No `.ccl` file exists anywhere in Offline
     (`search/code?q=extension:ccl` → 0 results), so the generator input lives
     outside the repo and nothing here records that these two numbers were
     changed by hand.
   - Impact: the next regeneration from an unmodified `.ccl` silently reverts
     the fix, and there is no in-tree signal that it happened. Silent is the
     operative word — a reverted haunch produces a valid job with slightly wrong
     shielding.
   - Suggested fix: the neighbouring files already have a convention for exactly
     this — `psAreaCeilingN.txt` carries `3349.22, // P0 mod AH was 3327.4`.
     Matching it here (`-4749.8, // RHCS11 mod was -4140.2, George's
     measurement`) costs nothing. Separately, please confirm whether the
     upstream `.ccl` has been updated too.

3. ⚪ [S3] `STM_v10.txt` records no reason or reference for the removal
   - Evidence: the new file is `#include STM_v09` plus
     `bool stm.shieldingHolePoly.build = false;` with no comment. Its parent
     `STM_v09.txt` cites "DocDB-49112v2" on essentially every dimension it sets.
   - Impact: "obsolete" appears only in #1932's title, which is not reachable
     from the geometry file a year from now.
   - Suggested fix: one comment line naming the request or DocDB reference.
     (This one really belongs on #1932 — noted here only because the file
     arrives in this diff.)

- 🟢 The haunch length is exact. The edge RHCS12→RHCS13 measures
  `|-11709.4 - (-4140.2)| = 7569.2 mm = 298.000″` before and
  `|-11709.4 - (-4749.8)| = 6959.6 mm = 274.000″` after; the shift is
  `609.6 mm = 24.000″`. Both figures land on the stated inch values to the
  micron, which is a strong sign the number came from the drawing rather than
  from a fit.
- 🟢 The new corner lands on an existing structural line. In absolute Mu2e
  coordinates the moved vertex goes to `(x, z) = (-3802.4, 2311.4)`, which is a
  corner shared by four other *active* hall volumes — `psAreaCeilingN`,
  `psAreaHatchLipN`, `psAreaHatchLower`, `psAreaHatchUpper`. The old value
  `(-3192.8, 2311.4)` coincided with nothing. I scanned all 275 active volumes
  reachable from `mu2eHall_v04.txt` to establish this.
- 🟢 No overlap is introduced or removed. I sampled the vacated patch
  (609.6 × 355.6 × 762 mm at floor+4419.6…5181.6) against every active hall
  volume: it was occupied by nothing but the strip before, and becomes hall air
  after. The retained strip interior overlaps nothing. Directly above the
  removed slab is `psAreaCeilingE`, which stays — so this shortens a haunch
  hanging below a ceiling that remains in place, rather than opening a hole
  through it. Net effect is ~0.165 m³ of `CONCRETE_MARS` removed.
- 🟢 The door is still fully spanned. `remoteHandlingDoorFrame` occupies
  x ∈ [-7721.6, -6121.4]; the shortened haunch still runs to -4749.8, clearing
  the door's near edge by 1.37 m.
- 🟢 Only `xPositions` changed. `yPositions`, `yHalfThickness` and the three
  offsets are untouched, so this is a pure in-plane vertex move.
- 🟢 The STM half is safe at the config level. `constructSTM.cc:248` reads
  `_config.getBool("stm.shieldingHolePoly.build", false)` and every
  sub-parameter (`gap`, `halfLength`, `rIn`, `rOut`, `material`) is read *inside*
  that `if`, so setting the flag false cannot trip a missing-parameter throw.
- 🟢 `geom_2021_PhaseI_v03.txt` is correctly left on `STM_v09`. It is the sixth
  file referencing v09 on main and the only one not bumped; it is reached solely
  by `geom_common_MDC2020.txt`, a frozen campaign configuration that should not
  move. Leaving it is right, not an oversight.
- 🟢 The default geometry does receive both changes:
  `standardServices.fcl` → `geom_common.txt` → `geom_run1_a_stickman.txt`, which
  this PR updates to `STM_v10` and which loads `mu2eHall_v04.txt`.

**Validation check**
- Build/tests run: CI is **pending** at this head — `mu2e/buildtest` and
  `mu2e/muDauSteps` are still running; `ceSimReco`, `ceDigi`, `ceSteps`,
  `cosmicSimReco`, `cosmicOffSpill`, `g4test_03MT`, `g4study`, `POT` and
  `transportOnly` are green. A run is already in progress, so I did not trigger
  another. #1932, which carries the STM half, is fully green at `d53f259f`.
- Config contract check: pass. All five geometry files that should point at
  `STM_v10` do; the one left at `v09` is deliberately frozen.
- Cross-repo consistency: n/a — geometry text files only, no code or FHiCL
  interface changes.

**Residual risk**
- My overlap analysis is an analytic 2-D polygon-plus-height test over the hall
  building volumes only. It does **not** cover dirt/berm volumes, detector
  geometry, or G4's own tolerance behaviour, and it does not reproduce the
  ExtrudedSolid placement rotation exactly. It is good evidence, not a
  substitute for the real check — see follow-up 2.
- The shielding effect of removing ~0.165 m³ of concrete above the remote
  handling door is not something I can quantify. It is small and local, but I
  have not checked it against any dose or background requirement.
- I reviewed the STM change only for configuration safety, not for physics
  justification. That belongs on #1932 and I have not assessed it there.

**Author follow-ups**
1. Add a line to the description saying this is stacked on #1932, and merge
   #1932 first.
2. Please run the geometry overlap check and report the outcome —
   `Mu2eG4/geom/geom_SurfaceCheck.txt` sets `g4.doSurfaceCheck = true` and
   includes `geom_common_current.txt`, so it covers exactly the chain this
   change lands in. Nothing in CI runs it, and a vertex move is precisely the
   change class it exists for.
3. Annotate the two hand-edited vertices in the style already used in
   `psAreaCeilingN.txt`, and confirm whether the upstream
   `geom_remote_ceilingStrip.ccl` has been updated so a future regeneration does
   not revert this.

oksuzian (commented):
### Correction to my review — Mu2e/Offline#1933

Reviewed at head `f2d98e8457998de19dce3af656223644d47d5907`. Correcting my own earlier review; the decision is unchanged (🟢 approve).

**I was wrong that CI does not check geometry overlaps. It does, and it already passed on this PR.**

`mu2e/buildtest` runs two overlap checks:

- **`g4surfaceCheck`** — runs exactly the job I asked you to run,
  `mu2e -c Offline/Mu2eG4/fcl/surfaceCheck.fcl`. It is graded properly, not on
  return code alone: `codetools/.../build.sh` counts
  `grep 'Checking overlaps for volume' | grep -v OK` and fails the job if that
  count is non-zero, with a `LEGAL > 0` guard so a run that checked nothing
  cannot pass vacuously.
- **`rootOverlaps`** — `codetools/bin/rootOverlaps.sh`, an independent check
  via ROOT's geometry checker.

Both are ✅ at `f2d98e8`, in FNALbuild's result table for build 3294. So the
geometry is covered by the project's own tooling over the whole geometry,
including the dirt and detector volumes my hand analysis did not reach and
G4's own overlap tolerances.

**Please disregard author follow-up 2** in my review — asking you to run
`geom_SurfaceCheck.txt` was redundant work. Follow-ups 1 (state the #1932
dependency in the description) and 3 (annotate the hand-edited vertices, and
confirm the upstream `.ccl`) still stand.

**Also updating the validation line:** at the time I posted, `mu2e/buildtest`
and `mu2e/muDauSteps` were pending. Both are now green, and the overall status
at `f2d98e8` is success.

**Why I got it wrong**, since it may be a useful trap to know about: these
checks have no separate commit-status context. The statuses API at this head
returns eleven contexts and none of them is an overlap check, so reading it
alone gives the wrong answer. `rootOverlaps`, `g4surfaceCheck`, `ceMix`,
`check_cmake`, FIXME/TODO, clang-tidy and the whitespace check all run *inside*
`mu2e/buildtest` and appear only as rows in FNALbuild's comment table — twenty
rows against eleven statuses. The table is the authoritative list of what ran.

PR #1937, opened 2026-08-22
‎RecoDataProducts/inc/CrvStatus: adjust data type and adding accessor for CRV status 

corrodis:
Tagging @ehrlich-uva for comments

### DIFF
--- a/CRVTools/src/CrvDigiDump_module.cc
+++ b/CRVTools/src/CrvDigiDump_module.cc
@@ -81,7 +81,7 @@ namespace mu2e {
       UChar_t status_dtcID_[maxROCs_];   // DTC ID (small number)
       UChar_t status_statusByte_[maxROCs_]; // Status byte (0-255)
       UChar_t status_linkStatus_[maxROCs_]; // Link status from subevent header
-      UChar_t status_linkLatency_[maxROCs_]; // Link latency from subevent header
+      UShort_t status_linkLatency_[maxROCs_]; // Link latency from subevent header
       Bool_t status_valid_[maxROCs_];    // Valid flag
   };
 
@@ -135,7 +135,7 @@ namespace mu2e {
     statusTree_->Branch("dtcID", status_dtcID_, "dtcID[nStatus]/b");          // UChar_t
     statusTree_->Branch("headerStatus", status_statusByte_, "headerStatus[nStatus]/b"); // UChar_t
     statusTree_->Branch("linkStatus", status_linkStatus_, "linkStatus[nStatus]/b"); // UChar_t
-    statusTree_->Branch("linkLatency", status_linkLatency_, "linkLatency[nStatus]/b"); // UChar_t
+    statusTree_->Branch("linkLatency", status_linkLatency_, "linkLatency[nStatus]/s"); // UShort_t
 
     // ROC header information (same indexing as status)
     statusTree_->Branch("wordCount", status_roc_controllerEventWordCount_, "wordCount[nStatus]/s"); // UShort_t
--- a/RecoDataProducts/inc/CrvStatus.hh
+++ b/RecoDataProducts/inc/CrvStatus.hh
@@ -65,8 +65,10 @@ namespace mu2e
       uint8_t GetLinkStatus() const                {return _linkStatus;}
       uint16_t GetLinkLatency() const               {return _linkLatency;}
       std::vector<CRVDataDecoder::CRVROCStatusPacketFEBII> &GetROCHeader() {return _rocHeader;}
+      bool HasROCHeader() const                    {return !_rocHeader.empty();}
 
       // --- MicroBunchStatus accessors (32-bit CRV ROC status word) ---
+      // Bit layout from ROC firmware: https://github.com/Mu2e/CRVFirmware/tree/main/ROC/FPGA1
       // The ROC header is pushed after construction, so these read lazily
       // from _rocHeader[0].  Returns 0 / false when no ROC header is present.
 
--- a/RecoDataProducts/inc/CrvStatus.hh
+++ b/RecoDataProducts/inc/CrvStatus.hh
@@ -63,8 +63,37 @@ namespace mu2e
       uint8_t GetDTCID() const                     {return _dtcID;}
       uint64_t GetSubeventEventWindowTag() const   {return _subeventEventWindowTag;}
       uint8_t GetLinkStatus() const                {return _linkStatus;}
-      uint8_t GetLinkLatency() const               {return _linkLatency;}
+      uint16_t GetLinkLatency() const               {return _linkLatency;}
       std::vector<CRVDataDecoder::CRVROCStatusPacketFEBII> &GetROCHeader() {return _rocHeader;}
+
+      // --- MicroBunchStatus accessors (32-bit CRV ROC status word) ---
+      // The ROC header is pushed after construction, so these read lazily
+      // from _rocHeader[0].  Returns 0 / false when no ROC header is present.
+
+      // Full 32-bit word
+      uint32_t GetMicroBunchStatus() const {
+        return _rocHeader.empty() ? 0 : _rocHeader[0].GetMicroBunchStatus();
+      }
+
+      // Bits [0:23] — per-port problem flags (3 groups of 8 ports)
+      uint32_t GetPortFlags() const        {return GetMicroBunchStatus() & 0x00FFFFFFu;}
+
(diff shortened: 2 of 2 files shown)


ehrlich-uva (approved):

oksuzian (commented):
### PR Review Summary — Mu2e/Offline#1937

Reviewed at head `73942219db98b5dab84b88e305f60b28f2d2028d`. First pass.

**Decision**
- 🟡 comment only — two S2 items below, and `mu2e/buildtest` is still pending at this head. Nothing here blocks; happy to see it merge once CI is green.

**Scope understood**
- `CrvStatus::_linkLatency` widened from `uint8_t` to `uint16_t`, with the matching branch type change in `CrvDigiDump`.
- Eight new inline accessors on `CrvStatus` that decode the 32-bit ROC `MicroBunchStatus` word from `_rocHeader[0]`.

**Findings**
1. 🟡 [S2] The new accessors return "no problem" when there is no ROC header.
   - Evidence: `GetMicroBunchStatus()` returns `0` on `_rocHeader.empty()`, so `GetFEBOverflow()`, `GetTruncation()`, etc. all return `false` for a link with no ROC attached. `CrvDigiDump` (lines 274-278 on main) deliberately writes `-1` for the same case so it is distinguishable from a clean status.
   - Impact: a downstream consumer that loops over `CrvStatusCollection` and counts error flags will silently count an absent ROC as healthy. Nothing in this PR consumes the accessors yet, so there is no wrong output today, but the convention is baked into a data-product interface.
   - Suggested fix: add `bool HasROCHeader() const {return !_rocHeader.empty();}` and either throw `cet::exception` from the accessors when it is false, or document at the call site that callers must check `HasROCHeader()` first. (`std::optional` would be the natural shape, but the header comment already records that it breaks the dictionary.)

2. 🟡 [S2] Bit assignments are unsourced.
   - Evidence: the bit meanings (24 = FEB µB mismatch, 25 = buffer, ... 31 = truncation) appear nowhere else in `Mu2e/Offline` or `Mu2e/artdaq_core_mu2e`; `CRVROCStatusPacketFEBII::GetMicroBunchStatus()` only assembles the word.
   - Suggested fix: one comment line naming the ROC firmware document / DocDB number the layout comes from, so the next person can check it when the firmware changes.

3. ⚪ [S3] `GetuBMatchError` breaks the capitalization pattern of its siblings (`GetFEBuBMismatch` keeps the `uB`, but the leading lowercase after `Get` reads as a typo). `GetMicroBunchMatchError` or `GetUBMatchError` would scan better. The PR title also carries a leading U+200E (left-to-right mark) before `RecoDataProducts`.

**Verified**
- 🟢 `uint16_t` is the right width: `DTC_SubEventHeader.h` declares every `linkN_drp_rx_latency` as a 16-bit field, so the previous `uint8_t` member truncated the value.
- 🟢 `_rocHeader[0]` is the only entry: `CrvDigisFromArtdaqFragmentsFEBII_module.cc` emplaces one `CrvStatus` per data block (line 187) and pushes exactly one ROC header into it (line 205).
- 🟢 Branch descriptor `linkLatency[nStatus]/s` matches the new `UShort_t` array.

**Validation check**
- Build/tests run: `mu2e/buildtest` pending at `73942219` (triggered by FNALbuild; not re-triggered). Not built locally.
- Config contract check: n/a — no FHiCL change.
- Cross-repo consistency: pass — `DAQ/`, `CRVTools/` and `EventNtuple` do not read `GetLinkLatency()` into a fixed-width type anywhere else.

**Residual risk**
- `mu2e::CrvStatus` has no `ClassVersion` in `classes_def.xml`; ROOT's automatic schema evolution converts `uint8_t` → `uint16_t` on read of older files, so this should be benign, but I did not test reading a pre-change `.art` file.
- PR body is "Tagging @ehrlich-uva for comments" — a one-line description of why the latency was truncated and what the new accessors are for would help the next reader of `git blame`.

**Author follow-ups**
1. Decide how an absent ROC header should surface through the new accessors (finding 1).
2. Cite the source of the `MicroBunchStatus` bit layout (finding 2).
3. Optional: rename `GetuBMatchError` and strip the stray U+200E from the title.

corrodis:
1) I don't think its really possible to have data without a ROC header in data. I think adding a HasROCHeader function and throwing an exception (or warning?) if the other functions are used sounds reasonable but also not too important (aka low priority)
2) I think this is THE place where we do the decoding so it might be ok what we have in here. This comes from firmware here: https://github.com/Mu2e/CRVFirmware/tree/main/ROC/FPGA1. Not the perfect documentation but maybe also not so bad as long as things might still change. 
3) seems too nitpicky. If anyone feels strongly about this, I am perfectly fine with changing it; "U+200E" in the PR title: I really don't care. If this is an issue sure.

oksuzian (approved):
### PR Review Summary — Mu2e/Offline#1937

Reviewed at head `73942219db98b5dab84b88e305f60b28f2d2028d`. Re-review of `73942219` — same head; this pass accounts for the author's replies and the now-complete CI run.

**Decision**
- 🟢 approve

**Scope understood**
- No code change since the first pass. `mu2e/buildtest` has completed green and the author answered all three findings.

**Prior findings**
- 🟡 [S2] Accessors return "no problem" when `_rocHeader` is empty — ANSWERED, deferred. The author reports that data without a ROC header does not occur in practice and agrees a `HasROCHeader()` guard is reasonable but low priority. Not a merge condition; the `-1` sentinel in `CrvDigiDump` remains the only place the distinction is visible.
- 🟢 [was 🟡 S2] Bit layout unsourced — RESOLVED on the thread. Source is the ROC firmware at https://github.com/Mu2e/CRVFirmware/tree/main/ROC/FPGA1, and this header is the one place the word is decoded in Offline. A one-line comment citing that path would still help the next reader, but it is optional.
- ⚪ [S3] `GetuBMatchError` naming / U+200E in title — WITHDRAWN as a finding; author's call, no strong preference here.

**Validation check**
- Build/tests run: `mu2e/buildtest` green at `73942219` (build #3300): build (prof), ceSimReco, ceMix, ceDigi, ceSteps, muDauSteps, trigger, check_cmake all pass; clang-tidy 0 errors 0 warnings; whitespace check clean; TODO/FIXME 0.
- Config contract check: n/a.
- Cross-repo consistency: pass — `uint16_t` matches `DTC_SubEventHeader.h` `linkN_drp_rx_latency : 16`; no other reader of `GetLinkLatency()` stores it in a narrower type.

**Residual risk**
- `CrvStatus` has no `ClassVersion` in `classes_def.xml`; reading pre-change `.art` files relies on ROOT's automatic `uint8_t` → `uint16_t` conversion. Not tested here.

**Author follow-ups**
- None required. Optional: `HasROCHeader()` and a firmware-path comment on the bit layout, whenever convenient.

ehrlich-uva (approved):

oksuzian (approved):
### PR Review Summary — Mu2e/Offline#1937

Reviewed at head `26020b3bbac1d5ca4bb28c8973416cc3f5c92220`. Re-review of `73942219`.

**Decision**
- 🟢 approve

**Scope understood**
- Delta since the approved head is one commit, `26020b3`, touching only `RecoDataProducts/inc/CrvStatus.hh`: a `HasROCHeader()` accessor and a comment citing the firmware source of the bit layout. Two added lines, no behaviour change.

**Prior findings**
- 🟢 [was 🟡 S2] No-ROC-header case — ADDRESSED in `26020b3`, verified: `bool HasROCHeader() const {return !_rocHeader.empty();}` gives callers the check. The accessors themselves still return `0`/`false` when the header is absent, which the author has said is the intended (low-priority) behaviour since such data does not occur; with the guard available that is fine.
- 🟢 [was 🟡 S2] Bit layout unsourced — FIXED in `26020b3`, verified: comment now points at `https://github.com/Mu2e/CRVFirmware/tree/main/ROC/FPGA1`.

**Validation check**
- Build/tests run: `mu2e/buildtest` is pending at `26020b3`; the previous head `73942219` was fully green (build #3300, clang-tidy 0/0, whitespace clean). The delta is a one-line inline accessor plus a comment, so I do not expect a different result. Not re-triggered — a run is already registered.
- Config contract check: n/a.
- Cross-repo consistency: pass, unchanged from the previous pass.

**Residual risk**
- Unchanged: `CrvStatus` carries no `ClassVersion`; reading pre-change files relies on ROOT's automatic `uint8_t` → `uint16_t` conversion, not tested here.

**Author follow-ups**
- None.

oksuzian:
@FNALbuild run build test

PR #1938, opened 2026-08-22
prepare runTool for agents

rlcee:
- allow json output for all switches, preferred for agents
- gate dbtables3 to subdetector configs only
- put a catch in the main so that agents can parse a throw more easily
Merge is very low stakes - only occasional use exists anywhere. CI and Validation are not sensitive to changes

### DIFF
--- a/DbService/inc/RunTool.hh
+++ b/DbService/inc/RunTool.hh
@@ -37,6 +37,17 @@ class RunTool {
  private:
   // read the database via query_engine
   DbReader _reader;
+
+  // Memoized flags() results -- online.run_type/online.run_transition_type
+  // are job-constant, so the first query's result is reused for the
+  // lifetime of this RunTool instance rather than re-querying on every
+  // call. Matters because flags() is called once or twice per run from
+  // runJson() (and, with -a, from printRun()) -- unmemoized, that's two
+  // extra database round-trips per run in the JSON run-listing path.
+  std::map<int, std::string> _runFlagsCache;
+  std::map<int, std::string> _transitionFlagsCache;
+  bool _runFlagsCached = false;
+  bool _transitionFlagsCached = false;
 };
 
 }  // namespace mu2e
--- a/DbService/inc/RunTool.hh
+++ b/DbService/inc/RunTool.hh
@@ -20,9 +20,19 @@ class RunTool {
 
   // fill argument with the names of all flag types
   std::map<int, std::string> flags(FlagType ftype);
+  // Both flag types (run, transition) as a single JSON object -- string
+  // return, same SWIG-compatibility reason as runJson(). Lives on the class
+  // rather than in runTool_main.cc so it's reachable from the Python/SWIG
+  // binding too, not just the CLI.
+  std::string flagsJson();
   RunInfo::RunVec listRuns(const RunSelect& runsel, bool configs = false,
                            bool transitions = false, bool subruns = false);
   void printRun(const RunInfo& info);
+  // Same content as printRun, as a JSON object (serialized text, not a
+  // nlohmann::json return type -- this class is SWIG-wrapped and
+  // nlohmann::json isn't a SWIG-recognized type, same reason
+  // RunInfo::dbTables2/3() and RunConfig::dbTables2/3() return std::string).
+  std::string runJson(const RunInfo& info);
 
  private:
   // read the database via query_engine
--- a/DbService/src/RunConfig.cc
+++ b/DbService/src/RunConfig.cc
@@ -1,11 +1,13 @@
 #include "Offline/DbService/inc/RunConfig.hh"
+
 #include "cetlib_except/exception.h"
-#include "nlohmann/json.hpp"
+
+#include <iostream>
 #include <string>
 #include <utility>
 #include <vector>
-#include <iostream>
 
+#include "nlohmann/json.hpp"
 
 namespace {
 
@@ -68,7 +70,7 @@ void walkCID(const json& node,
   }
 }
 
-}
+}  // namespace
 
 // ---------------------------------------------------------------------------
 // mu2e::RunConfig::dbTables2
--- a/DbService/src/RunInfo.cc
+++ b/DbService/src/RunInfo.cc
@@ -3,10 +3,11 @@
 #include "Offline/GeneralUtilities/inc/TimeUtility.hh"
 
 #include "cetlib_except/exception.h"
-#include "nlohmann/json.hpp"
 
 #include <ctime>
 #include <string>
+
+#include "nlohmann/json.hpp"
 
 using namespace mu2e;
 
@@ -36,10 +37,23 @@ std::string RunInfo::dbTables2(bool qjson) const {
 // requested the individual per-config objects are merged into a single JSON
 // object (an empty object "{}" if nothing is found).
 std::string RunInfo::dbTables3(bool qjson) const {
+  // TRG/Gateway/CFO/DQM configs are excluded from the cat-3 merge -- they
+  // don't carry relevant DBServiceTables content, and including them is what
+  // originally produced colliding keys across configs (see
+  // RUNINFO_DUPLICATE_TABLE below). Shared by both branches deliberately:
+  // this used to live only in the qjson branch below, which meant `runTool
+  // -q` (text) and `runTool -q -j` (JSON) silently reported different table
+  // sets for the same run -- a single predicate used by both means that
+  // can't happen again by the two conditions drifting apart.
+  auto excluded = [](const RunConfig& config) {
+    return config.subsystem() == "TRG" || config.subsystem() == "Gateway" ||
+           config.subsystem() == "CFO" || config.subsystem() == "DQM";
+  };
+
   if (qjson) {
     nlohmann::json merged = nlohmann::json::object();
     for (auto const& config : _configs) {
-      if (config.subsystem() == "TRG" || config.subsystem() == "Gateway" || config.subsystem() == "CFO" || config.subsystem() == "DQM") {
+      if (excluded(config)) {
         continue;
       }
       std::string tables = config.dbTables3(qjson);
@@ -53,8 +67,8 @@ std::string RunInfo::dbTables3(bool qjson) const {
       for (auto it = part.begin(); it != part.end(); ++it) {
         if (merged.contains(it.key())) {
           throw cet::exception("RUNINFO_DUPLICATE_TABLE")
-              << "RunInfo::dbTables3() found duplicate table key \""
-              << it.key() << "\" across run configs\n";
+              << "RunInfo::dbTables3() found duplicate table key \"" << it.key()
+              << "\" across run configs\n";
         }
         merged[it.key()] = it.value();
       }
@@ -65,6 +79,9 @@ std::string RunInfo::dbTables3(bool qjson) const {
   // Non-JSON: concatenate the per-config values, one per line.
   std::string result;
   for (auto const& config : _configs) {
+    if (excluded(config)) {
+      continue;
+    }
     std::string tables = config.dbTables3(qjson);
     if (!tables.empty()) {
       result += tables;
--- a/DbService/src/RunInfo.cc
+++ b/DbService/src/RunInfo.cc
@@ -39,6 +39,9 @@ std::string RunInfo::dbTables3(bool qjson) const {
   if (qjson) {
     nlohmann::json merged = nlohmann::json::object();
     for (auto const& config : _configs) {
+      if (config.subsystem() == "TRG" || config.subsystem() == "Gateway" || config.subsystem() == "CFO" || config.subsystem() == "DQM") {
+        continue;
+      }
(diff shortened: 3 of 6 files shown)


oksuzian (commented):
### PR Review Summary — Mu2e/Offline#1938

Reviewed at head `0078e97d2fc4b626091c0c14112f1ebe73b32649`. First pass.

**Decision**
- 🟡 comment only

I agree the merge is low-stakes: `runTool` is a standalone CLI plus the SWIG binding,
nothing in the art job path, and CI is green at this head (build 3302 — build, whitespace,
`check_cmake`, `g4surfaceCheck`, `rootOverlaps`, `trigger` and the six physics jobs).
Nothing below blocks. But four of the five S1s are places where the goal stated in the PR
body is only partly reached, so they are worth a second pass before merge.

**Scope understood**
- `-j` extended from `-q`/`-e` to the flags listing, the run listing and the `-b` blob,
  with the JSON shaping placed on `RunTool` so the SWIG binding can reach it too.
- `RunInfo::dbTables3` gains a subsystem filter; `pywrap.i` gains a module-wide
  `%exception`; `runTool_main.cc` gains a `try`/`catch` around the `-q` loop.

**Findings**

1. 🟠 [S1] `-b <subsystem> -j` aborts the process on an empty or malformed settings blob
   - Evidence: `runTool_main.cc:191` uses the throwing overload,
     `entry["settings"] = nlohmann::json::parse(settings);`. The same string is parsed
     twice elsewhere in this package with the *non*-throwing overload, and with a comment
     saying why — `RunConfig.cc:79-81` and `:120-122`: "tolerate malformed/empty content
     by returning an empty result rather than throwing",
     `nlohmann::json::parse(_settings, nullptr, false)`. `RunInfo.cc:49` does the same.
     `settings` is whatever `DbUtil::simplfyQeString()` returned at `RunTool.cc:230`, and
     that returns its input unchanged when it is empty or one character, so an empty
     column arrives as `""`. Against the same nlohmann as the build
     (`json_abi_v3_11_3`), `parse("")` and `parse("{\"a\":")` both give
     `terminate called ... parse_error.101` and exit 134, core dumped.
   - Impact: an agent calling `-b X -j` gets SIGABRT instead of the parseable error this
     PR is trying to provide. The text path has no such failure mode — the hand-rolled
     formatter prints whatever it is given.
   - Suggested fix: use the non-throwing overload and emit `"settings": null`, matching
     what `RunConfig`/`RunInfo` already decided for this same column; or let it throw and
     catch it properly, per finding 2.

2. 🟠 [S1] The `main` catch covers one branch, not `main`
   - Evidence: the `try`/`catch` at `runTool_main.cc:133-143` wraps only the `dbtables`
     loop. Still uncaught, roughly in order of how often they will fire:
     `tool.listRuns(...)` at `:116`, which throws `RUNTOOL_BAD_RUN_QUERY` /
     `RUNTOOL_BAD_CONFIG_QUERY` (`RunTool.cc:102-105`, `:217-221`) whenever the database
     is unreachable; `flagsJson()` at `:68` and `runJson` at `:288`, both reaching
     `RUNTOOL_BAD_FLAG_QUERY` (`RunTool.cc:40-43`); the two `nlohmann::json::parse` calls
     at `:191` and `:288`; and `std::stoi` at `:92` and `:96`, reachable as
     `runTool -n abc`. CI's own clang-tidy run at this head reports it:
     `runTool_main.cc:17:5: warning: an exception may be thrown in function 'main' which
     should not throw exceptions [bugprone-exception-escape]`.
   - Impact: the PR body describes "a catch in the main so that agents can parse a throw
     more easily", but the most likely error by far — the database being down — still
     ends in `terminate called` and exit 134.
   - Suggested fix: one `try { ... } catch (std::exception const& e)` around the body of
     `main`, replacing the inner one. I checked that this catches nlohmann's `parse_error`
     as well as `cet::exception`, so a single handler covers both.

3. 🟠 [S1] The `dbTables3` subsystem gate applies only to the JSON branch
   - Evidence: `RunInfo.cc:38-63` — the new `TRG`/`Gateway`/`CFO`/`DQM` filter at `:42`
     sits inside the `if (qjson)` opened at `:39`. The non-JSON path at `:66-77` still
     loops over all of `_configs` unfiltered.
   - Impact: `runTool -q` and `runTool -q -j` now report different cat-3 table sets for
     the same run. The text output is what a person checks to confirm what the agent saw,
     so this is the divergence that costs most later.
   - Suggested fix: hoist the filter above `if (qjson)` so both paths agree.

4. 🟠 [S1] `-q -j` and `-e -j` still emit concatenated JSON documents for a multi-run
   selection
   - Evidence: `runTool_main.cc:134-139` and `:152-157` print one `dump()` per run inside
     the loop. `dbTables3(true)` returns an object per run (`RunInfo.cc:62`) and
     `dbTables2(true)` an array (`RunInfo.cc:24,30` — and `"[]"` is non-empty, so it
     prints even for a run with no TRG config). Two runs give `{...}\n{...}`, which
     `json.load` rejects with `Extra data: line 4 column 1`. The comment at `:279-285`
     justifies this as "dbtables isn't meant to be queried as a list", but `-q` and `-e`
     are reached after the same `RunSelect` built at `:98`, so
     `runTool -r 100000-101000 -q -j` and `runTool -d 7 -e -j` are both accepted today.
   - Impact: `-j` is relabelled "output content in JSON format" at `:46-47` and the body
     says "allow json output for all switches", but two switches emit output an agent
     cannot parse as soon as the selection matches more than one run.
   - Suggested fix: assemble into an array as the run listing now does at `:286-290`, or
     make `-q -j`/`-e -j` reject a multi-run selection rather than emit something invalid.

5. 🟠 [S1] `runJson` issues two database queries per run
   - Evidence: `RunTool::flags()` runs a live `_reader.query(...)` on every call
     (`RunTool.cc:39`), uncached. `runJson` calls it twice unconditionally at
     `RunTool.cc:438-439`, and `runTool_main.cc:287-289` calls `runJson` once per run.
     `printRun` does not: its single `flags(FlagType::transition)` at `RunTool.cc:359` is
     inside `if (info.transitions().size() > 0)` at `:357`, so it does not fire at all
     without `-a`. For `runTool -n 200`: text path 0 flag queries, JSON path 400. With
     `-a`, 200 versus 400.
   - Impact: a regression on exactly the path this PR exists to serve, growing linearly
     with `-n`. `online.run_type` and `online.run_transition_type` are job-constant.
   - Suggested fix: memoize the two maps in `RunTool` on first use, so `runJson` and
     `printRun` both pay once per process.

6. 🟡 [S2] The subsystem gate is a hardcoded denylist for an allowlist intent
   - `RunInfo.cc:42` spells out four excluded names while the body describes the rule as
     "subdetector configs only", and `"TRG"` is now hardcoded twice in the file (also
     `:19`). A subsystem added to `online.config` later is silently treated as a
     subdetector, rejoins the merge, and can trip the `RUNINFO_DUPLICATE_TABLE` throw at
     `:55`. One named list or an `isSubdetector()` helper shared with `dbTables2` would
     give the rule a single home.

7. ⚪ [S3] Two small things on new lines
   - `RunInfo.cc:42` is 137 characters against `ColumnLimit: '100'` in the repo's
     `.clang-format` — the longest line in the file by 59 characters.
   - The new comment blocks (`RunTool.cc:420-434`, `runTool_main.cc:123-132`, `:179-186`,
     `:279-285`) are much longer than anything else in these files and mostly record what
     was decided rather than what the code does.

- 🟢 The `%exception` block is placed correctly. SWIG applies it to declarations parsed
  after the directive; it sits at `pywrap.i:21-29`, ahead of every header `%include`
  (`:48-57`), so it does cover `RunTool`, `RunInfo` and `RunConfig`. `exception.i` is
  included at `:19` for `SWIG_exception`, and `cet::exception` derives from
  `std::exception`, so the first handler catches it.
- 🟢 The `-b ... -j` path does need to be separate from the text formatter, as the comment
  at `:179-186` claims — that formatter substitutes real newlines for `\n` at `:198-200`,
  and an unescaped newline inside a JSON string is not valid JSON.

**Validation check**
- Build/tests run: not by me; CI is green at this head, and FNALbuild's table shows every
  job passing. Nothing in CI exercises `runTool` itself, so for finding 1 I compiled and
  ran a standalone reproduction of that one `nlohmann::json::parse` call against the same
  header from the Musing spack view.
- Config contract check: n/a, no FHiCL touched.
- Cross-repo consistency: pass. No companion PR
  (`gh search prs --owner Mu2e --author rlcee --state open`), and nothing outside
  `DbService` references these entry points.

**Residual risk**
- I could not exercise any of this against the real `run_info` database. Finding 1 rests
  on the guard `RunConfig`/`RunInfo` already apply to that same column, not on an observed
  bad row — if that column is never empty in practice, it drops to a robustness point.
- The SWIG binding is not exercised by CI, so the new `%exception` is unverified at
  runtime; it is correct as far as I can read it.

**Author follow-ups**
1. Finding 1: non-throwing parse with `"settings": null`, or a real handler.
2. Move the `try`/`catch` out to `main` — that also clears the clang-tidy warning.
3. Hoist the `dbTables3` filter above `if (qjson)`.
4. Wrap `-q -j`/`-e -j` in an array, or reject multi-run selections there.
5. Memoize the two `flags()` maps in `RunTool`.
6. Give "subdetector config" one named home shared with `dbTables2`.

rlcee:
1 addressed
2 addressed, but raises larger questions of error hanlding policy
3 addressed
4 addressed by limting reques to one run, which is more logical in user context
5 addressed small optimization
6 not addressed - this is more logical in larger context
7 addressed

oksuzian (approved):
### PR Review Summary — Mu2e/Offline#1938

Reviewed at head `d02cc0a10669fbbe41cb61521f90d07cfedb7d9c`. Re-review of `0078e97d2fc4b626091c0c14112f1ebe73b32649`.

**Decision**
- 🟢 approve

**Scope understood**
- Delta is one commit, `d02cc0a`, across the same six `DbService` files: the five S1s from the first pass are addressed, the S2 denylist is left as is by choice, and the rest is include reordering and a re-indent of `main` under the new `try`.

**Prior findings**
- 🟢 [was 🟠 S1] `-b <subsystem> -j` aborts on an empty or malformed settings blob — FIXED in `d02cc0a`, verified. `runTool_main.cc` now uses `nlohmann::json::parse(settings, nullptr, false)` and emits `"settings": null` when `is_discarded()`, matching the guard `RunConfig`/`RunInfo` already apply to that column.
- 🟢 [was 🟠 S1] The `main` catch covered one branch — FIXED in `d02cc0a`, verified. The whole body of `main` from `ParseCLI` construction to the final `return 0` is inside one `try { } catch (std::exception const& e)`, so `listRuns`, `flagsJson`, `runJson`, both `parse` calls and both `std::stoi` calls are covered. This should also clear the `bugprone-exception-escape` warning clang-tidy reported at the previous head; CI at this head is still pending, so that is not yet confirmed.
- 🟢 [was 🟠 S1] `dbTables3` subsystem gate applied only to the JSON branch — FIXED in `d02cc0a`, verified. `RunInfo.cc` now defines the `excluded` lambda above `if (qjson)` and both loops call it; `-q` and `-q -j` report the same table set.
- 🟢 [was 🟠 S1] `-q -j` / `-e -j` emitted concatenated JSON documents for a multi-run selection — FIXED in `d02cc0a`, verified, by the rejection route: both branches now exit 1 with a message on stderr when `runvec.size() > 1`. The message names the narrowing options, which is the right shape for an agent to act on.
- 🟢 [was 🟠 S1] `runJson` issued two flag queries per run — FIXED in `d02cc0a`, verified. `RunTool::flags()` memoises per `FlagType` in `_runFlagsCache`/`_transitionFlagsCache` with `bool` guards, so each map is fetched once per `RunTool` instance. The cache is returned by value, which is fine at these sizes.
- 🟡 [S2] Subsystem gate is a hardcoded denylist for an allowlist intent — NOT ADDRESSED, carried over as non-blocking. The author's reply ("more logical in larger context") is reasonable; the lambda at least gives the rule one home inside `dbTables3` now. `"TRG"` still appears separately in `dbTables2`.
- ⚪ [S3] Long line — FIXED (the 137-character condition became the lambda). The explanatory comment blocks remain and have grown; the author's choice, not a merge condition.

**Validation check**
- Build/tests run: not by me. `mu2e/buildtest` is pending at `d02cc0a` (status context registered; not re-triggered). The previous head `0078e97` was fully green (build #3302), and the delta is a handler hoist, a lambda, two early-return guards and a memoisation — none of it changes what CI exercises, since nothing in CI runs `runTool`.
- Config contract check: n/a.
- Cross-repo consistency: pass, unchanged — nothing outside `DbService` references these entry points.

**Residual risk**
- Unchanged from the first pass: none of this is exercised against the real `run_info` database or the SWIG binding here.
- The `"settings": null` degradation on a malformed blob is the package's existing convention rather than a fail-loud one; the author flagged this as a wider error-handling policy question, and it is outside this PR.

**Author follow-ups**
- None required. The denylist/allowlist item stays open as a later cleanup.

PR #1939, opened 2026-08-24
Add missing configuration check on calo info when regrowing

brownd1978:
Simple bug fix

### DIFF
--- a/Mu2eKinKal/fcl/prolog.fcl
+++ b/Mu2eKinKal/fcl/prolog.fcl
@@ -630,7 +630,7 @@ Mu2eKinKal : {
     }
     KKFitSettings: {
       @table::Mu2eKinKal.KKFIT
-      SampleSurfaces : ["ST_Outer","ST_Front","ST_Back"]
+      SampleSurfaces : ["ST_Outer","ST_Inner","ST_Front","ST_Back","TT_Mid","TT_Back"]
       AddHits : false
       SaveDomains : true
       SaveTrajectory : Full
--- a/Mu2eKinKal/inc/KKFit.hh
+++ b/Mu2eKinKal/inc/KKFit.hh
@@ -304,7 +304,7 @@ namespace mu2e {
       if(strawhits.back()->hitState().active())nactive++;
       // create the straw Xing for the associated straw, including the hit reference
     }
-    if(kseed.caloCluster()){
+    if(usecalo_ && kseed.caloCluster()){
       makeCaloHit(kseed.caloCluster(),calo,*ptraj,calohits);
     }
     for (auto const& paramhit : kseed.paramHits()){


oksuzian (approved):
### PR Review Summary — Mu2e/Offline#1939

Reviewed at head `a1a8f76ec458b92efdd79be612a6ec095f833646`. First pass.

**Decision**
- 🟢 approve

Correct, minimal, single-topic. It restores parity with the only other live
`makeCaloHit` call site, and I could not find any committed configuration whose
behavior it changes.

**Scope understood**
- `KKFit::regrowComponents` (`KKFit.hh:307`) created a `KKCaloHit` from the stored
  `KalSeed`'s cluster whenever that cluster was present, without consulting the fit
  configuration. The PR adds the missing `usecalo_ &&`.

**Verified**
- 🟢 `:307` really was the sole unguarded caller. `makeCaloHit` has three call sites
  outside `KKFit.hh` — `LoopHelixFit_module.cc:341`, `CentralHelixFit_module.cc:339`
  and `KinematicLineFit_module.cc:286` — and the latter two are commented out. The
  live one reads `if (kkfit_.useCalo() && hseed.caloCluster().isNonnull())`, so the
  fresh-fit path has always consulted the flag, and `extendTrack` pairs them the same
  way at `:392` (`addhits_ && usecalo_`). The two paths also do not differ on the
  cluster energy cut: `minCaloEnergy_` is used only in `addCaloHit` (`:609`), never in
  `makeCaloHit`.
- 🟢 Nothing downstream expects the calo hit that is now absent. `createSeed` fills
  `kseed._chit` only from `kktrk.caloHits()` (`KKFit.hh:735-753`), so an empty
  collection leaves `_chit` default-constructed rather than producing a half-filled
  seed.
- 🟢 No committed configuration changes behavior, which bounds the regression risk to
  zero. `UseCaloCluster` appears exactly once in `Mu2eKinKal/fcl/prolog.fcl` — line
  11, `true`, in `KKFIT` — and no `KKFitSettings` block there overrides it, including
  `RegrowLH` (`:631-637`), which inherits `@table::Mu2eKinKal.KKFIT` and changes only
  `AddHits`, `SampleSurfaces`, `SaveDomains` and `SaveTrajectory`. Searching
  `Mu2e/Production` for `UseCaloCluster` returns one hit,
  `JobConfig/recoMC/NoFieldRun1B.fcl:41`, which sets it on `TimeAndPhiClusterFinder` —
  a different module. So `usecalo_` is true for every regrow job that ships today, and
  this is a latent-bug fix rather than a behavior change.

**Findings**

1. ⚪ [S3] The two call sites now agree on `useCalo()` but not on the null test —
   observation only, outside this diff
   - `:307` tests the `art::Ptr` directly, and `art::Ptr::operator bool` is
     `isNonnull() && core_.isAvailable()` (canvas 3.17.00,
     `canvas/Persistency/Common/Ptr.h:154-157`). `LoopHelixFit_module.cc:341` tests
     `.isNonnull()` alone, which does not guarantee the product can be retrieved, while
     `makeCaloHit` dereferences it at `KKFit.hh:362` (`caloAxis(*cluster,calo)`). The
     regrow line you just fixed is the stricter of the two. Nothing to do here —
     flagging it in case the fresh-fit side is worth tightening later.

**Validation check**
- Build/tests run: not by me. CI is green at this head; every row of FNALbuild's table
  for build 3303 passed, including `build (prof)`, the six physics jobs,
  `g4surfaceCheck`, `rootOverlaps`, `check_cmake` and the whitespace check.
- Config contract check: pass. `usecalo_` is bound to `fitconfig.useCaloCluster()` at
  `KKFit.hh:177`, and the FHiCL key is `fhicl::Atom<bool> useCaloCluster
  { Name("UseCaloCluster"), ... }` at `KKFitSettings.hh:70`. No schema change.
- Cross-repo consistency: pass. No companion PR needed — see the third 🟢 above for the
  `Production` search.

**Residual risk**
- I did not run a regrow job or compare regrown tracks before and after. That would only
  be informative for a configuration with `UseCaloCluster : false`, and no such
  configuration exists in either repository today — the same reason the change is safe
  to merge as-is.

**Author follow-ups**
- None required.

PR #1941, opened 2026-08-25
Changes in DAQ/DigiGenerator and EventDisplay

ehrlich-uva:
- Replaced the exceptions in the DigiGenerator by entries in the CrvDAQErrorCollection (and some print outs). Prevents the DigiGenerator from crashing, if it encounters bad data.
- Changes in EventDisplay (added tracker hits, changed the thickness of the tracker hits).
- Changes in PedestalFinder.
- Added new DB txt files.
- Added a function that checks if an offline channel exists.


oksuzian (changes requested):
### PR Review Summary — Mu2e/Offline#1941

Reviewed at head `f30e5dcf123ae16040ce6ba2866ba27ed9f73141`. First pass.

**Decision**
- 🔴 request changes

The DigiGenerator change itself is the right shape — an unparseable fragment should
produce a record in a data product, not abort the job. Two things block: a new pedestal
quality cut that changes CRV calibration and is explained nowhere, and three new enum
values that the enum's own name table does not know about. Separately, the PR carries
five distinguishable changes while the description covers one.

**Scope understood**
- `DAQ/src/CrvDigisFromArtdaqFragmentsFEBII_module.cc`: two `cet::exception` throws
  replaced by `CrvDAQerrorCollection` entries, plus a new `invalidChannel` guard, with
  three new codes added to `CrvDAQerror`.
- Also in the PR, unmentioned in the body: a `CrvPedestalFinder` cut change in code and
  both era prologs; three new 4,650-line `CRVConditions/data` files; an EventDisplay
  `DbService` switch to `CRV_COMMISSIONING`; a new EventDisplay fcl; a `ContentSelector`
  fix; and a line-width change.

**Findings**

1. 🟠 [S1] A new pedestal quality cut is applied on the live path, with no justification
   - Evidence: `CrvPedestalFinder_module.cc:160` adds
     `if(fabs(iter->GetADCs().at(0)-iter->GetADCs().at(1))>=_maxADCspreadFirstTwoSamples) continue;`
     inside the `if(_firstSampleOnly)` branch, defaulting to `3` (`:47`) and set to `3` in
     both `CRVReco/fcl/prolog_v11.fcl:13` and `prolog_v12.fcl:13`. Both prologs set
     `firstSampleOnly : true` (`:8`), so this is the branch that actually runs, and it
     previously applied no spread cut at all — every hit's first sample was filled.
     The companion `maxADCspread` change from `5` to `20` (`:46`, prologs `:12`) is the
     opposite direction but is inert under both shipped prologs, since that cut only
     applies when `firstSampleOnly` is false. The comment that recorded exactly this
     — `maxADCspread : 5  #if firstSampleOnly: false` — was deleted in the same hunk.
   - Impact: which waveforms enter the pedestal histogram changes for every CRV pedestal
     job in both the v11 and v12 eras, so the pedestals themselves change, and through
     them CRV reconstruction. Nothing in the PR body, the commit messages or the code
     comments says why `3` is the right number or what motivated it.
   - Suggested fix: state the motivation and where `3` comes from — a plot, a run, a
     docdb entry — in the PR body or a comment, per the unsourced-numbers rule. Please
     also keep the applicability note that was on the `maxADCspread` line, since which of
     the two cuts is live depends entirely on `firstSampleOnly`.

2. 🟠 [S1] The three new `CrvDAQerror` codes are missing from `names()`
   - Evidence: `RecoDataProducts/inc/CrvDAQerror.hh:17` now declares `invalidChannel=7`,
     `invalidDtcId=8` and `zeroBlockError=9`, but `RecoDataProducts/src/CrvDAQerror.cc:14-23`
     still lists seven entries, ending at `byteCountMismatch`. That file is not in this PR.
     `EnumToStringSparse` treats the map as the authority: `isValid(int)` returns false
     when `names().find(cid) == end()` (`EnumToStringSparse.hh:110-119`), so the `int`
     constructor at `:62-65` throws `std::out_of_range` on any of the three; and `name()`
     at `:84-88` does `find()` and dereferences the iterator with no `end()` check, on the
     stated assumption that validity was checked at construction — which the `enum_type`
     constructor at `:60` does not do.
   - Impact: the three codes this PR exists to record cannot be constructed from an int
     and cannot be named. I checked the one in-repo consumer, `CrvDQMcollector_module.cc:488`,
     and it only compares the enum, so nothing crashes today — but the codes are invisible
     to anything that renders or validates them, which defeats the point of adding them.
   - Suggested fix: add the three entries to the map in `CrvDAQerror.cc`.

3. 🟠 [S1] The PR does five separable things and describes one
   - Beyond the DigiGenerator change, it carries the pedestal cut (finding 1), three new
     `CRVConditions/data/*_20260816.txt` files totalling 4,650 lines, an EventDisplay
     `DbService.purpose` switch from `Sim_best`/`v1_5` to `CRV_COMMISSIONING`/`v1_0` with
     `nearestMatch : true` in two fcl files, a new `EventDisplayExtractedCrvOnly.fcl`, and
     a `Straw.h` line-width change. Each may be individually fine; bundled, they cannot be
     reviewed or reverted independently, and the conditions switch in particular is a
     behaviour change for anyone running the extracted-position display. Per the
     collaboration's coding standard this is a blocking objection on its own. At minimum,
     please split out the pedestal change and describe the rest in the body.

**Smaller items**

- 🟡 `CRVOrdinal::offlineExists(CRVROC& channel) const` (`CRVOrdinal.hh:43`) takes a
  non-const reference to a parameter it only reads, so it cannot be called on a
  `const CRVROC`. Its sibling three lines down, `offline(const CRVROC& online)` (`:63`),
  gets this right. It compiles at the one call site only because `onlineChannel` is a
  non-const local. Make it `CRVROC const&`.
- 🟡 `zeroBlockError` and `invalidChannel` emit one record per hit, and every record is
  identical: `emplace_back(code, iFragment, iSubEvent, iDataBlock, header->GetPacketCount())`
  at `CrvDigisFromArtdaqFragmentsFEBII_module.cc:243` and `:254` uses only block-level
  values, then `continue`s to the next hit. A block whose hits are all on an unmapped FEB
  therefore writes N indistinguishable entries into a persisted collection, and the one
  piece of information that would make them useful — which channel — is printed to stderr
  and then discarded. Either record once per block, or add the channel to `CrvDAQerror`.
- 🟡 `dtcID`, `linkID` and `rocID` are recomputed from `header` on every iteration of the
  hit loop (`:219`, `:228-229`) although none depends on `crvHit`. Hoisting the
  `dtcID < _firstCrvDtcID` check above the loop would also make its `break` (`:226`, which
  exits the hit loop) unambiguous next to the `break` at `:215`, which exits the enclosing
  data-block loop.
- ⚪ Collapsed: the three new data files are referenced only from commented-out
  `services.DbService.textFile` lines in `EventDisplayExtractedReal.fcl` and
  `EventDisplayExtractedRealCrvOnly.fcl`, so nothing active reads them, and neither the
  files nor the PR body says what they are or who consumes them. Those same two files now
  carry two commented-out `textFile` lines each where `EventDisplayExtracted.fcl` had one
  removed in this PR — worth being consistent about which convention you want.

- 🟢 The enum extension is append-only — `invalidChannel=7`, `invalidDtcId=8`,
  `zeroBlockError=9` are appended after `byteCountMismatch=6` with the existing values
  unchanged, so data already written keeps its meaning. This is the part that would have
  been expensive to get wrong.
- 🟢 The `ContentSelector` change is a real fix, not a new feature. `_kalSeedHitVector` was
  already declared (`ContentSelector.h:54`) and already read under `case 4` of
  `getSelectedHitCollection` (`ContentSelector.cc:254-256`), with the
  `KalSeedCollection` template instantiation already present — it was simply never
  populated. The new `createNewEntries<mu2e::KalSeedCollection>(..., 4)` at `:73` uses the
  same `classID` as the existing track-side call at `:151` and as the reader's `case 4`.

**Validation check**
- Build/tests run: not by me. CI is green at this head — FNALbuild reports the build tests
  passed at `f30e5dcf`, and the combined commit status is `success`.
- Config contract check: pass. `maxADCspreadFirstTwoSamples` exists as a `fhicl::Atom` in
  `Config` (`CrvPedestalFinder_module.cc:47`) and both prologs spell it identically. I did
  not run `fhicl-dump` on the changed EventDisplay job configs.
- Cross-repo consistency: not checked. The `CRV_COMMISSIONING` / `v1_0` purpose-version
  pair in the two EventDisplay fcl files is a live database lookup I have no way to verify
  from here — please confirm that pair resolves.

**Residual risk**
- I have not assessed whether `3` ADC counts is the right spread cut, only that it is new,
  live and unexplained.
- I did not exercise the DigiGenerator against a fragment that triggers any of the three
  new codes, so the error records are reviewed by reading, not by running.

**Author follow-ups**
1. Add `invalidChannel`, `invalidDtcId` and `zeroBlockError` to the `names()` map in
   `RecoDataProducts/src/CrvDAQerror.cc`.
2. Say where the `maxADCspreadFirstTwoSamples : 3` cut comes from, and restore a note
   saying which cut applies under which `firstSampleOnly` setting.
3. Split the pedestal-finder change out, and describe the conditions-data and EventDisplay
   changes in the PR body.
4. Make `offlineExists` take `CRVROC const&`.
5. Decide whether the per-hit error records should be deduplicated per block or should
   carry the channel.

oksuzian:
@FNALbuild run build test

oksuzian (commented):
### PR Review Summary — Mu2e/Offline#1941

Reviewed at head `b92bd2e4fb01b23f7f7ac6bf0a15c2439e607cc8`. Re-review of `f30e5dcf`.

**Decision**
- 🟡 comment only

@ehrlich-uva — thanks for the quick turnaround on the pedestal value. Nothing open now
would fail a job or produce wrong numbers at merge, so this is a comment rather than a
change request. Two items still worth closing before it goes in: the new pedestal cut has
no stated justification, and three of the ten error codes this PR adds cannot be named.

**Delta reviewed**: `f30e5dcf`..`b92bd2e4`, one commit ("small bug fix for pedestal
finder"), two files, one line each — `maxADCspread` back to `5` in `CRVReco/fcl/prolog_v11.fcl:12`
and `prolog_v12.fcl:12`.

**Fixed since `f30e5dcf`**

- 🟢 [was part of S1] The `maxADCspread` change is reverted. Both prologs are back to `5`,
  matching `main`. That removes the confusing part of the original hunk: with
  `firstSampleOnly : true` set on line 8 of both files, `maxADCspread` governs a branch
  that never executes, so raising it to `20` looked like a threshold loosening while
  changing nothing.

**Findings still open**

1. 🟠 [S1] **Partly addressed** — the new live cut is still unexplained.
   - Evidence: with `maxADCspread` reverted, the only remaining behavior change in these
     two files is the one that is actually live. `CrvPedestalFinder_module.cc:158-163`:
     ```cpp
     if(_firstSampleOnly)
     {
       if(fabs(iter->GetADCs().at(0)-iter->GetADCs().at(1))>=_maxADCspreadFirstTwoSamples) continue;
       hist->Fill(iter->GetADCs().at(0));
       continue;
     }
     ```
     `maxADCspreadFirstTwoSamples : 3` is new in both prologs (`:13`), and `firstSampleOnly`
     is `true` in both (`:8`), so this branch is the one that runs. Before this PR it applied
     no spread cut at all — every digi's first sample went into the histogram. Now a digi is
     dropped whenever its first two samples differ by 3 ADC counts or more.
   - Impact: this changes which waveforms build the pedestal, so it changes the pedestal
     values themselves and the statistics behind each fit. That is a real calibration change,
     and it is the one thing in this PR that a future reader will want a reason for. The
     diff also deletes the comment that documented the interaction — `main` carries
     `maxADCspread : 5  #if firstSampleOnly: false`, and at this head that annotation is gone
     from both files, so nothing in the config records that one of the two cuts is inert.
   - Suggested fix: say in the PR body (or a comment on the new line) where `3` comes from —
     noise width, a scan, a number from the FEB-II data — and restore an annotation
     distinguishing the two cuts, since a reader now sees two similar keys with no way to
     tell which one is active.

2. 🟠 [S1] **Carried over, unaddressed** — three of the new error codes are missing from
   `names()`.
   - Evidence: `RecoDataProducts/inc/CrvDAQerror.hh:17` declares ten values through
     `zeroBlockError=9`, but `RecoDataProducts/src/CrvDAQerror.cc:14-23` still ends at
     `byteCountMismatch`. That file is not in the PR. `invalidChannel`, `invalidDtcId` and
     `zeroBlockError` are exactly the three codes this PR was written to record —
     `DAQ/src/CrvDigisFromArtdaqFragmentsFEBII_module.cc:225`, `:243` and `:254`.
   - Impact: bounded today, which is why this is not a blocker. `CrvDAQerror` stores a raw
     enum (`_errorCode` is `CrvDAQerrorCode::type`, `:41`), so no `EnumToStringSparse` object
     is built inside the product and nothing streams through the map. The gap bites the two
     paths that do go through it: `EnumToStringSparse::isValid` (`:110-119`) returns false
     for the three, so the `int` constructor at `:62-65` throws `std::out_of_range` via
     `isValidOrThrow`; and `name()` at `:84-88` does `Detail::names().find(_id)` and
     dereferences the result with no `end()` check — the comment there says "We checked for
     the validity of `_id` upon construction", which is true of the `int` constructor and not
     of the `enum_type` one at `:60`. The only consumer in Offline is
     `CrvDQMcollector_module.cc:488`, which compares rather than names, so nothing crashes;
     but any DQM display, ntuple or debug print that names these codes hits it.
   - Suggested fix: three lines in `CrvDAQerror.cc`. The enum itself is appended, not
     inserted, so existing data keeps its meaning and `classes_def.xml:306-311` needs no
     change.

3. 🟠 [S1] **Carried over, unaddressed** — the PR does five separable things and the body
   describes one. The body still reads only "Replaced the exceptions in the DigiGenerator by
   entries in the CrvDAQErrorCollection". Also in the diff: the pedestal-finder cut above,
   three new 4,650-line conditions files, the EventDisplay `DbService` switch to
   `CRV_COMMISSIONING`/`v1_0`, a new fcl and a line-width change. Each may well be fine; the
   ask is that the body name them, so a reviewer knows what to look at and a future bisect
   knows what this commit did.

**Carried over unchanged** — the delta touched none of these files, so they stay open by
definition. All re-verified at this head:

- 🟡 `CRVConditions/inc/CRVOrdinal.hh:43` takes `CRVROC& channel` by non-const reference
  while its sibling twenty lines down takes `const CRVROC& online` (`:63`), and
  `onlineExists` (`:36`) takes its argument by value. Nothing is modified through it —
  `const&` is the right signature.
- 🟡 The three per-hit error paths in `CrvDigisFromArtdaqFragmentsFEBII_module.cc` record
  only block-level fields. `:225`, `:243` and `:254` all emplace
  `(code,iFragment,iSubEvent,iDataBlock,header->GetPacketCount())`, so N bad hits in one
  block produce N identical records. The identifying detail — `rocID`, `rocPort`,
  `febChannel` at `:253` — goes to `std::cerr` and is dropped from the product.
- 🟡 `dtcID` (`:219`), `linkID` (`:228`) and `rocID` (`:229`) are recomputed for every hit
  from block-level values that cannot change inside the loop opened at `:217`; hoist them
  above it. Note also that the `break` at `:226` exits the hit loop while the one at `:215`
  exits the data-block loop — worth a look at whether `invalidDtcId`, which is a
  block-level condition, should abandon the block rather than the remaining hits.
- ⚪ The three new `CRVConditions/data/*_extracted_20260816.txt` files (2306 + 2306 + 38 =
  4,650 lines) are reached only from a commented-out line,
  `EventDisplay/fcl/EventDisplayExtractedReal.fcl:13`, since `:8-9` now take the same
  conditions live from `CRV_COMMISSIONING`/`v1_0`. Reading the comment above it
  ("current status and calibration until we can use the database"), they look like a
  deliberate offline fallback rather than leftovers — worth saying so in the PR body, so
  the next cleanup pass does not delete them.

**Verified correct**

- 🟢 The enum extension is append-only: `unknown=0` through `byteCountMismatch=6` keep their
  values and `7`/`8`/`9` are added after them, so data already written keeps its meaning.
- 🟢 The `ContentSelector` change completes a path that was already wired.
  `ContentSelector.cc:73` adds the `createNewEntries<mu2e::KalSeedCollection>` call for
  `_kalSeedHitVector`, which was already declared (`ContentSelector.h:54`) and already read
  under `case 4` — it was simply never populated.

**Validation check**
- Build/tests run: green at this head. There was no result at `b92bd2e4` when I started
  (`mu2e/buildtest` read "This test has not been triggered yet"), so I triggered a run and
  waited for it. The delta is two numeric values in a prolog, so this mostly confirms the
  files still parse.
- Config contract check: pass. `maxADCspread` and `maxADCspreadFirstTwoSamples` both match
  the `Config` atoms in `CrvPedestalFinder_module.cc`, and both prologs carry the same
  values.
- Cross-repo consistency: needs follow-up on finding 2 — `CrvDAQerror.cc` lives in this repo
  and can go in this PR.

**Residual risk**
- I have not run the pedestal finder on real FEB-II data, so I cannot say what fraction of
  digis the new cut removes. That is the question finding 1 asks you to answer.

**Author follow-ups**
1. Say where `maxADCspreadFirstTwoSamples : 3` comes from, and restore an annotation
   marking which of the two spread cuts is active under `firstSampleOnly : true`.
2. Add `invalidChannel`, `invalidDtcId` and `zeroBlockError` to `CrvDAQerror.cc`.
3. Expand the PR body to name the other four changes, or split them out.

ehrlich-uva:
Comments:
1. This additional cut is to remove ADC samples from the pedestal plot that are already part of the leading edge of a pulse. I added comments in the fcl files. The value of 3 is a best guess at the moment.
2. Fixed.
3. Fixed.

oksuzian:
@FNALbuild run build test

oksuzian (commented):
### PR Review Summary — Mu2e/Offline#1941

Reviewed at head `d4edd3707337f99c98b56dd530a2de1f87b841a8`. Re-review of `b92bd2e4`.

**Decision**
- 🟡 comment only

@ehrlich-uva — all three follow-ups from the last pass are closed, and the `dtcID` hoist
went further than I asked for and is the better fix. One thing I should have caught two
passes ago and did not: the `subEvent` field written into every `CrvDAQerror` record is
always `0`. It pre-dates this PR, but this is the PR that makes that field part of a
persisted product, so it is worth three characters here rather than a follow-up.

**Delta reviewed**: `b92bd2e4`..`d4edd370`, two commits ("small changes in DAQ/CRV and CRV
pedestalFinder", "changed print level for CRV zero-block error"), five files.

**Fixed since `b92bd2e4`**

- 🟢 [was S1 #2] The three new codes are in the name table. `RecoDataProducts/src/CrvDAQerror.cc:22-24`
  adds `invalidChannel`, `invalidDtcId` and `zeroBlockError`; I diffed the map against the
  enum at `CrvDAQerror.hh:16` and all ten values now have an entry, with the spellings
  matching. `EnumToStringSparse::isValid` and `name()` both work for the three from here on.
- 🟢 [was S1 #1, the part that was actionable] The annotation is back and now covers both
  keys: `CRVReco/fcl/prolog_v11.fcl:12-13` and `prolog_v12.fcl:12-13` say which cut applies
  to NZS data and which to ZS data, and the ZS one states what it is for ("a check that the
  first sample is not already part of the pulse"). Together with your answer on the thread
  this closes the question I asked. See the one nit below about recording that `3` is
  provisional.
- 🟢 [was S1 #3] The body now names all five changes rather than one. I am not pressing the
  split further.
- 🟢 [was S2] `CRVConditions/inc/CRVOrdinal.hh:43` is now `offlineExists(const CRVROC&)`,
  matching `offline(const CRVROC&)` at `:63`.
- 🟢 [was S2] The `dtcID` hoist is done, and better than I suggested. It is computed once per
  data block at `CrvDigisFromArtdaqFragmentsFEBII_module.cc:191`, the duplicate at the old
  `:311` is gone, and the `invalidDtcId` case now abandons the whole data-block loop instead
  of the hit loop — which is right, since the condition is a property of the block header,
  not of any hit. The only side effect I can find is that `totalSize` (`:152`) undercounts
  after an early `break`, and it feeds nothing but the `_diagLevel>1` print at `:357`; the
  pre-existing `break` at `:229` already had that property.
- 🟢 The new `header->GetPacketCount()>0` guard on `wrongSubsystemID` (`:186-190`) is a real
  improvement — an empty header from an unused DTC link is not an error, and the comment
  says so. The `continue` below it is unconditional, so a non-CRV block is still skipped
  either way; only the error record is suppressed.

**Findings**

1. 🟠 [S1] Every `CrvDAQerror` record reports `subEvent = 0`
   - Evidence: the subevent loop opens at `CrvDigisFromArtdaqFragmentsFEBII_module.cc:147`
     as `for(size_t iSubEvent=0; auto& dtcSubEvent : dtcSubEvents)` and `iSubEvent` is never
     incremented in its body — the only `++iSubEvent` in the file is at `:133`, a different
     variable in the byte-count loop that has already gone out of scope. All nine
     `crvDaqErrors->emplace_back(...)` calls pass `iSubEvent` as the third argument
     (`:160`, `:171`, `:186`, `:197`, `:216`, `:228`, `:248`, `:259`, and `:143` which
     passes a literal `0`), and every `std::cerr << "iSubEvent/iDataBlock: "` line prints it.
   - Impact: `iDataBlock` is only unique within a subevent, so `(subEvent, dataBlock)` is
     the pair that identifies where an error came from. With `subEvent` pinned to `0`, two
     errors in different subevents at the same block index are indistinguishable in the
     product and on the console. This is not new code — the same line is on `main` — but
     before this PR these paths threw, and now they write a record whose whole purpose is to
     say where the bad data was. It is also the field a DQM plot would bin on.
   - Suggested fix: `++iSubEvent;` at the end of the loop body (`:352`), matching `iFragment`
     at `:106`. Worth checking the sibling FEB-I module for the same pattern.

2. ⚪ Two nits, both one line
   - `decoder.PrintBlockFEBII(iDataBlock)` is now behind `_diagLevel>2` at `:247` but is
     still unconditional at `:227` (`errorUnpackingCrvHits`). Same call, same kind of site —
     worth gating both the same way, since the second commit's stated purpose was exactly
     this.
   - `linkID` (`:233`) and `rocID` (`:234`) are still recomputed on every hit from
     block-level values, and duplicated again at `:314-316` in the debug loop. Now that
     `dtcID` sits at `:191` these two can join it there.
   - `maxADCspreadFirstTwoSamples` carries a C++ default of `3` (`CrvPedestalFinder_module.cc:47`).
     Both shipped prologs set it, so nothing depends on the default; the convention here is that
     a physics-affecting key has no in-code fallback, so a job that forgets it fails validation
     rather than silently cutting. `maxADCspread` at `:46` has the same shape and predates you.

**Carried over, still open** — the delta did not touch these:

- 🟡 `zeroBlockError` (`:248`) and `invalidChannel` (`:259`) emit one record per hit, all
  identical, since the record carries only block-level fields. The identifying detail —
  `rocID`, `rocPort`, `febChannel` at `:258` — goes to `std::cerr` and is dropped. Either
  record once per block or extend `CrvDAQerror`. Compounds with finding 1: N identical
  records that also all say subevent 0.
- ⚪ The three `CRVConditions/data/*_extracted_20260816.txt` files (4,650 lines) are still
  reached only from a commented-out `services.DbService.textFile` line. The body now
  mentions them, which was the ask.

**Validation check**
- Build/tests run: not green at this head. `mu2e/buildtest` read "This test has not been
  triggered yet" for `d4edd370`, so I triggered a run
  (https://github.com/Mu2e/Offline/pull/1941#issuecomment-5417039007) and am posting without
  waiting for it. The last full green was `f30e5dcf`, two heads back. Please check the
  result before merging — the delta moves a scope-declared variable and a `break`, which is
  the shape of change a compiler catches and a reviewer does not.
- Config contract check: pass. The two prolog edits are comments only; both keys still match
  the `fhicl::Atom` names in `CrvPedestalFinder_module.cc:46-47`.
- Cross-repo consistency: not checked at this head. The `CRV_COMMISSIONING`/`v1_0`
  purpose-version pair in the EventDisplay fcl files is still a live database lookup I
  cannot verify from here.

**Residual risk**
- Finding 1 is a reading of scope, not a run. I did not execute the module.
- I could not read the `artdaq-core-mu2e` headers from this checkout, so my statement that a
  `DTCSubEvent` carries blocks from a single DTC — which is what makes the new `break` at
  `:198` safe — rests on the code's own structure, not on the library's contract. Worth a
  second of your time to confirm, since you know the format.

**Author follow-ups**
1. Increment `iSubEvent` in the subevent loop, and check the FEB-I module for the same.
2. Gate the `PrintBlockFEBII` call at `:227` on `_diagLevel` like the one at `:247`.
3. If `3` is still a best guess, say so in the fcl comment — a bare `3` beside a sourced
   explanation reads as measured to the next person.

ehrlich-uva:
Fixed 1. and 2.

oksuzian:
@FNALbuild run build test

oksuzian (approved):
### PR Review Summary — Mu2e/Offline#1941

Reviewed at head `6ac9cdd9edf77b17c73fbe705d9f54a74d7c242c`. Re-review of `d4edd370`.

**Decision**
- 🟢 approve

@ehrlich-uva — everything I raised across the three passes that was actionable in this PR
is now closed, and the build is green at this head. Two things are left over and neither
belongs in this PR: the duplicate per-hit error records, which is a data-product design
question, and the same `iSubEvent` bug in the FEB-I module, which this diff does not touch.

**Delta reviewed**: `d4edd370`..`6ac9cdd9`, one commit ("simplified the code"), one file,
+7/-9.

**Fixed since `d4edd370`**

- 🟢 [was S1] `iSubEvent` now advances. `CrvDigisFromArtdaqFragmentsFEBII_module.cc:147`
  is an index loop, `for(size_t iSubEvent=0; iSubEvent<dtcSubEvents.size(); ++iSubEvent)`,
  with `auto& dtcSubEvent=dtcSubEvents.at(iSubEvent);` at `:149`. The increment sits in the
  loop header, so the two `break`s inside — `:199` and `:234`, both of which exit the
  data-block loop, not this one — cannot skip it. Every `crvDaqErrors->emplace_back` and
  every `iSubEvent/iDataBlock` diagnostic now reports the subevent it came from, which is
  what makes `(subEvent, dataBlock)` a usable locator in the product.
- 🟢 [was S3] `PrintBlockFEBII` is gated at both sites: `_diagLevel>2` at `:232`
  (`errorUnpackingCrvHits`) as well as `:249` (`zeroBlockError`).
- 🟢 [was S3] `linkID` and `rocID` are hoisted. They are computed once per data block at
  `:211-213`, inside the `header->GetPacketCount()>0` block, and both copies that were
  inside the hit loops are gone — one declaration of each remains in the file, joining
  `dtcID` at `:192`. The `_useROC4asROC2` remap moved with them and its inputs are all
  block-level, so the value is unchanged.

**Still open, neither blocking**

1. 🟡 [S2] Carried over — the per-hit error records are still indistinguishable.
   `zeroBlockError` (`:250`) and `invalidChannel` (`:261`) emplace
   `(code,iFragment,iSubEvent,iDataBlock,header->GetPacketCount())` once per hit, so a block
   whose hits are all bad writes N identical entries. The detail that would separate them —
   `rocID`, `rocPort`, `febChannel`, printed at `:260` — is not in `CrvDAQerror`. Either
   record once per block or add the channel to the product. This is a data-product change,
   so a follow-up PR is the right place for it.

2. 🟡 [S2] The FEB-I module has the same `iSubEvent` bug you just fixed here.
   `DAQ/src/CrvDigisFromArtdaqFragments_module.cc:113` is still
   `for(size_t iSubEvent=0; auto& dtcSubEvent : dtcSubEvents)` with no increment anywhere in
   the body — the only `++iSubEvent` in that file is at `:101`, the byte-count loop. Its five
   `crvDaqErrors->emplace_back` calls (`:125`, `:135`, `:147`, `:163`, `:172`) all pass
   `iSubEvent`. Out of this PR's diff, so purely a follow-up; the fix is the same three-line
   shape you applied here.

3. ⚪ `maxADCspreadFirstTwoSamples : 3` — you said on the thread that `3` is a best guess for
   now. Worth putting that word in the fcl comment (`CRVReco/fcl/prolog_v11.fcl:13` and
   `prolog_v12.fcl:13`): the comment now explains what the cut does, which reads as though
   the value were measured.

**Validation check**
- Build/tests run: green at this head. `mu2e/buildtest` had not been triggered for
  `6ac9cdd9`, so I triggered it and waited: FNALbuild reports the build tests passed
  (job 3315), including `build (prof)`, `trigger` and the whitespace check.
- Config contract check: pass, unchanged from the last pass — the delta touches no fcl.
- Cross-repo consistency: the `CRV_COMMISSIONING`/`v1_0` purpose-version pair in the two
  EventDisplay fcl files is a live database lookup I still have not verified from here. It
  is the one thing in this PR I am approving without checking.

**Residual risk**
- I have not run the pedestal finder on FEB-II data, so the fraction of digis the new
  first-two-sample cut removes is still unmeasured. Your call whether that matters before
  merge.

**Author follow-ups**
1. Fix `iSubEvent` in `CrvDigisFromArtdaqFragments_module.cc` (FEB-I) in a follow-up.
2. Decide whether the per-hit error records should be deduplicated per block or should
   carry the channel.

Issue #1942, opened 2026-08-25
Trigger: static-review audit — wrong-histogram bugs, unreferenced module, convention findings


oksuzian:
### Preamble

The items below come from a static review of `Trigger/` on `main` at `3f6e789c319b463d1881bde0a65c4a45cc9252ae` — all 11 tracked files (2,603 lines) read in full. No build, no runtime test, no physics validation was performed. Every line cited was read directly from that commit.

Intent is a single triaged punch list. Items under **Bugs** are high-confidence; certainty decreases down the tiers.

Two items are already tracked and are not repeated below: #1208 (kutschke) on `PrescaleEvent`'s `e.event() % ps` prescaling, and #500 on `.at()` vs `operator[]`.

---

## Bugs

1. **`Trigger/src/ReadTriggerPath_module.cc:542`** — `if(track_passed) { //helix trigger rates`. The bin filled here is index 2 of `hGlobalTrigEff` / `hGlobalTrigRej`, labelled `"Helix"` (label list at `:469`, applied at `:471-472`). `helix_passed` is assigned once at `:509` and never read anywhere — because the assignment is `|=` it counts as a use, so no unused-variable warning fires. Visible symptom: the `Helix` and `Tracks` bins are numerically identical. Fix is `if(helix_passed)`.

2. **`Trigger/src/ReadTriggerPath_module.cc:507`** — `passed |= skip_exclusive;`. `skip_exclusive` is set at `:497-498` when a path name matches any entry of `ignorePaths`, whose default is `{"Path"}` (`:92`, commented "non-trigger paths typically contain 'Path'"). Every other category at `:508-512` requires `!skip_exclusive`. So `passed`, which drives bin 0 labelled `"Total"`, is true exactly when a *non-trigger* path fired and false when only real triggers fired. The `Total` bin is therefore the rate of the ignored bookkeeping paths. Probably wants `passed = true;` inside the `if(trigNavig.accepted(path))` block at `:488`.

   Both of the above affect only `Trigger/fcl/read_trig_paths.fcl`, which is the sole consumer of `ReadTriggerPath` anywhere in Offline, Production or mu2e-trig-config — it is not in `Production/Validation/trigger.fcl` or in any trigger menu. So nothing in production is currently producing these numbers, but anyone running the module by hand reads a wrong plot.

[...]

PR #1943, opened 2026-08-25
Trigger: fix wrong-histogram bins and audit findings from #1942

oksuzian:
Addresses the findings in #1942 (static-review audit of `Trigger/`).

Five commits, kept separate so the two live bug fixes can be read and reverted on their own:

| commit | what |
| --- | --- |
| `fix(Trigger): ReadTriggerPath filled two rate bins from the wrong flag` | the two defects that make a runnable job report a wrong number |
| `fix(Trigger): ReadTriggerPath cut-flow bound, menu check, dead code` | items 8, 9, 18, 23, 26, 30, 31 |
| `fix(Trigger): ReadTriggerInfo indexing, dead state and empty-input guards` | items 3, 4, 7, 15, 17, 19, 21, 22, 27 |
| `fix(Trigger): fail-loud inputs, validated types and cleanup in the remaining modules` | items 10, 12, 14, 24, 29, 31 and the safe part of 11 |
| `cleanup(Trigger): drop unused beginRun/endRun parameter names` | warning cleanup |

### The two live bugs

`ReadTriggerPath_module.cc:542` filled the `"Helix"` bin of `hGlobalTrigEff`/`hGlobalTrigRej` from `track_passed`, so it reported the track rate and was numerically identical to the `"Tracks"` bin. `helix_passed` was computed and never read — because the assignment is `|=` it counts as a use, so no unused-variable warning fires.

`:507` drove the `"Total"` bin from `passed |= skip_exclusive`, which is true exactly when a path matching `ignorePaths` fired — the paths every other category excludes.

Both are one-line changes. Their only consumer is `Trigger/fcl/read_trig_paths.fcl`, a hand-run job, so nothing in production was affected.

### Deliberately not in this PR

Four items from #1942 need a decision I cannot make from the source:

- **item 5** — `EvalWeightedEventCounts`: the two-batch cutoff normalization cancels itself and leaves the one-batch factor behind (`p2 = lumi·n2·c1·pdf2/pdf1`, 2.7% off). The algebra is unambiguous but the intended formula is not, so this needs its author.
- **item 11 (schema half)** — converting `DigiFilter` to validated FHiCL needs a companion `mu2e-trig-config` PR: the live configs also set `triggerPath`, a key this module has never read and which validated FHiCL would reject. The same coupling blocks removing `maxCaloEnergy`, which is read from FHiCL and never applied. The reject-every-event case is fixed here without touching the schema.
- **item 13** — `MergeTriggerInfo` collects `TriggerInfo` by type via `getMany`, so re-running the trigger over its own output double-counts. Fixing it means choosing between an explicit tag list and a process-name filter.
- **item 20** — whether `ReadTriggerInfo` should exist at all. It has no FHiCL consumer anywhere in Offline, Production or mu2e-trig-config, and its successor says so in its own header. I fixed it in place rather than deleting 1,113 lines on my own judgement; if it goes, commit 3 goes with it.

Two items are partial and say so in the commit messages: the `"Calib"` bin label is left in place (item 23) because removing it shifts every bin above it, and the `printf`-to-message-service conversion (item 25) is done only where a `cout`/`printf` was already being replaced for another reason — converting the ~60 remaining `printf` calls in the two analyzers would bury the rest of the diff.

### Validation

- All eight modules compile clean and **warning-free** under `g++ -std=c++20 -Wall -Wextra` against the Musing include paths. The four `-Wunused-parameter` warnings the package used to emit were verified to pre-date this branch by compiling the same files from `mu2e/main`.
- `Trigger/CMakeLists.txt`: `PrescaleEvent` no longer links `artdaq-core-mu2e::Data`, because the include that motivated it (`Data/EventHeader.hh`) was unused. It now names `Offline::DataProducts`, where `EventWindowMarker` and `PrescaleFilterFraction` live — the latter has a `.cc`, so this is the honest dependency.
- **Not run.** No art job was executed and no histogram was compared before and after. The behavioural claims are from reading the source. The two live bugs in particular deserve a before/after run of `read_trig_paths.fcl` by someone with a suitable input file.
- No change to any FHiCL key that `mu2e-trig-config` or `Production` sets, so no companion PR is required for what is here.

### Correction after the first CI run

The first build at `367c39db` failed four job tests — `trigger`, `ceDigi`, `ceMix` and `cosmicOffSpill` — all with the same message at module construction:

```
Prescale factor for mode OnSpill is -1; it must be >= 1, since 0 rejects every event
```

That guard was mine, and it was wrong. `prescale: -1` is how `mu2e-trig-config` turns a trigger path off — 35 sites set it, including `core/filters/trigCprFilters.fcl` and `data/physMenu.json` — so `main`'s `std::max(0,prescale)` clamp and the `prescale_ > 0` guard in `filter()` were encoding a sentinel, not being defensive. Commit `fix(Trigger): prescale <= 0 means disabled, not a bad configuration` restores `main`'s behaviour exactly and documents the sentinel where the type is declared. `build (prof)` was green throughout; this was a configuration rejection at runtime, which is why the compile-only check in the Validation section above did not catch it.

The other fail-loud guard in this PR, `DigiFilter` rejecting `useStrawDigi` and `useCaloDigi` both false, was checked against the same repo: every live site sets exactly one of the two true.

### DIFF
--- a/Trigger/fcl/read_trig_paths.fcl
+++ b/Trigger/fcl/read_trig_paths.fcl
@@ -6,7 +6,6 @@
 process_name : readTriggerPaths
 
 services : @local::Services.Reco
-services.GeometryService.bFieldFile : "Offline/Mu2eG4/geom/bfgeom_reco_v01.txt"
 source: { module_type: RootInput }
 services.TFileService.fileName : "nts.user.read_trig_paths.sequencer.root"
 
@@ -21,4 +20,5 @@ physics : {
     trigger_paths : [              ]
 }
 
+# NOTE: reaches into the Production repo, so this fcl needs Production on FHICL_FILE_PATH
 #include "Production/Validation/database.fcl"
--- a/Trigger/fcl/read_trig_paths.fcl
+++ b/Trigger/fcl/read_trig_paths.fcl
@@ -20,5 +20,4 @@ physics : {
     trigger_paths : [              ]
 }
 
-# NOTE: reaches into the Production repo, so this fcl needs Production on FHICL_FILE_PATH
 #include "Production/Validation/database.fcl"
--- a/Trigger/src/DigiFilter_module.cc
+++ b/Trigger/src/DigiFilter_module.cc
@@ -28,7 +28,7 @@ namespace mu2e
   public:
     explicit DigiFilter(fhicl::ParameterSet const& pset);
     virtual bool filter(art::Event& event) override;
-    virtual bool endRun( art::Run& run ) override;
+    virtual bool endRun( art::Run& ) override;
 
   private:
     art::InputTag   _sdTag;
@@ -119,7 +119,7 @@ namespace mu2e
     return retval;
   }
 
-  bool DigiFilter::endRun( art::Run& run ) {
+  bool DigiFilter::endRun( art::Run& ) {
     if(_debug > 0 && _nevt > 0){
       mf::LogInfo("DigiFilter") << moduleDescription().moduleLabel() << " passed " << _npass << " events out of "
                                 << _nevt << " for a ratio of " << float(_npass)/float(_nevt);
--- a/Trigger/src/DigiFilter_module.cc
+++ b/Trigger/src/DigiFilter_module.cc
@@ -15,11 +15,34 @@
 #include "Offline/RecoDataProducts/inc/CaloDigi.hh"
 // #include "RecoDataProducts/inc/TriggerInfo.hh"
 // c++
+#include <algorithm>
 #include <iostream>
 #include <memory>
 #include <string>
+#include <vector>
 
 using namespace std;
+
+namespace {
+  //Sum over every CaloDigi of the pedestal-subtracted peak sample. The baseline recipe is
+  //the one CaloHitMakerFast uses (CaloReco/src/CaloHitMakerFast_module.cc:144-149), so the
+  //two modules agree on what a digi's amplitude is. CaloDigi carries no calibrated energy,
+  //only the raw waveform, so this sum is in ADC counts rather than MeV.
+  float totalCaloSignal(const mu2e::CaloDigiCollection& digis) {
+    float total(0.);
+    for(const auto& digi : digis) {
+      const std::vector<int>& waveform = digi.waveform();
+      const int peak = digi.peakpos();
+      if(peak < 0 || size_t(peak) >= waveform.size()) continue; //no usable peak sample
+      const int nSamPed = std::min<int>(peak > 3 ? 4 : std::max(peak-1, 1), waveform.size());
+      float baseline(0.);
+      for(int i = 0; i < nSamPed; ++i) baseline += waveform.at(i);
+      baseline /= float(nSamPed);
+      total += float(waveform.at(peak)) - baseline;
+    }
+    return total;
+  }
+}
 
 namespace mu2e
 {
@@ -41,7 +64,7 @@ namespace mu2e
     int             _maxnsd;  //maximum number of StrawDigi required
     int             _minncd;  //minimum number of CaloDigi required
     int             _maxncd;  //maximum number of CaloDigi required
-    float           _maxcaloE;//NOTE: read from fhicl but never applied by filter() below
+    float           _maxcaloE;//maximum summed CaloDigi amplitude; negative disables the cut
 
     int             _debug;
     // counters
@@ -76,6 +99,7 @@ namespace mu2e
     // find the collection
 
     int         nsd(0), ncd(0);
+    float       caloSignal(0.);
 
     if (_useSD){
       auto sdH = event.getValidHandle<StrawDigiCollection>(_sdTag);
@@ -85,6 +109,8 @@ namespace mu2e
     if (_useCD){
       auto cdH = event.getValidHandle<CaloDigiCollection>(_cdTag);
       ncd   = (int)cdH.product()->size();
+      //only pay for the waveform loop when the cut is enabled
+      if (_maxcaloE >= 0.) caloSignal = totalCaloSignal(*cdH.product());
     }
 
     if (_useSD) {
@@ -95,8 +121,11 @@ namespace mu2e
     }
 
     if (_useCD) {
+      //a negative maxCaloEnergy means the amplitude cut is switched off, which is what every
+      //configuration in mu2e-trig-config currently sets
       if ( (ncd >= _minncd) &&
-           (ncd <= _maxncd) ){
+           (ncd <= _maxncd) &&
+           (_maxcaloE < 0. || caloSignal <= _maxcaloE) ){
         retvalCD = true;
       }
     }
--- a/Trigger/src/DigiFilter_module.cc
+++ b/Trigger/src/DigiFilter_module.cc
@@ -7,6 +7,8 @@
 #include "art/Framework/Principal/Event.h"
 #include "art/Framework/Principal/Handle.h"
 #include "fhiclcpp/ParameterSet.h"
+#include "cetlib_except/exception.h"
+#include "messagefacility/MessageLogger/MessageLogger.h"
 // mu2e
 // data
 #include "Offline/RecoDataProducts/inc/StrawDigi.hh"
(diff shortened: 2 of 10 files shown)


Trigger/fcl/read_trig_paths.fcl
```diff
@@ -21,4 +20,5 @@ physics : {
     trigger_paths : [              ]
 }
 
+# NOTE: reaches into the Production repo, so this fcl needs Production on FHICL_FILE_PATH
```
michaelmackenzie: Seems unnecessary to add this comment

Trigger/src/DigiFilter_module.cc
```diff
@@ -39,7 +41,7 @@ namespace mu2e
     int             _maxnsd;  //maximum number of StrawDigi required
     int             _minncd;  //minimum number of CaloDigi required
     int             _maxncd;  //maximum number of CaloDigi required
-    float           _maxcaloE;//maximum energy, from the sum of all caloDigi
+    float           _maxcaloE;//NOTE: read from fhicl but never applied by filter() below
```
michaelmackenzie: Implement this optional check

Trigger/src/PrescaleEvent_module.cc
```diff
@@ -45,7 +44,7 @@ namespace mu2e
       uint32_t prescale_;
       std::string name_;
       EventMode(EventWindowMarker::SpillType type, int prescale, std::string const& name ) : type_(type),
-      prescale_(static_cast<uint32_t>(std::max(0,prescale))), name_(name) {}
+      prescale_(static_cast<uint32_t>(prescale)), name_(name) {}
```
michaelmackenzie: This changes behavior, where prescale = -1 is no longer set to 0.

Trigger/src/PrescaleEvent_module.cc
```diff
@@ -86,11 +85,14 @@ namespace mu2e
       if(mode.eventMode() == "OffSpill") eventMode_.push_back(EventMode(EventWindowMarker::offspill, mode.prescale(),mode.eventMode()));
       else if(mode.eventMode() == "OnSpill") eventMode_.push_back(EventMode(EventWindowMarker::onspill, mode.prescale(),mode.eventMode()));
       else throw cet::exception("TRIGGER") << "Unknown prescale mode " << mode.eventMode();
+      if(mode.prescale() < 1) throw cet::exception("TRIGGER") << "Prescale factor for mode " << mode.eventMode()
```
michaelmackenzie: This is not an error, prescale < 1 is meant to indicate prescaling all events

Trigger/src/PrescaleEvent_module.cc
```diff
@@ -101,7 +103,7 @@ namespace mu2e
       if (spillType == ewm.spillType()){
         ++nevt_[imode];
     // Apply the prescale
-        bool retval = mode.prescale_ > 0 ? e.event() % mode.prescale_ == 0 : false;
+        bool retval = e.event() % mode.prescale_ == 0; //prescale_ >= 1 is enforced in the constructor
```
michaelmackenzie: This is also a change in the desired behavior

Trigger/src/PrescaleEvent_module.cc
```diff
@@ -116,10 +118,11 @@ namespace mu2e
       auto ff = std::make_unique<PrescaleFilterFraction>(mode.prescale_, nevt_[imode],npass_[imode]);
       subrun.put(std::move(ff),mode.name_,art::fullSubRun());
       if(debug_ > 0){
-        double frac = mode.prescale_ > 0 ? 1.0/double(mode.prescale_) : 0.0;
-        std::cout << moduleDescription().moduleLabel() << " mode " << mode.name_ << " passed " << npass_[imode] << " events out of " << nevt_[imode]
-          << " for a ratio of " << ((nevt_[imode] > 0) ? double(npass_[imode])/double(nevt_[imode]) : 0.f)
-          << " with an expected fraction of " << frac  << std::endl;
+        const double frac = 1.0/double(mode.prescale_);
```
michaelmackenzie: Also restore this

michaelmackenzie (changes requested):
Mostly it seems fine. But it changes the behavior of prescaling, which currently uses prescale = -1 to disable a path without removing it from the trigger paths list

oksuzian:
@FNALbuild run build test

Trigger/fcl/read_trig_paths.fcl
oksuzian: Removed in 0e5e1d804.

Trigger/src/DigiFilter_module.cc
oksuzian: Implemented in 0e5e1d804 rather than documented as dead. One thing I could not settle on my own, so flagging it here: `CaloDigi` carries no calibrated energy, only the raw waveform, so what I sum is the pedestal-subtracted peak sample of every digi, in ADC counts. The baseline recipe is copied from `CaloHitMakerFast` (`CaloReco/src/CaloHitMakerFast_module.cc:144-149`) so the two modules agree on what a digi's amplitude is. Converting it to MeV would mean giving `ADCToMeV` a second home in this module, which felt like your call rather than mine — say the word and I will do it that way. A negative `maxCaloEnergy` disables the cut and skips the waveform loop, so with every instance in mu2e-trig-config on `-1` this is inert until someone sets a threshold.

Trigger/src/PrescaleEvent_module.cc
oksuzian: You are right, and CI caught it too. Restored in f789ff15c along with the other three: the normalization is back, written as `prescale > 0 ? static_cast<uint32_t>(prescale) : 0u` in place of `std::max`, and the sentinel is now documented where `EventMode` is declared so the next reader does not repeat this.

Trigger/src/PrescaleEvent_module.cc
oksuzian: Agreed — the guard is gone in f789ff15c. I had read the clamp as defensive clutter without checking mu2e-trig-config, where 35 sites set `prescale:-1`. It failed `trigger`, `ceDigi`, `ceMix` and `cosmicOffSpill` at module construction on the first CI run, which is exactly what you predicted.

Trigger/src/PrescaleEvent_module.cc
oksuzian: Restored in f789ff15c: `bool retval = mode.prescale_ > 0 ? e.event() % mode.prescale_ == 0 : false;`, unchanged from main.

Trigger/src/PrescaleEvent_module.cc
oksuzian: Restored in f789ff15c: `const double frac = mode.prescale_ > 0 ? 1.0/double(mode.prescale_) : 0.0;`. The only remaining change on that line is that the print goes through `mf::LogInfo` instead of `std::cout`.

oksuzian:
@michaelmackenzie thanks — all six comments are addressed, in two commits.

`f789ff15c` restores the prescale behaviour exactly as it is on main: the normalization, the `prescale_ > 0` guard in `filter()`, and the guarded fraction in `endSubRun()`. The sentinel is documented where `EventMode` is declared, since the reason I broke it was reading `std::max(0,prescale)` as defensive clutter without checking mu2e-trig-config, where 35 sites use `-1` to disable a path. The first CI run failed `trigger`, `ceDigi`, `ceMix` and `cosmicOffSpill` on it.

`0e5e1d804` implements `maxCaloEnergy` instead of documenting it as dead, and drops the fcl note. One open question in the thread on `DigiFilter_module.cc`: `CaloDigi` has no calibrated energy, so the cut sums pedestal-subtracted peak samples in ADC counts using `CaloHitMakerFast`'s baseline recipe. If you would rather it be in MeV I will add the conversion, but that puts a second copy of `ADCToMeV` in this module and I did not want to make that call unilaterally. Negative disables the cut, so it is inert under every current configuration.

Both files compile clean under `-Wall -Wextra`; CI re-triggered below.

oksuzian:
@FNALbuild run build test

michaelmackenzie (approved):
I think this looks good to merge, thanks!

PR #1946, opened 2026-08-26
Update CaloDtsClusterFilter parameter

michaelmackenzie:
Rename NullFilter --> Enabled and flip the logic, to make it more readable. This is connected to the Production PR: https://github.com/Mu2e/Production/pull/573


oksuzian:
@FNALbuild run build test with Mu2e/Production#573

Issue #1947, opened 2026-08-26
Mu2eG4: static-review audit — wrong copy numbers on EMC edge virtual detectors, RackSteel built from aluminum, Hayman/OPA geometry arithmetic


oksuzian:
This is a static-review audit of `Mu2eG4/` at current `main` (`a432e8d9b18163fa11db1bc96ec27edc51115255`), in the same series as #1942 (Trigger) and #1944 (CaloReco). All 143 `.cc` files (47.6k lines) were reviewed in an agentic sweep (one finder plus one adversarial refuter per unit of 1–4 files, plus a ranking pass), with module liveness taken from an index of `Production`, `mu2e-trig-config` and `Offline` configs rather than from impressions of the code. The `g4run` prolog table makes `Mu2eG4` live in every production simulation stage, so most of this package has maximal blast radius.

The seven findings below marked S0/S1 were each re-verified by hand at the quoted lines, independent of the automated chain. Severity tags follow the usual scheme: S0 wrong persisted physics/data, S1 real bug in live code with bounded effect, S2 defect with low immediate risk, S3 nit.

---

## 🔴 S0 — EMC disk-edge virtual detectors are placed with the wrong copy number

`src/constructVirtualDetectors.cc:1264-1279`. Both disk-edge placements name the volume with `vdIdDiskEdge` but pass `vdIdDiskSurf` as the `copyNo` argument of `nestTubs`:

```cpp
if( vdg->exist(vdIdDiskEdge)){
  VolumeInfo vdInfo = nestTubs(VirtualDetector::volumeName(vdIdDiskEdge),
                               vdParamsInnerDisk,downstreamVacuumMaterial,0,
                               posInnerDisk,caloDisk,
                               vdIdDiskSurf,vdIsVisible,...);   // <-- copyNo = Surf id
  ++vdIdDiskEdge;
  VolumeInfo vdInfo2 = nestTubs(VolumeName(vdIdDiskEdge),
                                vdParamsOuterDisk,downstreamVacuumMaterial,0,
                                posInnerDisk,caloDisk,
                                vdIdDiskSurf,1,...);            // <-- again, and visible flag hardcoded to 1
```

[...]

PR #1948, opened 2026-08-26
Mu2eG4: fix the S0/S1 findings from the static-review audit (#1947)

oksuzian:
Fixes the S0 and all six S1 findings from #1947 (static-review audit of `Mu2eG4/`). Seven commits, one per finding, so each can be reviewed and reverted on its own. Every finding was hand-verified at the source before fixing; the issue carries the full evidence and arithmetic.

| commit | finding | what changes in output |
| --- | --- | --- |
| EMC disk-edge VD copy number | #1947 S0 | The four `EMC_Disk_*_Edge*` VD planes now write their own volumeId (77–80) instead of 75/77; the contaminated Disk_1_SurfIn / Disk_0_EdgeIn streams become pure. Any analysis that "used" edge-VD hits before was reading mislabeled data. |
| `saveSimParticleEnd` boundary | #1947 S1 | In events that hit `maxSimParticleCollectionSize`, the last admitted SimParticle now carries real end-of-track info instead of `endDefined()==false`. |
| RackSteel Fe/Al | #1947 S1 | STM shielding house and electronics racks become steel instead of aluminum-at-steel-density. **This shifts STM background rates — explicit STM-group sign-off requested; happy to drop this commit into its own PR if preferred.** |
| Hayman spoke rod-side offset | #1947 S1 | Back-ports the corrected expression `constructStickmanTarget` already uses (its comment: "seems to be a typo in the hayman"). Rod-side spoke-wire endpoints move by the tan²θ error factor; only the +1 mm padding is kept as-is. |
| OPA notch sagitta | #1947 S1 | The notch cut into OPA support ring 2 shrinks from 15.18 mm to the intended 3.76 mm extension (live v04 values). |
| `pConePlane::operator<` | #1947 S1 | The polycone duplicate-plane gate becomes a valid strict weak ordering. A geometry with genuine duplicate planes that previously slipped through will now fail loudly — that is the point. |
| ExtMon chiller rotation | #1947 S1 | The three hall pixel chillers get the `.inverse()` every sibling call site applies; their rotation flips to the intended sign. |

### Validation

- All seven touched files compile warning-free under `g++ -std=c++20 -Wall -Wextra -fsyntax-only` against the Musing include paths (the two `-Wunused-parameter` warnings in `constructExtMonFNALInfrastructure.cc:570` are in an untouched function and present on `main`).
- **Not run:** no art job, no GDML diff, no before/after geometry rendering. The behavioural claims come from the audited arithmetic. The geometry commits (VD, Hayman, OPA, chiller) deserve a `gdmldump` + overlap-check pass in CI, and the RackSteel commit a before/after STM rate check by someone with a suitable setup.
- No FHiCL or SimpleConfig key changes anywhere, so no companion `Production` or `mu2e-trig-config` PR is required.

### Deliberately not in this PR

The 29 S2 / 2 S3 findings from #1947 (dead code, dormant guards, hand-run-path bugs). Several want maintainer decisions (e.g. whether `sourceProfileReader`, `DuplicateLogicalVolumeChecker` and `constructStudyEnv_v00*` should exist at all) and none affects live production output.

### DIFF
--- a/Mu2eG4/src/ConstructMaterials.cc
+++ b/Mu2eG4/src/ConstructMaterials.cc
@@ -448,7 +448,7 @@ namespace mu2e {
     mat = uniqueMaterialOrThrow( "RackSteel" );
     {
       G4Material* RackSteel = new G4Material( mat.name, 8.05*CLHEP::g/CLHEP::cm3, 2);
-      RackSteel->AddMaterial(findMaterialOrThrow("G4_Al"),0.985);
+      RackSteel->AddMaterial(findMaterialOrThrow("G4_Fe"),0.985);
       RackSteel->AddMaterial(findMaterialOrThrow("G4_C"), 0.015);
     }
 
--- a/Mu2eG4/src/Mu2eG4TrackingAction.cc
+++ b/Mu2eG4/src/Mu2eG4TrackingAction.cc
@@ -607,7 +607,7 @@ namespace mu2e {
   // Append end of track information to the existing SimParticle.
   void Mu2eG4TrackingAction::saveSimParticleEnd(const G4Track* trk){
 
-    if( _sizeLimit>0 && _currentSize>=_sizeLimit ) return;
+    if( _sizeLimit>0 && _currentSize>_sizeLimit ) return;
 
     G4int trackingVerbosityLevel = fpTrackingManager->GetVerboseLevel();
 
--- a/Mu2eG4/src/constructProtonAbsorber.cc
+++ b/Mu2eG4/src/constructProtonAbsorber.cc
@@ -739,7 +739,7 @@ namespace mu2e {
 
             //extend the box so that it notches through at either end, not just the center
             const double notchExtension = (opaVersion > 2) ?
-              (rin - sqrt(rin*rin-notchWidth*notchWidth)) : 0.;
+              (rin - sqrt(rin*rin-(notchWidth/2.)*(notchWidth/2.))) : 0.;
 
             G4Box* notch = new G4Box("notch", notchWidth/2.*CLHEP::mm,
                                      (notchHeight+notchExtension)/2.*CLHEP::mm, hl*1.1 );
--- a/Mu2eG4/src/constructExtMonFNALInfrastructure.cc
+++ b/Mu2eG4/src/constructExtMonFNALInfrastructure.cc
@@ -62,8 +62,7 @@ namespace mu2e {
     //----------------------------------------------------------
     // Pixel Chiller Mother Volume
 
-    // finishNesting() uses the backwards interpretation of rotations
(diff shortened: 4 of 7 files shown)


oksuzian:
@FNALbuild run build test

Mu2eG4/src/Mu2eG4TrackingAction.cc
```diff
@@ -607,7 +607,7 @@ namespace mu2e {
   // Append end of track information to the existing SimParticle.
   void Mu2eG4TrackingAction::saveSimParticleEnd(const G4Track* trk){
 
-    if( _sizeLimit>0 && _currentSize>=_sizeLimit ) return;
+    if( _sizeLimit>0 && _currentSize>_sizeLimit ) return;
```
gaponenko, line 610: Good catch

oksuzian: @gaponenko, can you please also check ExtMon part?

YongyiBWu:
On "S1 — Hayman target spoke offset uses an inverted, dimensionally wrong formula":
Yes this would fix the Hayman geometry. The affected spoke wires shuold not make any big change, if at all.
On "STM Racksteel":
This affects the supporting structures of STM shield house. @jpeck607 may have more comments on this.
image

gaponenko (changes requested):
Please revert #61546d68b53c32b85e1a6224c84b08d0469c194e
It breaks geometry.

YongyiBWu:
On "STM Racksteel": Joey confirmed the materials are supposed to be steel.

YongyiBWu (approved):

oksuzian:
@gaponenko Reverted in 675ab3e0d — thank you for catching it.

For the record of why the original claim was wrong: `constructExtMonFNALChillersInHall` composes its `rotateY` angle from already-swapped wall-vector components (`rotAngle = -M_PI*0.5 - atan2(wallVector.x(), wallVector.y())`), i.e. the caller's angle is built for the callee's un-inverted use of the rotation, and the pair is self-consistent as it stands. The room-chiller path passes `IDENTITY`, so the change moved only the hall chillers — exactly the breakage you saw. I've corrected finding 7 on #1947 accordingly. The remaining six commits are untouched.

Mu2eG4/src/ConstructMaterials.cc
sdifalco, line 1: Checked that the change is effective

Mu2eG4/src/constructProtonAbsorber.cc
sdifalco, line 1: OPAsupport_2 

Image

sdifalco (approved):

gaponenko (approved):
Thank you

oksuzian:
@FNALbuild run build test

PR #1949, opened 2026-08-26
New standalone C++ class to create/fill DQM histograms for CRV.

RobMina:
New generic DQM histogram package and refactor CRV DQM to consolidate histogram definitions between online/offline.

The DQMHelpers package provides a generic interface for defining and managing DQM histograms for both online and offline. It also includes a fcl prolog to be included by subdetector-specific DQM jobs.

The CRVDQM package implements the CRV-specific DQM histograms, consolidating the existing functionality from otsdaq-mu2e-crv/ArtModules/CrvDQM_module.cc, otsdaq-mu2e-crv/ArtModules/CrvStatusMetrics_module.cc, and Offline/CRVReco/src/CrvDQMcollector_module.cc.

Companion PRs (this PR first):
- Offline DQM: https://github.com/Mu2e/DQM/pull/36 includes the offline DQM art modules and job fcls
- Online DQM: https://github.com/Mu2e/otsdaq-mu2e-dqm/pull/39 includes the online DQM art modules and fcl

Once the above land, additional PRs should be created to:
- retire the DQM modules from otsdaq-mu2e-crv
- retire Offline/CRVReco/src/CrvDQMcollector_module.cc


rlcee:
To start, this is an impressive PR, much more than I expected to start the DQMHelper concept.  I can't absorb it all, and there are things I don't really understand.  Here are few high-level comments
- The original concept was to do endjob-type work in the post-grid-job [metrics-maker code](https://github.com/Mu2e/DQM/blob/main/src/DqmMetrics_crv.cc) or [here](https://github.com/Mu2e/DQM/blob/main/python/crv_extractor.py).  We expect to add individual DQM files together to get, for example, the total DQM output for a run or a week of runs, and we'd re-run the metrics maker on these combined files.  Not sure if this design is compatible, to what extent
- I expected only the helpers (the online/offline shared code) to live in DQMHelpers, The modules and fcl make it more like a whole dqm system.  We need to decide what lives in DQM repo vs Offline
- do you have an example output file, from raw data, we could look at?

Pending any CRV group comments, if I had to manage this merge, I might just merge and see how it works out, since too much to actually look at or think about.

RobMina:
Thanks Ray! Sorry that you already started to review, I had a few more updates I wanted to make.

I was able to test this using modified versions of the online (https://github.com/Mu2e/otsdaq-mu2e-crv/blob/mu2e/ots_ops/otsdaq-mu2e-crv/ArtModules/CrvDQM_module.cc) and Offline (https://github.com/Mu2e/Offline/blob/main/CRVReco/src/CrvDQMcollector_module.cc) modules.

As an example, I ran over one file (raw.mu2e.trk.vst.124155_000001.art) from dataset raw.mu2e.trk.vst.art and the output histogram files are:
```
/exp/mu2e/data/users/ram2aq/crv_dqm/124155/crvDigiDQM_124155_000001.root
/exp/mu2e/data/users/ram2aq/crv_dqm/124155/crvStatusDQM_124155_000001.root
```
For comparison, I also ran over a cosmic MC file (dig.mu2e.CosmicCRYExtracted.MDC2025au_best_v1_5.001400_00001974.art) from (dig.mu2e.CosmicCRYExtracted.MDC2025au_best_v1_5.art):
```
/exp/mu2e/data/users/ram2aq/crv_dqm/MDC2025au/crvDigiDQM_MDC2025au_001400_00001974.root
/exp/mu2e/data/users/ram2aq/crv_dqm/MDC2025au/crvStatusDQM_MDC2025au_001400_00001974.root
```

CRVReco/fcl/prolog_v12.fcl
```diff
@@ -27,9 +26,15 @@ BEGIN_PROLOG
       module_type                            : CrvDQMcollector
       useDQMcollector                        : true
       crvDigiModuleLabel                     : "CrvDigi"
+      crvStatusModuleLabel                   : "CrvDigi"
       #crvRecoPulsesModuleLabel               : "CrvRecoPulses"
       crvCoincidenceClusterFinderModuleLabel : "CrvCoincidenceClusterFinder"
       crvDaqErrorModuleLabel                 : "CrvDigi"
+      crvDigiDQMDir                          : "CRVDigiDQM"
+      fillInclusiveDigiDQM                   : true
+      # KPP cabling: fold ROC4 onto ROC2. Set false once the full
```
ehrlich-uva: This is a fcl parameter of the DAQ digi generator. The current Pass1 scripts already change ROC 4 to ROC 2, so that all art data should see the correct ROC numbers.

CRVReco/src/CrvDQMcollector_module.cc
```diff
@@ -172,9 +178,13 @@ namespace mu2e
...
       fhicl::Atom<std::string> crvDigiModuleLabel{Name("crvDigiModuleLabel"), Comment("label of CrvDigi module")};
+      fhicl::Atom<std::string> crvStatusModuleLabel{Name("crvStatusModuleLabel"), Comment("label of CrvStatus module; empty uses crvDigiModuleLabel"), ""};
       //fhicl::Atom<std::string> crvRecoPulsesModuleLabel{Name("crvRecoPulsesModuleLabel"), Comment("label of CrvReco module")};
       fhicl::Atom<std::string> crvCoincidenceClusterFinderModuleLabel{Name("crvCoincidenceClusterFinderModuleLabel"), Comment("label of CoincidenceClusterFinder module")};
       fhicl::Atom<std::string> crvDaqErrorModuleLabel{Name("crvDaqErrorModuleLabel"), Comment("label of module that found the CRV-DAQ errors")};
+      fhicl::Atom<std::string> crvDigiDQMDir{Name("crvDigiDQMDir"), Comment("TFileService subdirectory for CRVDigiDQM histograms"), "CRVDigiDQM"};
+      fhicl::Atom<bool> fillInclusiveDigiDQM{Name("fillInclusiveDigiDQM"), Comment("also fill BarId/SiPM/ADC in CRVDigiDQM"), true};
+      fhicl::Atom<bool> crvDigiDQMkppReadout{Name("crvDigiDQMkppReadout"), Comment("KPP cabling: fold ROC 4 onto ROC 2 and book h1/h2_channels"), true};
```
ehrlich-uva: see comment made for fcl file

CRVReco/src/CrvDQMcollector_module.cc
```diff
@@ -230,35 +243,36 @@ namespace mu2e
     std::pair<int,int> _lastRunSubrun;
 
     std::vector<int>   _nCoincidences;       //for each sector
-    std::vector<int>   _nDigis;              //for each channel
-    std::vector<int>   _nDigisROC;           //for each channel
     std::vector<TH1F*> _histPEs;             //for each channel
     std::vector<TH1F*> _histPEsROC;          //for each channel
     std::vector<bool>  _notConnected;        //for each channel
 
     std::vector<TH1F*> _histDigisPerChannelAndEvent;
```
ehrlich-uva: Is this actually being stored or used anywhere (other than being filled later in the code)? If so, it should be moved to the helper class. If not, it should be removed.

rlcee:
can the two output files be created together as one file?
comparing to /pnfs/mu2e/persistent/datasets/phy-ntd/dqm/mu2e/cosmics_crv/crvreco-004-000-file/root/f2/62/dqm.mu2e.cosmics_crv.crvreco-004-000-file.122499_000011-001.root
I see some overlap, but not obviously intended to be the same.  Is this intentional and final (for now)?
I also see two layers of directories
crv_dirs

RobMina:
Thanks Ray and Ralf!

I merged the two fcls into one so the digi/status histograms live in the same output file. 
KPP data: /exp/mu2e/data/users/ram2aq/crv_dqm/124155/crvDQM_124155_000001.root
Cosmic MC: /exp/mu2e/data/users/ram2aq/crv_dqm/MDC2025au/crvDQM_MDC2025au_001400_00001974.root

The histogram directory in each is now one level deep instead of two (each file has two separate directories, one for digi and one for status histograms).

Regarding the comparison to the dqm.mu2e.cosmics_crv.crvreco file -- there should be some overlap (specifically, the digi histograms), but I have not yet implemented the reco-level histograms in a standalone C++ class. That would be the next step after the implementation details (for example, what lives in DQM and what lives in Offline) are ironed out.

I think I addressed Ralf's three comments.

rlcee:
We still need to decide where the modules live - here, in a CRV subdir, or in DQM repo.  The metric extractor is in the DQM repo which has to be co-developed.  There are several options. The general procedure is to get the stakeholders together: Yuri, Dave, CRV and DQM to hash it out.  With the CM next week, we won't have our usual meetings, so that's a big delay.  Not sure the best way forward.

tedeschi:
Another aspect to discuss, is will this program be extended to run in the online environment? 
In our slack conversation Ralf said:
Ralf Ehrlich  [12:24 PM]
[@Rob Mina](https://mu2e.slack.com/team/U0B7C9JN069) I could imagine something like CrvDQMHelper::Fill(mu2e::CrvDigiCollection& crvDigis, mu2e::CrvStatusCollection& crvStatus) that gets called for each event in Offline and Online, where the helper class object creates and holds all of the histograms.

RobMina:
Yes, the idea was that these same classes would be used by both the online and Offline DQM.

Here's the draft PR with the modified online modules: https://github.com/Mu2e/otsdaq-mu2e-crv/pull/30

oksuzian (commented):
### PR Review Summary — Mu2e/Offline#1949

Reviewed at head `dbb2f49`. First pass from me; carries forward @ehrlich-uva's three inline comments from `c652825`.

**Decision**
- 🟡 comment only

**Scope understood**
- A new `DQMHelpers` package with two GeometryService-free helper classes (`CRVDigiDQM`, `CRVStatusDQM`) that book and fill CRV DQM histograms, intended to be shared with the otsdaq online monitor (companion draft: Mu2e/otsdaq-mu2e-crv#30).
- Two new standalone art analyzers plus `CRVDQM.fcl`, and `CrvDQMcollector` rewired to delegate its per-channel digi counting to `CRVDigiDQM`.
- The architectural question @rlcee raised — whether the modules belong here, in a CRV subdirectory, or in the DQM repo — is unresolved and is the real gate on this PR, not anything below. Everything I found is fixable within whichever home you land on.

**Carried forward from @ehrlich-uva**

- 🟢 [was the ROC4→ROC2 objection, ×2] — **FIXED**, verified. `crvDigiDQMkppReadout` no longer folds anything; I traced every use (`CRVDigiDQM.cc:104`, `:229`, `:312`, `.hh:72`) and none remaps a ROC. The Comment at `CrvDQMcollector_module.cc:184` now says "ROC4->ROC2 is the unpacker's job" and the warning text at `CRVDigiDQM.cc:222` points the reader at `useROC4asROC2`.
- ⚪ [S3, carried over] — `_nCoincidences` is dead, and the answer to your question is "no, remove it". `CrvDQMcollector_module.cc:242` declares it, `:364` resizes it, and nothing anywhere writes or reads it. It is dead on `main` too, so this is a one-line delete rather than something this PR introduced.

**Findings**

1. 🟠 [S1] `CRVDigiDQM` uses two different FEB numberings, so the same physical FEB is labelled differently on the timing plots and the rate plots
   - Evidence: `CRVDigiDQM.cc:22` is `globalFebId = (roc-1)*kNFebSlotsPerROC + feb` with `kNFebSlotsPerROC = 25` and a 1-based `feb` (the module's own guard at `:264` accepts `febIdRaw >= 1 && <= CRVId::nFEBPerROC`). `CRVDigiDQM.cc:271` computes `portIndex = (rocId-1)*CRVId::nFEBPerROC + febIdRaw - 1` — the dense, 0-based form used everywhere else in Offline (`CrvTimeOffsets_module.cc:406`, and `CrvDQMcollector_module.cc:345` on `main`). The first feeds `dtVsFeb` and `dtOutOfRangePerFeb`; the second feeds `crvDigiRates`.
   - Impact: ROC 1 FEB 3 is "Global FEB ID 3" on one plot and "FEB port 2" on the other, and the offset grows by one per ROC. Someone cross-reading the two plots to confirm which FEB has slipped its clock — which is exactly what `dtVsFeb` exists for — reads off the wrong FEB.
   - Suggested fix: use `CRVId::nFEBPerROC` for both, and drop `kNFebSlotsPerROC`. If the 25-slot stride is deliberate, the axis title needs to say which numbering it is.

2. 🟠 [S1] The KPP FEB axis cannot hold the FEB IDs `globalFebId` produces, and the counters that would tell you have no reader
   - Evidence: with the arithmetic above, ROC 1 gives 1–24 and ROC 2 gives 26–49. `nFebIdBins()` (`CRVDigiDQM.hh:70-73`) returns `kNGlobalFebBins + 1` = 33, and the guard at `CRVDigiDQM.cc:209` rejects anything ≥ 33. So any ROC 2 FEB above port 7 lands in overflow on `dtVsFeb`, `h1_channels` and `h2_channels`. `kNGlobalChannelBins = 2112` (33×64) has the same problem against a maximum `globalChannelId` of 49×64+63 = 3199. The two axes that are supposed to share this geography do not even agree with each other: `dtVsFeb` is 33 bins over `[-0.5, 32.5]` (`:62-64`) while `h2_channels`'s FEB axis is 32 bins over `[0.5, 32.5]` (`:126-128`).
   - Impact: this is the configuration that every committed caller uses — `CrvDQMcollector` appears only in `Production/Pass1/Wideband/CrvPass1_extracted.fcl`, `CrvPass1_widebandExtracted.fcl` and the PassN extracted configs, all of which take `crvDigiDQMkppReadout : true` from `prolog_v12.fcl`. The single symptom is one `mf::LogWarning` per job, because `nFebIdOutOfAxis()`, `maxFebIdSeen()`, `nDtOutOfRange()` and `maxAbsDtSeen()` (`CRVDigiDQM.hh:139-142`) are incremented but have zero callers in the repo — the module's `endJob` prints none of them.
   - Suggested fix: derive the axis size from the same expression that generates the IDs, so the two cannot drift; and either print the four counters in `endJob` or delete them.

3. 🟠 [S1] A missing input product is silently replaced with an empty collection, and the output then looks like a healthy run
   - Evidence: `CRVStatusDQMAnalyzer_module.cc:120-126` builds a `const CrvStatusCollection emptyStatus;` and uses it when the handle is invalid, with no message at any `diagLevel`. `CRVDigiDQMAnalyzer_module.cc:228-234` and `CrvDQMcollector_module.cc:433-441` do the same. `CRVDigiDQMAnalyzer_module.cc:216-226` does log a missing digi collection, but only at `diagLevel > 1`, and the shipped `CRVDQM.fcl:30` sets `diagLevel : 1`.
   - Impact: a typo in `crvStatusTag` yields a complete DQM file reporting `nRocHeaders` 0, `eventHasError` 0, `eventHasDaqError` 0 for every event — byte-indistinguishable from a run with no errors. For a monitoring module whose entire purpose is to notice that something is wrong, a silently-empty input is the one failure mode that must not be silent.
   - Suggested fix: throw. These modules have exactly one input each; a missing one is a configuration error, not a data condition.

4. 🟠 [S1] Handling of an out-of-range ROC or FEB id changed from fatal to silent
   - Evidence: on `main`, `CrvDQMcollector_module.cc:432` is `++_nDigisROC.at(channelOnline);` — a bad `GetROC()`/`GetFEB()` throws `std::out_of_range` and stops the job. The delegated path guards instead (`CRVDigiDQM.cc:262-266`) and drops the digi with no counter and no message.
   - Impact: corrupt or mis-cabled channel ids used to stop pass-1 and get investigated; now they vanish. This is the same "a warning is not error handling" concern as finding 3, one step worse because there is not even a warning.
   - Suggested fix: keep a hard failure, or at minimum count the drops and report them in `endJob`.

5. 🟡 [S2] `CRVStatusDQM` books per-link histograms from an unbounded id, and its ROC census can disagree with its own counters
   - `latencyHistFor(dtcId, linkId)` (`CRVStatusDQM.cc:170`) keys a map on the raw pair with no range check — unlike the two other ROC-indexed paths at `:231` and `:291`, which both test against `kNRocBins`. `dtcId` is a `uint8_t` off the data, so a corrupt value books another `TH1F` of `nBinsLatency` (default 1024) bins, without limit. It is also called at `:215`, five lines *above* the `HasROCHeader()` guard at `:220`, so links with no ROC attached get a `linkLatency_dtc%u_roc%u` histogram of their own, and they are also folded into `latencySumThisSubrun_` at `:210` — which makes `meanLatencyPerSubrun` move when the populated-link fraction changes rather than when latency does.
   - Separately, `++nRocHeadersTotal_` and `seenRocs_.insert` (`:228-229`) are outside the `ybin < kNRocBins` guard at `:231`, so `rocCensus` and `errorBitsVsRoc` can be empty while the scalar counters look healthy. `CRVDigiDQM` latches an `mf::LogWarning` for exactly this case; `CRVStatusDQM.cc` includes no MessageLogger at all.

6. 🟡 [S2] `cfTime` takes its threshold from the global maximum but its time from the first crossing
   - `CRVCFTime.hh:37-56`: `std::max_element` over the whole window sets `threshold`, then the loop returns at the first `i` with `adcs[i] >= threshold && adcs[i-1] < threshold`. On a double-pulse waveform an earlier, smaller pulse that reaches 20% of the later peak wins, so `time_ns` describes one pulse while `peak`/`baseline` describe another — and that time is what feeds the inter-FEB Δt at `CRVDigiDQM.cc:295`, i.e. `dtVsFeb`. Worth confirming against a real waveform sample before deciding whether it matters at CRV rates.
   - The four failure modes (`size() < 3`, `amplitude <= 0`, `amplitude < minAmplitude`, no crossing) all return the same default `CFResult` and the caller drops the hit with no counter, so a run where every waveform sits below `minAmplitude` gives empty timing plots that look like a run with no hits.

7. 🟡 [S2] The output does not survive `hadd`, which answers @rlcee's question about combining per-job DQM files
   - Both helpers book `TGraph`s keyed on EWT or subrun (`CRVDigiDQM.cc:131`, `:140`, plus `std::map<uint8_t,TGraph*> g_ubStatusVsEwt_`; `CRVStatusDQM.cc:157`, `:162`). `hadd` concatenates graph points rather than merging a series, so the combined graph is not a time series unless the files happen to be added in X order. Worse, `h1_channelsLastEwt_` and `h_dtOutOfRangePerFebLastEwt_` are last-window *snapshots* stored as `TH1`, so `hadd` will silently sum two jobs' unrelated final windows into a number that means nothing. Also relevant to the same question: `CRVDigiDQM.cc:399`, `:409`, `:439`, `:446` use `AddBinContent` rather than `Fill`, so those two histograms carry `fEntries == 0` and read as "not filling" in the ROOT stats box.

8. 🟡 [S2] Existing histograms move one directory deeper
   - `CrvDQMcollector_module.cc:305` books the helper into `tfs->mkdir(_crvDigiDQMDir)`, default `"CRVDigiDQM"`. `crvDigiRates`, `crvDigiRates_ROC*` and `crvDigisPerChannelAndEvent_CRVsector*` were at the module's top-level TFileService directory on `main` and are now nested. Any consumer that opens them by path breaks silently. Worth calling out explicitly in the PR description even though it is intentional.

9. 🟡 [S2] `README.md:44` documents a DQM-repo consumer that does not exist
   - It states that `crv_status_extractor.py --dir CRVStatusDQM` keeps working. There is no `crv_status_extractor.py` anywhere in the Mu2e organisation (`search/code` returns 0 hits), and `Mu2e/DQM` has a single `main` branch with no open PRs. This is the co-development gap @rlcee named — worth either writing that script or marking the paragraph as describing intended future behaviour.

10. ⚪ [S3] Small things, collected
   - Physics-affecting parameters carry C++ defaults: `cfFraction` 0.20 and `minAmplitude` 10 (`CRVDigiDQMAnalyzer_module.cc:65-66, 75-76`) decide which digis get a constant-fraction time at all, so they decide the contents of every dt histogram. `crvDigiTag`/`crvStatusTag` defaulting to `"CrvDigi"` combines with finding 3 to make a wrong tag produce a plausible empty file.
   - Constants that `CRVId.hh` already owns, in a file that includes it: `kNChanPerFEB = 64` (`CRVDigiDQM.hh:61`) vs `CRVId::nChanPerFEB`, used symbolically 100 lines later at `.cc:264`; `febChannel / 16` (`.cc:296`) vs `CRVId::nChanPerFPGA`; `fpgaA * 4 + fpgaB` (`.cc:520`) vs `CRVId::nFPGAPerFEB`; `5503.5` (`.cc:149`) vs `CRVId::nBars - 0.5`.
   - `CRVStatusDQM.hh:42-44` hardcodes the same three counts with the comment `// CRVId::nROC; no GeometryService`. That justification is not right — `DataProducts/inc/CRVId.hh` is a header-only `constexpr` class that needs no service, which is why the sibling `CRVDigiDQM.hh:7` includes it directly. `kNPortFlags = 24` is also re-spelled as a bare literal at `CRVStatusDQM.cc:100` and `:235`.
   - `CRVCFTime.hh:15` re-declares `kDigitizationPeriodNs = 12.5`, which already exists as `CRVDigitizationPeriod` in `CRVConditions/inc/CRVDigitizationPeriod.hh:6`. Same value today; a drift hazard, and the number appears twice more as prose in the new package (`CRVDigiDQM.cc:97`, `.hh:98`).
   - `errorsPerSubrun` is booked `50, 0, 50` (`CRVStatusDQM.cc:143`) and filled with a whole subrun's error-event count, so any subrun past 50 errors — the regime the plot exists for — piles into overflow. It is the only histogram in `Book()` whose range is not exposed through `Config`.
   - `ewtMismatch` (`CRVStatusDQM.cc:137`) uses 201 bins over `[-100, +100]` for an integer quantity, so +100 lands in overflow and the top bin can never fill. For integers the booking wants `[-max-0.5, max+0.5]`.
   - `h1_digisPerEvt` spans `[0.5, maxDigisPerEvt+0.5]` (`.cc:82-86`) but is filled with `nDigis` including zero (`.cc:306`), so a CRV that has gone dead produces no visible entries — which on an online monitor looks the same as no data flowing.
   - Unused includes `TH1F.h` and `TString.h` (`CRVDigiDQMAnalyzer_module.cc:25-26`); `standardServices.fcl` included but unused (`CRVDQM.fcl:6`). Relatedly, `CRVDQM.fcl` supplies no GeometryService, so the `fillSectorOccupancy` switch it offers at line 50 cannot actually be turned on with this config.

**Validation check**
- Build/tests run: not by me. `mu2e/buildtest` is **pending at `dbb2f49`** as I write this — the green on the PR is from `913a6a9`, which FNALbuild already marked out of date. I did not trigger a second run.
- Config contract check: pass. I compared every `Name("…")` in both new `Config` structs against every key in `CRVDQM.fcl` — no orphan keys either way, and both modules use `art::EDAnalyzer::Table<Config>`.
- Cross-repo consistency: needs follow-up. `CRVReco/src/SConscript:32` and `CRVReco/CMakeLists.txt` both gained the new library, so the two build systems agree there. The DQM-repo side (finding 7) has no companion.

**Not checked**
- Nothing was compiled or run; every finding above is static.
- `art::TFileDirectory dir = *tfs;` (`CRVDigiDQMAnalyzer_module.cc:182`, `CRVStatusDQMAnalyzer_module.cc:111`) slices `TFileService` to its base to get the module directory. It is the only occurrence of that idiom in Offline, and I could not confirm from the art headers that the sliced copy points at the per-module directory rather than the file root. If it resolves to the file root the two analyzers' histograms collide by name — worth one run to confirm, since you have the setup.
- Whether a KPP ROC 2 actually populates FEB ports above 7, and what the real range of CRV `dtcId` is. Findings 2 and 5 are arithmetic — the constants are mutually inconsistent regardless — but whether you see them in data depends on cabling I have no source for.
- Anything in `artdaq-core-mu2e`'s `CRVDataDecoder.hh`, which is not in the checkout. So I could not confirm that `GetActiveFEBFlags()` fits the `std::bitset<24>` at `CRVStatusDQM.cc:235`, or that `GetEventWindowTag()` is the same width as the DTC EWT it is compared against at `:270` — if the ROC field is narrower it rolls over independently and `ewtMismatch` grows a spurious tail.

**Author follow-ups**
1. Reconcile the two FEB numberings (finding 1) and derive the axis extent from the ID expression (finding 2).
2. Decide what should happen when an input product is missing, and make all the call sites do that same thing (findings 3 and 4).
3. Bound `latencyHistFor` the way the other two ROC-indexed paths are bounded, and move it below the `HasROCHeader()` guard (finding 5).
4. Say in the PR description that `crvDigiRates*` and `crvDigisPerChannelAndEvent_*` move into a subdirectory, and note whether the per-job files are meant to be `hadd`-ed (finding 7).
5. Either add `crv_status_extractor.py` to the DQM repo or reword `README.md:44`.

RobMina:
I added a new C++ class for the reconstructed CRV pulses and coincidences, combining the histograms from the CrvDQMcollector module and the DQM repo's DqmCrv module (https://github.com/Mu2e/DQM/blob/main/src/DqmCrv_module.cc).

I also moved the standalone analyzer modules (and fcl) out of this PR. They're still on a branch of my Offline fork, so I can move them to wherever we decide they should land: https://github.com/RobMina/Offline/tree/feature/CRVDQMAnalyzers

The CrvDQMcollector module currently contains the functionality of all three analyzer modules.

corrodis:
From the discussion on slack: What I think we need to figure out is that, online, there are some histograms (like h1_channelsLastEwt ) in which we want to show the last N events. and some that we want to show integrated. I think this concept of "the last" is less relevant for Offline. So I think we might want to have that as an option or so in the helper? How does that sound?

I think we need a flag/settings in the helper that manage the "segmentation" of histograms. I think online we want:
- some that integrate over the full run
- some that only use the last N events (like h1_channelsLastEwt) 
   - potentially the ability to store "older segments" over time
   
If we want a common helper the helper needs to support such a schema - I think.

oksuzian (commented):
### PR Review Summary — Mu2e/Offline#1949

Reviewed at head `24f2092`. Re-review of `dbb2f49`.

**Decision**
- 🟡 comment only

The code findings on their own would be an approve: every 🟠 from my last pass is fixed or reduced, and the ~1600 lines of new segmentation machinery hold up. I am not approving because `mu2e/buildtest` has never completed at this head, and because the question @rlcee raised — what lives in `DQMHelpers` versus the `DQM` repo — is still open and is not mine to settle.

**What changed since `dbb2f49`**
- The two standalone analyzers and `CRVDQM.fcl` are gone (`8d64a66`); `CrvDQMcollector` is again the only module, and it now drives all three helpers.
- New `CRVRecoDQM` helper (`91e11a4`) folding in the DQM repo's `DqmCrv` histograms and the per-channel Landau-Gauss PE fits.
- New `DQMSegmentation` / `DQMHist` / `DQMSegmentationConfig` (`eae872d`, `779d961`): FHiCL-driven job / subrun / rolling-window copies of any histogram, plus a `publish`/`group` labelling scheme for the online consumer.

**Carried forward from my review of `dbb2f49`**

- 🟢 [was S1 #1] Two FEB numberings — **FIXED**, verified. You took the second branch of the fix: `CRVDigiDQM.hh:72-73` documents the 25-slot stride as the online convention, and the axis titles now name which numbering they are — `dtVsFeb` is `;Global FEB ID;` (`CRVDigiDQM.cc:100`) while `crvDigiRates` is `;FEB channel;FEB port` (`:197`) and `crvDigiRates_ROC*` is `;Online channel in ROC;` (`:190`). A reader cross-checking two plots can now tell they are different axes.
- 🟢 [was S1 #5] `latencyHistFor` unbounded — **FIXED**, verified. `CRVStatusDQM.cc:229-231` returns an empty handle unless `rocIndexed(dtcId, linkId)` passes, which is the same `linkId < kNLinksPerDTC` and `y < kNRocBins` test the other two ROC-indexed paths use. `MessageLogger` is now included (`:8`) and `noteUnindexedRoc` (`:56-70`) reports the skip once per job.
- 🟢 [was S2 #7] `hadd` — **FIXED**, verified. `fillLivePlots` now defaults false (`CRVDigiDQM.hh:68`) and the shipped prolog does not set it, so the `TGraph`s exist only for the online monitor. `README.md` has a per-copy-type `hadd` table.
- 🟢 [was S2 #9] `crv_status_extractor.py` — **FIXED**. That claim is gone from the README; the two remaining references are to "the KPP extractor" generically.
- 🟢 [was S3 #10, most of it] `kNChanPerFEB`, `kNPortFlags` and the three hardcoded `CRVStatusDQM` counts now derive from `CRVId` (`CRVStatusDQM.hh:55-57`); `h1_digisPerEvt` spans `-0.5` so a zero-digi event is visible (`CRVDigiDQM.cc:120-123`); `errorsPerSubrun` and `ewtMismatch` are now configurable (`CRVStatusDQM.hh:203-206`). `CRVRecoDQM.hh:76-81` derives its own constants from `CRVId` rather than re-spelling them, which is the pattern I was asking for.

Still open:

1. 🟡 [S2, was S1 #2] The KPP FEB axis still cannot hold what `globalFebId` produces; the change is that it now says so.
   - `globalFebId` is unchanged (`CRVDigiDQM.cc:21-24`), so ROC 2 FEB 24 is still 49 against the 33-bin axis `nFebIdBins()` returns for `kppReadout` (`CRVDigiDQM.hh:83-86`), and `kNGlobalChannelBins` is still 2112 against a maximum `globalChannelId` of 3199. What is new is that `CRVDigiDQM.cc:245-260` warns once per job and `WriteGraphs` warns again at end of job when `maxFebIdSeen_` is off-axis, so this can no longer pass unnoticed — which was the substance of the finding. Leaving the arithmetic as it is is a defensible call.
   - Two small gaps: the per-job warning names `dtVsFeb` and `h2_channels` but not `h1_channels`, which overflows for the same digis; and `nFebIdOutOfAxis_` and `nDtOutOfRange_` are still incremented with no reader anywhere (`maxFebIdSeen_` and `maxAbsDtSeen_` now have one).

2. 🟡 [S2, was S1 #3] A missing input product is now announced in the log but still not in the output.
   - `CrvDQMcollector_module.cc:314-345` adds a one-per-job `mf::LogWarning` for each of the five products, and the reco-pulse one correctly fires only when the label is non-empty. The empty-collection substitution itself is unchanged, so the `.root` file a metrics extractor reads is still byte-indistinguishable from a healthy run — the diagnosis lives only in a log the DQM file does not travel with. For digis and status, where there is no "configured off" case, a `cet::exception` is still the right answer.

3. 🟡 [S2, was S2 #6] `cfTime` — the loop is now bounded but the ambiguity is not resolved.
   - `CRVCFTime.hh:49` now runs `i` only up to `peakIdx`. That rules out a crossing found *after* the peak, but the case I raised was an earlier, smaller pulse that reaches 20% of the later global peak: that crossing is before `peakIdx`, so it still wins, and `time_ns` still describes a different pulse from `peak`/`baseline`. Worth one look at a real double-pulse waveform to decide whether it matters at CRV rates; if it is the intended leading-edge-vs-global-peak definition, one line saying so closes it.

4. ⚪ [was S1 #4] Out-of-range ROC/FEB — resolved along the branch I offered: `nCrvIdOutOfRange_` counts them and `CRVDigiDQM.cc:306-316` warns once. Unlike `maxFebIdSeen_`, the count is not echoed at end of job.

5. ⚪ [was S2 #8] `crvRecoDQMDir : ""` keeps `crvPEsMPV*` and `crvCoincidencesClusters` in the module directory, which is the half that mattered. `crvDigiRates`, `crvDigiRates_ROC*` and `crvDigisPerChannel*` do still move under `CRVDigiDQM/`. The README layout block documents it; the PR description, which is still empty, does not.

6. ⚪ [was S3 #10, remainder] `kDigitizationPeriodNs = 12.5` (`CRVCFTime.hh:15`) still duplicates `CRVConditions/inc/CRVDigitizationPeriod.hh:6`, and `5503.5` reappears as a bare literal in the new file (`CRVRecoDQM.cc:221`) where `CRVDigiDQM.cc:180` writes `CRVId::nBars - 0.5`.

**New findings on the delta**

7. 🟡 [S2] A FHiCL rule that omits `"job"` makes every public accessor on the helpers return `nullptr` while the histogram is alive and filling.
   - Evidence: `DQMSegmentationConfig.cc:86-91` sets `rule.job = false` for any `modes` list without `"job"`, and the "selects no modes" throw at `:120` only fires when all three are off — so `modes : [ "window" ]` is a legal, documented configuration. `DQMSegmentation::addEntry` then leaves `Entry::job` null (`DQMSegmentation.cc:218-220`), and `book1`/`book2` hand that null to the handle as `job_`. `DQMHistBase::get()`, `operator H*()` and `operator->()` all return it (`DQMHist.hh:41-43`), so `CRVDigiDQM::h1_channels()`, `CRVStatusDQM::ewtMismatch()` and the other ~30 raw-pointer accessors return `nullptr` for that histogram, even though `Fill` still reaches the window copies normally.
   - Impact: those accessors exist for the otsdaq consumer in Mu2e/otsdaq-mu2e-crv#30 — the one place likely to configure `modes : [ "window" ]` in the first place. This is the "returns null when undefined" convention the coding standard names: the callers that skip the check core-dump once it can actually return null. Nothing in Offline hits it today, because the shipped fcl configures no segmentation at all.
   - Suggested fix: either make `job` unconditional and let `jobPersist : false` mean "keep it in memory only", or have the accessors fall back to the live window copy when there is no job copy.

8. 🟡 [S2] The README's opening claim is not true of the package as it now stands, and that claim is what @rlcee's open question turns on.
   - Evidence: `README.md:6-7` says "This package is helper **classes only** — no art plugins, no FHiCL, and so no GeometryService or ProditionsService dependency anywhere in its build." But `DQMHelpers/fcl/CrvDQMcollector.fcl` is a complete job configuration for a module that lives in `CRVReco`, and it sets `services.GeometryService.inputFile` and three `ProditionsService`/`DbService` blocks; `CMakeLists.txt:23` installs it with `install_fhicl`; all three helper headers define `fhicl::Atom` config structs (e.g. `CRVStatusDQM.hh:203-206`); and `CMakeLists.txt` links `fhiclcpp::fhiclcpp` and `fhiclcpp::types`.
   - Impact: the "what belongs in DQMHelpers vs the DQM repo" decision @rlcee and @tedeschi are waiting on will be made partly from this README, and it currently describes a narrower package than the one in the diff. The library itself is still service-free, which is the part that matters for the DAQ build — it is the "no FHiCL" and the shipped job fcl that are off.
   - Suggested fix: either move `CrvDQMcollector.fcl` next to the module in `CRVReco/fcl`, or reword the paragraph to say the package holds the helper classes and their FHiCL schemas plus one reference job config.

9. ⚪ [S3] `CrvDQMcollector.fcl:71-73` pins `DbService.purpose : Sim_best`, `version : v1_5` in a file whose header offers run 124155 — KPP data — as its first example. The CRV tables come from the two `textFile` entries either way, so this only sets everything else, but a simulation calibration set is a surprising default for a data DQM job. If it is only there to give `DbService` something valid to resolve, a comment saying that would help.

**Validation check**
- Build/tests run: **no completed CI at this head.** `mu2e/buildtest` has been `pending` at `24f2092` since 2026-09-07T21:42Z, and the last finished build is #3322 at `913a6a9` — six commits and roughly 3500 added lines ago, and already marked out of date by FNALbuild at the time. I did not re-trigger, since a run is nominally still in flight; if it does not report, it is worth asking FNALbuild for a fresh one.
- Config contract check: pass. Every key `CrvDQMcollector.fcl` and `prolog_v12.fcl` set resolves to a `Name(...)` in the corresponding `Config` struct, and the three helper blocks are correctly nested one level down so `fillInclusive` in the digi and reco configs cannot collide.
- Cross-repo consistency: needs follow-up, unchanged. `CRVReco/CMakeLists.txt` and `CRVReco/src/SConscript` both gained `DQMHelpers`, and the new `DQMHelpers/CMakeLists.txt` matches how `CRVReco` handles the transitive `artdaq-core-mu2e` dependency (via `RecoDataProducts`, not spelled out) — so the two build systems agree. The `Mu2e/DQM` side still has no companion PR.

**Verified in the new code**
- 🟢 `DQMSegmentation`'s ring arithmetic. Fills land in both `windowLive` and `subBlocks[cur]` (`refreshTargets`, `:166-176`), so the live copy is exact between rotations; `rebuildLive` re-sums rather than subtracting, which keeps bin errors and the entry count right; `archiveSpan` fires on `completedSubBlocks % nSub == 0`, which is also correct for `subdivisions : 1`; and every `Reset("ICES")` is followed by a `label()` call that puts the `dqmSegment` stamp back. `DQMHistTargets` caps at 4, which is exactly the number `refreshTargets` can add.
- 🟢 An unconfigured job's output is unchanged. `label()` returns immediately when an entry has neither a subrun nor a window copy (`DQMSegmentation.cc:496-499`), so with the default job-only rule no title is annotated and no `TNamed` is added — the existing histograms come out byte-identical.
- 🟢 `CRVRecoDQM::fitChannel` does not dereference an empty container: `bins.size() < 4` at `:533` guards the `bins.rbegin()` on the next line, and `bins` can legitimately be empty when every bin falls below `kFitMinBinCenter`.
- 🟢 `CRVRecoDQM::spectrum` cannot return null to its two callers: `h_PEsOffline_` and `h_PEsOnline_` are `assign`ed at `Book` (`:199-200`) regardless of `writePerChannelPE`, which only decides whether `spectraDir_` exists, and both call sites range-check first (`:454`, `:473`).

**Not checked**
- Nothing compiled or run; all of the above is static.
- `art::TFileDirectory dir = *tfs;` (`CrvDQMcollector_module.cc:181`) still slices `TFileService` to its base to get the module directory, and the PR now depends on that resolving to the module directory rather than the file root for the whole reco block. I could not confirm it from the art headers, but your example output files settle it — a one-line confirmation that `crvPEsMPV` lands beside `crvMetaData` in `crvDQM_124155_000001.root` would close it for the next reader.
- Whether a KPP ROC 2 populates FEB ports above 7, which is what decides whether finding 1 is visible in data.
- Anything in `artdaq-core-mu2e`, so I still cannot confirm that `GetActiveFEBFlags()` fits `std::bitset<24>` or that the ROC and DTC event-window tags are the same width where `ewtMismatch` differences them (`CRVStatusDQM.cc:330`).

**Author follow-ups**
1. Decide what a rule with `modes` omitting `"job"` should do to the raw-pointer accessors (finding 7) — this is the one the online PR is most likely to hit.
2. Reconcile `README.md:6-7` with the shipped fcl and the FHiCL config structs (finding 8), since that paragraph is feeding the DQMHelpers-vs-DQM-repo discussion.
3. Make the missing-product path fatal for digis and status (finding 2), or say in the README why a warning is the right level for a monitoring module.
4. Add a PR description: intent, the fact that `crvDigiRates*` and `crvDigisPerChannel*` move into `CRVDigiDQM/`, and the validation you have already done on run 124155 and MDC2025au.
5. If `mu2e/buildtest` stays pending at `24f2092`, ask FNALbuild for a fresh run — there is no completed build covering `CRVRecoDQM` or `DQMSegmentation`.

brownd1978 (commented):
I suggest keeping the DQM code in its own package, unless there's a tight coupling between the CRV reco algorithms and the DQM code.

oksuzian:
@FNALbuild run build test

oksuzian (commented):
### PR Review Summary — Mu2e/Offline#1949

Reviewed at head `8cc3ae1`. Re-review of `24f2092`.

**Decision**
- 🟡 comment only

The three commits since `24f2092` (`6198397`, `f874c00`, `8cc3ae1`) are contained: an `archiveGroup` label for the segmentation registry, `ResetForNewRun()` on both helpers, and a `fillLinkPlots`/`fillLinkGraphs` block in `CRVStatusDQM` that ports the per-link status plots of the online `CrvDQM` module. `CrvDQMcollector`, `CRVCFTime.hh`, `CRVRecoDQM` and the shipped fcl are untouched, so every finding from my last pass on those files stays as it was. `mu2e/buildtest` is now green at this head (the first completed build since `913a6a9`, fourteen commits ago), so the one code-side reason I gave last time for not approving is gone. On the code alone this would be an approve with the two 🟡 items below as follow-ups; I am leaving it at comment only because the DQMHelpers-vs-DQM-repo placement question is still with @rlcee and @brownd1978 and is not mine to settle.

**What changed since `24f2092`**
- `DQMSegmentation`: `Rule::archiveGroup` routes the `_prevN` and per-subrun archive copies to their own publish label; `ResetContents()` empties every copy and restarts the window ring without re-booking.
- `CRVDigiDQM`: `ResetForNewRun()`; the `ubStatusVsEwt` graphs are gone (the online companion Mu2e/otsdaq-mu2e-crv#30 at `178e2f7` does not reference them, so nothing breaks).
- `CRVStatusDQM`: `daqErrorCode` binning now derives from `CrvDAQerrorCodeDetail::names()`; new per-link latency, link-status, ROC-status, ROC-group and port-flag histograms plus optional step graphs vs EWT.

**Carried forward from `24f2092`** — all still open, none of the files involved changed:

1. 🟡 [S2] KPP FEB axis vs `globalFebId` (`CRVDigiDQM.cc:21-24`, `CRVDigiDQM.hh:83-86`) — carried over. One correction to my last note: `nFebIdOutOfAxis()` and `nDtOutOfRange()` do have accessors (`CRVDigiDQM.hh:153-155`); what is still missing is an end-of-job echo of those counts alongside `maxFebIdSeen_`.
2. 🟡 [S2] Missing digi/status input product is substituted with an empty collection and only logged (`CrvDQMcollector_module.cc:314-345`) — carried over.
3. 🟡 [S2] `cfTime` leading-edge vs global-peak ambiguity (`CRVCFTime.hh:49`) — carried over.
4. 🟡 [S2] A rule with `modes` omitting `"job"` makes the raw-pointer accessors return `nullptr` (`DQMSegmentationConfig.cc:86-91`, `DQMSegmentation.cc:218-220`, `DQMHist.hh:41-43`) — carried over. The online PR now sets `fillLinkPlots` and calls `ResetForNewRun()`, so it is tracking this branch closely; worth settling before that PR configures a window-only rule.
5. 🟡 [S2] `README.md:6-7` still says "no art plugins, no FHiCL" while the package ships `CrvDQMcollector.fcl` and three `fhicl::Atom` config structs — carried over.
6. ⚪ [S3] `kDigitizationPeriodNs` duplicate, `5503.5` literal, `Sim_best` default in a data DQM fcl, `crvDigiRates*` moving into `CRVDigiDQM/` undocumented in the PR body — carried over.

**New findings on the delta**

7. 🟡 [S2] A per-link step graph never shows the span from the first status to the first change.
   - Evidence: `CRVStatusDQM.cc:306-344` (`stepPoint`). On first sight of a link the graph gets one point `(ewt0, y)`. Every following event with an unchanged value takes the "slide the last point forward" branch, `g->SetPoint(g->GetN() - 1, x, y)`, which overwrites that only point, so after N unchanged events the graph holds a single point at `(ewtN, y)` and `ewt0` is gone. The first change then appends `(ewtN, last)` and `(ewtN, new)`. The re-seed after `ResetForNewRun()` (`:325-330`) has the same shape.
   - Impact: `g_linkLatencyVsEwt_link*`, `g_rocStatus_link*` and `g_portFlags_link*` draw a step that starts at the moment of the first change, not at the start of the run — which is the opposite of what a "value vs EWT" monitor is for. A link whose status never changes draws as one point.
   - Suggested fix: seed two points at creation and after re-seed, `(x, y)` and `(x, y)`, so the slide moves only the second one; or in the unchanged branch append when `GetN() == 1`.

8. 🟡 [S2] The per-link plots are keyed on `linkId` alone, so with more than one DTC the links of different DTCs share a histogram.
   - Evidence: `fillLink` (`CRVStatusDQM.cc:373-469`) reads `status.GetLinkID()` and never `GetDTCID()`; the maps `h_latencyUs_`, `h_linkStatusSummary_`, `g_linkLatency_` and the rest are `std::map<uint8_t, ...>`, and `h2_rocStatusSummary_` puts `linkId` on a `kNLinkBins = 6` axis. The same class already has `rocBin(dtcId, linkId)` (`:61-64`) and books `errorBitsVsRoc` on `dtcId * nROCPerDTC + linkId` for exactly this reason, and `CRVId::nROC = 18` with `nROCPerDTC = 6` means three DTCs in the full detector. `fillLink` also runs before the `rocIndexed` check at `:519`, so a `linkId >= 6` lands in the overflow bin without the `noteUnindexedRoc` warning the other paths give.
   - Impact: fine for KPP with one DTC, which is what the online layout was written for (`:32`), but `kppReadout : false` is already a supported configuration of the sibling helper, and the first full-CRV run would silently merge ROCs 0/6/12 into "link 0".
   - Suggested fix: key the maps and the 2D y-axis on `rocBin(dtcId, linkId)` with `kNRocBins`, or gate `fillLink` on `rocIndexed` and say in the `Config` comment that the per-link plots assume a single DTC.

9. ⚪ [S3] Literals that re-spell a constant a few lines away: `kNLinkBins = 6` (`CRVStatusDQM.hh:63`) is `kNLinksPerDTC` = `CRVId::nROCPerDTC`; `kNPortFlagBits = 24` (`CRVStatusDQM.cc:49`) is the class's own `kNPortFlags` = `CRVId::nFEBPerROC`. `kLatencyTickToUs = 0.064` (`:64`) has no source comment — a pointer to the DTC register description would do. `status.GetROCHeader().front().GetMicroBunchStatus()` (`:415`) is what `CrvStatus::GetMicroBunchStatus()` already returns after the `HasROCHeader()` check. `fillLinkGraphs : true` without `fillLinkPlots` is silently ignored (`graphDir_` is only made under `fillLinkPlots`, `:266-268`); the `Comment` says "needs fillLinkPlots", which is documentation where a validation throw would be the house pattern.

**Validation check**
- Build/tests run: **`mu2e/buildtest` passed at `8cc3ae1`** (FNALbuild, 2026-09-18 03:43 UTC, merged onto `665dc45`): build (prof), ceSimReco, ceMix, ceDigi, cosmicSimReco, trigger, check_cmake, g4surfaceCheck, rootOverlaps all ✅; whitespace check "no whitespace errors found"; clang-tidy reports 1 error / 254 warnings, which I did not read. The previous run at `24f2092` had been stuck `pending` since 2026-09-07, so this is the first completed build covering `CRVRecoDQM`, `DQMSegmentation` and the per-link code. I triggered it; nothing here is compiled or run by me.
- Config contract check: pass on the delta. `fillLinkPlots`, `fillLinkGraphs`, `maxLinkGraphPoints` (`CRVStatusDQM.hh:259-268`) and `archiveGroup` (`DQMSegmentationConfig.cc:76`, `:85`) each resolve to a `Config` member with a default, so the shipped `prolog_v12.fcl`, which sets none of them, is unchanged in effect.
- Cross-repo consistency: needs follow-up, unchanged. Mu2e/otsdaq-mu2e-crv#30 (draft, `178e2f7`) consumes `ResetForNewRun()`, `fillLinkPlots` and `publishedCopies()` from this head and does not use the removed `ubStatusVsEwt()`. Mu2e/DQM still has no companion PR. @brownd1978's comment (keep the DQM code in its own package unless it is tightly coupled to CRV reco) is consistent with `DQMHelpers` as it stands; the modules and fcl are the part @rlcee asked about.

**Verified in the new code**
- 🟢 `DQMSegmentation::ResetContents()` (`DQMSegmentation.cc:581-616`) resets every copy `copies()` would return except `subrunArchive`, which the comment says is deliberate, clears all six range structs and the ring counters, and re-labels — so a reset run does not inherit a stale `dqmSegment` stamp. `publishedCopies()` (`:655-684`) visits the five copy classes in the same order as `copies()` (`:618-641`), so an unset `archiveGroup` publishes exactly as before.
- 🟢 `nDaqErrorCodes()` (`CRVStatusDQM.cc:616-620`) reads `CrvDAQerrorCodeDetail::names()`, which exists on `main` (`RecoDataProducts/inc/CrvDAQerror.hh:19`, returns a `const&` to a static map), and the bin labels are set from the same map, so a code added upstream gets a labelled bin without touching this file. No remaining reference to the deleted `kNDaqErrorCodes`.
- 🟢 The `const DQMHist1<TH1F>` handles returned by `linkHist` can be filled: `DQMHist1::Fill` is `const` (`DQMHist.hh:78`, `:89`). Every `graphDir_` dereference is behind `config_.fillLinkGraphs && graphDir_.has_value()` (`:378`).

**Not checked**
- Whether the KPP ROC actually reports link IDs 0-5 and one DTC ID, which decides whether finding 8 is visible before the full detector.
- Whether `0.064 us` is the right DTC latency tick.
- Anything in the online module beyond grepping it for the accessors this head removes or adds.

**Author follow-ups**
1. Seed the step graphs with two points (finding 7) — a one-line change in `stepPoint`'s two seed paths.
2. Decide whether the per-link plots are single-DTC by design, and either key them on `rocBin` or say so (finding 8).
3. Items 4 and 5 from last time — the null accessor under a window-only rule, and the README paragraph — are the two that the online PR and the package-placement discussion respectively depend on.
4. The clang-tidy "1 error" in the FNALbuild table is worth a look, since none of the earlier runs covered these files.

Issue #1958, opened 2026-09-01
Mu2eKinKal: exception raised in KinKal::TimeRange constructor for regrow_kl.fcl


RobMina:
I encountered an exception when running Production/JobConfig/reco/regrow_kl.fcl, it was "Invalid Time Range" in KinKal::TimeRange constructor:
#1  0x00007ffff214af99 in KinKal::TimeRange::TimeRange (this=this@entry=0x7fffffff4e60, begin=<optimized out>, end=end@entry=11456.136692011529, ordered=true) at /cvmfs/mu2e.opensciencegrid.org/spackages/241207/spack/var/spack/environments/muse-al9-prof-e29-p106/.spack-env/view/include/KinKal/General/TimeRange.hh:17
#2  0x00007ffff2197f41 in mu2e::KKExtrap::toTrackerPerimeter<KinKal::KinematicLine>(mu2e::KKTrack<KinKal::KinematicLine>&) const::{lambda(KinKal::TimeDir)#1}::operator()(KinKal::TimeDir) const (__closure=__closure@entry=0x7fffffff50b0, tdir=<optimized out>, tdir@entry=KinKal::TimeDir::backwards) at /cvmfs/mu2e.opensciencegrid.org/spackages/241207/spack/var/spack/environments/muse-al9-prof-e29-p106/.spack-env/view/include/KinKal/General/TimeRange.hh:23
#3  0x00007ffff219858f in mu2e::KKExtrap::toTrackerPerimeter<KinKal::KinematicLine> (this=this@entry=0xde9650, ktrk=...) at ./Offline/Mu2eKinKal/inc/KKExtrap.hh:174
#4  0x00007ffff21c465b in mu2e::KKExtrap::extrapolate<KinKal::KinematicLine> (this=0xde9650, ktrk=...) at ./Offline/Mu2eKinKal/inc/KKExtrap.hh:89
#5  0x00007ffff215b030 in mu2e::RegrowKinematicLine::produce (this=0xde4680, event=...) at Offline/Mu2eKinKal/src/RegrowKinematicLine_module.cc:239

The Offline call is in toTrackerPerimeter here: https://github.com/Mu2e/Offline/blob/0ef02a66db09fda50f2cc055cf88af43032e185d/Mu2eKinKal/inc/KKExtrap.hh#L159

brownd1978:
[...]

PR #1959, opened 2026-09-01
Fix for Invalid Time Range exception in regrow_kl.fcl

RobMina:
Clamp to valid time range (using t0) to address Invalid Time Range exception in regrow_kl.fcl.

Addresses issue #1958 

Tested as bit-identical for standard LoopHelix, CentralHelix jobs. No exception raised in regrow_kl.fcl.

### DIFF
--- a/Mu2eKinKal/inc/KKExtrap.hh
+++ b/Mu2eKinKal/inc/KKExtrap.hh
@@ -152,7 +152,7 @@ namespace mu2e {
     // first. Anchor the range and time origin on the tracker interior, and intersect the whole trajectory:
     // a calo-cluster track is already extrapolated to the calo, so its end piece lies beyond TT_Back.
     double tref = midinter.good() ? midinter.time_ : ftraj.t0();
-    if(!midinter.good()) ftraj.range().forceRange(tref);
+    ftraj.range().forceRange(tref);
     auto recordFace = [&](TimeDir tdir){
       if(!ktrk.extrapolate(tdir,toPerim)) return;
       double const sgn = KinKal::timeDirSign(tdir);
--- a/Mu2eKinKal/inc/KKExtrap.hh
+++ b/Mu2eKinKal/inc/KKExtrap.hh
@@ -151,7 +151,8 @@ namespace mu2e {
     // The track meets the perimeter at both ends; extend each time direction and keep the face crossed
     // first. Anchor the range and time origin on the tracker interior, and intersect the whole trajectory:
     // a calo-cluster track is already extrapolated to the calo, so its end piece lies beyond TT_Back.
-    double const tref = midinter.good() ? midinter.time_ : ftraj.t0();
+    double tref = midinter.good() ? midinter.time_ : ftraj.t0();
+    if(!midinter.good()) ftraj.range().forceRange(tref);
     auto recordFace = [&](TimeDir tdir){
       if(!ktrk.extrapolate(tdir,toPerim)) return;
       double const sgn = KinKal::timeDirSign(tdir);


Mu2eKinKal/inc/KKExtrap.hh
```diff
@@ -151,7 +151,8 @@ namespace mu2e {
     // The track meets the perimeter at both ends; extend each time direction and keep the face crossed
     // first. Anchor the range and time origin on the tracker interior, and intersect the whole trajectory:
     // a calo-cluster track is already extrapolated to the calo, so its end piece lies beyond TT_Back.
-    double const tref = midinter.good() ? midinter.time_ : ftraj.t0();
+    double tref = midinter.good() ? midinter.time_ : ftraj.t0();
+    if(!midinter.good()) ftraj.range().forceRange(tref);
```
brownd1978: A more complete solution is to always forceRange(tref). It will be a no-op in most cases.

brownd1978 (commented):
I agree with the solution, but suggest to make it broader.

RobMina:
I agree. Test with Dave's suggested change showed no changes in outputs.

oksuzian (approved):
### PR Review Summary — Mu2e/Offline#1959

Reviewed at head `c4673ba`. First pass by me; accounts for @brownd1978's review at `56c62d0`.

**Decision**
- 🟢 approve

**Scope understood**
- Two lines in `Mu2eKinKal/inc/KKExtrap.hh`: `tref` loses its `const`, and `ftraj.range().forceRange(tref)` clamps it into the fit range before `recordFace` builds the search windows at `:160`. This removes the "Invalid Time Range" throw reported in #1958 for `regrow_kl.fcl`.
- Single topic, and the PR body states both the intent and the validation the author ran.

**Prior review**
- 🟢 [was a change request] "A more complete solution is to always `forceRange(tref)`. It will be a no-op in most cases." — FIXED in `c4673ba`, verified. The head commit drops the `if(!midinter.good())` guard that `56c62d0` had, so the clamp is now unconditional. @brownd1978's "no-op in most cases" is exactly right, and provably so; see the first verified item below.

**Verified**

- 🟢 The unconditional form cannot change behaviour on the `midinter.good()` branch, so the author's bit-identical LoopHelix and CentralHelix results carry over from `56c62d0` unchanged. `midinter` comes from `KinKal::intersect(ftraj, trk.middle(), ftraj.range(), intertol_)` at `:148`, and `intersect` restricts every candidate piece to the range it was passed (`ParticleTrajectoryIntersect.hh`, `srange.restrict(trange)`) before returning. So `midinter.time_` is inside `ftraj.range()` by construction and the clamp is a genuine no-op there, not merely one in practice.
- 🟢 Clamping before the first `recordFace` call is sufficient — the value cannot fall back outside the range afterwards. `recordFace` calls `ktrk.extrapolate(tdir, toPerim)` before reading `ftraj.range()`, and both branches of `KinKal::Track::extrapolate` (`Fit/Track.hh:791`) only widen: the geometric branch sets `TimeRange(begin, end+maxDtStep)` forwards and `TimeRange(begin-maxDtStep, end)` backwards, and the field-corrected branch appends domains through `addDomain`. `extendTraj` likewise takes a min/max of the two ranges. Since the range only grows, `tref` stays inside it for both the forwards and the backwards window.
- 🟢 `:160` is the only place in the file that needs this. Of the fourteen `ftraj.range()` uses in `KKExtrap.hh`, every other `TimeRange` construction (`:277`, `:295`, `:312`) is built from a `tstart` that was itself read off `ftraj.range().begin()` or `.end()`, so it is in range by construction. `extrapolateOPA` is the one that takes its `tstart` as a parameter, and its only caller (`:116`) captures it from the range at `:91`.
- 🟢 The call is well-formed through a const reference: `TimeRange::forceRange(double&) const` mutates its argument, not the range.

**Findings**

1. 🟡 [S2] The same "t0 outside the fit range" assumption is still live 65 lines earlier, at `KKExtrap.hh:87`
   - `auto dir0 = ftraj.direction(ftraj.t0());` runs at the top of `KKExtrap::extrapolate`, before `toTrackerPerimeter` is called at `:89`. `PiecewiseTrajectory::nearestIndex` clamps rather than throws (`time <= range().begin()` returns piece 0), so this evaluates the front piece extrapolated back to `t0` instead of raising anything. The traceback in #1958 shows `end=11456.14` on the backwards branch, meaning `t0` was *before* `range().begin()` for that event — so this line was already extrapolating outside the front piece on the very track that motivated the PR.
   - For `KinematicLine` the direction is constant along a piece, so nothing goes wrong and `regrow_kl.fcl` will be fine. For `LoopHelix` and `CentralHelix` the direction rotates with time, and `dir0.Z() > 0` on the next line picks the `TimeDir` that drives the entire upstream chain — IPA, ST, TSDA, tracker and OPA. A `t0` far enough outside the range could flip it silently.
   - Not introduced by this PR and not a reason to hold it. Worth a follow-up in the same spirit as the change already made here, since it is the same root cause in the same function.

2. 🟡 [S2] Nothing records that the clamp did anything, and the comment above it now describes the other path
   - The comment at `:152` says the code will "Anchor the range and time origin on the tracker interior". That is true when `midinter.good()`. When it is false and the clamp fires, `tref` is a trajectory endpoint, which by construction is not the tracker interior — and `dt >= 0.0` at `:165` then measures "first face crossed" outward from that endpoint instead. This is the case @brownd1978 described on #1958 as the deeper problem, and it now produces no exception and no message of any kind.
   - The class already carries `debug_`. A line under `if(debug_ > 0)` when `!midinter.good()`, and a clause on the comment saying what `tref` falls back to, would make the degraded path visible to whoever next reads a surprising `TT_Front`/`TT_Back` assignment on a cosmic.
   - Keeping #1958 open after this merges is the right call, and the PR body's "Addresses" (rather than "Fixes") already avoids auto-closing it.

**Validation check**
- Build/tests run: none by me. `mu2e/buildtest` is **pending** at `c4673ba` as I write this, so I did not trigger anything. The previous run at `56c62d0` was green across all sub-tests including `ceSimReco`, `cosmicSimReco`, `ceMix` and `g4surfaceCheck`; the `:question:` on `trigger` there is pre-existing and appears on unrelated PRs. Since `c4673ba` only removes a conditional around an already-compiled call, I do not expect the build result to differ, but the tick should be confirmed before merge.
- Config contract check: not applicable — no FHiCL, no module parameters, no data products touched.
- Cross-repo consistency: pass. Header-internal change with no interface effect; the author has no other open Mu2e PRs.

**Not checked**
- I ran nothing. The claim that the outputs are bit-identical for LoopHelix and CentralHelix is the author's, and I verified only that the head commit cannot have changed it relative to `56c62d0` — not the underlying measurement.
- Why a `KinematicLine` track reaching `toTrackerPerimeter` fails to cross `TT_Mid` in the first place. That is issue #1958's question, and this PR does not answer it.

**Author follow-ups**
1. Confirm `mu2e/buildtest` goes green at `c4673ba` before merging.
2. Consider a debug line on the `!midinter.good()` path and a clause on the `:152` comment naming the fallback anchor.
3. Separately from this PR, look at `ftraj.direction(ftraj.t0())` at `:87` for the same out-of-range `t0`.

oksuzian:
@FNALbuild run build test

brownd1978 (approved):

PR #1962, opened 2026-09-05
https://github.com/Mu2e/Offline/pull/1960 sub PR #1

YongyiBWu:
LaBrTree module and detector-tree fixes. Independent of the score-based diffusion model fix

### DIFF
--- a/STMMC/CMakeLists.txt
+++ b/STMMC/CMakeLists.txt
@@ -43,6 +43,14 @@ cet_build_plugin(HPGeWaveformsFromStepPointMCs art::module
       Offline::STMConditions
       Offline::GlobalConstantsService
 )
+
+cet_build_plugin(LaBrTree art::module
+    REG_SOURCE src/LaBrTree_module.cc
+    LIBRARIES REG
+      art_root_io::TFileService_service
+      Offline::MCDataProducts
+)
+
 cet_build_plugin(PHTree art::module
     REG_SOURCE src/PHTree_module.cc
     LIBRARIES REG
--- a/STMMC/fcl/MakeTree.fcl
+++ b/STMMC/fcl/MakeTree.fcl
@@ -37,11 +37,13 @@ physics: {
       SimParticlemvTag : @local::SimplifyStage2Data.SimParticlemvTag
       Detector: @local::SimplifyStage2Data.DetectorName.HPGe
     }
+    # HPGeTree throws for Detector: LaBr as the LaBr response is undefined there, so the
+    # LaBr tree is made by LaBrTree instead. Note that its "E" branch is the raw ionizing
+    # energy deposited in the crystal, not a LaBr detector response - see LaBrTree_module.cc.
     Stage2LaBr : {
-      module_type : HPGeTree
+      module_type : LaBrTree
       StepPointMCsTag : @local::SimplifyStage2Data.StepPointMCsTagSTMDet
       SimParticlemvTag : @local::SimplifyStage2Data.SimParticlemvTag
-      Detector: @local::SimplifyStage2Data.DetectorName.LaBr
     }
     PHSpectra : {
       module_type : PHTree
--- a/STMMC/src/ConcatenateDigitizedWaveforms_module.cc
+++ b/STMMC/src/ConcatenateDigitizedWaveforms_module.cc
@@ -112,7 +112,7 @@ namespace mu2e {
   };
 
   void ConcatenateDigitizedWaveforms::endJob() {
-    mf::LogInfo log("ConcatenateDigitizedWaveforms summary");
+    mf::LogInfo log("ConcatenateDigitizedWaveforms");
     log << "=====ConcatenateDigitizedWaveforms summary=====\n";
     log << std::left << std::setw(25) << "\tNo. input events:  " << inputEvents  << "\n";
     log << std::left << std::setw(25) << "\tNo. output events: " << outputEvents << "\n";
--- a/STMMC/src/ConcatenationFilter_module.cc
+++ b/STMMC/src/ConcatenationFilter_module.cc
@@ -57,7 +57,7 @@ namespace mu2e {
   };
 
   void ConcatenationFilter::endJob() {
-    mf::LogInfo log("ConcatenationFilter summary");
+    mf::LogInfo log("ConcatenationFilter");
     log << "==========ConcatenationFilter summary==========\n";
     log << std::left << std::setw(25) << "\tNo. input events:"      << inputEvents     << "\n";
     log << std::left << std::setw(25) << "\tNo. kept events:"       << keptEvents      << "\n";
--- a/STMMC/src/DetectorEfficiency_module.cc
+++ b/STMMC/src/DetectorEfficiency_module.cc
@@ -75,7 +75,7 @@ namespace mu2e {
     };
 
     void DetectorEfficiency::endJob() {
-        mf::LogInfo log("Detector Efficiency");
+        mf::LogInfo log("DetectorEfficiency");
         log << "========= Data summary =========\n";
         log << "Accepted steps: " << acceptedSteps << "\n";
         log << "Rejected steps: " << rejectedSteps << "\n";
--- a/STMMC/src/LaBrTree_module.cc
+++ b/STMMC/src/LaBrTree_module.cc
@@ -1,5 +1,5 @@
-// Adapted from ReadVirtualDetector_module.cc
-// For StepPointMCs in STMDet, generates a TTree with energy in branch "E" and time in branch "time". For a chosen STM detector
+// Adapted from HPGeTree_module.cc
+// For StepPointMCs in STMDet, generates a TTree with energy in branch "E" and time in branch "time", for the LaBr detector
 //  - Iterate over the StepPointMCs, determine the associated SimParticle
 //  - Find the most parent particle of the associated SimParticle within the associated STMDet
 //  - Increment the energy associated with the parent particle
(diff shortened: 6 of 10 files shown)


oksuzian (approved):
### PR Review Summary — Mu2e/Offline#1962

Reviewed at head `5aef29b`. First pass; no prior reviews on the PR.

**Decision**
- 🟢 approve

**Scope understood**
- A new analyzer `STMMC/src/LaBrTree_module.cc` (254 lines) plus its `cet_build_plugin` entry, and `MakeTree.fcl`'s `Stage2LaBr` switched from `HPGeTree` to it with the `Detector` parameter dropped.
- `HPGeTree::topParent` gains a guard so a `SimParticle` with no parent is treated as its own top parent instead of dereferencing a null `art::Ptr`, with a count reported at end of job. The same guard is present in the new `LaBrTree`.
- `VirtualDetectorTree` gains `px`, `py`, `pz` branches, and six modules have their message-facility category renamed to match the module name.
- Blast radius is small: nothing outside `STMMC` references any of these modules, and `MakeTree.fcl` is a hand-run script (`end_paths` is `@nil`), not part of any production chain.

PR hygiene: this is four independent changes in one PR — a new module, a bug fix in `HPGeTree`, new branches in `VirtualDetectorTree`, and a logging rename sweep. The coding standard asks for these to be split. I am not holding the PR for it, since everything is inside one analysis package and the description does name the pieces, but a one-line body is thin for a PR that adds a module.

**Verified**

- 🟢 The `Detector` parameter removal from `Stage2LaBr` is correct and complete. `LaBrTree::Config` declares exactly `StepPointMCsTag` and `SimParticlemvTag`, so validated FHiCL would reject a leftover `Detector` key rather than ignore it, and no other `.fcl` in the repository instantiates `HPGeTree` — `grep -rn "HPGeTree" --include=*.fcl` returns only the two entries in `MakeTree.fcl`.
- 🟢 The position cut was copied correctly. `LaBrTree` keeps steps with `x >= xBeamCentre`, which is byte-for-byte the condition `HPGeTree` applies on its `detector == "LaBr"` branch, and matches the new comment that the LaBr crystal sits on the `+x` side.
- 🟢 The message-category renames break nothing. Grepping every `.fcl` in Offline for the six old category strings ("Detector tree", "PH tree summary", "Virtual detector tree summary", "Detector Efficiency", "ConcatenationFilter summary", "ConcatenateDigitizedWaveforms summary") returns no hits, so no `services.message.destinations.*.categories.*` block is keyed on them. The new names are unique and match the module names, which is the improvement the standard asks for.
- 🟢 The `HPGeTree` parent guard cannot trip the check that follows it. When the guard fires it returns `particle.id()`, and `analyze` has just inserted exactly that id into `SimParticleIds` from `SimParticles.at(step.trackId()).id()`, so the `LogicError` throw at the "precautionary check" is unreachable on this path.
- 🟢 Sub-tests are green at this head so far — `ceMix`, `ceDigi`, `ceSteps`, `ceSimReco`, `cosmicSimReco`, `cosmicOffSpill`, `muDauSteps`, `g4study`, `g4test_03MT`, `POT`, `transportOnly` all report success. See the validation section for the umbrella job.

**Findings**

1. 🟡 [S2] `LaBrTree` is a near-verbatim copy of `HPGeTree` that could have been three deletions in `HPGeTree`, and the copy leaves `HPGeTree`'s LaBr path permanently dead
   - Normalising the class names and diffing the two files at this head, they differ only in comments and in three places: the `Detector` config atom with its two constructor throws, the position-cut branch, and the two `if (detector == "HPGe")` guards around the `EDeps` accumulate and emplace. Everything else — `topParent`, `parentTime`, the whole of `analyze`, the fill loop, the member layout — is identical, roughly 200 duplicated lines.
   - Deleting the `detector == "LaBr"` throw in the `HPGeTree` constructor and removing the two `if (detector == "HPGe")` guards would make `HPGeTree` with `Detector: "LaBr"` produce exactly the tree this new module produces. Those guards are true whenever `detector == "HPGe"`, so removing them cannot change HPGe behaviour — which means the FIXME's stated reason for the split ("this module exists to fill that gap without changing HPGeTree's behaviour") does not require a separate module. Nor does the new module define a LaBr response; both write `step.ionizingEdep()`.
   - The cost is visible inside this PR: the null-parent fix had to be written twice, in two files, identically. It will have to be maintained twice from now on.
   - Meanwhile `HPGeTree` can now only ever be constructed with `"HPGe"`, so its `Detector` atom, the `detectors` vector, both constructor throws, the `else if (detector == "LaBr")` cut and the two `detector == "HPGe"` guards are dead, and `SimplifyStage2Data.DetectorName.LaBr` in `STMMC/fcl/prolog.fcl:166` is no longer referenced anywhere in the repository.
   - Suggested fix: either de-guard `HPGeTree` and drop `LaBrTree` and its CMake entry, or, if the separate module is wanted for the future response model, strip the now-dead LaBr handling out of `HPGeTree` in the same PR so there is one authoritative place for each detector.

2. 🟡 [S2] The parent guard re-implements accessors `SimParticle` already provides, and its comment describes a mechanism Mu2e does not use
   - `HPGeTree_module.cc:118` and the same line in `LaBrTree_module.cc` test `parent.isNull() || !parent.isAvailable()`. `SimParticle` already has `hasParent()` (`MCDataProducts/inc/SimParticle.hh:186`) and `isTruncated()` (`:201`), and `hasParent()` is the repo-wide idiom for exactly this walk — `CalPatRec/src/HlPrint_ComboHit.cc:87`, `CaloDiag/src/CaloExample_module.cc:332`, `Analyses/src/FilterVDHits_module.cc:156` and about a dozen others all write `while (p->hasParent()) p = p->parent();`.
   - The comment says the guard covers "a compressed SimParticleCollection [that] can leave a non-null Ptr whose pointee was dropped". That is not how Mu2e truncates a genealogy: `Compression/src/CompressDetStepMCs_module.cc:572` sets `newsim.parent() = art::Ptr<SimParticle>()`, a null Ptr, and `SimParticle::isTruncated()` is defined on `_parentSim.isNull()`. So `isNull()` is doing all the work for both cases the comment names. `isAvailable()` is a product-level check, not an element-level one — `Analyses/src/PointerCheck_module.cc:267-273` treats `isNonnull()`, `isAvailable()` and a successful `get()` as three separate conditions.
   - This is harmless as written, but the comment will mislead the next reader. Suggested fix: use `particle.hasParent()` and `parent->hasParent()`, and correct the comment to name a primary and `isTruncated()`. `isTruncated()` would also let the end-of-job line separate genuine primaries from a truncated genealogy, which is the distinction the message is currently trying to make in words.

3. 🟡 [S2] `xBeamCentre = -3904.0` is now hardcoded in two modules
   - `LaBrTree_module.cc` and `HPGeTree_module.cc` each carry `double xBeamCentre = -3904.0;` with no comment on where the number comes from, and the two crystals are separated by comparing against it. If the DS axis or either crystal moves, the two files must be edited together or the HPGe and LaBr selections will overlap or leave a gap, silently.
   - The geometry already knows this: `GeomHandle<STM>` exposes `getHPGeDetectorPtr()` and `getLaBrDetectorPtr()` (`STMGeom/inc/STM.hh:64-65`), each with `originInMu2e()`. `GeometryService/src/KinKalGeomMaker.cc:633` is the in-repo precedent for deriving exactly this kind of split "from config so it tracks the geometry instead of a literal".
   - The literal is pre-existing, but this PR is what makes it a duplicate, which is the point at which it becomes a maintenance hazard rather than a style issue.

4. ⚪ [S3] New-file hygiene in `LaBrTree_module.cc`
   - Three includes are carried over unused: `art/Framework/Principal/Run.h`, `art/Framework/Principal/Handle.h` (the module uses `event.getProduct`, no handles) and `fhiclcpp/ParameterSet.h` (it uses `fhicl::Atom`). Meanwhile `std::set`, `std::map`, `std::vector`, `std::tuple` and `std::find` are all used with only `<limits>` included, so the file relies on transitive includes.
   - `SimParticleIds` is a `std::set`, but all three lookups use `std::find(begin, end, ...)`, which is a linear scan; `count()` or `contains()` is the `std::set` idiom and is what the type was chosen for. `topParent` does up to two of these per step and `analyze` a third.
   - `VirtualDetectorTree_module.cc:78-80` labels the new `px`/`py`/`pz` branches `// MeV`; momentum is MeV/c. The neighbouring `E` comment is correct.

**Validation check**
- Build/tests run: none by me. `mu2e/buildtest` is **pending** at `5aef29b` as I write this and I did not trigger anything, since a run is already in progress at the current head. The eleven sub-contexts that have reported are all green; the umbrella result and the sub-tests with no status context (`g4surfaceCheck`, `rootOverlaps`, `check_cmake`, the whitespace check) still need to be read off FNALbuild's table before merge.
- Config contract check: pass. `Stage2LaBr`'s keys match `LaBrTree::Config` exactly, `Stage2HPGe` is unchanged, and no `.fcl` outside `STMMC` touches any module in this diff.
- Cross-repo consistency: pass. Nothing in `Production` or `mu2e-trig-config` references these modules, and the author has no companion PR for this one (#1960 is the separate score-based diffusion change, which the body says this is independent of).

**Not checked**
- I ran nothing, so I have not seen either tree produced. In particular, `Stage2HPGe` and `Stage2LaBr` both call `tfs->make<TTree>("ttree", ...)`, and this is the first configuration in which the two coexist — on `main` the LaBr instance threw at construction, so only one was ever made. art's TFileService gives each module label its own TDirectory, so `Stage2HPGe/ttree` and `Stage2LaBr/ttree` should not collide, but I did not confirm that on a real output file.
- Whether the LaBr side of the `x` cut actually selects the LaBr crystal and nothing else in `STMDet`. I checked only that the condition matches what `HPGeTree` already used for LaBr; I did not compare `-3904.0` against the crystal envelopes in the STM geometry.
- The physics claims in the new file's header comment about LaBr3:Ce non-proportionality and resolution. They read as reasonable and they only justify the warning, but they are the author's, not verified by me.
- Whether the null-parent case the guard fixes is a genuine primary or a mixing artefact. `MakeTree.fcl` reads `compressSTMDet:` products that come from an upstream job I do not have here, and `STMMC/fcl/Mix.fcl` remaps `SimParticle` Ptrs through `MixBackgroundFrames`, so I could not establish which of the two produces the parents being counted.

**Author follow-ups**
1. Confirm `mu2e/buildtest` goes green at `5aef29b` and read the sub-test table, not just the eleven status contexts.
2. Decide between the two shapes in finding 1: either de-guard `HPGeTree` and drop `LaBrTree`, or keep `LaBrTree` and strip the dead LaBr handling out of `HPGeTree` and `SimplifyStage2Data.DetectorName.LaBr` out of `STMMC/fcl/prolog.fcl` in the same PR.
3. Replace the raw `art::Ptr` tests with `SimParticle::hasParent()`, and correct the comment about how compression truncates a genealogy.
4. On the first real run, check that `Stage2HPGe/ttree` and `Stage2LaBr/ttree` both appear in the output file and that the LaBr entry count is what you expect.

PR #1964, opened 2026-09-05
https://github.com/Mu2e/Offline/pull/1960 sub PR #3

YongyiBWu:
VDResampler: coordinate transforms, naming, pTotal resampler

### DIFF
--- a/MachineLearningTools/inc/ScoreBasedDiffusionModel.hh
+++ b/MachineLearningTools/inc/ScoreBasedDiffusionModel.hh
@@ -1,4 +1,4 @@
-// Module for training and using score-based diffusion model
+// Module for training and using (variance preserving) score-based diffusion model
 // Added by Yongyi Wu
 // Mar. 2026
 
@@ -7,6 +7,7 @@
 #include <vector>
 #include <algorithm>
 #include <cmath>
+#include <limits>
 #include <string>
 #include <iomanip>
 #include <fstream>
@@ -14,6 +15,7 @@
 #include <iostream>
 #include <cctype>
 #include <cassert>
+#include <utility>
 
 #include "CLHEP/Random/RandomEngine.h"
 #include "CLHEP/Random/RandFlat.h"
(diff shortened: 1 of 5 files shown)


YongyiBWu:
The build failed due to mismatches when PR was splitted. Padded intermdiate commit [ecbdf8e](https://github.com/Mu2e/Offline/pull/1964/commits/ecbdf8e41a00210ffdbd2826b4d97378849d2383) to allow compilation. These temporary changes are reverted at later sub PRs.

YongyiBWu:
@FNALbuild run build test

YongyiBWu:
@FNALbuild run build test

oksuzian (changes requested):
### PR Review Summary — Mu2e/Offline#1964

Reviewed at head `2aafe5967`. First pass, reviewing the **increment** over #1963 (`a32e69eb8`), not the cumulative diff GitHub renders.

**Decision**
- 🔴 request changes — **on the base, not on the code.** Nothing in the increment needs changing; finding 1 is a 🟢. The single ask is finding 2: this PR targets `main`, so a maintainer can merge it on its own, and doing that lands a known 🔴 [S0] on `main`. Retargeting the base to #1963's branch removes the hazard and costs nothing. Say the word that the six sub-PRs merge as one unit and I will convert this to 🟢 immediately — the code is ready either way.

**Scope understood**
- Increment over #1963's head = 3 files, +1351/-102: `STMMC/inc/VDResamplerNameHelper.hh` (new, 194 lines), `STMMC/inc/VDResamplerPtotResampler.hh` (new, 297), `STMMC/inc/VDResamplerTransforms.hh` (+860/-102). Body: "VDResampler: coordinate transforms, naming, pTotal resampler".
- Because the base is `main` rather than #1963, the rendered diff is 4802/-456 — the whole stack to this point. I diffed `a32e69eb8...2aafe5967` instead.

**Findings**

1. 🟢 [verified] The increment's three files are byte-identical to the code already approved at `13f716223`.
   - Evidence: blob shas at this head and at #1968's head match exactly — `0aaf4fd04` `VDResamplerNameHelper.hh`, `792513ad8` `VDResamplerPtotResampler.hh`, `25cc44453` `VDResamplerTransforms.hh`.
   - So the transforms (mutual invertibility of the `atanh`/`tanh`, `atanh(sqrt)`/`tanh²`, `atanh(ρ²)`/`sqrt(tanh)` and `asinh`/`sinh` pairs, and the orthonormal local-polar rotation), the name helper and the pTotal resampler were all read line by line in the 2026-09-22 review of #1968 and are not re-litigated here. No new finding in this increment.

2. 🔴 [S0] **Merging this PR before #1967 puts a silent argument rebind on `main`, and six other known defects with it.**
   - Evidence: at `2aafe5967`, `ScoreBasedDiffusionModel::generateSample` has already gained its two extra parameters, but all six call sites are still positional — `VDResamplerGenerateFromModel_module.cc:263,274,284` and `VDResamplerGenerateMix_module.cc:376,387,397`, each `generateSample({}, useHeun_, diffusionSteps_)` or the `stage2Condition` variant. `int` → `bool` converts silently, so `useHeun_` and `diffusionSteps_` bind to the wrong parameters with no warning and CI stays green. Verified: CI *did* pass at this head on 2026-09-05. The call sites are rewritten only at `53ffc662d` (#1967), where they move behind `VDResamplerGenerateCommon.hh`.
   - Six further defects from my 2026-09-08 review of #1963 are also still live in `ScoreBasedDiffusionModel` at this head (`.cc` blob `fdef81ac5`), each fixed at `996d154d4` (#1968):

     | at `2aafe5967` | evidence at this head | fixed at |
     |---|---|---|
     | 🟠 `SBDMGeneratedSample` implicitly converts to its `value` member, so a caller written for the old API silently receives unnormalized values where the new API distinguishes `zscore` from `value` | `ScoreBasedDiffusionModel.hh:41` `operator const std::vector<double>&() const { return value; }` | `13f716223`, struct made plain |
     | 🟠 LOGSIG bounds unvalidated — `logSigMax > 1` drives `alphabar` to 0, so the data term vanishes and the model trains on noise over part of the schedule | only mention of LOGSIG validation is the v-prediction override at `.cc:324`; no bound throws | four throws added at `.cc:314-330` |
     | 🟡 CSV loader rejects every checkpoint written by then-current `main` | `.cc:3039` `throw ... << "Normalization parameter size mismatch"` | throw replaced by a named-sizes message plus a real check for a missing `[DATA_NORMALIZATION]` section |
     | 🟡 clip statistics cumulative but printed per epoch | `clipCount_` initialised in the ctor at `.cc:297`, incremented at `:980`, printed at `:1473`, never reset | reset at the top of each epoch |
     | 🟡 `firstLayerBlockMagnitudes` documents a per-sample gradient magnitude and computes the norm of the mean, so a healthy feature reads as dead at convergence | `.hh:74-75` and `:537` describe it as the gradient over the block; only `gradL2` exists | `gradRmsL2` added beside `gradL2`, both labelled |
     | 🟡 `dimWeights_` applied whether or not the controller is enabled, so switching it off does not stop the skew | `.cc:1379` and `:1675` multiply by `dimWeights_[i]` unconditionally | both gated: `useDimWeightController_ ? dimWeights_[i] : 1.0` |

   - Impact: none of these is a defect *of this PR* — this is the stack's second link and every one of them is fixed later in the same stack. They only matter if #1964 lands on its own. **This is the same open question I raised on #1963 (`issuecomment-5770880650`) and it is still unanswered.** If the six sub-PRs merge as one unit, ignore this finding entirely. If they merge one at a time, #1964 must be rebased onto #1967 first, and then this is 🔴.
   - Suggested fix: state the intended merge order in the PR body, or retarget the base to #1963's branch so GitHub shows the increment and enforces the order.

**Validation check**
- Build/tests run: FNALbuild passed at `2aafe5967` on 2026-09-05T12:56Z. No new run triggered — nothing in this review needs one.
- Config contract check: n/a, the increment adds no fhicl keys.
- Cross-repo consistency: n/a for this increment. The Production-side keys this workflow needs were checked at the tip under #1968.

**Residual risk**
- Not checked: I did not re-derive the transform algebra or the pTotal resampler here; that was done at `13f716223` and the files are identical, which is a stronger statement than a re-read would be.
- The rendered diff on this PR is the stack, so anyone reading #1964 without the increment diff will see 4802 lines and reasonably assume they are unreviewed.

**Author follow-ups**
1. Answer the merge-order question from #1963: as a unit, or one at a time?
2. If one at a time, retarget each sub-PR's base to its predecessor. That makes GitHub render the increment, enforces the order, and removes finding 2 outright.

oksuzian (approved):
### PR Review Summary — Mu2e/Offline#1964

Reviewed at head `2aafe5967ab2ddf8e6ed3645b6d01f2d2a86c70b`. Correction to my review of the same head ([review 5278641338](https://github.com/Mu2e/Offline/pull/1964#pullrequestreview-5278641338)); no new code has been read.

**Decision**
- 🟢 approve

**Why this replaces the request for changes**

My earlier review asked for changes without asking for any change to this PR's code. Its finding 1 verified the increment over #1963 as clean, and its finding 2 was, in its own words, not a defect of this PR: the argument rebind in `generateSample` and the other inherited items come from #1963 and are fixed later in the same stack, at #1967 and #1968. That is a merge-order note, and I should not have graded it 🔴 [S0]. Nothing in this PR needs to change, so I am withdrawing the block.

**Merge order (note, not a finding)**
- This PR targets `main` and carries the whole stack up to this point. Merge the stack as one unit, or merge this PR only after #1967, so that the rebind never lands on `main` by itself. The per-item evidence stays in my earlier review.
- The substantive request-changes review on #1963, where those defects are introduced, still stands.

PR #1965, opened 2026-09-05
https://github.com/Mu2e/Offline/pull/1960 sub PR #4

YongyiBWu:
VDResampler: training modules and curriculum

### DIFF
--- a/MachineLearningTools/inc/ScoreBasedDiffusionModel.hh
+++ b/MachineLearningTools/inc/ScoreBasedDiffusionModel.hh
@@ -1,4 +1,4 @@
-// Module for training and using score-based diffusion model
+// Module for training and using (variance preserving) score-based diffusion model
 // Added by Yongyi Wu
 // Mar. 2026
 
@@ -7,6 +7,7 @@
 #include <vector>
 #include <algorithm>
 #include <cmath>
+#include <limits>
 #include <string>
 #include <iomanip>
 #include <fstream>
@@ -14,6 +15,7 @@
 #include <iostream>
 #include <cctype>
 #include <cassert>
+#include <utility>
 
 #include "CLHEP/Random/RandomEngine.h"
 #include "CLHEP/Random/RandFlat.h"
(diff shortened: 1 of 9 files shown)


YongyiBWu:
@FNALbuild run build test

oksuzian (changes requested):
### PR Review Summary — Mu2e/Offline#1965

Reviewed at head `7d7d64fc1`. First pass, reviewing the **increment** over #1964 (`2aafe5967`), not the cumulative diff GitHub renders.

**Decision**
- 🔴 request changes — **on the base, not on the code.** Nothing in the increment needs changing; finding 1 is a 🟢. The single ask is finding 2: this PR targets `main`, so a maintainer can merge it on its own, and doing that lands a known 🔴 [S0] on `main`. Retargeting the base to #1964's branch removes the hazard and costs nothing. Say the word that the six sub-PRs merge as one unit and I will convert this to 🟢 immediately — the code is ready either way.

**Scope understood**
- Increment over #1964's head = 7 files, +3044/-386: the training modules and their shared header — `STMMC/src/VDResamplerTrain_module.cc`, `STMMC/src/VDResamplerTrainFromRoot_module.cc`, `STMMC/inc/VDResamplerTrainCommon.hh`, `STMMC/src/SConscript`, `STMMC/CMakeLists.txt`, plus `MachineLearningTools/inc/ScoreBasedDiffusionModel.hh` and `src/ScoreBasedDiffusionModel.cc`. Body: "VDResampler: training modules, curriculum and planner".
- Base is `main`, so the rendered diff is 7795/-791 — the whole stack to this point. I diffed `2aafe5967...7d7d64fc1` instead.

**Findings**

1. 🟢 [verified] Every line of this increment has already been reviewed, and the differences from the approved tip are confined to later additions.
   - Three of the seven files are byte-identical to `13f716223`: `STMMC/src/VDResamplerTrain_module.cc`, `STMMC/src/VDResamplerTrainFromRoot_module.cc`, `STMMC/src/SConscript`. Those carry the substance of this increment — the curriculum, the planner and the two training modules — and were read line by line in the 2026-09-22 review of #1968.
   - The four that differ from the tip differ *only* by work that arrives later in the stack, which I checked hunk by hunk:
     - `STMMC/inc/VDResamplerTrainCommon.hh` (+29/-14 to tip): entirely the `gradRmsL2` branch, histogram and log-line plumbing (`:1731,1739,1746,1759,1766,1774`), plus comments. No behavioural difference otherwise.
     - `STMMC/CMakeLists.txt` (+20/-0 to tip): purely additive — `Offline::GeneralUtilities` on plugins that do not yet need it at this head, and the `VDResamplerReconfigure` / `VDResamplerRegeneratePlots` blocks for modules that do not exist yet. **Checked and correct at this head:** the four modules that include `Offline/GeneralUtilities/inc/ParameterSetFromFile.hh` at `7d7d64fc1` (`VDResamplerConfigure`, `VDResamplerReconfigure`, `VDResamplerTrain`, `VDResamplerTrainFromRoot`) all carry `Offline::GeneralUtilities` in `LIBRARIES REG`, and the two generators, which do not include it yet, correctly do not.
     - `ScoreBasedDiffusionModel.hh` / `.cc`: the older blob, covered by finding 2.
   - No new finding in this increment.

2. 🔴 [S0] **Merging this PR before #1967 puts a silent argument rebind on `main`, and five other known defects with it.**
   - Evidence: at `7d7d64fc1` the six `generateSample` call sites are still positional — `VDResamplerGenerateFromModel_module.cc:263,274,284` and `VDResamplerGenerateMix_module.cc:376,387,397` — while the signature has already gained two parameters. `int` → `bool` converts silently, so `useHeun_` and `diffusionSteps_` bind to the wrong parameters and CI stays green; it passed at this head on 2026-09-05. The call sites are rewritten only at `53ffc662d` (#1967).
   - `ScoreBasedDiffusionModel` is at blob `531afd47f` here — one generation newer than #1964's `fdef81ac5`, but still short of the tip's `996d154d4`. Five defects from my 2026-09-08 review of #1963 are live:

     | at `7d7d64fc1` | evidence at this head | fixed at |
     |---|---|---|
     | 🟠 `SBDMGeneratedSample` implicitly converts to its `value` member, hiding the `zscore`/`value` distinction from a caller written for the old API | `.hh:41` `operator const std::vector<double>&() const { return value; }` | removed at `53ffc662d` (#1967) |
     | 🟠 LOGSIG bounds unvalidated — `logSigMax > 1` drives `alphabar` to 0, so the data term vanishes and part of the schedule trains on pure noise | no bound throws; the only LOGSIG check is the v-prediction override | four throws at `.cc:314-330` in `13f716223` |
     | 🟡 CSV loader rejects every checkpoint written by then-current `main`, and accepts one with no `[DATA_NORMALIZATION]` section at all | `"Normalization parameter size mismatch"` throw still present | replaced by a named-sizes message plus a missing-section throw |
     | 🟡 clip statistics cumulative but printed per epoch, so `ClipRatio` / `AvgClipScale` are running averages mislabelled as per-epoch | counters initialised in the ctor and never reset | reset at the top of each epoch |
     | 🟡 `dimWeights_` applied whether or not the controller is enabled — switching it off stops the adaptation but not the skew | both gradient sites multiply by `dimWeights_[i]` unconditionally | both gated: `useDimWeightController_ ? dimWeights_[i] : 1.0` |

     The sixth, `firstLayerBlockMagnitudes` reporting the norm of the mean gradient while documenting a per-sample magnitude, is also still live here — that is the same `gradRmsL2` work listed under finding 1, seen from the other side: this increment adds the *plumbing* for a diagnostic whose second measure does not exist yet.
   - Impact: none of these is a defect of this PR, and all are fixed later in the same stack. They matter only if #1965 lands alone. **Same unanswered question as on #1963 (`issuecomment-5770880650`).** Merge as a unit and this finding is moot; merge one at a time and #1965 must follow #1967, at which point it is 🔴.
   - Suggested fix: state the merge order in the PR body, or retarget the base to #1964's branch.

**Validation check**
- Build/tests run: FNALbuild passed at `7d7d64fc1` on 2026-09-05T13:10Z. No new run triggered.
- Config contract check: n/a for the increment — the training fhicl interface it introduces was checked at the tip, where it is finding 1 of the #1968 review (95 keys spelled twice across the two trainers).
- Cross-repo consistency: n/a at this head.

**Residual risk**
- Not checked here: the training curriculum's numerical behaviour, the peak-importance sampling windows and the EMA promotion logic. Those were read at `13f716223`, and the files carrying them are byte-identical at this head.
- The rendered diff is the whole stack, so #1965 looks like 7795 unreviewed lines to anyone who does not diff against `2aafe5967`.

**Author follow-ups**
1. Answer the merge-order question from #1963: as a unit, or one at a time?
2. If one at a time, retarget each sub-PR's base to its predecessor — GitHub then renders the increment, enforces the order, and finding 2 disappears.

oksuzian (approved):
### PR Review Summary — Mu2e/Offline#1965

Reviewed at head `7d7d64fc1f0d532110e6a9ce4a02535a982952a3`. Correction to my review of the same head ([review 5278641525](https://github.com/Mu2e/Offline/pull/1965#pullrequestreview-5278641525)); no new code has been read.

**Decision**
- 🟢 approve

**Why this replaces the request for changes**

My earlier review asked for changes without asking for any change to this PR's code. Its finding 1 verified the increment over #1964 as clean, and its finding 2 was, in its own words, not a defect of this PR: the argument rebind in `generateSample` and the other inherited items come from #1963 and are fixed later in the same stack, at #1967 and #1968. That is a merge-order note, and I should not have graded it 🔴 [S0]. Nothing in this PR needs to change, so I am withdrawing the block.

**Merge order (note, not a finding)**
- This PR targets `main` and carries the whole stack up to this point. Merge the stack as one unit, or merge this PR only after #1967, so that the rebind never lands on `main` by itself. The per-item evidence stays in my earlier review.
- The substantive request-changes review on #1963, where those defects are introduced, still stands.

PR #1966, opened 2026-09-05
https://github.com/Mu2e/Offline/pull/1960 sub PR #5

YongyiBWu:
VDResampler: job configuration and fcl emission

PR #1967, opened 2026-09-05
https://github.com/Mu2e/Offline/pull/1960 sub PR #6

YongyiBWu:
VDResampler: generation and validation plots


YongyiBWu:
@FNALbuild run build test

oksuzian (approved):
### PR Review Summary — Mu2e/Offline#1967

Reviewed at head `53ffc662d458e26fac96191d2bdaf61324133111`. First pass.

**Decision**
- 🟢 approve

**Scope understood**
- Reviewed this PR's **own increment**, `06546d891...53ffc662d` — six files, +2707/-236: `VDResamplerValidationPlots.hh` (new, 1466 lines), `VDResamplerGenerateCommon.hh` (new, 289), the two generate modules rewritten (+336/-94 and +619/-137), two lines of `CMakeLists.txt`, and five deleted lines of `ScoreBasedDiffusionModel.hh`. GitHub displays this PR as 19 files and +12,033 because its base is `main` rather than #1966's branch; the earlier 13 files were reviewed on #1968 at the stack tip, and four of the six files here are byte-identical between this head and that one.
- **This is the PR that makes the stack safe.** It deletes the temporary conversion operator and replaces all six positional `generateSample` call sites. Findings 1 and 3 of the #1963 review are resolved *here*, not there — see the merge-order note below.

**Findings**

1. 🟡 [S2] The photon-line zoom energies are hardcoded in C++ and unsourced, while the line the training actually tags is a FHiCL parameter
   - Evidence: `VDResamplerValidationPlots.hh:600-602` fixes `const double kLines[] = {1809., 844., 347.};` with `kHalf = 15.` keV, commented only as "keV, in STM order of interest". On the training side the same physics lives in configuration: `VDResamplerTrainingPlan.fcl:407-408` sets `SBDMpeakTagCenters : [1.809]` and `SBDMpeakTagHalfWidths : [0.001]`.
   - Impact: the validation view cannot follow the configuration it is validating. The plan file explicitly contemplates working on other lines — `:363-364` notes "347keV really hard to sample here (~60/1e6) ... should consider a standalone 347 keV" — and an analyst who retrains on a different line still gets plots zoomed on this fixed set, needing a C++ edit and a rebuild to see their own line. Separately, the coding standard asks a physics constant to carry its origin; a reader cannot tell which transitions these three are, and the ±15 keV window has no stated basis.
   - Suggested fix: derive the zoom windows from the configured peak tags where they exist, or expose them as a FHiCL sequence with the same units as `SBDMpeakTagCenters`. Either way name the source of the three energies in a comment. The differing half-widths (±1 keV for tagging, ±15 keV for plotting) are defensible as different quantities — worth saying so explicitly rather than leaving the reader to infer it.

2. 🟡 [S2] 1755 lines of non-trivial implementation added as header bodies
   - Evidence: this increment adds `VDResamplerValidationPlots.hh` at 1466 lines and `VDResamplerGenerateCommon.hh` at 289, both implementation-only headers with no paired `.cc`. `computeDistributionMetrics` (`:131-198`), `finalize` (`:809-960`) and the spec builders are substantial functions, and the header is pulled into three translation units.
   - Impact: every consumer recompiles all of it, and a change to a metric forces a rebuild of all three modules. `STMMC/src/` already has a `mainlib` these could live in.
   - Suggested fix: move the function bodies to a `.cc` in `STMMC/src/`, leaving declarations and the small structs in the header. Line counts are measured at this head; the compile-time saving is an estimate. Non-gating, and not a reason to hold the PR.

3. ⚪ [S3] Commented-out alternative axis titles at `VDResamplerValidationPlots.hh:590` and `:607-608` — leftover variants of the `per 1ns` / `per 1keV` labels. The standard asks for these to be deleted rather than kept beside the live versions.

- 🟢 The temporary shim is removed exactly as planned and as the #1963 review asked. The diff deletes `operator const std::vector<double>&() const { return value; }` from `SBDMGeneratedSample`, whose own comment said it would be "removed by the PR that replaces the generation modules". That is this PR, and it does it.
- 🟢 All six positional call sites are gone. `VDResamplerGenerateFromModel_module.cc:263,274,284` and `VDResamplerGenerateMix_module.cc:376,387,397` previously read `generateSample({}, useHeun_, diffusionSteps_)` and assigned to a `std::vector<double>`; generation now goes through `VDResamplerGenerateCommon.hh:86,127,134,152,192`, each passing all six arguments explicitly and reading `zscore` or `value` by name. Counted at every head in the stack: six such calls at #1963, #1964, #1965 and #1966, zero from here on.
- 🟢 The distribution metrics are correctly implemented, which matters because these are the numbers a reader will judge the model by. Total variation is `0.5*sum|p-q|`; Jensen-Shannon is the base-2 divergence with `sqrt` applied, so it lands in [0,1]; KS is `sup|CDF_p - CDF_q|`; W1 integrates `|CDF_p - CDF_q|` against the bin width. Both histograms are scaled to unit area immediately before the call (`:866-867`) and were booked with `Sumw2` (`:705-706`), so `chi2` is computed on normalized contents with correctly propagated errors and is a shape statistic rather than a restatement of the sample-size difference. `chi2ndf()` guards `ndf > 0`.
- 🟢 The paired comparison cannot drift apart. `mother_` and `generated_` are booked from the same `DimSpec` (`:697-707`), so they share bin count and range by construction; `finalize` applies the same integer `Rebin` factor to both (`:848-849`), and the same `fx`/`fy` to the 2D pair (`:923-924`). The target bin count is driven by `min(nMother_, nGenerated_)` — the thinner side — which is the right choice and is documented as such.
- 🟢 Diagnostics behave. A low-statistics run warns rather than reporting confident metrics, an empty generated set returns early with an explanation, and the per-dimension table is accumulated into a single `LogInfo` instead of one message per row.
- 🟢 The new `Offline::GeneralUtilities` dependency is real (`ParameterSetFromFile.hh`, included by both modules) and both build systems carry it — `CMakeLists.txt` gains it here and `STMMC/src/SConscript` already lists `mu2e_GeneralUtilities`.

**Validation check**
- Build/tests run: none by me. CI is **green at this head** — FNALbuild reports "The build tests passed at 53ffc662d", prof build 4 min 23 s, and the individual contexts (`ceSimReco`, `ceDigi`, `ceMix`, `g4test_03MT`, `POT`, `transportOnly`, `cosmicSimReco`, `cosmicOffSpill`, `g4study`, `muDauSteps`, `ceSteps`) are all success.
- FNALbuild also reports `whitespace check` found errors, `clang-tidy` 12 errors / 111 warnings, and `TODO (1)`. The TODO is in a deliberately generated fcl string, not a leftover marker.
- Config contract check: not applicable, no fcl in this increment.
- Cross-repo consistency: pass. No companion needed.
- Not checked: whether the metric values on real output actually indicate a good model, the 2D correlation views' physics content, and the `DimSpec` axis ranges per particle species.

**Residual risk**
- **Merge order.** This PR is what clears the silent `generateSample` rebind, and the four heads below it (#1963, #1964, #1965, #1966) each still carry all six old call sites. Merging any of those without this one lands the rebind on `main`; merging this one or later does not. Either take the stack as a unit, or append the two new booleans to the end of `generateSample`'s parameter list so the lower PRs are independently safe — raised on #1963.
- The hardcoded line energies of finding 1 are the kind of number that gets copied. If a second home appears before it is made configurable, they will disagree eventually.

**Author follow-ups**
1. Make the photon-line zoom windows follow the configured peak tags, or expose them in FHiCL, and cite the origin of the three energies (finding 1).
2. Move the header function bodies into `STMMC/src/` (finding 2) — at your convenience, not a blocker.
3. Delete the commented-out axis-title variants (finding 3).
4. Confirm the merge order for this PR relative to #1963-#1966.

PR #1968, opened 2026-09-05
https://github.com/Mu2e/Offline/pull/1960 sub PR #7

YongyiBWu:
VDResampler: standalone plot regeneration for grid jobs


YongyiBWu:
@FNALbuild run build test

brownd1978 (approved):
It's impossible to review a submission of this size, either for content or code style. I suggest to allow it as new code, then review the correctness of the output separately before putting it into production (maybe Yongyi has already done that).

oksuzian (approved):
### PR Review Summary — Mu2e/Offline#1968

Reviewed at head `13f716223cda375eee83da88a862eed2b65e97f1`. First pass on this PR, and a re-review of the eight findings raised on #1963.

**Decision**
- 🟢 approve

**Scope understood**
- This PR's own increment is two files: `STMMC/src/VDResamplerRegeneratePlots_module.cc` and `STMMC/fcl/VDResamplerRegeneratePlots.fcl`. The other 19 files arrive from the stack below it. The sub-PRs are strictly nested by file set — #1963 (2 files) ⊂ #1964 (5) ⊂ #1965 (10) ⊂ #1966 (15) ⊂ #1967 (19) ⊂ #1968 (21) ⊂ #1960 (31), with nothing dropping out at any step — so I reviewed the union here once rather than the same code six times. That is also the practical answer to @brownd1978's point about size: the reviewable unit per PR is the increment, and the 21-file diff is an artifact of stacking.
- `MachineLearningTools/src/ScoreBasedDiffusionModel.cc` is *not* frozen across the stack. It has three distinct blobs: `fdef81ac5` at #1963/#1964, `531afd47f` at #1965-#1967, `996d154d4` here. So the #1963 findings had to be re-derived at this head rather than carried forward, and most of them are fixed here.
- Blast radius: no prodtools campaign JSON names `VDResamplerTrainingPlan.fcl`, `VDResamplerRegeneratePlots.fcl` or `VDResamplerReconfigure.fcl`. This is an analyst-run workflow producing user datasets, not a production campaign path, which is what caps the findings below at 🟡.

**Findings**

1. 🟡 [S2] The training FHiCL interface is spelled twice, once per trainer
   - Evidence: `VDResamplerTrain_module.cc` declares 95 keys in `struct Config` at `43-159` (117 lines, 30% of a 388-line file); `VDResamplerTrainFromRoot_module.cc` declares 95 keys at `40-157` (118 lines of 393). 93 of the 95 are identical. The only difference is the input source: `SimParticlemvTag`/`StepPointMCsTag` against `InputRootFile`/`TreeName`. `VDResamplerTrainCommon.hh` already owns the shared post-parse struct (`ModelBuildParams`, `:315`, commented "SBDM constructor arguments, filled from fhicl Config") but contains zero `Name("` declarations, so the shared home stops just short of the layer that actually duplicates.
   - Impact: adding or renaming a training parameter is a two-file edit with nothing to catch a miss, and the two trainers can drift into accepting different spellings of the same knob.
   - Suggested fix: move the shared 93 into a `TrainConfig` struct in `VDResamplerTrainCommon.hh` and have each module hold it as a `fhicl::Table<TrainConfig>` alongside its own two source keys. (Line counts are measured at this head; the saving is an estimate.)

2. 🟡 [S2] V1 generation conditions stage 2 on stage 1's *data* z-scores, where V2 beside it normalizes explicitly
   - Evidence: `VDResamplerGenerateCommon.hh:133` builds the stage-2 condition as `{s1.zscore[0], s1.zscore[1], s1.zscore[2]}` — stage 1's own data-block z-scores. The V2 path at `:187` instead calls `stage2Model.normalizeCondition(0, ptot_t)`, which applies stage 2's stored *condition* normalization. Training fills `s2.cond = {t_trans, x_trans, y_trans}` un-z-scored (`VDResamplerTrainCommon.hh:1196`) and lets `normalizeData` z-score the condition block with stage 2's own statistics (`:2256`).
   - Impact: the V1 path is correct only because the two statistics coincide — both training sets are filled from the same events in the same loop (`:1183`, `:1196`) and truncated to the same `trainingSize` (`:2220`, `:2241`), so stage 1's data mean/stdev for (t,x,y) equal stage 2's condition mean/stdev for (t,x,y). Nothing asserts or documents that, and it is the sort of invariant an independent change to either stage's sample selection would break silently, leaving stage 2 conditioned on mis-scaled inputs.
   - Suggested fix: route the V1 condition through `stage2Model.normalizeCondition` as V2 does, so the coincidence stops being load-bearing; or state the invariant where the two training sets are built.

3. ⚪ [S3] A comment asserts an output the inversions cannot produce
   - `VDResamplerRegeneratePlots_module.cc:453-456` says "A generated sample CAN come out with pz <= 0 (nothing bounds the model's output)", and `:457` drops such samples and counts them on the grounds that "a large fraction would itself be a finding about the model". Both inversions make `pz` positive by construction: `invertGeneratedSampleV1` sets `pz = p0 * std::exp(pzTrans)` (`VDResamplerTransforms.hh:733`) and `invertMomentumV2FromPtot` sets `pz = pTot / std::sqrt(1.0 + ur*ur + uphi*uphi)` with its own comment ">0; denom >= 1, no guard needed" (`:674`). The branch is therefore reachable only through `exp` underflow, needing `pzTrans` below roughly -745, and `skippedNonPositivePz` cannot report the condition the comment describes. Either drop the claim and keep the guard as pure defence, or say it guards underflow.

4. ⚪ [S3] Two small consistency items, grouped
   - `VDResamplerConfigure_module.cc:211` and `:218` carry commented-out `mf::LogWarning` calls; the coding standard asks for these to be deleted rather than kept for reference.
   - Seven of the eight new headers open with `#pragma once`, while `VDResamplerConfigureCommon.hh` uses a conventional `STMMC_VDResamplerConfigureCommon_hh` guard. GitHub code search reports `#pragma once` in 2 `.hh` files across Offline against 1060 using `#ifndef`, so the odd one out here is the majority of this PR. Worth settling one way inside the PR.

**Carried forward from the #1963 review, still open**

5. 🟡 [S2] The binary checkpoint format still ships readers for versions that never existed in the repository. `saveModel` writes version 9 (`ScoreBasedDiffusionModel.cc:1821`) and the loader still carries 15 `version >=`/`version <=` branches. `grep 'SBDM"'` on the base commit `f6e43507` returns no hits, so no in-tree code has ever written a pre-9 file and every one of those branches is compatibility with out-of-tree checkpoints. Non-gating, and unchanged from the #1963 review.

6. 🟡 [S2] The 30-parameter positional constructor remains, spelled out positionally at two call sites inside `loadModel`. Since the callers were rewritten anyway in this stack, a `struct Config` passed by `const&` would have been cheap here. Non-gating.

**Fixed since the #1963 review, verified at this head**

- 🟢 [was S0] `generateSample`'s two middle-inserted booleans no longer silently rebind anything. All five call sites pass the full six arguments explicitly (`VDResamplerGenerateCommon.hh:86,127,134,152,192`), and the old positional `(condition, useHeun_, diffusionSteps_)` calls are gone.
- 🟢 [was S1] `saveModel`'s format no longer contradicts the configuration. The FHiCL comments now read ".dat" for outputs and "(.dat, or legacy .bin/.csv)" for checkpoint inputs (`VDResamplerTrain_module.cc:47-52`).
- 🟢 [was S1] The `SBDMGeneratedSample` conversion operator is gone; it is a plain struct of `zscore` and `value` (`ScoreBasedDiffusionModel.hh:33-36`), so the dangling-reference spelling is impossible. Callers read the member they mean, and the V1 stage-2 condition correctly takes `zscore` while the physical outputs take `value`.
- 🟢 [was S1] LOGSIG now validates for every prediction target, not just V: `logSigMin > 0`, `logSigMax > 0`, `logSigMax > logSigMin` and `logSigMax <= 1` all throw (`ScoreBasedDiffusionModel.cc:328-341`).
- 🟢 [was S2] The CSV loader's unconditional "Normalization parameter size mismatch" throw is gone, so legacy CSVs load.
- 🟢 [was S2] The gradient-clipping counters are reset per epoch (`:1276-1278`).
- 🟢 [was S2] `firstLayerBlockMagnitudes` now reports both quantities and labels them: `gradL2 = ||sum_k g_k||/n` beside `gradRmsL2 = sqrt(mean_k ||g_k||^2)` (`:1755-1756`), so a feature whose per-sample gradients cancel no longer reads as dead.
- 🟢 [was S2] The per-dimension gradient weight is now gated on the controller being enabled (`:1414`), and `trainingSampleSize_` is assigned rather than serialized as a permanent zero (`:1107`).
- 🟢 The stale "Only ODE solver, no noise added" comment above the Heun block now reads "for both the ODE and SDE".

- 🟢 Transform pairs are mutually inverse, checked algebraically rather than by inspection. All three radial maps invert exactly — `atanh(rho)`/`tanh`, `atanh(sqrt(rho))`/`tanh^2`, `atanh(rho^2)`/`sqrt(tanh)`; the local-polar rotation inverts exactly by substitution, since `(rx,ry)` and `(-ry,rx)` are orthonormal and `rx^2+ry^2 = 1`; `forwardTimeAsinh`/`invertTimeAsinh` compose to the identity through `sinh(asinh(u)) = u`. This matters because the header comment in `VDResamplerRegeneratePlots_module.cc` rests the whole replay argument on that pair being analytic and mutually inverse, and it is.
- 🟢 The `pz > 0` precondition the slope division needs is enforced at every input feed, not just some: `VDResamplerTrain_module.cc:335`, `VDResamplerTrainFromRoot_module.cc:347`, `VDResamplerConfigure_module.cc:216`, and `VDResamplerRegeneratePlots_module.cc:457`. `pz` is also floored before the extrapolation's divide in both forward transforms, with the fallback counted and reported once in `endJob` rather than per hit.
- 🟢 No dead FHiCL keys. Across the seven modules, 246 declared keys (8, 17, 15, 4, 12, 95, 95) all reach an accessor. All seven use the validated `Table<Config>` pattern.

**Validation check**
- Build/tests run: none by me. CI is **not green at this head** — `mu2e/buildtest` is pending at `13f716223`, and FNALbuild's last passing table is for `e31c742c`, a different sha. I did not trigger a run because one is already queued at the head. The approval below is contingent on it.
- FNALbuild's table at the last run reports `whitespace check` found errors, `clang-tidy` 14 errors / 113 warnings, and `TODO (1)`. The TODO is `VDResamplerConfigureCommon.hh:592`, inside a generated fcl string that deliberately emits `@nil # TODO set the training ROOT file`, so it is content rather than a leftover marker. `trigger` returns code 1 here as it does on the sibling PRs, so it does not look specific to this diff.
- Config contract check: pass for the three new fcl files against their modules' validated tables.
- Cross-repo consistency: no companion needed. The STMMC callers that #1963 would have broken are updated within this stack.
- Not checked: whether the trained models actually reproduce the STM momentum spectrum, the numerical soundness of the curriculum and peak-importance-sampling schedule, and the contents of the 14 findings from the 2026-09-19 STMMC sweep that fall outside this diff.

**Residual risk**
- Merge order still matters, and it is the one thing I would not want lost. Findings 1-4 of the #1963 review are fixed at *this* head, not at #1963's own head `a32e69eb`. If #1963 merges by itself, `main` gets the `generateSample` signature change with the old STMMC call sites, which is the silent rebind that review opened with. #1963's outstanding CHANGES_REQUESTED should be resolved by advancing it to this code or by closing it in favour of the later PR, not by merging it as it stands.
- The `.dat` checkpoints written before the variance-preserving change are not distinguishable from later ones by anything in the file, and the forward process changed. That is a property of the format, not of this PR.

**Author follow-ups**
1. Fold the 93 shared training keys into one `fhicl::Table<TrainConfig>` in `VDResamplerTrainCommon.hh` (finding 1).
2. Route the V1 stage-2 condition through `normalizeCondition`, as V2 already does, or document the equal-statistics invariant where the two training sets are built (finding 2).
3. Correct or reframe the `pz <= 0` comment in `VDResamplerRegeneratePlots_module.cc` (finding 3).
4. Delete the two commented-out `mf::LogWarning` calls, and pick one include-guard style across the new headers (finding 4).
5. Confirm the merge order for #1963 relative to this stack (residual risk).

oksuzian:
Not a new review — **the 🟢 approval from 2026-09-22 stands**, and nothing below changes it.

Context: a static sweep of `STMMC` at main `665dc45db` (2026-09-19) produced findings that were triaged as "in flight in #1960" purely on a file-list overlap. That triage explicitly did **not** read this PR, so "in flight" meant "the file is being rewritten", not "fixed". I have now checked each of those findings against this PR's head `13f716223`. Three of them are fixed here, which is worth recording; one survives.

### Fixed by this PR — verified at `13f716223`

- **`ModelFileDir` was mandatory and no fcl supplied it** (was S0). `ModelFileDir` now defaults to `""`, the new `ModelFileList` catalog atom is declared at `VDResamplerGenerateMix_module.cc:134-141`, and `:386-395` throws when neither is set and warns when both are. The construction failure is gone and the `ModelFileList` key that Production `prolog.fcl` already carried now exists in the `Config`.
- **Generated model paths omitted `dataSourceTag`, so a per-category rerun silently overwrote the previous model** (was S1). Paths now come from `VDResampler::modelFileName(stage, versionTag, virtualDetectorID, sourceTag, pdg, runNumber)` — `VDResamplerNameHelper.hh`, called at `VDResamplerGenerateMix_module.cc:684,747,758` and `VDResamplerConfigureCommon.hh:527-532`. Two categories sharing a pdgID no longer collide.
- **`trainingPlanFile` was set at three Production fcl sites and declared by no `Config`** (was S2). It is now declared and read at `VDResamplerConfigure_module.cc:81`, `VDResamplerGenerateMix_module.cc:164`, `VDResamplerRegeneratePlots_module.cc:167` and `VDResamplerReconfigure_module.cc:63`, and `STMMC/fcl/VDResamplerTrainingPlan.fcl` is added by this PR. The comment at `VDResamplerGenerateMix_module.cc:143-146` — `VirtualDetectorID`, `VDz0` and `VDr` are no longer module parameters because a local override that disagreed with training would silently rescale every generated sample — is exactly the right call, and it closes the hole rather than papering over it.

### One finding survives — 🟡 [S2], and it does not gate anything

**The `"E"` branch holds kinetic energy, sitting next to `px`/`py`/`pz`.**

- Evidence at `13f716223`: `VDResamplerGenerateMix_module.cc:566` books `outTree_->Branch("E", &E_gen_, "E/D")` immediately after `px`, `py` and `pz` in the tree created at `:555`; `:873-874` fill it as `const double eTotal = std::sqrt(momParticle.mag2() + mass_gen_*mass_gen_); E_gen_ = eTotal - mass_gen_;` — while the `GenParticle` four-momentum at `:875` correctly uses `eTotal`. Same pattern at `VDResamplerGenerateFromModel_module.cc:498,560-561`.
- Impact: a tree carrying `px, py, pz, E` reads as a four-momentum, so anyone comparing `E` against the training input's total-energy spectrum is low by the rest mass — 0.511 MeV for e-, 105.7 MeV for mu-. The branch is written in the deployed configuration: `doROOTDump` and `doValidationPlots` default `false` in the `Config` now (`:198-201`), but Production main `JobConfig/pileup/STM/prolog.fcl:54-55` sets both `true` on this `module_type`. Diagnostic output only, though — the art output keeps only the `GenParticles` products, so no dataset carries it.
- Suggested fix: fill `eTotal`, or rename the branch `KE`. Either is a one-line change and neither needs to happen in this PR.

### Coverage of this check

I re-derived each finding at this PR's head by reading the lines quoted above. I did **not** re-run the sweep, and this check covered only the nine findings the sweep had triaged against #1960 — it is not a second review of the PR. Four other sweep findings are in files this PR does not touch (`DetectorEfficiency`, `ConcatenateDigitizedWaveforms`, `PHTree`, `HPGeTree`); those are main-branch issues and belong in their own issue, not here.

oksuzian:
@FNALbuild run build test

PR #1973, opened 2026-09-08
Added measured LRU values for calorimeter crystals

giro94:


bechenard:
did you check that the code is actually calculating the LRU as you measured it? Just checking...

simonagiov (approved):

giro94:
> did you check that the code is actually calculating the LRU as you measured it? Just checking...

@bechenard Simona calculated these values with the formula written in the code

bechenard (approved):
Looks good

PR #1976, opened 2026-09-10
P107 for root 6.38 (update2)

rlcee:
Also updates python to 3.12.13, fhicl-cpp to 4.19.02, canvas-root-io custom build, and other minor updates


oksuzian (approved):
### PR Review Summary — Mu2e/Offline#1976

Reviewed at head `315557955c56d260445567927a5615b016e74ac7`. First pass.

**Decision**
- 🟢 approve

**Scope understood**
- Bumps the build environment from p106 to p107 (`.muse`) for ROOT 6.38, plus python 3.12.13, fhicl-cpp 4.19.02, and a custom canvas-root-io build (per PR body).
- Adapts `CosmicReco/src/MinuitDriftFitter.cc` to a Minuit2 API change in ROOT 6.38: `MnMigrad` no longer forwards `SetLimits`/`Fix`/`SetPrecision`/`Release` directly, so callers now go through `MnUserParameters` (before constructing the fitter) or `MnMigrad::State()` (after).
- One whitespace-only fix (tab → spaces) in `DAQ/src/EventHeaderFromCFOFragment_module.cc`, already covered by the whitespace check.

**Findings**

None. This is a mechanical, CI-validated envset migration.

**Verification notes**

- `MnMigrad`/`MnApplication::State()` returns a live reference to the internally-held `MnUserParameterState` (`fState`), not a copy, so `migrad.State().Fix(4)` / `.Release(4)` / `.SetPrecision(...)` mutate the same state object the minimizer will use — behavior-preserving relative to the old direct `migrad.Fix(4)` etc. Confirmed against the ROOT master/6.38 header source.
- The other two call sites move `SetLimits`/`Fix` from post-construction (`migrad.SetLimits(...)`) to pre-construction (`params.SetLimits(...)` before `MnMigrad migrad(fit, params, ...)`). `MnUserParameters` has its own native `SetLimits`/`Fix` overloads (confirmed against the class reference), and since `MnMigrad` copies `params` by value at construction with nothing in between that depends on the limits, the two orderings are equivalent.
- Build (prof) passed at this head, and no lingering `migrad.SetLimits/Fix/SetPrecision/Release` call sites broke the build elsewhere in the tree — that's the strongest signal no other spot needed the same fix.
- `cosmicSimReco` and `cosmicOffSpill` (the CI targets that exercise `MinuitDriftFitter`) both passed cleanly.

**Validation check**
- Build/tests run: CI — build (prof), ceSimReco, g4test_03MT, transportOnly, POT, g4study, cosmicSimReco, cosmicOffSpill, ceSteps, ceDigi, muDauSteps, ceMix, rootOverlaps, g4surfaceCheck, check_cmake, whitespace check — all green at this head.
- Config contract check: n/a (no FHiCL schema touched).
- Cross-repo consistency: n/a for this diff.

**Residual risk**
- `trigger` CI target returned code 1 ("Job completed with warning") rather than passing cleanly. The log shows only sub-1-sigma drifts in several trigger-path pass counts (delta/sqrt(reference) ≤ 1.03) and a timing-scale note — consistent with small numerical drift from the Minuit2/ROOT 6.38 upgrade rather than a real behavior change, but the reference counts for this test likely need a follow-up bump once p107 is adopted. Not a blocker for this PR.
- clang-tidy reports 1 error, but it's in an external header (`cetlib/LibraryManager.h`, a `constexpr` return-type diagnostic) unrelated to this diff — an environment/toolchain interaction from the p107 bump, not something introduced by the changed files. The magic-number warnings on `MinuitDriftFitter.cc` are on literals that already existed pre-PR (only relocated), not new.

**Author follow-ups**
- Confirm whether the `trigger` test's reference counts need updating in a follow-up now that p107 is the default envset, and whether the `cetlib` constexpr clang-tidy error is tracked separately.

PR #1988, opened 2026-09-15
Add multi-disk illumination support to CaloCalibGun and fixed piping bug

HJafree:
Changed the disk numbers for vectors, to allow both disks to be illuminated together, and fixed bug in the piping generation


oksuzian (changes requested):
### PR Review Summary — Mu2e/Offline#1988

Reviewed at head `8432b6b078ee25188973ab81cdc97cf4ac8b0162`. First pass.

**Decision**
- 🔴 request changes

Illuminating several disks in one event is a reasonable addition, and the per-photon code is untouched. Two parts of the interface change need fixing before merge, though. Removing `nDisk` breaks the calibration group's own simulation configs. The new default silently changes what every existing config generates. Also, the piping bug named in the title is not fixed in this commit, and there is a real one in the file.

**Scope understood**
- `CaloCalibGun` replaces `nDisk` (an int, default 1) with `disks` (a sequence, default `[0,1]`), and loops over those disks with one pipe centre per disk.
- The PR also adds a `multiphotons` flag, which the description does not mention. When it is set, each disk draws a Poisson number of photons (mean 2.2467) instead of exactly one.
- According to the title and description, it also "fixed bug in the piping generation".

**Findings**

1. 🟠 [S1] **Removing `nDisk` breaks the CaloCalibration simulation configs, and there is no companion PR.**
   - Evidence: `Mu2e/CaloCalibration` `SourceCalib/simulation/RunCaloCalibGun-sim.fcl:31` and `RunCaloCalibGun-simdig.fcl:35` both set `nDisk : 0 //Change me to run over other disk`. `CaloCalibGun` uses validated FHiCL (`art::EDProducer::Table<Config>`), and `nDisk` is no longer in `Config`, so art rejects both files at configuration time as an unsupported parameter. The author has no open CaloCalibration PR (`gh search prs --owner Mu2e --author HJafree --state open` returns only this one).
   - Impact: the source-calibration simulation stops running as soon as a musing picks up this Offline.
   - Suggested fix: open a CaloCalibration PR that changes both files to `disks : [0]`, and note the merge order here. The failure is at least loud, which is better than a silent one.

2. 🟠 [S1] **The `[0,1]` default silently changes the output of every config that does not set `disks`.**
   - Evidence: none of the in-tree users set a disk. That covers `EventGenerator/fcl/prolog.fcl:137`, `CaloMC/test/RunCaloCalibGun.fcl:28`, and the `CaloCalibGun` tables in `Mu2e/pbar2m` and `Mu2e/pipenu` `fcl/prolog.fcl`. All of them used to get one photon per event on disk 1 (the old C++ default, `nDisk = 1`). After this PR they get two photons per event, one on each disk, with no warning and no change to their fcl.
   - Impact: anyone reproducing an earlier `CaloCalibGun` sample gets twice the photons, spread over both disks, and nothing in the job tells them.
   - Suggested fix: give `disks` no C++ default, because it determines the physics output, and set it explicitly in `EventGenerator/fcl/prolog.fcl`. While there, validate it in the constructor with a `cet::exception`. At the moment an empty list silently produces events with no photons, a repeated disk doubles that disk, and an out-of-range index only fails at `disks_.at()` in `beginRun`, with a bare `std::out_of_range` message.

3. 🟠 [S1] **The piping-generation fix is not in the diff, and the file still contains a pipe-position bug.**
   - Evidence: `git diff -w` of the event loop against the base changes only the loop structure and `pos += _zPipeCenters[d]`. Every per-photon pipe formula is identical to `main`. Meanwhile `beginRun` still reads `xsmall = _cal->G4Info().get<double>("radSmTor");` (line 157), which sets `xsmall` to 41 mm. The geometry defines `calorimeter.xsmall = 71.0` separately (`Mu2eG4/geom/calorimeter_CsI_v2.txt:52`), `DiskCalorimeterMaker.cc:123` loads it into `G4Info`, and `Mu2eG4/src/constructDiskCalorimeter.cc:315,377` builds the small torus at `xsmall + xdistance * ipipe` using 71 mm.
   - Impact: photons generated on the small torus and the straight section sit about 30 mm in |x| away from the pipes Geant4 actually builds. This is older than this PR and has been there since at least the `caloInfo()` era, but it matches the kind of bug the title describes.
   - Suggested fix: `xsmall = _cal->G4Info().get<double>("xsmall");`. If you had a different bug in mind, please say which one, since this commit does not change any pipe formula.

4. 🟡 [S2] **The Poisson mean is a hard-coded literal, and the comments disagree with it.**
   - Evidence: the constructor passes `2.246666666642` (line 131, which clang-tidy also flags as a magic number), with a comment deriving it as 33.33 photons/crystal/s × 674 crystals × 100 µs. The loop comments say "Poisson(mean=4)" and "mean at 4 per disk" twice. The 100 µs gate is also the `tmax - tmin` window that the CaloCalibration configs use (`tmax : 100000.`). So the mean depends on the configured time window, but the code fixes it for one particular window.
   - Suggested fix: make the source rate a fcl parameter, with its origin cited. Compute the mean as rate × crystals-per-disk × (`tmax - tmin`), taking the crystal count from the geometry, and correct the comments.

5. ⚪ [S3] **Change-history comments.** Please drop the `// NEW:` markers, the "Multi-disk edits author" line and the `} // NEW: end loop over disks` comment. The coding standard leaves change history to git. Also indent the body of the new disk loop, so that the two nested loops can be read. Renaming the member `_multi` to match its fcl name, `multiphotons`, would help too.

- 🟢 Old samples stay reproducible with `disks : [1]` and `multiphotons : false`. The per-photon random draws are unchanged in order and count (`git diff -w` above), and `_randPoisson_dist` is fired only when `_multi` is true.
- 🟢 The pipe centre per disk uses the same expression as before, applied to each entry of `_disks`, and the index `d` addresses `_zPipeCenters`, which `beginRun` fills in the same order.

**Validation check**
- Build/tests run: `mu2e/buildtest` passed at this head (FNALbuild, 2026-09-15T21:07Z), with whitespace clean and `trigger` ✅. The one clang-tidy error is in art's `LibraryManager.h`, not in this file. None of the buildtest jobs runs `CaloCalibGun`. I did not build or run the module myself.
- Config contract check: fail. See findings 1 and 2.
- Cross-repo consistency: needs a follow-up in `Mu2e/CaloCalibration` (finding 1). `pbar2m` and `pipenu` still parse, but their output changes (finding 2). Production has no `CaloCalibGun` user.

**Residual risk**
- Not checked: the 33.33 photons/crystal/s source rate, or any physics validation of simultaneous two-disk illumination. The description gives no validation evidence.

**Author follow-ups**
1. Open the CaloCalibration companion PR (`nDisk : 0` → `disks : [0]`) and link it here.
2. Remove the C++ default for `disks`, set it in `EventGenerator/fcl/prolog.fcl`, and validate the list in the constructor.
3. Say which piping bug the PR fixes. If it is `xsmall`, the one-line change above is still needed.
4. Move the Poisson rate to fcl with a cited source, derive the mean from the time window, and correct the comments.
5. Add to the description what `multiphotons` does, and what you ran to validate the two-disk output.

HJafree:
Thanks for the detailed review — here's what's been addressed.

The nDisk removal breaking CaloCalibration is fixed with a companion PR:Mu2e/CaloCalibration#49. Both RunCaloCalibGun-sim.fcl and RunCaloCalibGun-simdig.fcl now use disks : [0], multiphotons : false, which reproduces the exact old nDisk : 0 behavior — one disk, one photon per event, no physics change. This PR and the companion PR need to merge together, or the companion one first, since merging this alone would break those configs.

The silent default change is fixed too. disks no longer has a C++ default, so any config that leaves it unset now fails loudly at configuration time instead of quietly changing output. I added an explicit disks : [1] (matching the old nDisk : 1 default) to the two in-tree configs that were relying on the implicit default: EventGenerator/fcl/prolog.fcl and CaloMC/test/RunCaloCalibGun.fcl. I also added validation in the constructor and beginRun: an empty disks list, a duplicate disk index, or an out-of-range index all now throw a cet::exception instead of failing silently or with a bare std::out_of_range.

The piping bug you found is the one I meant, and it's fixed: xsmall now reads the "xsmall" key instead of "radSmTor".

The Poisson mean is no longer a hard-coded literal. There's a new fcl parameter, sourceRate (photons/crystal/sec, default 33, cited to docdb:54585), and the mean is now computed per disk in beginRun as sourceRate times crystal count times gate width — the crystal count comes from the geometry (_cal->disk(disk).nCrystals()) instead of the hard-coded 674, and the gate width is the actual configured tmax - tmin instead of a fixed window. I also fixed the comments that disagreed with the old value.

On multiphotons, which I should have described originally: when it's set to true, each disk in the disks list independently draws a Poisson-distributed photon count (using the mean above) instead of generating exactly one photon. It defaults to false, which keeps the original fixed-one-photon behavior for anyone who doesn't set it.

I also cleaned up the change-history comments and indentation you flagged, and renamed the internal _multi variable to _multiphotons to match the fcl parameter name — the fcl key itself didn't change, so no existing config is affected by that rename.

oksuzian (changes requested):
### PR Review Summary — Mu2e/Offline#1988

Reviewed at head `e55a705331d00f4c2df4866817e4da32c9804557`. Re-review of `8432b6b078ee25188973ab81cdc97cf4ac8b0162`.

**Decision**
- 🔴 request changes

Thank you for the thorough response — all four substantive findings from the first pass are genuinely fixed, and I verified each one rather than taking the commit message for it. The `xsmall` fix is correct, the companion CaloCalibration PR is merged, `disks` is now a required parameter with real validation, and the Poisson mean is derived from the geometry and the configured gate. One new problem came in with the fix commit: the member initializer list no longer matches the declaration order, and Mu2e builds with `-Wall -Werror`, so this will not compile.

**Scope understood**
- `nDisk` (int) becomes `disks` (a required `fhicl::Sequence<int>`), and `produce` loops over the requested disks, using one pipe centre per disk.
- New `multiphotons` switch: when true, each disk draws a Poisson number of photons with a mean computed in `beginRun` from the new `sourceRate` parameter, the disk's crystal count and the `tmax - tmin` gate.
- `beginRun` now reads the `"xsmall"` geometry key instead of `"radSmTor"`.

**Findings**

1. 🔴 [S0] **The constructor initializes members out of declaration order, which is an error under `-Werror`.**
   - Evidence: members are declared `_randFlat` (line 108), `_randomUnitSphere` (110), `_randPoisson_dist` (111), but initialized `_randFlat`, `_randPoisson_dist`, `_randomUnitSphere` (lines 133-135). Both Mu2e builds turn this into an error: the Muse/scons build passes `-Wall -Werror` (`muse-4.17.0/python/sconstruct_helper.py:205-212`) and the CMake build sets `WERROR` plus an explicit `-Wreorder` (`Offline/CMakeLists.txt:12-33`). I confirmed with a minimal reproducer under `g++ -std=c++20 -Wall -Werror` that the diagnostic is `error: ... will be initialized after [-Werror=reorder]`. This is new since `8432b6b`, where the three were in matching order and the build was green.
   - Impact: the PR does not build.
   - Suggested fix: move `, _randPoisson_dist{_engine}` after `, _randomUnitSphere{...}` in the initializer list, or move the declaration above `_randomUnitSphere`.

2. 🟡 [S2] **`sourceRate` is a physics-affecting parameter with a C++ default.**
   - Evidence: `fhicl::Atom<double> sourceRate{Name("sourceRate"), Comment("..."), 33.}` (line 67), and `multiphotons` defaults to `false` in code (line 68). Neither appears in `EventGenerator/fcl/prolog.fcl`. Long-standing Mu2e practice — this is convention, not the wiki standard — is that only debug knobs carry code defaults, which is exactly the reasoning behind dropping the `disks` default in this commit. Note also that `33` is not the `33.33` the replaced literal was derived from; a 1% difference matters only for new `multiphotons` samples, but the cited value should be the one from docdb:54585.
   - Suggested fix: drop the default on `sourceRate` and set `sourceRate` and `multiphotons` explicitly in the prolog table next to `disks`.

3. 🟡 [S2] **Two satellite repos keep their own `CaloCalibGun` prolog copies that now fail validation.**
   - Evidence: `Mu2e/pbar2m` `fcl/prolog.fcl:256-264` and `Mu2e/pipenu` `fcl/prolog.fcl:311-319` are standalone copies of the Offline table with no disk parameter at all. Now that `disks` is required, any job using those tables is rejected at configuration time. I could not confirm that either repo has a job that actually references its table, so this may be dormant — but it is the same cross-repo shape as the CaloCalibration one, and a one-line `disks : [1]` in each keeps them working.

4. ⚪ [S3] **Leftovers from the previous round.** The body of the new disk loop is still unindented (lines 191-193 sit at the same level as the `for` that opens them), so the two nested loops do not read as nested; the inner loop is fine. `_randPoisson_dist` also mixes `bouncingCaps` with an embedded underscore — `_randPoisson` matches the recommendation and the other members here.

5. ⚪ [S3] **Pre-existing, and only because the line moved in this diff:** `beginRun` reads `_nPipes` from the geometry (line 168) and nothing ever uses it, while the pipe index is drawn with a literal `5` (line 208). If you want it, `int(_nPipes*_randFlat.fire())` ties the draw to the geometry; otherwise the member is dead. Not something I would hold the PR for.

**Prior findings**

- 🟢 [was S1] Removing `nDisk` breaks the CaloCalibration configs — FIXED. `Mu2e/CaloCalibration#49` is merged and changes both `RunCaloCalibGun-sim.fcl` and `RunCaloCalibGun-simdig.fcl` to `disks : [0]` with `multiphotons : false`. Because the companion merged first, those two configs are broken on CaloCalibration's `main` until this PR lands, so this is now worth merging promptly rather than a merge-order risk.
- 🟢 [was S1] Silent default change — FIXED. `disks` has no C++ default (line 66), so an unset config fails validation instead of changing output; `EventGenerator/fcl/prolog.fcl` and `CaloMC/test/RunCaloCalibGun.fcl` both set `disks : [1]`, matching the old `nDisk = 1`. The constructor rejects an empty or duplicated list and `beginRun` rejects an out-of-range index, each with a `cet::exception`. Caveat in finding 3 above.
- 🟢 [was S1] The `xsmall` bug — FIXED and verified. `beginRun` now reads `_cal->G4Info().get<double>("xsmall")` (line 176). That is the same key `GeometryService/src/DiskCalorimeterMaker.cc:123` fills from `calorimeter.xsmall` and the same one `Mu2eG4/src/constructDiskCalorimeter.cc:315` uses to place the pipes, so generation and construction now agree. `radSmTor` is still read separately and still used as the small-torus radius, which is correct.
- 🟢 [was S2] Hard-coded Poisson mean — FIXED. `_photonMean` is filled per disk in `beginRun` as `sourceRate * disk(d).nCrystals() * (tmax - tmin) * 1e-9` (line 166), with the units spelled out in the comment, and the stale "mean=4" comments are gone.
- 🟡 [was S3] Change-history comments — PARTIAL. The `// NEW:` markers and the author line are gone and `_multi` is renamed; indentation and the new member name remain (finding 4).

**Validation check**
- Build/tests run: none at this head. `mu2e/buildtest` has been `pending` on `e55a705` since 2026-09-16T18:54Z with no FNALbuild result comment, so I did not re-trigger. Given finding 1 I expect it to fail to compile; the reproducer I ran is a cross-check of the flags, not of this file. I did not build Offline myself.
- Config contract check: pass in-tree. `disks` is required and every in-tree table sets it; `fhicl::Sequence<int>` accepts the `[1]` and `[0]` spellings used.
- Cross-repo consistency: CaloCalibration is done (merged). `pbar2m` and `pipenu` need the one-line fix in finding 3.

- 🟢 Old single-disk samples stay reproducible with `disks : [1]`, `multiphotons : false`. A whitespace-insensitive diff of `produce` against `main` shows only the loop scaffolding, `pos += _zPipeCenters[d]` replacing `pos += _zPipeCenter`, and the removal of a commented-out `xmanifold` line — every per-photon `_randFlat` draw is unchanged in order and count, and `_randPoisson_dist.fire` is reached only when `multiphotons` is true.
- 🟢 `CLHEP::RandPoissonQ::fire(double mean)` exists (CLHEP 2.4.7.1, `RandPoissonQ.h:92`), so passing a per-disk mean to an instance built with the default mean is valid. Both means in play here (about 2.2 for the 100 µs calibration gate, about 0.03 for the 1194 ns prolog gate) are far below the mean where the class falls back to the Gaussian approximation.

**Residual risk**
- The 33 photons/crystal/s rate and any physics validation of simultaneous two-disk illumination are unchecked — the description still gives no validation evidence for the multi-disk output.
- With the prolog's 500-1694 ns gate, `multiphotons : true` would give a mean of about 0.03 photons per disk, so most events would be empty. That is what the numbers say rather than a defect, but it is worth confirming the intended gate for multi-photon running.

**Author follow-ups**
1. Fix the initializer order so the module compiles, and let buildtest report on the new head.
2. Drop the `sourceRate` code default and set `sourceRate`/`multiphotons` in the prolog table; use the exact rate from docdb:54585.
3. Add `disks : [1]` to the `CaloCalibGun` tables in `Mu2e/pbar2m` and `Mu2e/pipenu`, or confirm they are unused.
4. Indent the disk-loop body, and consider `_randPoisson` for the new member.
5. Add what you ran to validate two-disk output to the description.

PR #1989, opened 2026-09-19
Energy deposition module for optimization studies

michaelmackenzie:
This is the module used in the "fast sim" optimization workflow for Run 1 studies. It evaluates average energy depositions and energy losses that are used as inputs to sensitivity expectation and relative detector doses evaluations. The intent is to move this into Offline so standard SimJob musings can be easily used in optimization workflows.


michaelmackenzie:
@FNALbuild run build test

oksuzian (commented):
### PR Review Summary — Mu2e/Offline#1989

Reviewed at head `7509119cad5b52f35c195fa78a4fc75b75fc2124`. First pass.

**Decision**
- 🟡 comment only

The three 🟠 [S1] items are all quantity or normalization questions inside the module's
headline outputs. None of them breaks a build or changes anything that runs today, so they
do not block the merge, but they are worth settling before the numbers are used as
optimization inputs.

**Scope understood**
- Adds `Analyses/src/EdepAna_module.cc`, an `art::EDAnalyzer` that sums calorimeter and
  tracker energy depositions, tracks per-primary energy loss to the tracker front face,
  writes a flat TTree plus four selection-binned histogram sets, and fits a Landau to the
  energy-loss distributions in `endJob`.
- Adds the `cet_build_plugin(EdepAna ...)` entry and a driver `Analyses/test/edep.fcl`.
- Single topic. Nothing in `Production` or `mu2e-trig-config` is affected, and no campaign
  configuration references the module — it is user-run code today.

**Findings**

1. 🟠 [S1] The tracker-front "energy" is a momentum magnitude, differenced against a total energy
   - Evidence: `h_trk_front_energy_diff_` is filled with
     `front_trk_sp->momentum().mag() - primsim->startMomentum().e()`, and
     `h_trk_front_energy_edep_diff_` with `edep - front_trk_sp->momentum().mag()`.
     `StepPointMC::momentum()` returns a `CLHEP::Hep3Vector`, so `.mag()` is |p|, while
     `SimParticle::startMomentum()` returns a `CLHEP::HepLorentzVector` whose `.e()` includes
     the rest mass. The tree branch `primary_trk_front_energy` and both histogram titles call
     the |p| value an energy.
   - Impact: for an electron primary the mismatch is m²/2p ≈ 1.3 keV and is irrelevant. For a
     muon primary — and `h_primary_pdg_` is booked over PDG ±25, so muons are in range — |p|
     and E differ by about 40 MeV at 100 MeV/c, far larger than the energy loss being measured.
   - Suggested fix: use the same quantity on both sides. Either difference against
     `primsim->startMomentum().vect().mag()`, or rebuild the energy at the tracker front as
     `sqrt(p² + m²)` from `primsim->startMomentum().m()`. Then rename the branch and the axis
     titles to match what is stored.

2. 🟠 [S1] A missing `GenEventCount` silently normalizes the per-gen-event numbers to zero
   - Evidence: `beginSubRun` calls `subrun.getHandle<GenEventCount>("genCounter")` and, on an
     invalid handle, writes one line to `std::cerr` and leaves `ngen_` unchanged. `endJob` then
     prints `"Average calo energy deposition per gen event: 0 MeV"`, indistinguishable from a
     real measurement of zero. The PR body describes these per-gen-event averages as the inputs
     to the sensitivity and dose evaluations.
   - Impact: a job over a stage whose subruns no longer carry `GenEventCount`, or with a
     producer labelled anything but `genCounter`, reports 0 for every per-gen-event quantity
     and for the output rate, with only a `std::cerr` line to say so.
   - Suggested fix: three parts. (a) Follow the existing idiom in
     `CommonMC/src/GenEventCountReader_module.cc` and `CaloDiag/src/CaloNeutron_module.cc` —
     `sr.getMany<GenEventCount>()` rather than a hardcoded label, and throw a `cet::exception`
     when the record is absent. If you prefer a label, make it a `fhicl::Atom<art::InputTag>`
     rather than the string literal. (b) Add `consumes<GenEventCount, art::InSubRun>(...)`
     beside the four event-level `consumes` calls. (c) Mark `beginSubRun` `override` — `analyze`
     and `endJob` already are, and most Offline modules implementing this hook do. Without it an
     art signature change would silently disable the accumulation and produce exactly the
     zero-normalization failure above, with no compiler diagnostic.

3. 🟠 [S1] Calorimeter sums use the Birks-quenched visible energy, which is not a dose
   - Evidence: every calorimeter sum uses `css.energyDepBirks()` — `edepBySim`, the `totalCaloE`
     loop in `analyze`, and `h_step_energy_` / `h_step_energy_vs_time_`. `CaloShowerStep` also
     carries `energyDepG4()`, the unquenched deposit. The tracker side uses
     `step.ionizingEdep()`, which is not quenched.
   - Impact: the PR body names "relative detector doses evaluations" as a use. Absorbed dose is
     the deposited energy; Birks quenching describes scintillation light yield and suppresses
     heavily-ionizing contributions by a dE/dx-dependent factor. Using the quenched value biases
     a dose estimate low by an amount that varies with the particle mix, and leaves the calo and
     tracker numbers as two different quantities that cannot be compared.
   - Suggested fix: if the use really is dose, sum `energyDepG4()`. If the quenched value is
     wanted for a calorimeter-signal estimate that is a legitimate choice, but then please say so
     in a comment and in the histogram titles — "calo energy deposition" reads as the raw
     deposit. This one I am asking rather than asserting; you know the downstream use.

4. 🟡 [S2] The `break` in `simTrkFrontStep` makes the earliest-time selection dead code
   - Evidence: `if(!front_trk_sp || sp.time() < front_trk_sp->time()) front_trk_sp = &sp; break;`
     sits inside the matching branch, so the `break` fires on the first match and
     `front_trk_sp` is always null when the comparison runs. The function returns the first
     matching step in collection order, never the earliest in time.
   - Suggested fix: drop the `break` if earliest-time was intended. If first-in-collection is
     fine, drop the comparison — but note that the result then rests on `CompressDetStepMCs`
     preserving input ordering, which it does but does not promise.

5. 🟡 [S2] Dead code and unused includes
   - Evidence: `landau_crystal_ball` and `landau_crystal_ball_func` are never called — the only
     fit path uses `landau_func`. `Info_t::weight_` is set to `1.` and nothing else, so
     `data_.weight` and every `* info_.weight_` in the summations are no-ops. `TString.h` and
     `fhiclcpp/ParameterSet.h` are included but unused.
   - Suggested fix: delete them. Dropping `landau_crystal_ball` also removes the citation to a
     personal GitHub repository, which the coding recommendations ask us not to use as
     documentation.

6. 🟡 [S2] The solenoid x-offset is a bare literal
   - Evidence: the primary start radius is computed as
     `std::sqrt(std::pow(primsim->startPosition().x()+3904.,2) + std::pow(primsim->startPosition().y(),2))`.
   - Suggested fix: `GeomHandle<DetectorSystem> det;` then
     `det->toDetector(primsim->startPosition()).perp()`; the fcl already configures
     `GeometryService`. The literal appears at roughly fifteen other sites in Offline, including
     `Analyses/src/CaloCalib_module.cc`, so this is not a convention you are breaking — a new
     file is just the cheap place to stop propagating it.

7. 🟡 [S2] Primary 0 has its calorimeter sum and tracker-front scan computed twice per event
   - Evidence: `analyze` sets `info_.primsim_edep = edepBySim(info_.primsim, true)` and
     `info_.front_trk_sp = simTrkFrontStep(info_.primsim)` for
     `primarySimParticles().front()`, and the loop directly below repeats both calls for
     `iprim == 0`. `edepBySim` walks the whole `CaloShowerStepCollection` and, per step, the
     SimParticle parent chain; `simTrkFrontStep` walks the whole `StepPointMCCollection`.
     Separately, `fillHistograms` calls `primsim->startMomentum()` six times, and that accessor
     builds a fresh `CLHEP::HepLorentzVector` from the stored `XYZTVectorF` on every call
     (`MCDataProducts/inc/SimParticle.hh:192`) rather than returning a reference.
   - Suggested fix: compute each primary's values once in the loop and point `info_` at the
     `iprim == 0` results; hoist `startMomentum()` into a local at the top of the `if(primsim)`
     block.

8. ⚪ [S3] Error reporting does not follow the standard
   - `bookHistograms` and `fillHistograms` throw `std::runtime_error`; the standard requires
     `cet::exception` with a category unique to the call site, and the one `cet::exception`
     present uses the generic category `"Analysis"`.
   - The `std::cerr` writes in `beginSubRun` and `fit_landau` belong in `mf::LogWarning` /
     `mf::LogError`. The `catch(...)` around the FWHM calculation swallows the failure and
     leaves `fwhm` at 0, which is then printed as `FWHM = 0.00 MeV`.

9. ⚪ [S3] Naming and small cleanups, collected
   - `CaloShowerStepCollection`, `StrawGasStepCollection` and `StepPointMCCollection` are FHiCL
     parameter names that start with capitals and name the product type rather than the role,
     while `primary` and `debugLevel` in the same `Config` are lowercase. Same for the
     `fillHistograms(Hist_t* Hist, const double Weight)` parameters.
   - `new TF1("landau", ...)` collides with ROOT's built-in `landau` formula name in the global
     function list; something like `edepLandauFit` avoids the ambiguity.
   - `hists_` is a 100-entry raw pointer array of which four are used, and the `new Hist_t`
     allocations are never freed; `std::array<Hist_t, 4>` of values removes both the magic size
     and the leak. The bare `hists_[2]` in `endJob` would read better as a named constant.
   - `Hist_t`'s pointer members are uninitialized at declaration — prefer `TH1* h_... = nullptr;`.
   - `endJob` prints `"Primary energy - Edep fit"` for `h_primary_energy_edep_diff_`, whose own
     title correctly says `"Primary Edep - energy"`; the printed label has the sign backwards.

**Validation check**
- Build/tests run: FNALbuild `mu2e/buildtest` is green at `7509119c` (job 3376) — build (prof),
  `ceSimReco`, `ceSteps`, `ceDigi`, `ceMix`, `muDauSteps`, `g4surfaceCheck`, `rootOverlaps`,
  `trigger`, `check_cmake` and the whitespace check all pass. The `check_cmake` failure at the
  earlier head `b363e75d` is resolved. I ran nothing myself.
- Config contract check: pass. Every `Config` field is supplied by `Analyses/test/edep.fcl` with
  matching names and types, the physics-affecting input tags correctly carry no C++ default, and
  `debugLevel` defaulting to 0 is the permitted debug-knob exception.
- Cross-repo consistency: pass. Nothing in `Production` or `mu2e-trig-config` refers to this
  module, and no campaign configuration references `edep.fcl` or `EdepAna`.
- CMake entry: checked. Alphabetical placement between `DiskCal00` and `EMFCSimpleDumper` is
  correct, and omitting explicit `ROOT::Hist` / `ROOT::Tree` matches local practice — only one of
  the 69 plugins in `Analyses/CMakeLists.txt` names a ROOT target.

**Not checked**
- The clang-tidy log (2 errors, 137 warnings at this head). The artifact is on buildmaster and
  clang-tidy is advisory in this job.
- I did not run `edep.fcl` over a real file, so I have not confirmed that `compressDetStepMCs` and
  `FindMCPrimary` are the right labels for the stage you intend to analyze, nor that its subruns
  carry `GenEventCount`.

**Residual risk**
- The module is user-run code with no campaign consumer, so the blast radius of everything above
  is confined to the optimization workflow the PR describes.
- Findings 1 to 3 each change the value of a number that workflow consumes.

**Author follow-ups**
1. Decide whether the tracker-front quantity is |p| or E, and make both sides of every difference
   the same quantity.
2. Make the `GenEventCount` lookup label-free or configurable, fail rather than warn when the
   record is absent, add the `InSubRun` `consumes`, and mark `beginSubRun` `override`.
3. Confirm whether `energyDepBirks()` or `energyDepG4()` is what the dose evaluation wants.
4. Remove either the `break` or the time comparison in `simTrkFrontStep`.
5. Delete the unused Landau-crystal-ball functions, the unused includes, and the always-one
   `weight_`.
6. What did you run to validate this — which dataset, and do the printed averages agree with what
   the existing fast-sim workflow produces today? A line in the description would help the next
   reader.

michaelmackenzie:
I believe I addressed the main AI comments (1-5):

1. I updated to get the actual energy using the sim mass
2. I set the defaults to -1 if `ngen_` is 0
3. I switched to energyDepG4 for dose-like values and visible for edep-based selections
4. I deleted `break;`
5. I deleted the dead code, but left in the weight value to make an update easier to set the weight
6. I left this for now
7. I left this for now, a slight inefficiency
8. I left this for now
9. I left this for now

michaelmackenzie:
@FNALbuild run build test

michaelmackenzie:
@FNALbuild run build test

oksuzian (commented):
### PR Review Summary — Mu2e/Offline#1989

Reviewed at head `dc83453867da6b2451d01815efe4ecf72e7617d0`. Re-review of `7509119cad5b52f35c195fa78a4fc75b75fc2124`.

**Decision**
- 🟡 comment only

Thank you for the point-by-point reply — it made this pass quick. Findings 1, 4 and 5 are fixed
and verified. Finding 2 is partly fixed. Finding 3 is the one worth another look: the
`energyDepG4` / `energyDepBirks` split is applied correctly to the averages and the selections,
but three histograms and both `endJob` fits were left on the visible energy while the quantities
they are differenced against became total energies. That is now the largest remaining item, and
it lands on the module's headline printed numbers.

**Scope of this pass**
- The delta is a single file, `Analyses/src/EdepAna_module.cc`, +57/-62. `Analyses/test/edep.fcl`
  and `Analyses/CMakeLists.txt` are unchanged since the last review, so the config-contract and
  cross-repo conclusions from that review still hold.
- Every prior finding is accounted for below, including the five you deferred.

**Findings**

1. 🟠 [S1] The Birks/G4 split is incomplete, and the part left on the visible energy is what
   `endJob` fits
   - Evidence: `info_.primsim_edep` is set with `edepBySim(info_.primsim, true, true)`, so the
     `edep` local in `fillHistograms` is the Birks-quenched value. It then fills
     `h_primary_edep_`, and is differenced against two unquenched total energies:
     `h_primary_energy_edep_diff_->Fill(edep - primsim->startMomentum().e(), ...)` and
     `h_trk_front_energy_edep_diff_->Fill(edep - step_energy, ...)`. `endJob` Landau-fits
     `hists_[2]->h_primary_energy_edep_diff_` and prints it as
     `"Primary energy - Edep fit: ... mean = ..."`, and reads the MPV of
     `hists_[2]->h_trk_front_energy_edep_diff_` for
     `"Primary Edep - tracker front StepPointMC energy distribution"`.
     Two related mismatches come from the same split: `h_total_calo_energy_` is now filled with
     `info_.calo_total_edep`, which is the `energyDepG4()` sum, while `h_step_energy_` and
     `h_step_energy_vs_time_` in the same histogram set still fill `css.energyDepBirks()` — so
     integrating the step histogram no longer reproduces the total histogram. And the tree's
     `primary_calo_edep` is `edepBySim(sim, true)`, the G4 value, while `h_primary_edep_` is the
     visible value for that same particle when `iprim == 0`.
   - Impact: the two fitted responses are the numbers the optimization consumes, and each is a
     difference between a quenched calorimeter deposit and an unquenched energy. The bias is the
     quenching factor of the shower, which depends on the particle mix, so it does not cancel in
     the difference. This is the same category error as the original finding 1 — one side of a
     subtraction measuring something the other side does not — in quenched-versus-unquenched
     form rather than |p|-versus-E.
   - Suggested fix: decide per histogram rather than per accessor. If the energy-balance plots
     are meant to answer "how much of the primary's energy reached the calorimeter", all three
     want `energyDepG4()`; set `info_.primsim_edep` with `do_vis = false` and add a separate
     visible member if you also want the quenched distribution. If `h_primary_edep_` is
     deliberately the calorimeter-signal estimate, please say so in a comment and keep the two
     difference histograms on G4. Either way `h_step_energy_` should match whatever
     `h_total_calo_energy_` uses, and the tree and histogram spellings of "primary calo edep"
     should agree or be named differently.

2. 🟡 [S2] Finding 2, partly fixed — the label is configurable now, but the record is still not
   declared or required
   - Fixed and verified: the tag is `fhicl::Atom<art::InputTag> genCountTag`, `beginSubRun` uses
     `subrun.getHandle<GenEventCount>(gen_count_tag_)`, and the `endJob` averages fall back to
     `-1.` rather than `0.`, so a missing record is no longer indistinguishable from a real
     measurement.
   - Remaining: the constructor still registers only the four event-level products; there is no
     `consumes<GenEventCount, art::InSubRun>(gen_count_tag_)`. `beginSubRun` is declared
     `virtual void beginSubRun(const art::SubRun& subrun);` without `override`, while `analyze`
     and `endJob` on the two adjacent lines both have it. And `genCountTag` carries the C++
     default `"genCounter"`, which is the one input tag in this `Config` that has one — the other
     three deliberately do not, and this parameter decides whether the per-gen-event
     normalization is right.
   - Suggested fix: add the `InSubRun` `consumes`, add `override`, and drop the string default so
     a job config has to name the producer.

3. 🟡 [S2] Finding 5, partly fixed — the unused includes are still there
   - `landau_crystal_ball` and `landau_crystal_ball_func` are gone, and with them the citation to
     a personal repository. Keeping `weight_` as a hook for event weights is a reasonable call
     and I withdraw that part.
   - `TString.h` and `fhiclcpp/ParameterSet.h` remain included and remain unused.

4. 🟡 [S2] Finding 6, carried over — the solenoid x-offset is still the bare literal `3904.` in
   the `h_primary_start_r_` fill. `GeomHandle<DetectorSystem> det;` and
   `det->toDetector(primsim->startPosition()).perp()` is the one-line replacement.

5. 🟡 [S2] Finding 7, carried over, and the cost has gone up rather than down
   - Removing the `break` was the right fix, but it means `simTrkFrontStep` now always walks the
     whole `StepPointMCCollection` instead of stopping at the first match, and it is still called
     twice for primary 0 — once for `info_.front_trk_sp` and again at `iprim == 0`. The
     Birks/G4 split adds a third full `CaloShowerStepCollection` walk for primary 0:
     `edepBySim(info_.primsim, true, true)`, then `edepBySim(sim, true)` and
     `edepBySim(sim, true, true)` in the loop, each of which also walks the parent chain per step.
   - Still not a blocker, and still a one-line fix: compute each primary's values once in the
     loop and point `info_` at the `iprim == 0` results.

6. ⚪ [S3] Findings 8 and 9, carried over, plus two small things this delta added
   - Unchanged from the last review: `std::runtime_error` rather than `cet::exception` in
     `bookHistograms` and `fillHistograms`; the generic `"Analysis"` category; the `std::cerr`
     writes that belong in `mf::LogWarning`/`mf::LogError`; the `catch(...)` that leaves `fwhm` at
     0 and prints it; the capitalised type-named FHiCL keys; `TF1("landau", ...)` colliding with
     ROOT's built-in formula; the 100-entry raw pointer array with four used and never freed; the
     uninitialized `Hist_t` pointers; and the backwards sign in the
     `"Primary energy - Edep fit"` label.
   - New: `h_trk_front_p_` is one more uninitialized pointer in `Hist_t`. And
     `energyFromStepPoint` is a non-`const`, non-`static` member that reads no member data, while
     `isDescendant` and `edepBySim` beside it are `const` — it belongs in the anonymous namespace
     in the `.cc`, or at least `static`.

**Verified fixed**

- 🟢 [was S1] Finding 1 — `energyFromStepPoint(sim, step)` returns
  `sqrt(p² + m²)` from `sim->startMomentum().m()`, and both `h_trk_front_energy_diff_` and
  `h_trk_front_energy_edep_diff_` now use it, so each difference is between two total energies.
  The momentum is preserved rather than lost: `h_trk_front_p_` and the new `primary_trk_front_p`
  branch keep |p| available. The muon case that motivated the finding — about 40 MeV at
  100 MeV/c — is correct now. Verified line by line at this head.
- 🟢 [was S2] Finding 4 — the `break` is gone from `simTrkFrontStep`, so the
  `sp.time() < front_trk_sp->time()` comparison is now reachable and the function returns the
  earliest step rather than the first in collection order.
- 🟢 [was S1, the accounting part of finding 2] The `-1.` sentinel in the five `endJob` averages
  removes the specific failure I reported, which was a missing `GenEventCount` printing as a
  credible `0`.

**Validation check**
- Build/tests run: FNALbuild `mu2e/buildtest` is green at `dc834538` (job 3379) — build (prof),
  `ceSimReco`, `ceSteps`, `ceDigi`, `ceMix`, `muDauSteps`, `g4test_03MT`, `g4study`,
  `transportOnly`, `POT`, `cosmicSimReco`, `cosmicOffSpill`, `g4surfaceCheck`, `rootOverlaps`,
  `trigger`, `check_cmake`, FIXME/TODO and the whitespace check all pass. I ran nothing myself.
- Config contract check: partial. `edep.fcl` is unchanged, so it does not set the new
  `genCountTag` key and the job takes the C++ default — which works, and is the subject of
  finding 2.
- Cross-repo consistency: pass, unchanged. Nothing in `Production` or `mu2e-trig-config` refers
  to this module.

**Not checked**
- The clang-tidy log, now 2 errors and 136 warnings at this head. It is advisory in this job.
- I still have not run `edep.fcl` over a real file, so I have not confirmed the
  `compressDetStepMCs` and `FindMCPrimary` labels for your intended stage, nor that its subruns
  carry `GenEventCount`.

**Residual risk**
- Unchanged: the module is user-run code with no campaign consumer, so everything above is
  confined to the optimization workflow the PR describes.
- Finding 1 changes the value of the two fitted responses that workflow consumes.

**Author follow-ups**
1. Decide which histograms want `energyDepG4()` and which want `energyDepBirks()`, and make
   `h_step_energy_` agree with `h_total_calo_energy_`. The two `endJob` fits are the ones that
   matter most.
2. Add `consumes<GenEventCount, art::InSubRun>`, mark `beginSubRun` `override`, and drop the
   `"genCounter"` C++ default.
3. Remove the now-unused `TString.h` and `fhiclcpp/ParameterSet.h` includes.
4. Still open from the last pass and entirely your call: the `3904.` literal, the duplicated
   per-event work for primary 0, and the S3 list.

michaelmackenzie:
I think I addressed the main remaining issues, skipping some that do not seem important to address at this stage.

michaelmackenzie:
@FNALbuild run build test

oksuzian (commented):
### PR Review Summary — Mu2e/Offline#1989

Reviewed at head `e57767be04ac5f2caa397c24e0bdbdda5c224dea`. Re-review of `dc83453867da6b2451d01815efe4ecf72e7617d0`.

**Decision**
- 🟡 comment only

Most of the last round is closed and verified. One thing is still open — the visible-versus-G4
question on the two fitted histograms — and the changes in this delta have made the histogram
titles and one comment describe something the code no longer does, which is worth a minute even
if you keep the current behaviour.

**Scope of this pass**
- `Analyses/src/EdepAna_module.cc` +40/-34 and `Analyses/test/edep.fcl` +2/-2.
- The FHiCL keys were renamed to leading capitals (`Primary`, `GenCountTag`, `DebugLevel`
  joining the three collection keys) and `edep.fcl` was updated to match. I checked the tree at
  this head: `Analyses/test/edep.fcl` is the only file that configures this module, so nothing
  else needs the rename.

**Findings**

1. 🟠 [S1] Carried over — the two fitted responses still difference a visible energy against a
   total energy, and the labels now say otherwise
   - Evidence: `info_.primsim_edep_vis` is set from `edepBySim(info_.primsim, true, true)`, so
     `edep` in `fillHistograms` is the `energyDepBirks()` sum. It still fills
     `h_primary_energy_edep_diff_->Fill(edep - startMomentum.e(), ...)` and
     `h_trk_front_energy_edep_diff_->Fill(edep - step_energy, ...)`, and `endJob` Landau-fits the
     first and takes the MPV of the second. Renaming the member to `_vis` made the code honest
     about which value it holds without changing which value is used.
     Two labels are now wrong as a result of this delta. `h_step_energy_` and
     `h_step_energy_vs_time_` were retitled "Visible CaloShowerStep energy", but
     `h_total_calo_energy_` is filled from `info_.calo_total_edep`, the `energyDepG4()` sum, and
     is still titled "Total Calo energy from steps" — so in one output directory the only "steps"
     histogram is the visible one and the "total from steps" is not its integral. The comment
     above the event loop still reads `// Calo shower steps - sum energyDepBirks across the
     collection`, while the loop below sums both and the total it feeds is the G4 one.
   - Impact: unchanged from the last round — the quenching factor depends on the particle mix, so
     it does not cancel in a difference against an unquenched energy, and those two differences
     are the numbers the optimization consumes. The new labelling makes it harder for the next
     reader to notice.
   - Suggested fix: at minimum, correct the `h_total_calo_energy_` title and the loop comment so
     the visible/total split is legible. If you want the fits on the energy balance rather than on
     the calorimeter signal, set `info_` from `edepBySim(..., false)` for those two histograms and
     keep `primsim_edep_vis` for `h_primary_edep_`.

2. 🟡 [S2] `GenCountTag` still carries the C++ default `"genCounter"`
   - The `consumes<GenEventCount, art::InSubRun>` and the `override` on `beginSubRun` are both in
     and verified. What remains is that this is the only input tag in the `Config` with a C++
     default, and `edep.fcl` does not set it — so the producer label that decides the per-gen-event
     normalization is fixed in code. Dropping the default is a one-line change plus one line in
     `edep.fcl`.

3. ⚪ [S3] Small things, collected
   - `fit_landau` changed from `h->Fit(f, "SR")` to `"QNSR"`. The parameters are still updated, so
     the printed results are unchanged — but `N` means the fitted `TF1` is no longer attached to
     the histogram, so the fit curve will not be in the output ROOT file. If that was the intent,
     fine; flagging it because it was not part of anything requested.
   - `data_.prim_calo_edep[iprim] = edepBySim(sim, true)` is still a second full walk of the
     `CaloShowerStepCollection` for `iprim == 0`. Caching the G4 value in `info_` beside
     `primsim_edep_vis` would take primary 0 down to one walk, the same way the `_vis` and
     `front_trk_sp` reuse already do.
   - The FHiCL keys are now consistently leading-capital, which settles the mixed spelling I
     raised. The Mu2e recommendation is the other direction — types capitalised, everything else
     lowercase — so lowercasing the three collection keys would have been the other way to make
     them consistent. Your call; consistency was the substance of the finding.
   - Deferred from the last round and unchanged: the `"Analysis"` exception category is still
     shared by all three throw sites rather than unique per site; `hists_` is still a 100-entry
     raw pointer array with four used and never freed; the `Hist_t` pointers are still
     uninitialized at declaration; `endJob` still prints `"Primary energy - Edep fit"` for a
     histogram titled `"Primary Edep - energy"`; the `std::cerr` writes and the `catch(...)` that
     leaves `fwhm` at 0 are unchanged; and `energyFromStepPoint` is still a non-`const`,
     non-`static` member that reads no member data.

**Verified fixed**

- 🟢 [was S2] The `3904.` literal is gone. `GeomHandle<DetectorSystem> det;` plus
  `det->toDetector(primsim->startPosition()).perp()` is exactly equivalent —
  `DetectorSystem::toDetector` returns `v - _origin` and `DetectorSystemMaker` builds the origin
  as `(-config.getDouble("mu2e.solenoidOffset"), 0., ...)`, so the x term is `x + solenoidOffset`
  — and it now tracks the geometry configuration instead of restating it.
- 🟢 [was S2] `TString.h` and `fhiclcpp/ParameterSet.h` are removed. Better than asked:
  `cetlib_except/exception.h` is now included, and it was previously missing while
  `cet::exception` was already used at the `nprim` check, so that had been relying on a transitive
  include.
- 🟢 [was S2] `consumes<GenEventCount, art::InSubRun>(gen_count_tag_)` is registered beside the
  four event-level products, and `beginSubRun` is `override`.
- 🟢 [was S2] The duplicated per-event work is largely gone: `iprim == 0` now reuses
  `info_.primsim_edep_vis` and `info_.front_trk_sp`, and `startMomentum()` / `startPosition()` are
  hoisted into locals in both `fillHistograms` and the primary loop.
- 🟢 [was S3] Both `std::runtime_error` throws are now `cet::exception`, and the fit function is
  renamed `landau_func`, which removes the collision with ROOT's built-in `landau` formula.

**Validation check**
- Build/tests run: `mu2e/buildtest` is **in progress** at `e57767be` — you triggered it at 20:37
  UTC and it had not reported when I wrote this. The previous head `dc834538` was green
  (job 3379). I ran nothing myself. The one thing worth watching in this run is the FHiCL key
  rename, since `edep.fcl` is the only consumer and it was updated in the same commit.
- Config contract check: pass. Every `Config` field is supplied by `Analyses/test/edep.fcl` with
  matching names, except `GenCountTag`, which takes its C++ default — finding 2.
- Cross-repo consistency: pass, unchanged. Nothing in `Production` or `mu2e-trig-config` refers to
  this module.

**Not checked**
- The clang-tidy log at this head.
- I still have not run `edep.fcl` over a real file, so the `compressDetStepMCs` and
  `FindMCPrimary` labels and the presence of `GenEventCount` in your intended stage remain
  unconfirmed from my side.

**Residual risk**
- Unchanged: no campaign configuration runs this module, so everything above is confined to the
  optimization workflow the PR describes.

**Author follow-ups**
1. Fix the `h_total_calo_energy_` title and the `// sum energyDepBirks across the collection`
   comment, which no longer describe what that code does. Then decide whether the two fitted
   differences should be on the G4 energy.
2. Drop the `"genCounter"` C++ default and set `GenCountTag` in `edep.fcl`.
3. The S3 list is entirely optional and none of it gates anything.

oksuzian (approved):
### PR Review Summary — Mu2e/Offline#1989

Reviewed at head `23186674d22d7a84a0aebe35b4bbb1a6b120609e`. Re-review of `e57767be04ac5f2caa397c24e0bdbdda5c224dea`.

**Decision**
- 🟢 approve

The visible/G4 question is settled — you have made the calorimeter side visible consistently, and
I am withdrawing that finding rather than raising it a fourth time. What the switch left behind is
two members that are now written and never read, and one of them means the printed summary and the
histograms are reporting different quantities. Both are small and neither gates anything.

**Scope of this pass**
- One commit, `23186674d`, +5/-5 in `Analyses/src/EdepAna_module.cc`. `h_total_calo_energy_` is
  renamed `h_total_calo_energy_vis_`, retitled "Total visible Calo energy from steps", and filled
  from `info_.calo_total_edep_vis`; the two `h_ref` uses in `endJob` follow the rename.

**Withdrawn**

- 🟠 [was S1] The visible-versus-G4 finding. I raised this in all three previous rounds. At this
  head the calorimeter side is visible throughout: `h_step_energy_`, `h_step_energy_vs_time_` and
  now `h_total_calo_energy_vis_` all use `energyDepBirks()`, and the three selection cuts at
  `:493-495` are on `totalCaloEVis`. Differencing that against the primary's total energy is then
  a coherent measurement of how much of the primary's energy does not become calorimeter signal,
  which is a reasonable thing for the optimization workflow to want. The reason I kept raising it
  was that the labels said one thing and the code did another; that is no longer true. Treating
  your consistent conversion as the design answer it is.

**Findings**

1. 🟡 [S2] `Info_t::calo_total_edep` is now written and never read
   - Evidence: declared at `:74`, assigned at `:433` from `totalCaloE`, and that is all — this
     delta removed its only reader when `h_total_calo_energy_->Fill(info_.calo_total_edep, ...)`
     became `h_total_calo_energy_vis_->Fill(info_.calo_total_edep_vis, ...)` at `:370`. The G4 sum
     itself is still used, at `:446` and `:481`, just not through `info_`.
   - Suggested fix: delete the member and the assignment.

2. 🟡 [S2] The printed averages are the G4 sums while every histogram and cut is now visible
   - Evidence: `total_calo_edep_vis_` is declared at `:238` and accumulated at `:482`, and nothing
     reads it. The summary at `:554-555` computes `averageCaloEdep` and `averageCaloEdepGen` from
     `total_calo_edep_`, the G4 accumulator, and prints them as "Average calo energy deposition
     per event" and "per gen event". Meanwhile `events_above_50_mev_` at `:483` and the three
     `fillHistograms` cuts at `:493-495` are all on `totalCaloEVis`.
   - Impact: the two numbers the PR body describes as the inputs to the sensitivity and dose
     evaluations are on a different quantity from everything plotted beside them, and nothing in
     the printout says which is which. An unread `_vis_` accumulator sitting next to them reads
     like the switch stopped halfway rather than like a decision.
   - Suggested fix: pick one. G4 is the defensible choice for anything called a deposition or a
     dose — in which case delete `total_calo_edep_vis_` and say "total" in the two printed labels.
     If you want these to match the plots, use `total_calo_edep_vis_` and say "visible". Either
     way one of the two accumulators should go.

3. ⚪ [S3] Carried over, unchanged by this delta
   - `GenCountTag` still carries the C++ default `"genCounter"` at `:46` and `edep.fcl` does not
     set it, so the producer label that decides the normalization is still fixed in code.
   - The comment at `:390`, `// Calo shower steps - sum energyDepBirks across the collection`, now
     sits above a loop that sums both (`:424-425`).
   - The rest of the round-3 S3 list is untouched and remains optional: the shared `"Analysis"`
     exception category, `hists_` as a 100-entry raw array with four used, the uninitialized
     `Hist_t` pointers, the `"Primary energy - Edep fit"` printed label, the `std::cerr` writes and
     the `catch(...)` that leaves `fwhm` at 0.

**Verified**
- 🟢 The rename is complete. `h_total_calo_energy_` appears nowhere at this head; both `endJob`
  efficiency denominators at `:526` and `:543` use `h_total_calo_energy_vis_`, and the histogram's
  ROOT name is unchanged at `total_calo_energy`, so existing plotting scripts still find it.
- 🟢 [was S1, round 1] The "reports 0 as if it were a measurement" problem is gone. `:554-555` now
  yield `-1.` when `total_events_` or `ngen_` is zero, so a job with no `GenEventCount` prints a
  value that cannot be mistaken for a real average. The `std::cerr` warning at `:302` is still a
  warning rather than a throw, which is the part of that finding I am leaving closed as your call.

**Validation check**
- Build/tests run: `mu2e/buildtest` is **pending** at `23186674d`; the parent `e57767be` was green
  (job 3381). I did not trigger a run, since one is already in flight, and I ran nothing locally.
  Nothing above depends on the build — the delta is a rename plus one member swap.
- Config contract check: pass, unchanged. `edep.fcl` supplies every `Config` field except
  `GenCountTag`, which takes its C++ default.
- Cross-repo consistency: pass, unchanged. Nothing in `Production` or `mu2e-trig-config` refers to
  this module.

**Not checked**
- The clang-tidy log at this head, and I still have not run `edep.fcl` over a real file, so the
  `compressDetStepMCs` and `FindMCPrimary` labels and the presence of `GenEventCount` in your
  intended stage remain unconfirmed from my side.

**Author follow-ups**
1. Delete `Info_t::calo_total_edep` and its assignment.
2. Decide whether the printed averages should be the total or the visible sum, delete the
   accumulator you do not use, and put the word in the two printed labels.

oksuzian:
@FNALbuild build run test

oksuzian:
@FNALbuild run build test

PR #1990, opened 2026-09-19
update TrkPanelMap to 18 statiions, cosmetic changes to the digi maker

pavel1murat:
changes described in the title


oksuzian (commented):
### PR Review Summary — Mu2e/Offline#1990

Reviewed at head `9b6fbe817669c557a7688b34eaf9f6e407b55829`. First pass.

**Decision**
- 🟡 comment only

**Scope understood**
- `TrackerConditions/data/TrkPanelMap.txt`: the `121700-200000` IOV block is rewritten from 24 planes to the full 36 (18 stations), plus two `ppid` label corrections and one mnid swap.
- `DAQ/src/StrawDigisFromArtdaqFragments_module.cc`: `print_` gains a severity argument and dispatches to five message-facility macros; the message category is renamed; `diagLevel` is marked obsolete.
- `DAQ/fcl/prolog.fcl`: the `makeSD` default for `keyOnMnid` is changed from `false` to `true`. This is a behavior change, not a cosmetic one, and the PR title and body do not mention it.

**Findings**

1. 🟠 [S1] The `keyOnMnid` default flip changes how straw digis are addressed, and is not described anywhere in the PR
   - Evidence: `DAQ/fcl/prolog.fcl:14` now reads `keyOnMnid : true ## false`. In the module, `keyOnMnid_` selects the lookup key: `StrawDigisFromArtdaqFragments_module.cc:388` takes `panel_map_by_online_ind(dtc_id,link_id)` when it is false, and `:441` takes `panel_map_by_mnid(mnid)` when it is true, where `mnid` comes from the hit payload (`:438`). The `fhicl::Comment` at `:82` still says "default:false", and `DAQ/test/digi_to_frag.fcl:46` still sets `false` explicitly.
   - Impact: every job that takes the `DAQ` prolog defaults — `DAQ/test/make_straw_digis.fcl` and the other test configs, which are the real-data decoding path — switches from a hardware-authoritative key (DTC, link) to a payload-authoritative one. For 214 of the 216 rows the two keys resolve to the same row, so the output is unchanged; for `dtc 15` links 2 and 5 they now disagree, because this PR also swaps the mnids on those two links (`TrkPanelMap.txt:393,395`). Concretely, hits carrying mnid 225 land on plane 23 panel 5 instead of panel 2, and mnid 274 on panel 2 instead of panel 5.
   - Suggested fix: state the flip and its motivation in the PR body, update the `fhicl::Comment` at `:82` to match the new default, and drop the `## false` — the previous value is in git and the house rule is not to record history in comments. Please also confirm the IOV choice: `121700-200000` is a window that already contains data, so the mnid correction is applied retroactively. If the swap was a mapping error present since run 121700 that is correct; if the panels were re-cabled at a known run, this needs an IOV split at that run instead.

2. 🟠 [S1] With `keyOnMnid` true, the mnid/hardware consistency check can never fire, so `StrawDigiFlag::corrupted` is never set
   - Evidence: `StrawDigisFromArtdaqFragments_module.cc:454` tests `tpm != nullptr && tpm->mnid() != mnid`. When `keyOnMnid_` is true, `tpm` was obtained at `:441` as `panel_map_by_mnid(mnid)`, so `tpm->mnid() == mnid` identically and the branch is unreachable. The flag it sets at `:461` is the only place `StrawDigiFlag::corrupted` is raised in this module.
   - Impact: in the new default mode nothing cross-checks the payload channel ID against the hardware address it arrived on. A corrupted mnid that happens to collide with another panel's mnid is silently decoded onto that other panel, with no flag on the digi. That is the "detect a bad input and continue" pattern the coding standard rules out.
   - Suggested fix: in the `keyOnMnid` branch, compare the row against the hardware address instead — `tpm->dtc() != dtc_id || tpm->link() != link_id` — and raise the same `corrupted` flag. The existing check can then stay as-is for the `keyOnMnid == false` path.

3. 🟠 [S1] Routing `e_DEBUG` through `MF_LOG_TRACE` silently disables all `debugMode` output in the standard build
   - Evidence: `StrawDigisFromArtdaqFragments_module.cc:188` uses `MF_LOG_TRACE("MAKE_DIGI_NT")` for `e_DEBUG`, replacing the unconditional `mf::LogVerbatim`. In `messagefacility/MessageLogger/MessageLogger.h`, `NDEBUG` or `MF_NDEBUG` defines `MF_SUPPRESS_LOG_DEBUG`, under which `MF_LOG_TRACE(id)` expands to `mf::NeverLogger_{}`. Offline configures `CMAKE_BUILD_TYPE : RelWithDebInfo` (`CMakePresets.json`), which carries `-DNDEBUG`.
   - Impact: every `print_(e_DEBUG, ...)` call — the `-- START`/`-- END` markers, the fragment and ROC dumps, the waveform dump — produces nothing in a prof build, and the `debugMode` and `debugBits` parameters become inert. Nothing warns the user that the knob no longer does anything.
   - Suggested fix: use `MF_LOG_VERBATIM` for `e_DEBUG` as well, or gate on `MF_DEBUG`, or state in the PR that debug output is now intended to be a debug-build-only facility.

4. 🟡 [S2] Renaming the message category from `MAKE_SD` to `MAKE_DIGI_NT` orphans the throttling configuration in six fcl files
   - Evidence: all five macros at `:186-206` now emit under `MAKE_DIGI_NT`. `DAQ/test/dump_fragments.fcl:53`, `filter_straw_digis.fcl:43`, `analyzeDigis.fcl:27`, and also `analyzeTrkFragments.fcl`, `filter_straw_hits.fcl` and `dump_fragments_pre_107247.fcl` still configure `services.message.destinations.log.categories.MAKE_SD` with `reportEvery : 1, limit : -1` under the comment "make makeSD print everything".
   - Suggested fix: rename the category in those six files in this PR, so the intent does not silently stop working.

5. 🟡 [S2] Two `zface` values in the rewritten table break the pattern the other 34 planes follow
   - Evidence: across the 216 new rows every plane assigns `zfc` by panel parity, in one of exactly two phases — even planes use `(0,1,0,1,0,1)` or `(1,0,1,0,1,0)` over panels 0..5, odd planes `(2,3,2,3,2,3)` or `(3,2,3,2,3,2)`. Two rows do not fit: `TrkPanelMap.txt:393` gives plane 23 panel 5 `zfc 3` where the plane's own phase `(3,2,3,2,3,·)` requires 2, and `:421` gives plane 28 panel 3 `zfc 2`, which is not even a legal face for an even plane. The result is 53/54/54/55 rows per face instead of 54 each. Both values are carried over unchanged from the current table, so this is not a regression, but the PR rewrites the block.
   - Suggested fix: confirm against the installation record and correct both rows, or add a comment saying why these two panels are exceptions.

6. 🟡 [S2] Inline comments on two data rows are outside the file format
   - Evidence: `TrkPanelMap.txt:393` and `:395` end with `# fixed swap with 274` / `# fixed swap with 225`. `DbUtil::readString` only recognises a comment when `#` is the first character after trimming, so these are parsed as data; the last field survives only because `TrkPanelMap::addRow` uses `std::stoi`, which stops at the first non-digit. The comment text becomes part of the table's CSV payload and of the hash computed by `hashCsv()`.
   - Suggested fix: move both notes to a `#` line above the rows.

7. 🟡 [S2] `diagLevel` is marked obsolete in a dated, initialled comment rather than removed
   - Evidence: `StrawDigisFromArtdaqFragments_module.cc:77` now reads `fhicl::Comment("2026-04-25 PM: OBSOLETE, WILL BE REMOVED SOON")`, and `DAQ/fcl/prolog.fcl:9` keeps `diagLevel : 0 ## not used yet`. The parameter is read into a member but never used.
   - Suggested fix: delete the parameter and its prolog entry here rather than annotating it; git carries the history, and dated initials in comments are what the coding standard asks reviewers to reject.

8. ⚪ [S3] Collected nits
   - `:261` introduces a typo: `n_fragment_collections):{}`, stray `)`.
   - The `"WARNING: "` and `"ERROR: "` string prefixes at `:198,202,206` duplicate the severity that message-facility already applies through `MF_LOG_PRINT`/`MF_LOG_PROBLEM`.
   - The rewritten table drops the `# root [N] print_trk_panel_map(N)` provenance lines that the earlier IOV blocks carry, and the replacement `#---- slot N` markers appear for only two of the eighteen stations.

**Verified**
- 🟢 The new `121700-200000` block is internally consistent: 216 rows, all 36 planes present with 6 panels each, mnid unique, (DTC, link) unique, (plane, panel) unique, 36 distinct DTC IDs 1..36 one-to-one with planes, and 36 distinct `ppid` values 1..36 used once each.
- 🟢 The change is strictly additive apart from the two label fixes: 12 DTCs (9, 10, 19-28) and 12 planes (0, 1, 8-17) are added, and no (DTC, link) key is removed.
- 🟢 The two `ppid` corrections are right. Production plane 17 is geometric plane 0 in all three earlier IOV blocks (mnids 120/244/126/210/269/191 on DTC 21); the current `121700-200000` block wrongly reused `ppid 17` for the DTC-30 plane. The PR restores production plane 17 to plane 0 — now on DTC 20 — and gives the DTC-30 plane its own `ppid 2`; plane 3 moves from `ppid 4` to `ppid 7` for the same reason. All 36 production-plane IDs are then used exactly once.
- 🟢 The 225/274 change moves only the mnid attached to `dtc 15` links 2 and 5; panel and zface stay with the link, so decoding by (DTC, link) is bit-for-bit unchanged and only the `keyOnMnid` path is affected.
- 🟢 CI is green at this head, including `build (prof)`, `ceSimReco`, `ceDigi`, `trigger`, `check_cmake`, the whitespace check and `g4surfaceCheck`/`rootOverlaps`.
- 🟢 No `Production` or `mu2e-trig-config` configuration instantiates `StrawDigisFromArtdaqFragments`, and no production campaign runs it, so the prolog default flip does not reach campaign output. (The `makeSD` label in the campaign configs is the MC digitiser, a different module.)

**Not checked**
- Whether mnids 225 and 274 are physically on links 5 and 2 — that is hardware truth this review cannot reach, and it is the whole basis of finding 1.
- Whether the plane-0 DTC change from 21 to 20 matches the installation record.
- Whether data already taken in the `121700-200000` window has been reconstructed with the current map.

**Validation check**
- Build/tests run: no local build; relied on the FNALbuild result at this head.
- Config contract check: partial — the `keyOnMnid` default now contradicts its own `fhicl::Comment`, and six fcl files configure a message category the module no longer emits.
- Cross-repo consistency: pass — no consumer outside `Offline`.

**Residual risk**
- The decoding key change reaches every job that takes the `DAQ` prolog defaults, and nothing in CI exercises artdaq fragment decoding, so a mistake here would not be caught before it reaches data.
- With the mnid/hardware cross-check unreachable, corrupted channel IDs are decoded silently rather than flagged.

**Author follow-ups**
1. Describe the `keyOnMnid` flip in the PR body, update the `fhicl::Comment` to match, and remove the `## false`.
2. Confirm the IOV: should the 18-station map and the mnid swap apply retroactively from run 121700, or open a new IOV at the run where the configuration changed?
3. Restore a mnid/hardware cross-check in the `keyOnMnid` path, comparing `tpm->dtc()`/`tpm->link()` with the arrival address.
4. Confirm that `debugMode` output is still expected to appear in a prof build; if so, `MF_LOG_TRACE` is the wrong macro.
5. Rename the `MAKE_SD` message category in the six `DAQ/test` fcl files.
6. Check the two off-pattern `zface` values at `TrkPanelMap.txt:393` and `:421`, and move the inline row comments onto their own lines.

oksuzian (changes requested):
### PR Review Summary — Mu2e/Offline#1990

Reviewed at head `36f3d57e976394bd6a5da9d8931aeecdbfaa027c`. Re-review of `9b6fbe817669c557a7688b34eaf9f6e407b55829`.

**Decision**
- 🔴 request changes

**Scope of this pass**

Two commits landed since the first review: `584d9104d` ("obsolete") and `36f3d57e9` ("more
cleanups"). They delete `DAQ/src/StrawDigiFilter_module.cc` (430 lines) and make further changes
to `StrawDigisFromArtdaqFragments_module.cc`. `TrackerConditions/data/TrkPanelMap.txt` and
`DAQ/fcl/prolog.fcl` are untouched by the delta, so every finding against them stands.

One prior finding is fixed, six are unaddressed, the delta introduces one blocker and two new
major items, and one 🟢 from the previous review turns out to have been wrong — see the
correction below, which enlarges the blast radius of the `keyOnMnid` finding.

**Correction to the previous review**

My first pass recorded a 🟢 saying "no `Production` or `mu2e-trig-config` configuration
instantiates `StrawDigisFromArtdaqFragments`". That is wrong, and I am withdrawing it.
`mu2e-trig-config/core/trigProducers.fcl:12` does `#include "Offline/DAQ/fcl/prolog.fcl"`, and
`core/producers/trigTrkProducers.fcl:4-6` defines `TrigMakeSD : { @table::DAQ.producers.makeSD }`
with no overrides. `TrigMakeSD` is the first module in both `trkPrepareHits` and
`trkPrepareHitsOffSpill` (`core/trigRecoSequences.fcl:6-7`), and `keyOnMnid` does not appear
anywhere in `mu2e-trig-config`. So the prolog default flip in `DAQ/fcl/prolog.fcl:14` propagates
straight into the trigger reconstruction sequences. Carried-forward finding 1 below should be
read with that in mind.

**New findings at this head**

1. 🔴 [S0] `StrawDigiFilter` is deleted but `DAQ/test/filter_straw_digis.fcl` still runs it, so that job can no longer start
   - Evidence: the module source is gone and its `cet_build_plugin` block is commented out
     (`DAQ/CMakeLists.txt:199-212`), so no `StrawDigiFilter` plugin is produced. But
     `DAQ/fcl/prolog.fcl:28` still defines `StrawDigiFilter : { module_type : StrawDigiFilter ... }`,
     `DAQ/test/filter_straw_digis.fcl:65` pulls that table in with `filters : { @table::DAQ.filters }`,
     `:67` puts the label on the trigger path (`p1 : [ makeSD , StrawDigiFilter ]`), and `:89-92`
     still override four of its parameters.
   - Impact: `mu2e -c Offline/DAQ/test/filter_straw_digis.fcl` fails at JobSetup with
     `Library specification "StrawDigiFilter" does not correspond to any library in CET_PLUGIN_PATH
     of type "module"`. No `DAQ/test` configuration is in the buildtest job list, so a green CI run
     does not cover this — it surfaces only when someone runs the VST decoding chain by hand.
   - Suggested fix: in this PR, remove the `StrawDigiFilter` block from `DAQ/fcl/prolog.fcl`, and
     either delete `DAQ/test/filter_straw_digis.fcl` or rewrite its path and overrides without the
     filter. `grep -rn StrawDigiFilter` over the tree at this head returns only those two files
     plus the commented CMake block, so the cleanup is bounded.

2. 🟠 [S1] The ROC loop bound changed from the sub-event byte count to `frag->dataSizeBytes()`, which is what the comment deleted in the same edit warned against
   - Evidence: `StrawDigisFromArtdaqFragments_module.cc:308` now sets
     `uint8_t* last_address = fdata+frag->dataSizeBytes();` from the unadvanced `fdata`, and the
     old computation is commented out at `:340-341`:
     `// ushort* buf = (ushort*) fdata;` / `// int nbytes = buf[0]; // frag.dataSizeBytes() includes extra 24 bytes`.
     Previously `buf` pointed at the `DTC_SubEventHeader`, because `fdata` had already been advanced
     by `sizeof(DTCLib::DTC_EventHeader)` at `:321`, so `last_address` was the end of the sub-event.
     The two agree only if `dataSizeBytes() == sizeof(DTC_EventHeader) + subevent_bytes`. The removed
     comment asserts the opposite — that `dataSizeBytes()` carries extra bytes — and this PR edits
     that number from `0x20` to `24` while commenting the line out.
   - Impact: `while (roc_data < last_address)` at `:344` governs how far the decoder walks. If the
     removed comment was right, the loop now runs past the last real ROC block and parses a
     `RocDataHeaderPacket_t` out of trailer bytes. The `roc_data+packet_size >= last_address` guards
     at `:360` and `:424` catch the tail of that, not a garbage header with `packetCount > 1`. The
     difference is larger again when `missingDTCHeaders_` is true (runs <= 107246), because `fdata`
     is then never advanced and the two bounds differ by the DTC event header as well. Nothing in CI
     decodes artdaq fragments.
   - Suggested fix: say in the PR body why the new bound is safe, and put an accurate comment next
     to `:308` rather than deleting the one that explained the old choice. If the intent was to fix
     the 16-bit truncation of `buf[0]`, that is a good change and worth stating as one; it is not
     cosmetic.

3. 🟠 [S1] Container fragments now fall into the DTC decoding path instead of being unpacked
   - Evidence: `:273-282` comment out the `if (handle->front().type() == artdaq::Fragment::ContainerFragmentType)`
     branch together with its `else {`, and `:552` comments out the matching closing brace, so the
     body that used to be the `else` now runs for every handle. The two vectors the container branch
     filled are commented out at `:261-262`.
   - Impact: the old branch was already broken — in the pre-PR file neither `fragments` nor
     `containerFragments` is read anywhere after being filled, so container input was silently
     dropped. The change replaces "dropped" with "handed to the tracker decoder", where `frag->type()`
     is `ContainerFragmentType`, the CFO skip at `:304-305` does not catch it, and the
     `DTC_SubEventHeader` at `:328` is read out of container payload. The subsystem test at `:329-334`
     will usually reject it, but on arbitrary bytes that is luck, not a check.
   - Suggested fix: settle the question rather than commenting it out. If container fragments no
     longer occur, delete the branch and the two vectors outright and say so in the PR body. If they
     do occur, the branch has to actually feed the decoder.

4. 🟡 [S2] Commented-out code carrying dated initials, now in four places
   - Evidence: `DAQ/CMakeLists.txt:199-212` (the whole `StrawDigiFilter` build rule), and
     `StrawDigisFromArtdaqFragments_module.cc:261-262`, `:273-282` and `:552`, all prefixed
     `# 2026-09-21 PM` or `// 2026-09-21 PM`. The dead-code rule and the "do not use comments to
     mark your changes or record code history" rule both apply. This is the same pattern as
     carried-forward finding 7, which was one site at the previous head and is now five.
   - Suggested fix: delete the blocks; git has them.

5. 🟡 [S2] The debug-bit parse still trusts `sscanf`, and a bad index is logged rather than rejected
   - Evidence: `:160` calls `sscanf(key,"bit%i:%i",&index,&value)` without checking the return value.
     `index` and `value` are initialized to 0 at `:158`, so a `debugBits` entry that does not match
     the format silently sets `debugBit_[0] = 0` and passes the new range check at `:161`. An
     out-of-range index takes the `else` at `:165`, logs `e_ERROR` and continues.
   - Suggested fix: check `sscanf(...) == 2` and throw a `cet::exception` on either failure. This is
     a configuration error, which the standard treats as fatal; `debugBits` is read at construction
     so throwing is cheap, and a typo currently disappears into the log.

6. 🟡 [S2] The PR now does two things, and the title no longer describes it
   - Evidence: the title is "update TrkPanelMap to 18 statiions, cosmetic changes to the digi maker"
     (with a typo in "stations") and the body is "changes described in the title". The PR deletes a
     430-line filter module, changes the decoder's loop bound, changes container-fragment handling,
     and fixes a digi-flag write — none of that is cosmetic or related to the panel map.
   - Suggested fix: split the `StrawDigiFilter` removal into its own PR, and describe the decoder
     changes in the body.

7. ⚪ [S3] `mf_stream_` at `:124` holds a message-facility category, not a stream; `mfCategory_`
   would say what it is.

**Fixed since the last review**

- 🟢 [was 🟡 S2] The message-category rename is reverted. `:124` introduces
  `const std::string mf_stream_{"MAKE_SD"}` and all five macros use it — `MF_LOG_TRACE` at `:194`,
  `MF_LOG_VERBATIM` at `:198`, `MF_LOG_PRINT` at `:204`, `MF_LOG_PROBLEM` at `:208`,
  `MF_LOG_ABSOLUTE` at `:212` — so the `services.message.destinations.log.categories.MAKE_SD`
  blocks in the six `DAQ/test` fcl files work again. Verified at all five call sites.

**Also fixed in the delta, not previously reported**

- 🟢 `:506-507` changes `auto digi = straw_digis->back(); digi.digiFlag() = digi_flag;` to
  `auto digi = &straw_digis->back(); digi->digiFlag() = digi_flag;`. `StrawDigi::digiFlag()` has both
  a const and a non-const overload (`RecoDataProducts/inc/StrawDigi.hh:57-58`), so `auto` previously
  deduced `StrawDigi` by value and the flag was written to a discarded copy. This is a real fix:
  `StrawDigiFlag::corrupted` was never actually stored on any digi before. It also makes
  carried-forward finding 2 more consequential rather than less — that check is now the only thing
  between a corrupted channel ID and an unflagged digi, and it is unreachable in the new default mode.
- 🟢 `:161` adds the `(index >= 0) and (index < kNDebugBits)` guard around `debugBit_[index] = value`.
  `debugBit_` is `int debugBit_[kNDebugBits]` with `kNDebugBits = 100` (`:103`, `:111`) and the index
  came straight from `sscanf`, so a `debugBits` entry such as `"bit200:1"` was an out-of-bounds write
  into the module object. Finding 5 covers what is still open on that line.

**Carried over from the review at `9b6fbe81` — all unaddressed**

Full evidence is in that review; these are re-verified at this head, not restated.

- 🟠 [S1] carried over, blast radius corrected upward — the `keyOnMnid` default flip.
  `DAQ/fcl/prolog.fcl:14` still reads `keyOnMnid : true ## false`, the `fhicl::Comment` at
  `StrawDigisFromArtdaqFragments_module.cc:81` still says `default:false`, and neither the title nor
  the body mentions it. Per the correction above, this also changes the trigger's `TrigMakeSD`. The
  IOV question from the first review is still open.
- 🟠 [S1] carried over — the mnid/hardware consistency check is unreachable in the new default mode.
  `:460` tests `tpm->mnid() != mnid` on a `tpm` that `:447` obtained via `panel_map_by_mnid(mnid)`.
  The `StrawDigiFlag::corrupted` it sets at `:467` is the flag whose write was just fixed at `:507`.
- 🟠 [S1] carried over — `e_DEBUG` still routes through `MF_LOG_TRACE(mf_stream_)` at `:194`, which
  expands to `mf::NeverLogger_{}` under `NDEBUG`, so `debugMode` and `debugBits` produce nothing in
  the prof build.
- 🟡 [S2] carried over — the two off-pattern `zface` values at `TrkPanelMap.txt:393` and `:421`.
- 🟡 [S2] carried over — the inline `# fixed swap with ...` comments on data rows at
  `TrkPanelMap.txt:393,395`, which `DbUtil::readString` parses as data.
- 🟡 [S2] carried over — `diagLevel` is still annotated
  `fhicl::Comment("2026-04-25 PM: OBSOLETE, WILL BE REMOVED SOON")` at `:76` instead of removed, and
  `DAQ/fcl/prolog.fcl:9` still carries it.
- ⚪ [S3] carried over — the `n_fragment_collections):{}` typo is still at `:266`, and the
  `"WARNING: "` / `"ERROR: "` / `"SEVERE: "` prefixes at `:204,208,212` still duplicate the severity
  that `MF_LOG_PRINT` / `MF_LOG_PROBLEM` / `MF_LOG_ABSOLUTE` already apply.

**Validation check**
- Build/tests run: `mu2e/buildtest` is **pending** at `36f3d57e` (triggered 2026-09-22 01:22 UTC);
  `jenkins/ghprb` is green. The last complete result is at the previous head `9b6fbe81` and covers
  none of the delta. I did not trigger a run, since one is already in flight, and I ran nothing
  locally.
- Config contract check: **fail** — finding 1, and `keyOnMnid` still contradicts its own
  `fhicl::Comment`.
- Cross-repo consistency: **needs follow-up**, and the previous review got this wrong.
  `mu2e-trig-config` consumes `DAQ.producers.makeSD` unmodified, so the prolog default reaches
  `TrigMakeSD`. `Production` has no reference to either module.

**Not checked**
- Whether the build accepts `MF_LOG_TRACE(mf_stream_)` and the other four macros with a
  `std::string` category argument — the pending run will settle that.
- `sizeof(DTCLib::DTC_EventHeader)` and the real layout of `DTC_SubEventHeader`, which is what would
  turn finding 2 from a question into a demonstration. DTCLib is not available in this session.
- Whether the mnid-keyed addressing is what the trigger wants, now that the flip is known to reach
  `TrigMakeSD`.
- The hardware questions from the first review: whether mnids 225 and 274 are physically on `dtc 15`
  links 5 and 2, and whether plane 0 moved from DTC 21 to DTC 20 in the installation record.

**Residual risk**
- Nothing in CI decodes artdaq fragments or runs any `DAQ/test` configuration, so findings 1, 2 and
  3 all sit outside what a green build proves.
- The decoder changes arrive in a PR described as cosmetic, which is the state in which they are
  least likely to be read carefully by the next person.

**Author follow-ups**
1. Remove the `StrawDigiFilter` entry from `DAQ/fcl/prolog.fcl` and fix or delete
   `DAQ/test/filter_straw_digis.fcl`, so the deletion leaves no configuration referring to the module.
2. Explain the `last_address` change and restore an accurate comment; confirm it holds for the
   `missingDTCHeaders_` path too.
3. Settle the container-fragment question and delete the branch outright if they no longer occur.
4. Delete the commented-out blocks in `DAQ/CMakeLists.txt` and in the module rather than prefixing
   them with a date and initials.
5. Split the `StrawDigiFilter` removal out, and give the PR a body listing the decoder changes.
6. Confirm that flipping `keyOnMnid` in the shared `DAQ` prolog is intended to change `TrigMakeSD`
   as well, and the rest of the items carried over from the previous review.

oksuzian:
@FNALbuild run build test

oksuzian:
@FNALbuild run build test

oksuzian (approved):

oksuzian:
Post-merge correction to my review above — one finding was wrong and should not stand on the record.

**Withdrawn: the `last_address` finding (item 2).** I claimed the change from `fdata+buf[0]` to `fdata+frag->dataSizeBytes()` was an unexplained loosening of the ROC loop bound, on the strength of the deleted comment saying `dataSizeBytes()` "includes extra 0x20". That reading was wrong. `DTC_EventHeader` is three `uint64_t` bitfield words, so `sizeof(DTCLib::DTC_EventHeader)` is **24**, and `fdata` has already been advanced by exactly that much before the old `buf[0]` read. The two bounds therefore land on the same byte, and the new form additionally drops the 16-bit truncation of `inclusive_subevent_byte_count`, which the old one silently suffered above 64 KiB. The `0x20` → `24` edit corrected a comment that had the wrong number in it.

So that was a fix, not a regression, and it closes the first item of #1997. Apologies for the noise.

**For the record, what this PR fixed that #1997 had open:** the loop bound above, the `digiFlag()` write to a discarded copy at what is now `:506`, and the unchecked `debugBits` index at `:161`. The `StrawDigiFilter` items in that issue are moot now that the module is gone.

**Still open**, moved to #1997 rather than reopened here:

- `// 2026-09-21 PM` commented-out container handling at `:261-282` and `:552`, and the underlying question — with the branch commented out, container fragments now reach the DTC decoder instead of being unpacked
- `diagLevel` still carries `fhicl::Comment("2026-04-25 PM: OBSOLETE, WILL BE REMOVED SOON")` at `:76`
- `TrkPanelMap.txt:393,395` — the trailing `# fixed swap with ...` notes sit after the last field, and `DbUtil::readString` only treats `#` as a comment in column 1, so they become part of the CSV and of `hashCsv()`

On the two answers given in Slack: the readout-level MN225/MN274 swap settles the IOV question, and routing `e_DEBUG` through `MF_LOG_TRACE` is a deliberate separation of debug output from error diagnostics. Both taken as answered.

#2003 is the `forceOfflineAddressing` cleanup we discussed.

PR #1991, opened 2026-09-20
TrkHitReco: fix ComboHit v-variance and the combine chi-square

oksuzian:
`CombineStrawHits::combineHits` writes two quantities incorrectly for every **multi-straw** ComboHit. Single-straw hits are unaffected — they keep the correct values from `StrawHitRecoUtils`.

Found by a static sweep of the package at `665dc45db`. **Not observed in a running job**, and both changes alter reconstruction output, so this wants validation against a reference sample before it is merged. Flagging that up front.

## 1. `_vvar` was assigned the *time* variance

```cpp
combohit._timevar    = 1.0/twtsum;
...
combohit._vvar       = 1.0/twtsum;   // identical value
```

`twtsum` accumulates `1.0/ch.timeVar()`, so this is a time variance in ns². But `ComboHit` declares `_uvar`, `_vvar`, `_wvar` as the diagonals of the **position** covariance, and `vVar()` is consumed as a position variance in mm²:

- `TrkReco/src/RobustHelixFit.cc:1424,1441` and `TrkReco/src/MergeHelices_module.cc:200,202` form `uVar()*sin²θ + vVar()*cos²θ`, mixing it with `uVar` in a rotation;
- `CalPatRec/src/CalLineTimePeakFinder_module.cc:239` takes `sqrt(transVar() + wireVar())` as "uncertainty in the hit position in the x-y plane";
- `TrkHitReco/src/Chi2Clusterer.cc:245` and `TrkHitReco/src/TNTClusterer.cc:389` form `sqrt(1 - vVar()/uVar())` as a cluster eccentricity, which needs both terms in the same units and `vVar <= uVar`.

`StrawHitRecoUtils.cc:188` sets `_wvar = _vvar = strawOuterRadius²/3` for single hits, which is the correct form.

The constituents of a combo all come from the same panel (`panels[ch.strawId().uniquePanel()]`), so their `uDir` are parallel and their v variances simply average — matching how `midpos`, `_edep` and the end times are already combined in this function.

## 2. The quality chi-square dropped a square

```cpp
double chisq = wacc2 - wacc/wwtsum;
```

With `wacc = Σw·x` and `wacc2 = Σw·x²`, the weighted chi-square is `wacc2 - wacc²/wwtsum`. As written it is dimensionally inconsistent (`Σw·x²` minus a bare `x`) and systematically understates the quality. The correct identity is already used ten lines further down, in the `_checkWres` resolution scaling.

`ndof = nsh - 1` is safe: `combineHits` is only reached from `if (nch > 1)`, so there is no zero-dof `TMath::Prob` call.

## Testing

`g++ -std=c++20 -fsyntax-only -Wall -Wextra` against the spack view headers — patched and pristine both exit 0, no warnings attributed to the file. **No build and no art job were run**, and no reference comparison was made.


oksuzian:
Follow-up on the "wants validation" note in the description — I went and bounded the impact, and it is more specific than I first wrote. Summary: no physics cut moves, but one of the two changes feeds a trained network.

### Existing data is unaffected

Everything already on disk was produced with the current code. Nothing is retroactively invalidated; this only affects future reprocessing.

### The `_vvar` change is numerically small — which is why the bug survived

- Old: `1.0/twtsum`, where `timeVar = tres²` and `tres` is the TOT drift-time error, order 1–2 ns, so roughly 0.5–2 for a two-hit combo.
- New: `strawOuterRadius²/3` = 2.5²/3 = **2.083** (2.122 for `tracker_v7`, radius 2.5235).

Same order of magnitude. Dimensionally wrong, numerically close, so nothing ever blew up.

The consumers barely move. In `RobustHelixFit.cc:1424` and `MergeHelices_module.cc:200`, `e2 = uVar·sin²θ + vVar·cos²θ`, and `uVar` is the wire-direction variance, order (30–100 mm)² ≈ 10³–10⁴ mm². A shift of ~1 in `vVar` is a per-mille effect except where `cos²θ → 1`. Similarly for the eccentricity `sqrt(1 - vVar/uVar)`: the ratio is ~2e-4 to 2e-3 either way, so the value sits at ~1.000 before and after and changes in the third decimal.

### The chi-square change is the one that matters, and it lands on an MVA input

`qual()` has **no fcl cut anywhere** in Offline. Its only consumers are:

- `Validation/src/ValComboHit.cc:40` — a histogram
- `Print/src/ComboHitPrinter.cc` — printing
- `TrkHitReco/src/TNTClusterer.cc:394` and `TrkHitReco/src/Chi2Clusterer.cc:250` — `sqrSumQual += pow(qual(),2)`

That last one is the catch. Both clusterers assemble these into `kerasvars` and call `sofiePtr_->infer(...)` (`TNTClusterer.cc:432`) against the checked-in weights `Offline/TrkHitReco/data/TrainBkgDiag.dat`. Because `chisq_new < chisq_old`, `TMath::Prob` rises, so `qual()` increases systematically — which shifts a feature distribution under a network whose weights were fit on the **buggy** values.

So the concrete risk is not a changed selection, it is that background rejection could measurably degrade until `TrainBkgDiag.dat` is retrained, even though the code is now correct.

Two things that limit the exposure:

- `sqrSumQual` is gated on `StrawHitFlag::sline`, set only by `MakeStereoHits_module.cc:270`. `makeSTH` appears in 7 fcl/json configs, and some swap in `dummyMakeSTH` (`RecoNullProducer`), so this feature is inactive in most configurations.
- `kerasvars[11]` (the eccentricity) is always active but, per the numbers above, barely moves.

### Suggested validation, rather than a generic "please check"

A background-rejection comparison before and after on a sample where `makeSTH` is active, looking at the `kerasQ` distribution and the resulting background flags — and a decision from whoever owns `TrainBkgDiag.dat` on whether a retrain is wanted alongside this. I am happy to split the chi-square fix into its own PR if it is preferable to land the `_vvar` correction first, since that one is nearly inert numerically.

Also worth a second opinion from whoever owns `ComboHit`: I inferred the intended meaning of `_vvar` from its consumers rather than from documentation. If it was deliberately carrying something time-like, the right change would be to rename the field instead.

brownd1978 (approved):
These look like correct fixes

michaelmackenzie:
I just want to point out this failed the trigger CI test where the TPR helix trigger results changed by > 3 sigma:
```
>>> Path tpr_HlxDe_70m50p has a major change!
Local counts    : [20000, 1236, 18764, 0]
Reference counts: [20000, 1350, 18650, 0] (delta/sqrt(reference) = 3.10)
>>> Path tpr_HlxUe_50m30p has a major change!
Local counts    : [20000, 1619, 18381, 0]
Reference counts: [20000, 1778, 18222, 0] (delta/sqrt(reference) = 3.77)
```

I updated the reference log to use this new reference point.

PR #2002, opened 2026-09-21
Stage normalization

michaelmackenzie:
This adds stage normalization information to the art data, which is useful for resampling efficiencies in cases where the dataset isn't declared. This can help catch cases where the wrong `SimEfficiences2` table was used in an analysis, as well as automate efficiency calculation in studies that are not declaring test datasets (e.g. a stopping target optimization study that goes through hundreds of iterations). This is paired with a Production PR of the same name, to be posted here.


oksuzian (commented):
### PR Review Summary — Mu2e/Offline#2002

Reviewed at head `4e07f92aa307ae29594d343dab2f893ed2d54fb3`. First pass. Most of the testing was done at `fd4a36b5`; I read the one commit since, rebuilt, and repeated the resampling tests at this head.

**Decision**
- 🟡 comment only — the design is sound and I would like to see it merged. Nothing here changes a job that does not opt in, so nothing blocks; findings 1 and 2 should be settled before the product-forwarding form is switched on.

**Scope understood**
- New SubRun product `StageNormalization` (origin-equivalent generated count, events kept, chain depth), a `StageNormalizationCounter` producer for the end of each trigger path, and an opt-in `stageNormMixer` block in `Mu2eProductMixer` that either forwards the pool's product per draw (`moduleLabel`) or takes stated pool totals (`poolGenCount` + `poolEventCount`).
- Companions: Mu2e/Production#579 (configs) and Mu2e/prodtools#74 (totals from SAM). I built this head against the current `main` backing build and ran it together with #579; the numbers below come from those runs.

**Findings**

1. 🟠 [S1] The product-forwarding form reads low by the generated events of every pool subrun that kept no events
   - Evidence: a mix op is only ever called with the subrun of an event that was drawn, so a subrun with zero survivors contributes nothing, while its generated events are part of the true denominator. Test pool made with this PR's counter: three subruns with 100, 100 and 50 generated and 27, 2 and 0 kept (250 generated, 29 kept). Resampling it for 290 draws, exactly ten passes over the pool:
     ```
     moduleLabel : "poolCounter"                  -> 290 events representing 2000 origin-generated
     poolGenCount : 250  poolEventCount : 29      -> 290 events representing 2500 origin-generated
     ```
     The truth is 2500; the forwarded form is short by exactly the 50 generated in the empty subrun. The result is the same whether the empty subrun sits inside a concatenated file or in a zero-event file of its own.
   - Impact: the expected shortfall is about exp(-mu), where mu is the mean number of kept events per pool subrun. For most pools it is nothing: `sim.mu2e.TargetStops.Run1Ban-001.art` has no empty file in 5000. For `sim.mu2e.IPAStops.Run1Ban-001.art`, 111 of 5000 files are empty and they hold 2.22 % of the generated events (mu = 3.7, exp(-mu) = 2.4 %), so an IPA-stop primary would come out 2.2 % high in efficiency with nothing in the file to show it. The use case in the PR description, many small undeclared test productions, is where mu is smallest.
   - Suggested fix: the mixer cannot recover subruns it never sees, so the unbiased number has to come from the pool as a whole. The stated-totals form already does that, and the totals are available without SAM: the SubRuns tree of the pool files holds every subrun's product, empty ones included (I read `GenEventCount` that way for `MuminusStopsCat.Run1Ban` and it reproduces SAM exactly, 2000000000 / 79203). So either treat pool totals as the primary form and let prodtools fill them from the files, or keep forwarding as it is and state the exp(-mu) bias in `StageNormalization.hh` next to the existing caveat about reuse.

2. 🟠 [S1] `fromOrigin` is lost at the next forwarded stage: false becomes true
   - Evidence: the last commit records whether a normalization reaches the origin, and `resample()` carries the flag, but nothing calls `resample()`. The forwarding branch of `Mu2eProductMixer::endSubRun` still constructs `StageNormalization(perEventSum_, 0, upstreamStages_ + 1)`, which takes the constructor's default `fromOrigin = true`, and `mixStageNormalization` never reads the drawn product's flag. Two stages at this head, the first bootstrapped with the default `poolFromOrigin : false`, the second forwarding the first one's counter:
     ```
     stage 2 (totals stated)   2500 generated, nStages 2, fromOrigin False
     stage 3 (moduleLabel)     2500 generated, nStages 3, fromOrigin True
     ```
   - Impact: the flag exists to stop a stops-pool bootstrap from being read as a rate per proton (the factor of about 78 named in the new comment). Any chain with one more resampling stage after the bootstrap, for example RMC gamma stops made from a legacy stops pool and then resampled for conversions, gets exactly that label back.
   - Suggested fix: accumulate the flag per draw as `upstreamStages_` is accumulated (false if any draw is false) and pass it to the constructor, or build the product with `resample()`. A constructor default of `true` for a flag whose safe value is `false` is what let this through; I would drop the default.

3. 🟡 [S2] The legacy-pool bootstrap described in four places does not exist
   - Evidence: `EventMixing/test/stageNormResampling.fcl` tells the reader to replace `moduleLabel` with `genCounterLabel : "genCounter"` and `poolEventCount : 735`. Doing so stops at configuration: `Unsupported parameters: mu2e.products.stageNormMixer.genCounterLabel`. The same earlier design survives in the `moduleLabel` comment in `Mu2eProductMixer.hh` ("use the genCounterLabel bootstrap"), in the `endSubRun` exception text ("Configure genCounterLabel + poolEventCount"), and in `StageNormalization::resample` ("no GenEventCount fallback was available").
   - Suggested fix: name `poolGenCount` + `poolEventCount` in all four. All four are still there at this head.

4. 🟡 [S2] `StageNormalizationCounter` turns a missing input into a warning and an unseeded product
   - Evidence: an invalid `upstreamTag` or `genCountTag` handle logs a warning and writes `nStages = 0`. With #579's defaults that is every resampling job: the counters are on and `stageNormMixer` is not, so each subrun logs the warning and each output file gains a product that says nothing. A later stage that names that product as `moduleLabel` then fails in `endSubRun`, after all the events have been processed.
   - Suggested fix: a missing input the configuration named should throw; where absence is legitimate, let the configuration say so (an `OptionalAtom` tag, or a `required` flag) and write nothing. In the mixer, throw on the first draw that carries no usable normalization rather than at the end of the subrun, so a day-long job does not find out at the last step.

5. ⚪ [S3] Smaller items, together
   - `StageNormalization.cc` throws `std::runtime_error`; the coding standard asks for `cet::exception`, which `DataProducts` already uses (`StrawId.cc`).
   - `StageNormalization::resample`, `operator+=` and `operator+` have no caller: the mixer accumulates its own sum and the counter constructs directly. Finding 2 is what that costs: the fix went into the function nobody calls. Either use `resample` where its arithmetic is redone, or drop all three.

**Verified**
- 🟢 Apart from empty subruns, the per-draw sum is an unbiased estimator: a draw lands in subrun s with probability P_s / P, and contributes G_s / P_s, so the expectation per draw is the sum of G_s over non-empty subruns divided by P. The test above gives exactly that (2000 = 10 x 200).
- 🟢 `endSubRun` ordering is safe. art 3.15 runs subrun transitions path by path and in path order (`TriggerPathsExecutor::process`), not by module label; a counter named `aCounter` placed after `genCounter` was seeded correctly.
- 🟢 Run with #579 on this build: `POT.fcl` (200 events) writes `beamCounter` 200 / 11 and `neutralsCounter` 200 / 109, matching the 11 and 109 events in the two files; `Validation/ceSteps.fcl` (50 draws) writes 140970.91 = 50 x 28400000 / 10073; cosmic `S1DSStopsCRY.fcl` writes 300 generated with 19 and 12 kept, so `countAsGenerated` does what it says.
- 🟢 The bootstrap no longer assumes its own depth: with the totals stated and nothing else, the output is `nStages 2, fromOrigin False`.
- 🟢 A zero-event pool file does not upset `ResamplingMixer` with the subrun op declared: the three-file pool including one ran to completion through ten wraps.
- 🟢 Giving both or neither of `moduleLabel` and the totals, or only one total, is refused at construction with the intended message; I ran all three.

**Validation check**
- Build/tests run: FNALbuild was green at `fd4a36b5` and is pending at this head. Locally: `DataProducts`, `EventMixing` and `UtilityModules` of this head built against the `main` backing build, and the jobs listed above.
- Config contract check: pass against Production#579 for parameter names, instance name and counter labels.
- Cross-repo consistency: #579's CI is green when built together with this PR (at `fd4a36b5`).

**Residual risk**
- Any stage that selects events 1:1 without a counter of its own passes the upstream product through unchanged, and the next resampler then uses a per-event factor that belongs to a different set of events. The product cannot detect that; it is raised against #579, where the stop selectors are such a stage.

**Author follow-ups**
1. Decide how empty pool subruns are handled: pool totals as the main route, or a documented bias.
2. Carry `fromOrigin` through the forwarding branch.
3. Replace the `genCounterLabel` text in the four places.
4. Make a missing counter input an error, and fail on the first unnormalized draw.

oksuzian (approved):
### PR Review Summary — Mu2e/Offline#2002

Reviewed at head `24ac35ae14a697d40dd2f86dd02516b13dc01f0c`. Re-review of `4e07f92aa307ae29594d343dab2f893ed2d54fb3`.

**Decision**
- 🟢 approve — every finding of the first review is fixed or answered, and I verified each on a build of this head. CI is still pending at this head; the last green run was at `fd4a36b5`.

**Scope understood**
- Delta: `9b576cc2`, `945b082a`, `24ac35ae`. `fromOrigin` is accumulated per draw and the constructor defaults are gone; the bootstrap goes through `resample()`; the counter throws on a missing input, or writes nothing with `required: false`; the mixer throws on the first unseeded draw; `cet::exception` throughout; the stale `genCounterLabel` text is gone; the empty-subrun bias is documented in the header and the `moduleLabel` comment.

**Findings**
- None open.

**Verified**
- 🟢 [was S1] `fromOrigin` carried through forwarding — FIXED in `9b576cc2`, verified: a stage bootstrapped from stated totals writes `nStages 2, fromOrigin False`; the next stage forwarding that counter now writes `nStages 3, fromOrigin False` (it wrote `True` at `4e07f92a`). The counts are unchanged (2500 for 290 draws in both).
- 🟢 [was S1] empty-subrun bias — answered by documentation rather than by a code change, which was one of the two options I offered: `StageNormalization.hh` now states the exp(-mu) shortfall with the Run1B numbers, and the `moduleLabel` comment says to state totals for a sparse pool. I re-ran the three-subrun test at this head: forwarded 2000, stated totals 2500, unchanged.
- 🟢 [was S2] `genCounterLabel` text — FIXED, verified: no occurrence left in the four places, and the test fcl now shows `poolGenCount` + `poolEventCount`.
- 🟢 [was S2] missing input — FIXED in `9b576cc2`/`945b082a`, verified: with the mixer off and the counter at its default, the job fails in `endSubRun` with the `BADCONFIG` text naming `upstreamTag`; with `required: false` it runs to completion and the output carries no product for that label (art accepts the declared-but-not-put SubRun product; the file's branch is present and empty). A draw from an unseeded product now throws at the draw rather than at the end of the subrun (read from the code; not exercised, since no test pool carries an `nStages 0` product any more).
- 🟢 [was S3] `cet::exception` in `StageNormalization.cc`, and `resample()` now has a caller (the bootstrap branch). `operator+=` and `operator+` still have none; they are the merge for a future concatenation stage and I do not mind them staying.
- 🟢 `S1DSStopsCRY.fcl` from Production#579 at `c22aa8cf` runs on this build: 300 generated, 19 and 12 kept, and `GenEventCountReader` accepts both output files. `POT.fcl` runs.
- 🟢 The bad-configuration refusals (both forms, neither, one total) are unchanged.

**Validation check**
- Build/tests run: FNALbuild pending at this head. Locally: `DataProducts`, `MCDataProducts`, `EventMixing` and `UtilityModules` of this head built against the `main` backing build, and the jobs above.
- Not checked: `Validation/ceSteps.fcl` at this head. In my partial build the loader picks the backing build's `libmu2e_DataProducts.so` ahead of mine and fails on the new `resample` symbol, which is an artefact of building four packages against a full backing build, not of the PR; the same code path ran in the bootstrap test above, and CI's `ceSteps` will cover the config.
- Config contract check: pass against Production#579 at `c22aa8cf` (`required`, `poolStages`, `poolFromOrigin`, `srOutInstance`).

**Residual risk**
- `poolEventCount` is a `double` atom and is truncated to `uint64_t` when the pool normalization is built; a non-integer value is accepted silently. prodtools#74 always writes an integer, so this only matters for a hand-written override.

**Author follow-ups**
- None required.

oksuzian:
@FNALbuild run build test

Issue #1997, opened 2026-09-21
Straw DAQ modules: unchecked ROC loop bound, corrupted flag set on a copy, and StrawDigiFilter/StrawHitFilter bugs


oksuzian:
Part of #1993, split out so each owner gets a short list. cc @pavel1murat — most commits on these files since 2024, by `git log`; please re-assign if that is wrong.

Context: a static review of `Offline/DAQ` at main `ef31f95df`, against the rules for pass1 on real data (also the Mu2e coding standard for errors and output):

1. Error prints go through message facility, not `std::cout`/`std::cerr`/`printf`, so they can be rate-limited.
2. Debug prints are gated by a fcl level and default off.
3. Fatal errors (assert, throw) are only for memory corruption or a missing resource, never for bad input data.
4. Corrupt data is always reported and skipped — counted, not silently dropped, and not turned into a product.

Every item below was re-read by hand at `ef31f95df`. Nothing was built or run.

## `StrawDigisFromArtdaqFragments` — runs in the trigger through `TrigMakeSD: { @table::DAQ.producers.makeSD }`

Open PR #1990 rewrites this file. The first two items are still present at its head `9b6fbe817` (`:336` and `:500-501`), so that PR may be the easiest place to fix them.

- [ ] **`:311-316`** — rule 4. `int nbytes = buf[0]; ... uint8_t* last_address = fdata+nbytes;` takes the loop bound from a payload word and never checks it against `frag.dataSizeBytes()`. Every later bound trusts it, so a corrupt first word walks the ROC loop past the fragment.
- [ ] **`:478-479`** — rule 4. `auto digi = straw_digis->back(); digi.digiFlag() = digi_flag;` sets the flag on a copy. `StrawDigiFlag::corrupted` never reaches the stored digi. Needs `auto&`.
- [ ] **`:246-254`** — container fragments are unpacked into a vector that is never decoded.
- [ ] **`:281-293`** — no fragment-size check at all when `missingDTCHeaders` is true.
- [ ] **`:148-152`** — the `debugBits` index from `sscanf` is not range-checked before it indexes the array.

## `StrawDigiFilter`, `StrawHitFilter` — test tools; nothing in Production, mu2e-trig-config or PassN runs them

- [ ] **`StrawDigiFilter_module.cc:354-357`** — `panel_map_by_online_ind(plane, panel)` is called with the offline plane and panel from `strawId()`; that accessor is keyed by (dtc, link). The row returned is null or the wrong panel.
[...]

PR #2003, opened 2026-09-22
DAQ: one addressing choice instead of three parallel booleans

oksuzian:
`StrawDigisFromArtdaqFragments` decides how to turn `(dtc_id, link_id, payload MnID)` into a `StrawId` using three independent FHiCL booleans — `keyOnMnid`, `forceOfflineAddressing` and `allowOfflineFallbackWhenPanelMapMissing`. That is eight spellable combinations for what is really one three-valued choice, and `forceOfflineAddressing` has to be re-tested at four separate sites, three of which exist only to switch off work that the fourth then discards.

This replaces them with one choice and one policy:

```
addressing   : byLink | byMnid | geographic
onMissingRow : skip   | geographic
```

| before | after |
|---|---|
| `forceOfflineAddressing: true` | `addressing: geographic` |
| `keyOnMnid: true` | `addressing: byMnid` |
| `keyOnMnid: false` | `addressing: byLink` |
| `allowOfflineFallbackWhenPanelMapMissing: true/false` | `onMissingRow: geographic/skip` |

Combinations that used to be expressible but meaningless — `forceOfflineAddressing` together with `keyOnMnid`, for instance — can no longer be written down. An unrecognised value throws a `cet::exception` rather than silently selecting a mode.

### No behaviour change

I modelled the five decision points, old logic against new, and compared them over every reachable state: **30 states, 0 mismatches.** Two of the five simplify as a consequence rather than by choice:

- The payload-vs-hardware cross-check becomes `addressing_ == byLink && tpm && tpm->mnid() != mnid`. In `byMnid` the row was fetched *by* mnid, so `tpm->mnid() == mnid` holds by construction (`TrackerPanelMap.cc:40` — `_tpm_by_mnid[r->mnid()] = r`); in `geographic` there is no row. The old guard could never fire in either mode, so restricting it to `byLink` changes nothing.
- The `StrawId` falls back to `(dtc,link,straw)` on `tpm == nullptr` alone. The old `forceOfflineAddressing_ ||` term was redundant, since neither lookup runs in that mode.

Worth stating plainly: in the `byMnid` mode nothing cross-checks the payload channel ID against the hardware address it arrived on. That is pre-existing and this PR does not change it — the refactor only makes it visible instead of hiding it behind a guard that looked live. Tracked in #1997.

### Also fixed

`DAQ/test/digi_to_frag.fcl` set `geography : []`. The module has declared no such parameter for some time, so validated FHiCL rejects it and **that job cannot start today**. No `DAQ/test` configuration runs in CI, which is why it went unnoticed. Removed here.

### Scope

Only `DAQ/fcl/prolog.fcl` and `DAQ/test/digi_to_frag.fcl` set these keys. `mu2e-trig-config` sets none of them — `TrigMakeSD` is `@table::DAQ.producers.makeSD` with no overrides, so updating the prolog carries it. `Production` does not reference the module.

Rebased onto `main` after #1990 merged (`5b9d048f`). The prolog line is now `addressing : "byMnid"`, which is the translation of `keyOnMnid: true` as that PR set it — not a new decision.

Happy to close this if @pavel1murat would rather make the change himself; it comes out of the #1990 discussion, where he agreed the three-boolean logic needs fixing.

### What I did not check

I could not build locally — muse is not set up in this session, so the only compile evidence is CI. The new enum and parse helpers do build standalone under `-std=c++20 -Wall -Wextra -Werror`, but that is not the same as compiling the module. I also have not run any job over real fragment data; nothing in CI decodes artdaq fragments.


oksuzian:
@FNALbuild run build test

PR #2005, opened 2026-09-22
changed cerr to mf::LogError in CRV code of DAQ

ehrlich-uva:
Changes recommended by AI.
Changed the cerrs to mf::LogErrors in the CRV part of Offline/DAQ.
Removed two outdated CRV modules from Offline/DAQ. Should also be done artdaq-core-mu2e.
Changed continue to break in CrvDigisToFragments_module to skip NZS digi collections with just one error message, since all digis of the collection will be NZS.
Changed the default of the CrvDigi comparison: It will fail on a mismatch.


oksuzian (changes requested):
### PR Review Summary — Mu2e/Offline#2005

Reviewed at head `f483b2a7e603e953334fcb845de15b8a71082ddc`. First pass.

**Decision**
- 🔴 request changes

The 🔴 is a one-line-per-module omission and should be quick. The 🟠 is a behaviour
change in `CrvDigisToFragments` that is not described in the PR and may not be the one
that was intended.

**Scope understood**
- `CrvDigisFromArtdaqFragmentsFEBII` and `CrvDigisToFragments`: `std::cerr` and the
  `std::cout` lines that accompanied errors become `mf::LogError`.
- `CrvDigisFromArtdaqFragments_module.cc` (257 lines) and
  `CrvGRdataFromArtdaqFragments_module.cc` (271 lines) are deleted as outdated.
- `CrvDigiCollectionsComparator`: the `failOnMismatch` default changes from `false` to `true`.

**Findings**

1. 🔴 [S0] Both deleted modules still have their `cet_build_plugin` entries, so the build cannot configure
   - Evidence: `DAQ/CMakeLists.txt:214-224` still declares
     `cet_build_plugin(CrvDigisFromArtdaqFragments art::module / REG_SOURCE src/CrvDigisFromArtdaqFragments_module.cc ...)`
     and `:239-248` does the same for `CrvGRdataFromArtdaqFragments`. Both source files are
     deleted by this PR, and `DAQ/CMakeLists.txt` is not among the changed files.
   - Impact: `check_cmake` fails with
     `File .../Offline/DAQ/src/CrvDigisFromArtdaqFragments_module.cc defined in CMakeLists.txt but not found in directory!`
     — this is exactly what happened on #1990 two days ago, where the same omission turned the
     job red. There the block had at least been commented out so the build itself survived;
     here the rule is live, so I would expect the build to fail as well rather than just the
     cmake check.
   - Suggested fix: delete both blocks — `DAQ/CMakeLists.txt:214-224` and `:239-248`. Deleting
     is the fix, not commenting out; `check_cmake` reads the filename through a comment prefix.

2. 🟠 [S1] In `CrvDigisToFragments`, `continue` became `break`, which truncates the collection rather than skipping it
   - Evidence: `CrvDigisToFragments_module.cc:171-175`. The NZS test changed from
     `if(digi.IsNZS()) continue;` to a block that logs
     `"This seems to be a CrvDigiCollection with NZS data. Skipping this collection."` and then
     `break`. The `break` leaves the digi loop at `:167-204`, and control falls straight into the
     `for(auto& [dtcID, linkMap] : byDtcAndLink)` loop at `:206`, which builds DTC events out of
     whatever `byDtcAndLink` accumulated before the NZS digi was reached.
   - Impact: if the first NZS digi is not the first digi in the collection, the digis before it
     are still encoded and written out, and `_totalDigisEncoded` counts them. The collection is
     not skipped — a truncated version of it is produced, and the log line says the opposite.
     Nothing downstream can tell a truncated fragment collection from a complete one. If the
     first digi is NZS the result is an empty `byDtcAndLink` and nothing is produced, which is
     presumably the intended case, so the behaviour depends on where in the collection the NZS
     digi sits.
   - Suggested fix: decide which is meant. To skip the collection, test for NZS before the
     encoding loop — a scan, or the first digi if NZS is a per-collection property — and return
     without populating `byDtcAndLink`, so nothing partial is emitted. If instead individual NZS
     digis should be dropped, keep `continue` and reword the message. Either way the message and
     the control flow should agree.

3. 🟡 [S2] The `failOnMismatch` default flips from `false` to `true`, and the only consumer does not set it
   - Evidence: `CrvDigiCollectionsComparator_module.cc:33` changes the `fhicl::Atom<bool>`
     default to `true`. The only configuration that instantiates this analyzer,
     `DAQ/test/compareCrvDigiCollections_extracted.fcl:43-49`, sets `diagLevel`, `referenceTag`
     and `candidateTag` but not `failOnMismatch`, so that job now throws on the first mismatch
     instead of reporting them.
   - Suggested fix: this may well be what you want for a validation tool, but it is a behaviour
     change in a PR whose title is about logging. Please mention it in the PR body, and consider
     setting the value explicitly in the test fcl so the intent is visible where the job is read.

4. 🟡 [S2] Three unrelated changes in one PR
   - Evidence: the logging conversion, the removal of two modules, and the `failOnMismatch` flip
     are independent. The title and body describe only the first, and the module removal is
     mentioned in the body but not the default change.
   - Suggested fix: not worth re-cutting at this point, but the module removal in particular
     would have been easier to review on its own — it is the part carrying the blocker.

**Verified**
- 🟢 The `mf::LogError` conversion preserves the error accounting. `CrvDAQerrorCode`
  `emplace_back` sites: 9 on `main`, 9 at this head. `std::cerr` goes 19 → 0, and the 19 lines
  become 8 multi-line `mf::LogError` calls, so several related lines are now one message rather
  than several — which is the right shape for a rate-limited destination.
- 🟢 The `std::cout` that remain are diagnostics, not error paths. All the ones I sampled in
  `CrvDigisFromArtdaqFragmentsFEBII` sit inside `if(_diagLevel>N)` blocks — `:118`, `:182`,
  `:301`, `:314` — and the ones in `CrvDigisToFragments` at `:180` and `:189` likewise. That is
  what the coding standard asks for, so leaving them is correct and not an incomplete conversion.
- 🟢 Neither deleted module has a FHiCL consumer. `module_type` for
  `CrvDigisFromArtdaqFragments` and `CrvGRdataFromArtdaqFragments` appears in no `.fcl` in
  `Offline` at this head or on `main`, none in `Mu2e/Production` at `main`, and none in
  `mu2e-trig-config`. Worth noting how I checked: my local `Production` checkout is on a personal
  branch and does contain two `Pass1/Wideband/CrvPass1_*.fcl` files that instantiate
  `CrvDigisFromArtdaqFragments`, which looked like a blocker until I pulled upstream `main` — that
  directory does not exist there. So the deletions are safe as far as committed configuration
  goes, but if those wideband configs are heading for Production, they will need updating first.

**Validation check**
- Build/tests run: **no CI result at this head.** FNALbuild triggered at `051fef0d`, the first of
  the two commits, so the run in flight covers the logging change but not the module deletions —
  it cannot see finding 1. I did not trigger a second run, because the failure is caused by the
  diff and re-running would only reproduce it; please re-run after fixing `DAQ/CMakeLists.txt`.
  I ran nothing locally.
- Config contract check: pass for the modules that remain; see finding 3 for the default that
  changed under an unchanged `.fcl`.
- Cross-repo consistency: pass, with a caveat. The PR body notes the same removal "should also be
  done artdaq-core-mu2e" — that is a follow-up in another repo, not a blocker here.

**Not checked**
- I did not read the two deleted modules to confirm they are genuinely superseded by the FEBII
  versions; that is your call as the author and the CRV owner.
- I have not run any job over real CRV fragment data, and nothing in `buildtest` decodes artdaq
  fragments, so none of the runtime behaviour above is covered by CI either way.

**Residual risk**
- Finding 2 is the one that can reach data: it changes what gets written, and the current log
  line would make a truncated collection look like a skipped one in the logs.

**Author follow-ups**
1. Delete the two stale `cet_build_plugin` blocks at `DAQ/CMakeLists.txt:214-224` and `:239-248`,
   then re-run the build test.
2. Decide whether the NZS case should skip the collection or skip the digi, and make the control
   flow and the message agree.
3. Mention the `failOnMismatch` default change in the PR body, and consider setting it explicitly
   in `DAQ/test/compareCrvDigiCollections_extracted.fcl`.

oksuzian (commented):
### PR Review Summary — Mu2e/Offline#2005

Reviewed at head `55fae9489a100e780479d4c778be6e99016335c8`. Re-review of `f483b2a7e603e953334fcb845de15b8a71082ddc`.

**Decision**
- 🟡 comment only

The blocker is fixed and verified. Two items from the last round are still open and neither is
a blocker, so I am not asking for changes on them — but please note that a comment does not lift
my earlier request, so the PR will keep showing as blocked until I approve. I will do that as soon
as the NZS question has an answer, even if the answer is "the current behaviour is what I want".

**Scope of this pass**
- One commit, `55fae9489`, −23 lines in `DAQ/CMakeLists.txt` and nothing else.

**Verified fixed**

- 🟢 [was 🔴 S0] Both stale `cet_build_plugin` blocks are gone — `CrvDigisFromArtdaqFragments` and
  `CrvGRdataFromArtdaqFragments` — and they were deleted rather than commented out, which is what
  `check_cmake` needs. `git grep` for either name over the whole tree at this head returns nothing
  outside the surviving `...FEBII` module, so no build rule, FHiCL file or source now refers to
  either one. Verified in the diff, not from the commit message.

**Carried over, both unaddressed — this delta touched only `DAQ/CMakeLists.txt`**

1. 🟠 [S1] carried over — the `continue` → `break` in `CrvDigisToFragments` still truncates the
   collection rather than skipping it
   - Evidence: `CrvDigisToFragments_module.cc:171-175` is unchanged. The `break` leaves the digi
     loop and control falls into the DTC-event builder, so digis encountered before the first NZS
     digi are still encoded and written, while the message says the collection was skipped.
   - This is the one item I would like a reply on rather than a change. If NZS is in practice a
     per-collection property — all digis in a run are NZS or none are — then the first digi decides
     and the current code is equivalent to skipping, in which case say so and I will close it. If
     a mixed collection is possible, the partial output is the problem.

2. 🟡 [S2] carried over — `failOnMismatch` still defaults to `true` at
   `CrvDigiCollectionsComparator_module.cc:33`, and `DAQ/test/compareCrvDigiCollections_extracted.fcl`
   still does not set it, so that job now throws on the first mismatch. Worth one line in the PR
   body, and ideally setting the value explicitly in the test fcl.

**Validation check**
- Build/tests run: `mu2e/buildtest` is **pending** at `55fae9489`. The only completed run on this
  PR is at `051fef0d`, the first commit, which predates both the module deletions and this fix — so
  there is still no green covering the deletions. Nothing to do but wait for this run; I did not
  trigger another, and I ran nothing locally.
- **The branch now conflicts with `main`** (`mergeStateStatus: DIRTY`). `DAQ/CMakeLists.txt` at this
  head still contains `cet_build_plugin(StrawDigiFilter ...)`, which #1990 removed from `main` when
  it merged earlier today. Same file, so the two edits collide. A merge or rebase onto current
  `main` (`5b9d048f`) should resolve it — just keep `main`'s removal of the `StrawDigiFilter` block
  alongside your removal of the two CRV blocks.
- Config contract check: pass, unchanged. Cross-repo consistency: pass; the `artdaq-core-mu2e`
  follow-up you noted in the PR body is a separate repo and not a blocker here.

**Not checked**
- Whether the two deleted modules are genuinely superseded by the FEBII versions — your call as the
  CRV owner, unchanged from last time.
- I have still not run anything over real CRV fragment data, and nothing in `buildtest` decodes
  artdaq fragments.

**Author follow-ups**
1. Merge or rebase onto `main` to clear the `DAQ/CMakeLists.txt` conflict.
2. Say whether a `CrvDigiCollection` can hold a mix of NZS and non-NZS digis. If it cannot, I will
   close finding 1 and approve.
3. Optional: mention the `failOnMismatch` default change in the PR body.

ehrlich-uva:
Comments:
1. Fixed.
2. Updated the comment above: Changed continue to break in CrvDigisToFragments_module to skip NZS digi collections with just one error message, since all digis of the collection will be NZS.
3. Updated the comment above.

ehrlich-uva:
A single CrvDigiCollection doesn't hold a mixture of NZS and ZS data. It is of course technically possible to write a module that puts NZS and ZS digis into the same collection, but none of the existing modules do that.

oksuzian (approved):
### PR Review Summary — Mu2e/Offline#2005

Reviewed at head `55fae9489a100e780479d4c778be6e99016335c8`. Re-review of `55fae9489a100e780479d4c778be6e99016335c8` — same head, closing out the NZS question answered above.

**Decision**
- 🟢 approve

Nothing further from me. Two things still stand between this and the merge button, but neither is
a review finding: the branch conflicts with `main`, and `mu2e/buildtest` has not reported at this
head yet.

**Closed**

- 🟢 [was 🟠 S1] The `continue` → `break` in `CrvDigisToFragments`. Answered: a single
  `CrvDigiCollection` does not hold a mixture of NZS and ZS digis, so the first digi decides and
  `break` is equivalent to skipping the collection. Withdrawn as a defect.
- 🟢 [was 🔴 S0] The two stale `cet_build_plugin` blocks, fixed in `55fae9489` and verified in the
  previous pass.

**Still open, neither gating**

1. ⚪ [S3] The equivalence above rests on a convention rather than on anything the code enforces —
   as you put it, a module *could* put both kinds in one collection, none currently does. If you
   ever want that to be self-enforcing, testing `crvDigis.front().IsNZS()` once before the loop and
   returning would say "this is a per-collection property" in the code rather than in the mailing
   list. Entirely optional, and the current behaviour is correct today.
2. 🟡 [S2] carried over — `failOnMismatch` still defaults to `true` at
   `CrvDigiCollectionsComparator_module.cc:33` while
   `DAQ/test/compareCrvDigiCollections_extracted.fcl` does not set it, so that job now throws on
   the first mismatch rather than reporting. Worth a line in the PR body.

**Before merging**
- The branch conflicts with `main` (`mergeStateStatus: DIRTY`): `DAQ/CMakeLists.txt` here still
  contains the `cet_build_plugin(StrawDigiFilter ...)` block that #1990 removed from `main` this
  morning. Rebase or merge onto `5b9d048f`, keeping both removals.
- `mu2e/buildtest` is pending at `55fae9489`; the only completed run is at `051fef0d`, which
  predates the module deletions. Worth confirming it comes back green before merging, since this is
  the first run that will actually exercise the deletions and the CMake change.

**Not checked**
- Whether the two deleted modules are genuinely superseded by the FEBII versions — your call as the
  CRV owner.
- Nothing here has been run over real CRV fragment data, and `buildtest` does not decode artdaq
  fragments, so the runtime behaviour of the logging change is not covered by CI either way.

oksuzian:
@FNALbuild run build test

PR #2006, opened 2026-09-22
Add straw status to trkstraw

bonventre:
Allows determining if straw xing missing hit is expected or not


oksuzian (approved):
### PR Review Summary — Mu2e/Offline#2006

Reviewed at head `718bc43f53f51779aa350ccffba7f1f06219cd47`. First pass.

**Decision**
- 🟢 approve — only minor findings; the build test is still running at this head.

**Scope understood**
- Adds a `dead` bit (value 4) to `StrawFlag`. `KKFit::createSeed` sets it on each `TrkStraw` whose straw is `noSignal()` or `suppress()` in `TrackerStatus`, so an analysis can tell an expected missing hit from an unexpected one. All six `createSeed` callers now fetch `TrackerStatus` and pass it in.
- Also adds `TrkStraw::_phi`, the angle of the track's POCA around the wire. The PR description does not mention this.
- Companion PR: Mu2e/EventNtuple#413 copies both fields into `TrkStrawMatInfo`. It has to merge after this one.

**Findings**
1. 🟡 [S2] `KKFit.hh` now uses `TrackerStatus` without including its header.
   - Evidence: `createSeed` takes a `TrackerStatus const&` (`Mu2eKinKal/inc/KKFit.hh:121`, `:676`) and calls `noSignal`/`suppress` (`:762`), but the header's include list (`:6-49`) has no `TrackerConditions/inc/TrackerStatus.hh`. It compiles only because all six including modules happen to include `TrackerStatus.hh` first (for example `LoopHelixFit_module.cc:21` before `:53`). No other header in the repo includes `TrackerStatus.hh` except `TrackerStatusMaker.hh` and `StrawHitRecoUtils.hh`, so there is no transitive path.
   - Suggested fix: add `#include "Offline/TrackerConditions/inc/TrackerStatus.hh"` next to the existing `StrawResponse.hh` include at `KKFit.hh:19`.

2. 🟡 [S2] `KKTruthSeed` constructs its `ProditionsHandle<TrackerStatus>` every event.
   - Evidence: `KKTruthSeed_module.cc:316` declares a local `ProditionsHandle<TrackerStatus> trackerStatus_h;` inside `produce`. The other five modules in this PR make it a class member (`trackerStatus_h_`), which is the convention for Proditions handles.
   - Suggested fix: make it a member `trackerStatus_h_`, as the other five modules do.

3. ⚪ [S3] `_phi`: the comment's "+v" does not describe what the code computes, and the PR description does not mention `_phi`.
   - Evidence: `TrkStraw.hh:39-42` picks `raddir = sdir × ẑ`, flips it to point radially outward, and computes `atan2(cperp·raddir, cperp·ẑ)`, so +π/2 means radially outward. The panel/ComboHit `v` is `ẑ × u` (`ComboHit.hh:37`), which is the opposite sign of `sdir × ẑ` before the flip. I did not check whether `ẑ × u` points outward for every panel. If it does not, the "+v" comment in both this PR and EventNtuple#413 is wrong for part of the tracker. The comment on `smid` also says "straw middle", but `sensorPoca()` is the POCA on the wire. The sign test still works because `raddir ⟂ sdir`.
   - Smaller points on the same lines: `static XYZVectorF zdir` is a mutable function-local static in a data-product constructor and should be `static const`. The 10-line geometry calculation would read better in `TrkStraw.cc` than inline in the header, since the coding standard asks data products to keep complicated functions out of the class. Files written before this PR read back `_phi = 0` and `dead() == false`, which cannot be told apart from real values. That is worth saying in the description.
   - Suggested fix: change the comment to "+pi/2 is radially outward", fix "straw middle", and add a line about `_phi` to the PR description.

**Verified**
- 🟢 `dead = 4` is appended after `drifthit = 3` (`StrawFlag.hh:17`), so bit values already written to files keep their meaning. `bitNames()` gains a matching `"Dead"` entry (`StrawFlag.cc:27`).
- 🟢 A `grep -rn 'createSeed('` over the whole repo at this head finds exactly the six call sites in the diff, and all six pass the new argument.
- 🟢 `trackerStatus : @local::TrackerStatus` is in the default ProditionsService table (`ProditionsService/fcl/prolog.fcl:15`), so jobs that already configure Proditions do not need new fcl.
- 🟢 Adding a `float` member with an in-class default to `TrkStraw` needs no `ioread` rule. The existing rules at `classes_def.xml:215-234` are keyed on the old checksum and are not affected.

**Validation check**
- Build/tests run: no local build. `mu2e/buildtest` was triggered by FNALbuild at `718bc43f` and is still running (pending), so there is no result yet.
- Config contract check: pass. No fcl keys were added.
- Cross-repo consistency: EventNtuple#413 depends on this PR. Merge this one first.

**Residual risk**
- Not checked: I ran no job and did no before/after comparison of KalSeeds. `_phi` and the `dead` bit are new outputs only. They do not feed back into the fit, so existing fit results should be unchanged.

**Author follow-ups**
1. Add the `TrackerStatus.hh` include to `KKFit.hh`.
2. Make the `KKTruthSeed` handle a class member.
3. Fix the `_phi` comment and mention `_phi` in the PR description.

RobMina (approved):
All changes look good to me! I ran some tests exercising the modules touched by this PR and found no issues.

